PR #14, opened 2017-08-22
moab_storage/index  implemented (first pass)

### LINKED ISSUE #5: PCC:  ReST call returns all druids found in (fixture) storage root
This may be split into multiple tickets -- one to get the druids, and another to present them via a ReST call?

Blocked by #2

ndushay:
- created `MoabStorageController` with `index` action that currently looks for druid entries in hard-coded relative location `spec/fixtures/moab_storage_root` and serves them up as JSON (default) or XML.
- also added 2 more fixture moab objects that are tiny ones from the moab_versioning gem

### JSON

json response

### XML

xml response

Resolves #5


SaravShah:
LGTM;  👍  to merge

PR #19, opened 2017-08-23
Naomi sarav pair current version dirty

SaravShah:
Figured out a way to get current version from a given druid.


SaravShah:
Rspec is not passing because I had to split up the directory from 'dg806ms0373' to 'dg/806/ms/0373/dg806ms0373'. The rspec test is checking to see if the list of druids includes 'dg806ms0373', instead its 'dg'.

SaravShah:
This isn't ready yet.

PR #29, opened 2017-08-24
Rubocop and comment out unneeded gems

### LINKED ISSUE #22: PCC: add rubocop
- [x] add rubocop gem with reasonable rubocop.yml defaults to project;
- [x] add rubocop rake task
- [x] add rubocop run to build used by travis

consider pegging the version of rubocop so we aren't churning on code because some folks decided they like certain syntax better than other syntax.

from https://github.com/bbatsov/rubocop
```
RuboCop's development is moving at a very rapid pace and there are often backward-incompatible changes between minor releases (since we haven't reached version 1.0 yet). To prevent an unwanted RuboCop update you might want to use a conservative version locking in your Gemfile:

gem 'rubocop', '~> 0.49.1', require: false
```

ndushay:
Resolves #22 

Also commented out gems related to html or js or css or windows, as this is going to be json or xml responses only.   Happy to move that commit to a separate PR.

We can twiddle our rubocop settings if these initial ones prove annoying ... or nominate your favorites in the comments!


Gemfile
```diff
@@ -15,16 +15,16 @@ gem 'sqlite3'
 # Use Puma as the app server
 gem 'puma', '~> 3.7'
 # Use SCSS for stylesheets
-gem 'sass-rails', '~> 5.0'
+# gem 'sass-rails', '~> 5.0'
```
tingulfsen: Why not just delete instead of comment out?

ndushay: Excellent question!   All of this stuff was added by the rails generator, including pinning to versions.  I was feeling tentative, so I commented stuff out.  I would be happy to remove cruft if we prefer -- we can always get it back from git history or another rails app if we need it.  What do you think?

tingulfsen: My opinion on this one is that we might as well remove this stuff from the Gemfile. These are fairly small changes that would be easy to reinstate, even just by - as you say - running the rails generator again to figure out what we're missing.

I'm less certain about #30 - I believe that removing some of those directories can come back to bite us later and it might be worth just leaving them in there?

PR #36, opened 2017-08-25
removed unused Rails scaffolding and unused gems 

### LINKED ISSUE #30: PCC: remove rails pieces we won't use
(we can always add them back later if we need them;  `rails new` created a lot of stuff we won't use)

- [x] app/assets folder
- [x] app/channels folder
- [x] app/views folder
- [x] config/cable.yml
~~- [ ] config/spring.rb~~  we 'need' this for development environment
- [x] config/initializers/assets.rb (? - might also require tweaks to `application.rb` to not use the rails piece, and then to `config/environments/xxx.rb` to not use the settings ...)

- [x] lib/assets folder
- [ ] lib/tasks folder
- [ ] spec/helpers folder
- [x] spec/views folder

less certain 
- [ ] app/helpers
- [x] app/jobs
- [x] app/mailers

ndushay:
Resolves #30


Gemfile
```diff
@@ -14,21 +14,9 @@ gem 'rails', '~> 5.1.3'
 gem 'sqlite3'
 # Use Puma as the app server
 gem 'puma', '~> 3.7'
-# Use SCSS for stylesheets
-# gem 'sass-rails', '~> 5.0'
-# Use Uglifier as compressor for JavaScript assets
-# gem 'uglifier', '>= 1.3.0'
-# See https://github.com/rails/execjs#readme for more supported runtimes
-gem 'therubyracer', platforms: :ruby
```
ndushay: this was mistakenly believed to be needed when we were adding Capistrano;  have no fear - the tests pass and the app works.

config/application.rb
```diff
@@ -1,6 +1,9 @@
 require_relative 'boot'
 
-require 'rails/all'
+# Select only parts we need from rails/all
+require 'rails'
+require 'active_record/railtie'
+require 'action_controller/railtie'
```
ndushay: it is my hope that the tests will run faster if they don't need to load so much stuff, most of which we weren't using!

PR #37, opened 2017-08-25
Add config gem so we can use Settings.yml etc.

### LINKED ISSUE #21: PCC: use settings.yml instead of hardcoding
we have a branch (`show-action-naomi`) with a clear need for some settings so this can be set up off that branch (or master, if PR #25 is merged)

e.g.  need config/settings.yml  with values to be used by `config/application.rb`

We want to do the same sort of configuration approach we use in most rails apps, with the `config` gem and `config/settings.yml` with defaults and `config/settings/[env].yml` also with defaults, with no sensitive info in `config/settings/production.yml`.   `dlme` project might be a good example.

This will set us up for #4, where we put VM specific or sensitive information into shared_configs and hook it up so rails app finds it as if it was in `config/settings/[env].yml`

tingulfsen:
Fixes #21

### DIFF
--- a/config/initializers/config.rb
+++ b/config/initializers/config.rb
@@ -1,0 +1,4 @@
+Config.setup do |config|
+  # Name of the constant exposing loaded settings
+  config.const_name = 'Settings'
+end
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ coverage/*
 config/settings.local.yml
 config/settings/*.local.yml
 config/environments/*.local.yml
+
+.ruby-gemset
+.ruby-version
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,8 @@ gem 'jbuilder', '~> 2.5'
 # Use ActiveModel has_secure_password
 # gem 'bcrypt', '~> 3.1.7'
 
+gem 'config'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: [:mri]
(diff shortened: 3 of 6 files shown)


config/initializers/config.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  # Ability to process variables names:
+  #   * nil  - no change
+  #   * :downcase - convert to lower case
+  #
+  # config.env_converter = :downcase
+
+  # Parse numeric values as integers instead of strings.
+  #
+  # config.env_parse_values = true
```
jmartin-sul: on the one hand, this feels like it runs afoul of what we talked about this morning where we don't include commented out boilerplate.  on the other hand, i'd have been inclined to leave it also.  i don't think i have an opinion?  what do others think?

ndushay:
In the end, I don't think we check in `development.yml` either after we have shared_configs, but we do need an example  .yml and README instructions on how to turn example.yml (or test.yml) into development.yml for local development.

config/initializers/config.rb
ndushay: given that we actually do use configs a lot and I have never noticed all this boilerplate or mucked about with these settings, I would be comfortable removing all this boilerplate.  But I don't have strong feels and we could leave it, too.

tingulfsen:
I think I probably addressed the concerns mentioned above, rebased on master. I left development.yml and test.yml in there for the moment. We can remove those too if you like, but maybe it's easier to leave them since we know we will need them but don't have any offending data yet.

ndushay:
👍  to merge.  @jmartin-sul  I'll leave it to you ...

PR #28, opened 2017-08-24
[WIP] initial schema

jmartin-sul:
a first pass at the tables needed for the object inventory store: preserved_objects (a master record covering all preserved copies for a given druid), endpoints (unique replication targets), and preservation_copies (individual preserved copies of preservation objects, on specific endpoints).

pairing w/ @SaravShah on implementation.  met w/ @LynnMcRae and @julianmorley to refine schema choices.

still need to write tests.

Resolves #16
Resolves #17
Resolves #31

PR #43, opened 2017-08-28
initial schema and database configuration

### LINKED ISSUE #16: PCC: create initial db schema
Based on https://docs.google.com/document/d/1XaRDlI0aSPtbNe7JPKiCjeIDHhipdJCLXP1ltO9YIAA/edit

we think a first pass at fields might be what's below:

Druid
Local disk location
Object size
Current Preservation Core version number
Moab ingest status
Timestamp for last Moab ingest change for the current object
For each archive copy of the current object:
Replication type (what is this?)
Replication status (what is this?)
Timestamp for last replication change
Latest audit findings (how do we best express this?)
Overall status (what is this?)

### LINKED ISSUE #17: PCC: add postgres configs to rails app
at least for dev db environment

- update README with instructions for how to configure laptop to use postgres

### LINKED ISSUE #31: Create tests for database
- Check uniqueness of druid objects
- Basic tests to check db

jmartin-sul:
A first pass at the tables and models needed for the object inventory store: preserved_objects (a master record covering all preserved copies for a given druid), endpoints (unique replication targets), and preservation_copies (individual preserved copies of preservation objects, on specific endpoints).

paired w/ @SaravShah on implementation.  met w/ @LynnMcRae and @julianmorley to refine schema choices (meeting w/ @julianmorley again this afternoon for further schema refinement).

supersedes #28 

Resolves #16
Resolves #17
Resolves #31

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,11 @@ language: ruby
 rvm:
   - 2.4.1 # make sure this matches our VMs
 cache: bundler
+services:
+  - postgresql
+before_script:
+  - psql -c 'CREATE USER preservation_core_catalog;' -U postgres
+  - psql -c 'ALTER USER preservation_core_catalog CREATEDB;' -U postgres
+  - psql -c 'CREATE DATABASE preservation_core_catalog;' -U postgres
+  - psql -c 'ALTER DATABASE preservation_core_catalog OWNER TO preservation_core_catalog;' -U postgres
+  - psql -c 'GRANT ALL PRIVILEGES ON DATABASE preservation_core_catalog TO preservation_core_catalog;' -U postgres
--- a/Gemfile
+++ b/Gemfile
@@ -5,28 +5,23 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
-# work with Moab Objects
-gem 'moab-versioning'
-
-# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 5.1.3'
-# Use Puma as the app server
-gem 'puma', '~> 3.7'
-
+# general Ruby/Rails gems
+# gives us Settings to manage configs on different instances
+gem 'config'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem 'jbuilder', '~> 2.5'
-# Use ActiveModel has_secure_password
-# gem 'bcrypt', '~> 3.1.7'
-
-gem 'config'
-# Use Capistrano for deployment
-# gem 'capistrano-rails', group: :development
-
+# use postgres as the database
+gem 'pg'
 # useful for debugging, even in prod
 gem 'pry-byebug' # Adds step-by-step debugging and stack navigation capabilities to pry using byebug
 gem 'pry-rails' # use pry as the rails console shell instead of IRB
+# Use Puma as the app server
+gem 'puma', '~> 3.7'
+gem 'rails', '~> 5.1.3'
 
-gem 'pg'
+# Stanford/Samvera gems
+# work with Moab Objects
+gem 'moab-versioning'
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,12 @@ gem 'jbuilder', '~> 2.5'
 # gem 'bcrypt', '~> 3.1.7'
 
 gem 'config'
+# Use Capistrano for deployment
+# gem 'capistrano-rails', group: :development
+
+# useful for debugging, even in prod
+gem 'pry-byebug' # Adds step-by-step debugging and stack navigation capabilities to pry using byebug
+gem 'pry-rails' # use pry as the rails console shell instead of IRB
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
--- a/Gemfile
+++ b/Gemfile
@@ -10,8 +10,6 @@ gem 'moab-versioning'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.1.3'
-# Use sqlite3 as the database for Active Record
-gem 'sqlite3'
 # Use Puma as the app server
 gem 'puma', '~> 3.7'
 
@@ -28,11 +26,12 @@ gem 'config'
 gem 'pry-byebug' # Adds step-by-step debugging and stack navigation capabilities to pry using byebug
 gem 'pry-rails' # use pry as the rails console shell instead of IRB
 
+gem 'pg'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: [:mri]
   # Call 'binding.pry' anywhere in the code to stop execution and get a pry console
-  gem 'pry-byebug', require: false
   gem 'rspec-rails', '~> 3.6'
   gem 'rails-controller-testing'
   gem 'coveralls'
@@ -48,6 +47,10 @@ group :development do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
+end
+
+group :test do
+  gem 'shoulda-matchers', git: 'https://github.com/thoughtbot/shoulda-matchers.git', branch: 'rails-5'
 end
 
 group :deploy do
--- a/README.md
+++ b/README.md
@@ -45,20 +45,23 @@ If you use homebrew you can install PostgreSQL by typing  `brew install postgres
 Make sure Postgres starts every time your computer starts up.
 `brew services start postgresql`
 
-Check to see if Postgres is running.
+Check to see if Postgres is installed:
 `postgres -V`
+
+and that it's accepting connections:
+`pg_isready`
 
 #### 2. Configuring Postgres
 
-The `psql` utility, which lets you carry out admin functions.
+Use the `psql` utility, which lets you carry out admin functions:
+* Enter `psql postgres` into the command line.
+* Create the preservation core user and grant it the privileges it needs by entering the following commands in psql:
+```sql
+CREATE USER preservation_core_catalog;
+ALTER USER preservation_core_catalog CREATEDB;
+CREATE DATABASE preservation_core_catalog;
+ALTER DATABASE preservation_core_catalog OWNER TO preservation_core_catalog;
+GRANT ALL PRIVILEGES ON DATABASE preservation_core_catalog TO preservation_core_catalog;
+```
 
-Enter `psql postgres` into the command line.
-Create your username and password by entering the following into the command line.
-`CREATE USER 'your_username' WITH PASSWORD 'your_password';`
-
-Next create the name of your database.
-`CREATE DATABASE 'name of database';`
-
-Grant superuser privileges so you can create and drop database.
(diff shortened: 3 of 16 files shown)


app/views/moab_storage/show.html.erb
```diff
@@ -0,0 +1 @@
+<h1> Test </h1>
```
ndushay: This shouldn't be here -- we have no views.  If someone wants this locally, they can add it to .gitignore

spec/models/endpoint_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    expect(Endpoint.new).not_to be_valid
+  end
+
+  it 'is valid with valid attributes' do
+    expect(endpoint).to be_valid
+  end
+
+  it 'enforces unique constraint on endpoint_name' do
+    expect do
+      Endpoint.create!(endpoint_name: 'aws', endpoint_type: 'cloud')
```
tingulfsen: How about changing the endpoint_type to be something other than 'cloud' just to emphasize the 'endpoint_name' requirement?

spec/models/endpoint_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe Endpoint, type: :model do
+
+  let!(:endpoint) { Endpoint.create(endpoint_name: 'aws', endpoint_type: 'cloud') }
+
+  it 'is not valid without valid attributes' do
```
tingulfsen: Can we add another test where just 1 of the 2 requirements are fulfilled? Or am I too anal?

app/models/endpoint.rb
```diff
@@ -0,0 +1,8 @@
+##
+# Metadata about a replication endpoint, including a unique human
+# readable name, and the type of endpoint it is (e.g. :online, :archive).
+class Endpoint < ApplicationRecord
+  has_many :preservation_copies
+  validates :endpoint_name, presence: true, uniqueness: true
+  validates :endpoint_type, presence: true
```
tingulfsen: Do we expect 'endpoint_type' to be a finite set and if so does it make sense to represent it as such?

spec/models/preservation_copy_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe PreservationCopy, type: :model do
+  let!(:preservation_object2) do
+    PreservedObject.create!(druid: 'ab123cd45679', version: 1, preservation_policy: 'keepit', size: 1)
```
tingulfsen: Do we expect that 'preservation_policy' will be a small finite set, and if so does it make sense to represent it as such?

db/migrate/20170823232731_create_preserved_objects.rb
```diff
@@ -0,0 +1,12 @@
+class CreatePreservedObjects < ActiveRecord::Migration[5.1]
+  def change
+    create_table :preserved_objects do |t|
+      t.string :druid, null: false, unique: true, index: true
+      t.integer :version
```
tingulfsen: When would we expect 'version' to be null or not present? Should this also be a case of 'null: false'?

spec/models/preservation_copy_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  let!(:endpoint2) { Endpoint.create!(endpoint_name: 'oracle', endpoint_type: 'cloud') }
+  let!(:preservation_copy) do
+    PreservationCopy.create!(
+      preserved_object_id: preservation_object2.id, endpoint_id: endpoint2.id, version: 0, status: 'fixity_error'
+    )
+  end
+
+  it 'is not valid without valid attributes' do
+    expect(PreservationCopy.new).not_to be_valid
+  end
```
tingulfsen: Same as per my comment above - is there a sane and non-anal way to test that the correct attributes are present without generating all possible permutations?

.travis.yml
```diff
@@ -1,4 +1,12 @@
...
\ No newline at end of file
+cache: bundler
+services:
+  - postgresql
+before_script:
+  - psql -c 'CREATE USER preservation_core_catalog;' -U postgres
+  - psql -c 'ALTER USER preservation_core_catalog CREATEDB;' -U postgres
+  - psql -c 'CREATE DATABASE preservation_core_catalog;' -U postgres
+  - psql -c 'ALTER DATABASE preservation_core_catalog OWNER TO preservation_core_catalog;' -U postgres
+  - psql -c 'GRANT ALL PRIVILEGES ON DATABASE preservation_core_catalog TO preservation_core_catalog;' -U postgres
```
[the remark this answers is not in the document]
jmartin-sul: yeah, a script would be reasonable.  see #48.

Gemfile
```diff
@@ -21,12 +19,19 @@ gem 'jbuilder', '~> 2.5'
 # gem 'bcrypt', '~> 3.1.7'
 
 gem 'config'
+# Use Capistrano for deployment
+# gem 'capistrano-rails', group: :development
+
+# useful for debugging, even in prod
+gem 'pry-byebug' # Adds step-by-step debugging and stack navigation capabilities to pry using byebug
+gem 'pry-rails' # use pry as the rails console shell instead of IRB
```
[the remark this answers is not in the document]
jmartin-sul: i've definitely used rails console a number of times on prod to investigate argo issues.  not a thing that i'd expect to do often, but something i'd expect to come up at least some.  e.g., you want to run load up some objects via ActiveRecord to because some error cropped up when a thing you expected to be there wasn't.  and in that case, the syntax highlighting and formatting and inter-session history and other niceties of pry are great.

README.md
```diff
@@ -45,20 +45,23 @@ If you use homebrew you can install PostgreSQL by typing  `brew install postgres
...
+Use the `psql` utility, which lets you carry out admin functions:
+* Enter `psql postgres` into the command line.
+* Create the preservation core user and grant it the privileges it needs by entering the following commands in psql:
+```sql
+CREATE USER preservation_core_catalog;
+ALTER USER preservation_core_catalog CREATEDB;
+CREATE DATABASE preservation_core_catalog;
+ALTER DATABASE preservation_core_catalog OWNER TO preservation_core_catalog;
+GRANT ALL PRIVILEGES ON DATABASE preservation_core_catalog TO preservation_core_catalog;
+```
```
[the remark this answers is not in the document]
jmartin-sul: could probably script this, let's say tentatively see #48 for this one too.

app/views/moab_storage/show.html.erb
jmartin-sul: yeah, not sure how this got added, will remove.

spec/models/endpoint_spec.rb
jmartin-sul: i think we're already going to revisit this (per #46) and turn `endpoint_type` into an FK to a table that has richer endpoint info.  so i'm inclined to leave this as-is, just to not tweak something that'll be changing in a different way anyway.

app/models/endpoint.rb
jmartin-sul: i think we do, but per #46, there will be a table containing a list of endpoint types.  otherwise, i'd agree that we should constrain the value to known good types.

spec/models/preservation_copy_spec.rb
jmartin-sul: similar to endpoint type, there'll be a policy definition table that this will point to (see #46).  so i'm inclined to leave this as-is and revisit it when we work on #46 (which should happen either later this week, or first thing next week).

spec/models/endpoint_spec.rb
tingulfsen: 👍

db/migrate/20170823232731_create_preserved_objects.rb
jmartin-sul: i'm not sure... i had a vague thought along these lines when we were first defining the schema last week.  but what, e.g., if you had a corrupted moab where you couldn't determine the version?  in that case, i think we'd want a catalog entry, but i don't know that we could put a valid integer in this field.  though this is not a use case i'm sure about at all.

spec/models/preservation_copy_spec.rb
tingulfsen: 👍

jmartin-sul:
@ndushay:  i think i've made the requested fixes (i added a commit that just generally cleaned up Gemfile so that its layout is organized more like our other projects).  for the things that you said we could ticket, i created a ticket.

@tingulfsen:  will discuss the testing suggestions w/ @SaravShah

db/migrate/20170823232731_create_preserved_objects.rb
julianmorley: What would a corrupted Moab look like? Presumably missing version directories, or missing expected files inside one or more version directories. Would we ever even ingest such a Moab into OIS, or violently throw up before we even got to that point? If a known-good, previously-sane Moab later has problems (v7 is fine, but v8 is added and is bad), would we record that Moab as being v7, but with problems, or set version to 'null' and and then handle the exception somehow?

julianmorley: That's a round-about way of saying I think requiring a value for version is probably desirable, at the very least as a way of tracking "the last known-good version" of a Moab if we run into subsequent problems.

jmartin-sul: i guess i was imagining something like... we create a top level moab directory for an object, and something dies silently before the first version subdirectory is created.  other processes expect a valid moab object to be there, but while there was a place made for it, a valid moab object was never fully created in the first place.

if this is completely unrealistic, and no similar situation can be imagined, i'd agree that this field should be required.

julianmorley: I think in that scenario, either the inventory process or ingest should throw up long before we try to write data to the database. We should take it as a point of pride that we don't ingest bad data and alert very noisily if that happens.

There's also the side-question of "what makes a directory a Moab object?"; we shouldn't pin that to 'the directory is in Moab name format', but rather that the contents of that directory contain everything we'd expect to see in a well-formed Moab, and nothing else.

LynnMcRae: Agreed. We don't have a scenario where we're building up a Moab object from an initially empty directory. It only exists to contain at least one version of the object, so anything short of that should fail to be created.

jmartin-sul: thanks for the info, sounds good, will add the constraint

jmartin-sul:
@tingulfsen:  we added an invalid permutation to the tests for each of the things you mentioned.  we didn't do all possible permutations, partially because that part of the schema is going to change, and partially because i don't think we typically do every possible permutation for that sort of thing?  i know there are cases where we've done tons of permutations (e.g. the cancan tests that we added to stacks during the recent IIIF work), but i think we typically just do a few representative cases, right?

@LynnMcRae and @julianmorley:  added the not null constraint to the version column (and added a corresponding test similar to the ones @tingulfsen suggested above).

Gemfile
```diff
@@ -5,28 +5,28 @@ git_source(:github) do |repo_name|
...
+gem 'pg'
+# useful for debugging, even in prod
+gem 'pry-byebug' # Adds step-by-step debugging and stack navigation capabilities to pry using byebug
+gem 'pry-rails' # use pry as the rails console shell instead of IRB
+# Use Puma as the app server
+gem 'puma', '~> 3.7'
+gem 'rails', '~> 5.1.3'
 
-gem 'config'
+# Stanford/Samvera gems
```
ndushay: Samvera???   ew.

spec/models/endpoint_spec.rb
ndushay: Did this get addressed by added specs below?

PR #55, opened 2017-08-31
update catalog for a single druid

### LINKED ISSUE #45: PCC: implement update action to store info for a single druid
once we have a db attached to our rails app (See #16, #17), we want (something) to trigger storing information in the db.

what is the trigger?  Perhaps @julianmorley can illuminate, but for now, we *could* have it be one of
- [ ] a before action on the moab_storage_controller  for `show` and `index` actions  and/or
- [ ] a separate action in the controller  and/or 
- [ ] a command line trigger   and/or
- [ ] ??? your idea here ???

- [ ] Advanced (separate ticket?**) - only write to db storage if db storage for given druid is older than xxx threshold

We want full blown tests for this using the test db and the fixture data.

jmartin-sul:
* add a CatalogController, which updates the Object Inventory Store (i.e. takes client requests and manipulates the database, based on the contents of moab objects on the file system).
* routes and tests for CatalogController
* change a couple rubocop defaults to allow common departmental practices

this is a first pass at implementing #45.  at the very least, it will require updating `CatalogController#moab_size` to get the actual moab size once sul-dlss/moab-versioning#21 is implemented.  for the purposes of getting a start on this, we've hardcoded a dummy value.

this assumes for now that callers to PCC will either ask to add a record for a moab to the object inventory store, or update an existing record, but that the caller will already know whether the object moab is cataloged in the OIS, and so will not ask for something like "create an entry if one doesn't already exist".  figuring out the particulars of that usage appears to be an open question, at least for us developers right now (see https://github.com/sul-dlss/preservation_core_catalog/issues/45#issuecomment-325750585, https://github.com/sul-dlss/preservation_core_catalog/issues/45#issuecomment-325757624, and https://github.com/sul-dlss/preservation_core_catalog/issues/45#issuecomment-325758392).

does this close out #45, or is there more work to be done in figuring out caller expectations?  if there's more to be done for that ticket, should it go in this PR, or in a follow on PR?

anyway, this
connects to #45

paired w/ @SaravShah on this

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -1,0 +1,40 @@
+require 'moab/stanford'
+
+##
+# CatalogController allows clients to manipulate the Object Inventory Catalog, e.g.
+# to add an existing moab object to the catalog, or to update an entry for a moab object
+# that's already in the catalog.
+class CatalogController < ApplicationController
+  before_action :set_id
+
+  def add_preserved_object
+    PreservedObject.create!(
+      druid: @id, size: moab_size(@id), current_version: Stanford::StorageServices.current_version(@id)
+    )
+  end
+
+  def update_preserved_object
+    preserved_obj = PreservedObject.find_by(druid: @id)
+    preserved_obj.update_attributes(
+      druid: @id, size: moab_size(@id), current_version: Stanford::StorageServices.current_version(@id)
+    )
+  end
+
+  private
+
+  def set_id
+    @id = catalog_params[:id]
+    head(:unprocessable_entity) if @id.blank?
+  end
+
+  def catalog_params
+    params.permit(:id)
+  end
+
+  def moab_size(id)
+    # TODO: make this actually get the size once there's a method for that.
+    # for now, just using the id to quiet rubocop complaint.
+    # https://github.com/sul-dlss/moab-versioning/issues/21
+    42 || id
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -53,10 +53,15 @@ Style/StringLiterals:
 Style/SymbolArray:
   Enabled: false
 
+RSpec/ExampleLength:
+  Max: 50
+
 RSpec/MultipleExpectations:
   Exclude:
     - 'spec/controllers/moab_storage_controller_spec.rb'
 
+RSpec/NamedSubject:
+  Enabled: false
 
 Bundler/OrderedGems:
   Exclude:
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -3,5 +3,8 @@ Rails.application.routes.draw do
 
   get 'moab_storage/show/:id', to: 'moab_storage#show'
 
+  post 'catalog/add_preserved_object', to: 'catalog#add_preserved_object'
+  post 'catalog/update_preserved_object', to: 'catalog#update_preserved_object'
+
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
 end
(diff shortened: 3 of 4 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    preserved_obj.update_attributes(
+      druid: @id, size: moab_size(@id), current_version: Stanford::StorageServices.current_version(@id)
+    )
+  end
+
+  private
+
+  def set_id
+    @id = catalog_params[:id]
+    head(:unprocessable_entity) if @id.blank?
```
tingulfsen: So this would respond with a 422 Unprocessable Entity. Why did you decide to use this instead of a regular 400?

https://tools.ietf.org/html/rfc7231#section-6.5.1

app/controllers/catalog_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def catalog_params
+    params.permit(:id)
```
tingulfsen: Is it worth using `require()` here?

jmartin-sul: if you do a `require`, you're required to have the specific params you permit in a hash named by the `require`.  e.g., if you do `params.require(:catalog_params).permit(:id)`, the callers are required to pass in a hash of params called `catalog_params`, and that hash can contain a field called `id`.  this seems more useful when you have a form to submit, and where the rails form builder stuff is handling that construction for you (using, IIRC, named forms, which will build that hash into the form structure).  since (IME) all our usage of strong params is with form submission, and since this is what the majority of examples i've seen online are dealing with, we're used to always seeing the `require` part.

but that seems overly circuitous for a REST call that's just going to be providing some parameters without being in the context of a form.  we just want to make sure that we only allow specific param names, but we don't really need to bother having them in a hash.  the rails docs do actually have instructions for this, they're just not the first thing described:
http://edgeguides.rubyonrails.org/action_controller_overview.html#permitted-scalar-values (section 4.5.1, my browser landed just past the anchor for that section when i reloaded that link, so you may want to scroll up if it doesn't land properly on that section).

we did originally implement it using `require`, but then the request in the unit tests looked like:
```
post :add_preserved_object, params: { catalog_params: { id: id } }
```

instead of:
```ruby
post :add_preserved_object, params: { id: id }
```

and we thought the latter would be preferable for real callers.

app/controllers/catalog_controller.rb
jmartin-sul: this was something that we weren't totally sure about, but it seemed like the most recommended error code from the few stack exchange posts and such that we found, e.g. https://stackoverflow.com/a/10323055:
> The 422 (Unprocessable Entity) status code means the server understands the content type of the request entity (hence a 415 (Unsupported Media Type) status code is inappropriate), and the syntax of the request entity is correct (thus a 400 (Bad Request) status code is inappropriate) but was unable to process the contained instructions. For example, this error condition may occur if an XML request body contains well-formed (i.e., syntactically correct), but semantically erroneous, XML instructions.

from https://tools.ietf.org/html/rfc4918#page-78

though the RFC you link is newer (2014 vs 2007).  and, now that i scroll up, i notice that maybe the spec we made that decision on is WebDAV specific.

400 felt a bit generic to me, and not quite right based on the short description in the RFC, but it did feel like the next most correct thing besides 422, so if you still feel that 422 is wrong, i think we're happy to switch over to 400.

tingulfsen: OK, sounds like this is really a toss-up.

PR #57, opened 2017-08-31
Moab object size method in show action

SaravShah:
Show action now shows object size of the moab directory. 
Note: Didn't add tests for private methods, since they're pretty simple and are used in public methods.

Note2: Another way to test object_size instead of using `be_between` maybe use the `\du`unix command and use the directory size number from there to compare.

paired w/ @jmartin-sul

### DIFF
--- a/app/controllers/moab_storage_controller.rb
+++ b/app/controllers/moab_storage_controller.rb
@@ -1,4 +1,6 @@
 require 'moab/stanford'
+require 'action_view'
+include ActionView::Helpers::NumberHelper
 
 # API to retrieve data from Moab Object Store
 class MoabStorageController < ApplicationController
@@ -11,7 +13,14 @@ class MoabStorageController < ApplicationController
   end
 
   def show
-    @output = { current_version: Stanford::StorageServices.current_version(params['id']) }
+    moab_path = path_from_object_id(params['id'])
+    object_size = directory_size(moab_path)
+    object_size_human = number_to_human_size(object_size)
+    @output = {
+      current_version: Stanford::StorageServices.current_version(params['id']),
+      object_size: object_size,
+      object_size_human: object_size_human
+    }
     respond_to do |format|
       format.xml { render xml: @output }
       format.all { render json: @output, content_type: 'application/json' }
@@ -27,4 +36,20 @@ class MoabStorageController < ApplicationController
     end
   end
 
+  def path_from_object_id(id)
+    storage_repo = Stanford::StorageRepository.new
+    storage_repo.find_storage_root(id).join(storage_repo.storage_trunk, storage_repo.storage_branch(id)).to_s
+  end
+
+  def directory_size(dirname)
+    size = 0
+    Find.find(dirname) do |path|
+      if FileTest.directory?(path)
+        Find.prune if File.basename(path)[0] == '.'
+      else
+        size += FileTest.size(path)
+      end
+    end
+    size
+  end
 end
--- a/app/models/preserved_object.rb
+++ b/app/models/preserved_object.rb
@@ -4,6 +4,9 @@
 # represent a specific stored instance on a specific node, but aggregates
 # those instances.
 class PreservedObject < ApplicationRecord
+  # NOTE: The size field stored in PreservedObject is approximate,as it is determined from size
+  # on disk (which can vary from machine to machine). This field value should not be used for
+  # fixity checking!
   has_many :preservation_copies
   validates :druid, presence: true, uniqueness: true
   validates :current_version, presence: true
--- a/spec/controllers/moab_storage_controller_spec.rb
+++ b/spec/controllers/moab_storage_controller_spec.rb
@@ -47,6 +47,7 @@ RSpec.describe MoabStorageController, type: :controller do
       get :show, params: { id: fixture_druid }
       expect(response).to have_http_status(:success)
     end
+
     context 'assigns @output correctly' do
       it 'Hash' do
         get :show, params: { id: fixture_druid }
@@ -56,6 +57,17 @@ RSpec.describe MoabStorageController, type: :controller do
         get :show, params: { id: fixture_druid }
         expect(assigns(:output)[:current_version]).to eq 10
       end
+
+      it 'object_size' do
+        get :show, params: { id: fixture_druid }
+        expect(assigns(:output)[:object_size]).to be_between(21_900_000, 22_000_000)
+      end
+
+      it 'object_size_human' do
+        get :show, params: { id: fixture_druid }
+        expect(assigns(:output)[:object_size_human]).to a_string_ending_with("MB")
+      end
+
     end
     it 'returns json by default' do
       get :show, params: { id: fixture_druid }


app/controllers/moab_storage_controller.rb
```diff
@@ -11,7 +11,9 @@ def index
   end
 
   def show
-    @output = { current_version: Stanford::StorageServices.current_version(params['id']) }
+    moab_path = path_from_object_id(params['id'])
+    object_size = directory_size(moab_path)
+    @output = { current_version: Stanford::StorageServices.current_version(params['id']), object_size: object_size }
```
ndushay: consider putting hash in this format:

```ruby
@output = {
  current_version: Stanford::StorageServices.current_version(params['id']), 
  object_size: object_size
}

File: app/models/preserved_object.rb [hunk 2]
```diff
@@ -4,6 +4,8 @@
 # represent a specific stored instance on a specific node, but aggregates
 # those instances.
 class PreservedObject < ApplicationRecord
+  # NOTE: The size of the object is the size on disk, so it may vary from machine to machine.
+  # It is not used for fixity checking.
```
ndushay: I'm glad to see this comment, but there's is nothing here about size, so it seems out of place.  I'm wondering how to improve this situation.  Maybe something like:

"NOTE: the size field stored in PreservedObject is approximate, as it is determined from size on disk (which can vary from machine to machine).  This field value should not be used for fixity checking!"

spec/controllers/moab_storage_controller_spec.rb
```diff
@@ -56,6 +56,12 @@
         get :show, params: { id: fixture_druid }
         expect(assigns(:output)[:current_version]).to eq 10
       end
+
+      it 'object_size' do
+        get :show, params: { id: fixture_druid }
+        expect(assigns(:output)[:object_size]).to be_between(21_909_525, 22_049_525)
```
ndushay: consider making these round numbers:  21_909_000, 22_049_000   ???

app/controllers/moab_storage_controller.rb
ndushay: also:  consider putting object size in K, M, G, etc.

spec/controllers/moab_storage_controller_spec.rb
SaravShah: Yeah sounds good, I just did a ±70,000 bytes from the directory file size given by the method. But i'll change that.

app/controllers/moab_storage_controller.rb
```diff
@@ -11,7 +11,13 @@ def index
   end
 
   def show
-    @output = { current_version: Stanford::StorageServices.current_version(params['id']) }
+    moab_path = path_from_object_id(params['id'])
+    object_size = directory_size(moab_path)
+    size_human = (object_size / 1_000_000.0)
```
jmartin-sul: i would be in favor of having `object_size` be in bytes (maybe called `object_size_in_bytes`, but i think it'd be good to assume that a field called `object_size` is in bytes unless otherwise specified).

and then, we could do the human conversion in a field called `object_size_human`.  or, in this case, i might call it `object_size_in_mb`.  there is also a rails method that puts out human readable size, given a number:  https://apidock.com/rails/ActionView/Helpers/NumberHelper/number_to_human_size

which is not as consistent or easily machine-readable as always having it in MB, but i think machine-readability is what size in bytes is for anyway.

personally, i'd be slightly inclined not to include a human readable size here at all, and leave that to a display layer that we know is for humans.  although maybe that's an argument for leaving it here, since developers will probably be reading lots of REST calls while developing.

SaravShah: I like the idea of having two variables object_size_in_mb and object_size_in_bytes. For now, I think i'll have the output be size in mb, but we can change that later if it is not needed. Using the ActionView Helper `number_to_human_size` converts bytes to mb however it adds "MB" after the number.

e.g 

`object_size_in_bytes = 6601408
object_size_in_mb = number_to_human_size(object_size_in_bytes)
p object_size_in_mb

>> "6.3 MB"`

jmartin-sul: including both the `_bytes` version and the human readable version of the field sounds good to me.   

re: `number_to_human_size`:  i think adding "MB" is fine in terms of functionality (and it should presumably switch to "GB" or "KB" or whatever as appropriate), since that's just supposed to be a human readable string.  but i don't think i have an informed opinion on the performance hit of `require 'action_view'` / `include ActionView::Helpers::NumberHelper`.  my gut feeling is that it won't be substantial at the volume of requests that we're dealing with.  but i leave it up to you.  if you did go with the helper method, you'd probably want to name the field something like `object_size_human` or something, since it won't necessarily be in MB.

slight preference for the functionality of using `number_to_human_size`, but i'd also have no objection to just dividing by 1,000,000 and having the field always be `_mb` and calling it a day.

ndushay: technically, since a kilobyte is 1024, the conversion is correct.  I think it's appropriate to have the version be "correct" rather than divide by 000s.  If it weren't bytes we were talking about, I would go with our manual division.    I think using the action_view helper will be easier in the long run as I'm guessing it will cope well with KB, MB, GB, TB, etc. and I suspect our objects may vary quite a bit.

I would go with labels `object_size` and `object_size_human` but don't feel strongly about the label of the latter if someone else has a better suggestion.

app/controllers/moab_storage_controller.rb
```diff
@@ -11,7 +11,13 @@ def index
...
+    @output = {
+      current_version: Stanford::StorageServices.current_version(params['id']),
+      object_size: size_human
```
ndushay: and yes, definitely output both ...

spec/controllers/moab_storage_controller_spec.rb
```diff
@@ -56,6 +56,12 @@
         get :show, params: { id: fixture_druid }
         expect(assigns(:output)[:current_version]).to eq 10
       end
+
+      it 'object_size' do
+        get :show, params: { id: fixture_druid }
+        expect(assigns(:output)[:object_size]).to be_between(21.5, 22.5)
+      end
```
ndushay: and of course you'd want a test for both fields ...

spec/controllers/moab_storage_controller_spec.rb
```diff
@@ -56,6 +57,22 @@
...
+
+      it 'object_size_human' do
+        get :show, params: { id: fixture_druid }
+        expect(assigns(:output)[:object_size_human]).to include("MB")
+      end
+
+      it 'object_size_human instance' do
+        get :show, params: { id: fixture_druid }
+        expect(assigns(:output)[:object_size_human]).to be_kind_of String
```
ndushay: one last quibble -- I would just put this expects statement in the previous test.  And, to get picky, I think it could actually be

`expect(assigns(:output)[:object_size_human]).to be a_string_ending_with("MB")`

check out https://relishapp.com/rspec/rspec-expectations/v/3-6/docs/composing-matchers

jmartin-sul: good suggestion.  i'd be inclined to just replace this and the `to include("MB")` test with the expectation naomi suggested.  that's a pretty neat composition -- i've done similar with expecting something `to be a_hash_including`, but i don't think i would've thought to do the `be a_string_ending_with` thing.

Issue #61, opened 2017-09-05
Update object size method in preservation core catalog after #63 (moab gem release)


SaravShah:
Once the Moab Versioning Gem is updated with the object size method, update PCC object size method.

PR #73, opened 2017-09-08
Change object_size method

SaravShah:
- Used updated moab-versioning gem for object_size method

connects to #61

### DIFF
--- a/app/controllers/moab_storage_controller.rb
+++ b/app/controllers/moab_storage_controller.rb
@@ -13,8 +13,7 @@ class MoabStorageController < ApplicationController
   end
 
   def show
-    moab_path = path_from_object_id(params['id'])
-    object_size = directory_size(moab_path)
+    object_size = Stanford::StorageServices.object_size(params['id'])
     object_size_human = number_to_human_size(object_size)
     @output = {
       current_version: Stanford::StorageServices.current_version(params['id']),
@@ -36,20 +35,4 @@ class MoabStorageController < ApplicationController
     end
   end
 
-  def path_from_object_id(id)
-    storage_repo = Stanford::StorageRepository.new
-    storage_repo.find_storage_root(id).join(storage_repo.storage_trunk, storage_repo.storage_branch(id)).to_s
-  end
-
-  def directory_size(dirname)
-    size = 0
-    Find.find(dirname) do |path|
-      if FileTest.directory?(path)
-        Find.prune if File.basename(path)[0] == '.'
-      else
-        size += FileTest.size(path)
-      end
-    end
-    size
-  end
 end


ndushay:
I just updated the tolerances on coveralls configuration to be 0.5%

Gemfile.lock
```diff
@@ -55,7 +55,7 @@ GEM
       bundler (~> 1.2)
       thor (~> 0.18)
     byebug (9.1.0)
-    capistrano (3.9.0)
+    capistrano (3.9.1)
```
ndushay: It's a good idea to split out gem updates into a commit separate from coding updates

Also, I would prefer if you update gems in separate commits of logical groupings, e.g. "update rails to 5.1.4" "update capistrano to 3.9.1" "update rubocop-rspec to 0.16.1" etc.

This is for two reasons:  

1. to easily see what has been updated in the commit history (when you are hunting for "what changed?  why isn't it working now?")  and  
2.  to make it easier to back out of some updates and refactor the PR if there are objections to some of the gem updates

app/controllers/moab_storage_controller.rb
```diff
@@ -13,8 +13,7 @@ def index
   end
 
   def show
-    moab_path = path_from_object_id(params['id'])
-    object_size = directory_size(moab_path)
+    object_size = Stanford::StorageServices.object_size(params['id'])
```
ndushay: yay!  smaller code here!

spec/models/preserved_object_spec.rb
```diff
@@ -11,7 +11,7 @@
   end
 
   it 'is not valid without a specified current_version' do
-    expect(PreservedObject.new(druid: 'ab123cd45678', preservation_policy: 'keepit', size: 1))
+    expect(PreservedObject.new(druid: 'ab123cd45678', preservation_policy: 'keepit', size: 1)).not_to be_valid
```
ndushay: I just noticed this myself while working on my ticket.  This is another thing that should be a separate commit.  As I am working on a PR right now refactoring this spec a bit, I suggest you leave this off altogether and see if you're comfy with how I do it in my PR.

ndushay: if my home connection will ever upload my changes ...  (See #74)

Gemfile.lock
SaravShah: So can I split this commit into 2 commits? 
1. Update Gem versions - i'm updating ~30 gems, and it seems like that would be a lot of commits if I were to create a commit for each gem. I may not be understanding you though. 

I could also just update Moab Versioning Gem and have that as 1 commit and not update anything else?

2. Update object size method

spec/models/preserved_object_spec.rb
SaravShah: Sounds good.

Gemfile.lock
ndushay: you can `bundle update rails`  which will take care of most of the gems as one commit.   Another commit after `bundle update moab-versioning` etc.

Or you can leave the non-moab-versioning updates off;  I can prepare a PR for those, if you like.

SaravShah: sure. that works for me. and then this PR will just have 1 commit having to do with updating the object size. Sounds good.

jmartin-sul:
it appears that @ndushay's concern has been addressed (by breaking out gem updates into #79), so i'm going to go ahead and merge this.

PR #76, opened 2017-09-09
preserved_object_spec: refactor to prepare for more tests

ndushay:
Given that more tests will find their way into this spec file, I thought it made more sense to refactor the tests under a "validation" context.  I also tried to distinguish between required and optional attributes, and fixed a spec that was only partially written.


SaravShah:
this looks good to me 👍

ndushay:
merge it - I would say this is not heady stuff needing lots of eyes.

Issue #74, opened 2017-09-08
Remove large fixture objects from git repo history


ndushay:
The upload speed from home is killing me.  

Now that Lynn found us some wonderful small objects, I want us to purge the beasts from our git repo and its history.

This explains how -- we're not removing sensitive data but we will be removing data from non-recent commits, so it applies
https://help.github.com/articles/removing-sensitive-data-from-a-repository/

ndushay:
https://rtyley.github.io/bfg-repo-cleaner/

atz:
Doesn't seem like this is a problem currently (e.g. CI is still reasonably fast).  Maybe the purge was already accomplished?

PR #77, opened 2017-09-09
Remove big fixture objects

ndushay:
This removes the 2 largest fixture objects from the directories.  It does NOT remove the large files from git history;  that will be covered by #74 (and this is a pre-req for that).


SaravShah:
again look good!

PR #78, opened 2017-09-11
schema updates

### LINKED ISSUE #46: PCC: Reconcile schema with new info
See:
sdr-pc ois v3
@jmartin-sul, @SaravShah and I hashed this out.

jmartin-sul:
this is an attempt to implement the schema improvements outlined in #46.  paired w/ @SaravShah on the implementation.

i'd initially figured i'd squash everything into one commit, but i was actually able to separate things out into somewhat logical groupings of changes that built on one another.  unfortunately, we were a bit messy about committing our WIP, and we did all the test updates after the implementation.  so while i'd ideally like each commit to be a working build, that's unlikely to be the case, since some of the test fixes were clumped together.  the different chunks of the schema migration and their accompanying model changes are pretty well grouped, though, hopefully.  happy to do more commit squashing or splitting if people have suggestions, and will try to be better about keeping tests in sync as work goes on future PRs like this.

re: the rubocop exception, i'm guessing rubocop's thinking is something along the lines of:  http://blog.flatironschool.com/why-you-dont-need-has-and-belongs-to-many/

but @SaravShah and i both felt like that was overkill in this case.  the join table is there, and if we need to add more columns to it, a migration and a model of its own (and the implied AR association and test updates) shouldn't be hard to add.  and the rails guide seems to suggest `has_and_belongs_to_many` for this use case:
http://guides.rubyonrails.org/association_basics.html#the-has-and-belongs-to-many-association
http://guides.rubyonrails.org/association_basics.html#the-has-many-through-association

@julianmorley, i realized friday afternoon when i was looking over these changes to see if we missed anything, that we had indeed missed getting rid of creation dates.  i swear we didn't do that on purpose, though i admit i was reticent to drop the column when we met about the schema changes week before last 😈 .  umm... would you mind if we file a ticket to get rid of it if we find that the column really does lead to space usage and/or performance issues?  it looks pretty easy to get rid of by using [`remove_timestamps`](http://edgeguides.rubyonrails.org/active_record_migrations.html#using-the-change-method).  but one nice thing is that active record magic manages creation and update timestamps on rows if the columns are there, and i'm not sure if having only one of the columns there would then also necessitate manually managing those timestamps.

i tried to follow the direction of @ndushay's refactoring when resolving the conflict that came up in `preserved_object_spec.rb` when rebasing, but didn't bother going and doing similar refactoring on the other specs (extant or added).  i'd propose saving that for a follow-on ticket/PR.

closes #46

### DIFF
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -5,4 +5,7 @@ class Endpoint < ApplicationRecord
   has_many :preservation_copies
   validates :endpoint_name, presence: true, uniqueness: true
   validates :endpoint_type, presence: true
+  validates :endpoint_node, presence: true
+  validates :storage_location, presence: true
+  validates :recovery_cost, presence: true
 end
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -3,6 +3,8 @@
 # readable name, and the type of endpoint it is (e.g. :online, :archive).
 class Endpoint < ApplicationRecord
   has_many :preservation_copies
+  belongs_to :endpoint_type
+
   validates :endpoint_name, presence: true, uniqueness: true
   validates :endpoint_type, presence: true
   validates :endpoint_node, presence: true
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -4,6 +4,7 @@
 class Endpoint < ApplicationRecord
   has_many :preservation_copies
   belongs_to :endpoint_type
+  has_and_belongs_to_many :preservation_policies
 
   validates :endpoint_name, presence: true, uniqueness: true
   validates :endpoint_type, presence: true
--- a/db/migrate/20170907215045_add_details_to_endpoints.rb
+++ b/db/migrate/20170907215045_add_details_to_endpoints.rb
@@ -1,0 +1,8 @@
+class AddDetailsToEndpoints < ActiveRecord::Migration[5.1]
+  def change
+    add_column :endpoints, :endpoint_node, :string, null: false
+    add_column :endpoints, :storage_location, :string, null: false
+    add_column :endpoints, :recovery_cost, :int, null: false
+    add_column :endpoints, :access_key, :string
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,6 +22,9 @@ AllCops:
 Rails:
   Enabled: true
 
+Rails/HasAndBelongsToMany:
+  Enabled: false
+
 Metrics/BlockLength:
   Exclude:
     - 'spec/**/*'
--- a/app/models/endpoint_type.rb
+++ b/app/models/endpoint_type.rb
@@ -1,0 +1,8 @@
+##
+# metadata about a specific replication endpoint
+class EndpointType < ApplicationRecord
+  has_many :endpoints
+
+  validates :type_name, presence: true, uniqueness: true
+  validates :endpoint_class, presence: true
+end
--- a/app/models/preservation_copy.rb
+++ b/app/models/preservation_copy.rb
@@ -3,7 +3,10 @@
 class PreservationCopy < ApplicationRecord
   belongs_to :preserved_object
   belongs_to :endpoint
+  belongs_to :status
+
   validates :preserved_object, presence: true
   validates :endpoint, presence: true
   validates :current_version, presence: true
+  validates :status, presence: true
 end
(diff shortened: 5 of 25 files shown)


jmartin-sul:
just rebased on current master

app/models/endpoint.rb
```diff
@@ -3,6 +3,12 @@
...
 class Endpoint < ApplicationRecord
   has_many :preservation_copies
+  belongs_to :endpoint_type
+  has_and_belongs_to_many :preservation_policies
+
   validates :endpoint_name, presence: true, uniqueness: true
   validates :endpoint_type, presence: true
+  validates :endpoint_node, presence: true
+  validates :endpoint_storage, presence: true
+  validates :recovery_cost, presence: true
```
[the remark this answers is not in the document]
jmartin-sul: IIRC, you can create an ActiveRecord object using `#new`, and validations won't run automatically.  if you use `#create`, i believe they will (and i'd be inclined to use `#create` at first, and then switch to new if perf needs indicate that we should).  i believe the object will always be validated at save time?  i'm a bit of an ActiveRecord noob, so i'm not totally sure about these things.  but i think we can get the sort of validation control you're asking about, at object creation time, without tweaking these declarations at all.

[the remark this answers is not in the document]
jmartin-sul: also, through the magic of RDBMS constraints, postgres will force these things to be true, and will prevent saving records that violate these constraints.  if our validation becomes more bespoke, that might no longer be true, but these are still simple `NOT NULL` and foreign key type things.  still, it's a bit rails-ier, and probably a bit more developer and user friendly to let ActiveRecord do its thing, if performance constraints allow it.

db/migrate/20170907215045_add_details_to_endpoints.rb
```diff
@@ -0,0 +1,8 @@
+class AddDetailsToEndpoints < ActiveRecord::Migration[5.1]
+  def change
+    add_column :endpoints, :endpoint_node, :string, null: false
+    add_column :endpoints, :endpoint_storage, :string, null: false
```
tingulfsen: This seems like a bad name - what IS endpoint_storage?

app/models/endpoint.rb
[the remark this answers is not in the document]
tingulfsen: Let's do a round of performance testing with this new schema, it is quite different to what we had before, and our first round didn't exercise all the tables anyway.

db/migrate/20170907215045_add_details_to_endpoints.rb
jmartin-sul: good question, i had to look that up again just now, which would sort of indicate an unclear name.  but it is this nebulous thing that could be a file path, or an S3 bucket, or a ??? for some different storage technology.  see @julianmorley's diagram here:  https://user-images.githubusercontent.com/8161286/29797985-0ab543f0-8c0f-11e7-836f-487c3d347a40.png

i'm not sure what would be a better name, though i'm open to suggestions.  one thought i had was that we could collapse `endpoint_node` and `endpoint_storage` into a single `endpoint_uri` column.  and so, julian's example might turn into something like `sul-sdr-services:/services-disk03/sdr2objects/` and `s3.us-east-2.amazonaws.com/sdr-bucket-01` if you were to convert the first couple rows of his example table to that approach.  the two things that make me hesitant to do that are:  1) is there a standard way to construct an S3 URI like that?  i imagine there is, but i don't know if offhand, since i'm unfamiliar w/ the workings of S3 and AWS.  2) it'd be harder to query that setup in a performant way (e.g., give me all the MOABs that are archived to sul-dlss.oraclecloud.com, and give me all the MOABs that have online copies on sul-sdr-services, would both require the query to do some text wrangling if the two cols were combined into one, and text wrangling is slow, though maybe that's the sort of info that gets shoved into argo's solr index anyway, and then you can do faceting or substring search or whatever in a more performant way without going directly to postgres).

jmartin-sul:
rebased on current master, pushed rebased branch

jmartin-sul:
sidenote: i hate that github sorts commits differently than the standard command line git log.  e.g., after squashing commits, the `updated schema` commit is actually supposed to fall last, and does locally for me, but github sorts it second to last.  at least github and travis know which commit is the head of the branch, so that the tests get run on the right code (and the green check mark does appear next to what should be the last commit).

db/migrate/20170907215045_add_details_to_endpoints.rb
jmartin-sul: after discussion w/ @tingulfsen, we decided that `endpoint_storage_location` was better.  i might even be inclined to shorten that to `storage_location`?

ndushay: I like `storage_location` and can live with `endpoint_storage_location`

jmartin-sul:
went with `storage_location`.  squashed in the fixes and re-pushed.

PR #79, opened 2017-09-11
Update gems

SaravShah:
Updated rails and moab-versioning gem.


greysteil:
@SaravShah @ndushay would you be up for automatically creating dependency update PRs? I built [Dependabot](https://dependabot.com) to do that at my last company and we made it free for open-source - could be a time saver?

ndushay:
@greysteil we often need to control our dependency updates but thanks for the info.

Issue #66, opened 2017-09-08
logging class: lightweight wrapper


ndushay:
purpose:  
- regularize the format of our messages for machine processing
- make it easy to hook up honebadger
- make it easy to use diff logging system (kafka - haha!)

takes in log level, druid, message
- [ ] logs at correct level
- [ ] outputs information in easily parsed, easily read format
- [ ] timestamps should be easily understood

consider `log_rage` for message formatting -- may not be what we want

Note:  #65 should ultimately use this for logging;  if that isn't true, ensure it become true (separate ticket, separate PR?)

ndushay:
See https://logmatic.io/blog/rails-logger-best-practices/  ?

ndushay:
See also:  http://guides.rubyonrails.org/configuring.html   there is a section on setting up logging in a rails app.

ndushay:
look at `PreservedObject.update` method -- it will need to be tweaked to work with new approach, I imagine.

ndushay:
See https://docs.google.com/document/d/1Gxi73V2TfZviOIc6yr4cEpcUWN55Jowcgu93VxoxrIY

ndushay:
met with Christina and Erin;  next will meet with TonyZ Tues 10/3 to learn about rails support for instrumentation in logs.

ndushay:
@tingulfsen can this ticket be closed now?

ndushay:
@tingulfsen can this ticket about logging (and the accompanying PR) be closed now?

Issue #65, opened 2017-09-08
preserved_object_model: update_entry method


SaravShah:
method to update db table entry:

if entry already exists:
- [x] updates entry with new info (including timestamp of check)
   - new info:
     - [x] timestamp of this check (and what is this check)  <-- this is truly the new data
     - [ ] size (check if diff? - separate ticket?)
     - [x] current_version (check if diff? - separate ticket?)
     - [ ] source of check (if  inventory|ingest is recorded in db )
- [x] log level "info" - stating check happened
  - rails log for now
  - possible future other logging instead (future ticket)
- [x] returns "ok"

if entry doesn't exist yet:
- [x] log level "warn": expected object xx but it wasn't there
  - rails log for now
  - possible future other logging instead (future ticket)
- [x] add entry  (assume no extra work needed - active record foo)
- [x] return "not ok"

Note:  see #66 for logging.  If that ticket isn't done yet, add a task to update the logging done here.

ndushay:
I started trying to write up the missing checklist tasks in ticket #81 but I think it needs work.

PR #80, opened 2017-09-12
[WIP] PreservedObject update method first pass

### LINKED ISSUE #81: PreservedObject.update:  write size for create and version updates only
Size will be used to consider things like "which endpoints get this object" or "which storage root should this be moved to" and therefore is not needed to be 100% accurate, and is also NOT NEEDED TO MATCH FOR EXISTENCE integrity checks.

If we're updating the db for current_version change or because record is new, we update the size; o.w. we ignore it.

note that callers will send size (always? whenever possible?)

old below this line
--------

#### if entry doesn't yet exist
  - include size in object (along with other info) before saving

#### if entry already exists in db
-  if incoming and db current_version matches
    - if no size, or if size matches 
        - update timestamp in db
    - if size doesn't match  
        - log a warning?  an error?
            - do what?

- if incoming version is more recent/greater than db version
    -  if have size
          - update size, current version and timestamp in db ?
    - if don't have size
           - update current version and timestamp in db ?

- if incoming version is older/less than db version
    - log a warning?  an error?
    -  if have size
          - do what?
     - if don't have size
          - do what?

### LINKED ISSUE #90: existence check (PreservedObject.update) - version diffs -  handle all cases
The below is vetted as the desired spec

- when incoming version and db version are the same
  - update timestamp to show it's been checked

- when incoming version is newer than db version:
   - log WARNING 
   - update db version and timestamp

- when incoming version is OLDER than db version
  - log ERROR
       - in new logging approach and/or  
       - in honeybadger and/or 
       - sending email and/or 
       - workflow status update
  - this is a serious problem;  manual intervention

ndushay:
Work now occurring on pres-obj-handler branch;  this PR will be closed without merging once a new PR is ready.

[WIP] - expect this to be superseded by a new PR coming soon.

Connected to #65 
Closes #90 (what to do when storage root moab is older than db)
Closes #81  (when to update size)

Implemented the most basic interpretation of method to update db table entry:

if entry already exists:
- [x] `updated_at` field is updated
- [x] deals with the three different cases of current_version (between incoming and db - see code) 
- [x] log level "info" - stating check happened;  rails log for now (see #66 for logging)
- [x] returns `true`

if entry doesn't exist yet:
- [x] log level "warn": expected object xx but it wasn't there;  rails log for now (see #66 for logging)
- [x] add entry  (as entry in single table)
- [x] return "not ok"


ndushay:
I started trying to write up the missing checklist tasks in ticket #81 but I think it needs work.

jmartin-sul:
may be superseded by a PR from @ndushay that combines PR #80, PR #111, and work that @ndushay and @tallenaz had in progress yesterday refining PR #80

ndushay:
Work now occurring on pres-obj-handler branch;  this PR will be closed without merging once a new PR is ready.

ndushay:
Superseded by #130

PR #83, opened 2017-09-12
[WIP] Initialize Postgres by Rake

### LINKED ISSUE #48: create a rake task or script for postgres db setup
create a script that performs the PG DB setup steps that are performed in .travis.yml.  bonus: have it take some flag as an arg to indicate whether it should stick `-U postgres` onto the commands for travis (presumably developers installing on laptop will be installing postgres under their regular user account, in which case no user needs to be specified for the `psql` command).

tingulfsen:
Fixes #48

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,4 @@ cache: bundler
 services:
   - postgresql
 before_script:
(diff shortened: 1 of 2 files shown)


tingulfsen:
Closing in favor of #85

Issue #58, opened 2017-09-01
catalog_controller: which HTTP request for update action


SaravShah:
Change POST to PUT/PATCH  since it is doing the update action (forget which one)

ndushay:
- [ ] investigate difference between PUT and PATCH
- [ ] confirm which is the current rails practice
- [ ] surface findings to devs

SaravShah:
Notes on PUT vs PATCH
	- PUT means resource creation or replacement at some given URL
	- PUT does not allow partial update to a resource
	- PATCH allows full and partial updates and side-effects on other resources
	- PUT means create/replace and it is idempotent. PATCH means update (full or partial) and it is not idempotent.
		○ Idempotent -  clients can make that same call repeatedly while producing the same result. making multiple identical requests has the same effect as making a single request

	- The existing HTTP PUT method only allows a complete replacement of a document.
   This proposal adds a new HTTP method, PATCH, to modify an existing
   HTTP resource. (https://tools.ietf.org/html/rfc5789)

        - So with PUT you are replacing or creating resource. PATCH is for update (https://github.com/rails/rails/pull/505)

- After reading through these links and some other research... I believe the HTTP method we should be using is PATCH.

ndushay:
PUT:
- pro
  - idempotency
- con
  - sending extra content via HTTP (could be pretty big extra content)

PATCH:
- pro
  - smaller data sent
  - more "rails standard"
- con
  - what if we need to re-run? (risk ??)
  - what if multiple updates at same time (risk:  very low)

tingulfsen:
If we're doing partial replacements, then it seems to me that PATCH is the way to go.

ndushay:
Note that the test is still doing PUT, not PATCH - please fix

jmartin-sul:
good catch, @ndushay (sorry for missing that!)

PR #84, opened 2017-09-12
Update to PATCH method

SaravShah:
Changed POST to PATCH

connects to #58


jmartin-sul:
i'm going to assume that coveralls isn't going to get updated, since travis went green about 30 min ago, but coveralls is still yellow.  travis reports [98.21% coverage, FWIW](https://travis-ci.org/sul-dlss/preservation_core_catalog/builds/274765770#L634).

PR #85, opened 2017-09-12
Set up Postgres by tasks

tingulfsen:


jmartin-sul:
i just pushed a commit with the suggestion i made, which i think technically means i'm not supposed to merge this.  and tommy and i did talk a bit about the implementation earlier, so i guess we kinda sorta paired asynchronously on it.  could we get @ndushay or @SaravShah to review/merge?

PR #89, opened 2017-09-14
[WIP] Add a simple logger

### LINKED ISSUE #66: logging class: lightweight wrapper

tingulfsen:
Fixes #66 

This provides a single function you may call in order to write a message and an optional druid to the Rails log to produce consistent output.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -17,6 +17,8 @@ module PreservationCoreCatalog
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+
+    config.eager_load_paths += %W[#{config.root}/lib]
   end
 end
 
--- a/lib/pcc_logger.rb
+++ b/lib/pcc_logger.rb
@@ -1,0 +1,7 @@
+# A basic logger so we can write log messages in a regular format.
+class PCCLogger
+  def self.log(severity, message, druid = '')
+    log_message = Time.now.iso8601 + " " + druid + " " + message
+    Rails.logger.add(severity, log_message)
+  end
+end
--- a/spec/lib/pcc_logger_spec.rb
+++ b/spec/lib/pcc_logger_spec.rb
@@ -1,0 +1,26 @@
+require 'rails_helper'
+require 'spec_helper'
+
+describe PCCLogger do
+  before do
+    allow(Rails.logger).to receive(:add)
+  end
+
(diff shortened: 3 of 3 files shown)


config/application.rb
```diff
@@ -17,6 +17,8 @@ class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+
+    config.eager_load_paths += %W[#{config.root}/lib]
```
ndushay: coul there be a comment explaining why we want this, and why we want it for all environments?  Or is it super obvious and I don't have the rails foo?

lib/pcc_logger.rb
```diff
@@ -0,0 +1,7 @@
+# A basic logger so we can write log messages in a regular format.
+class PCCLogger
+  def self.log(severity, message, druid = '')
+    log_message = Time.now.iso8601 + " " + druid + " " + message
```
ndushay: uh - the rails log doesn't default to using a timestamp?  Or doesn't have config options for such a thing?

ndushay:
I'm also wondering if there is a way to do this so our logging class "takes over" from the Rails.logger so we don't have to remember to use the special logging class as we continue to code (and for long term maintenance).

Is there a way to set the Rails logger to use a particular format and/or set it to a different class of logger?

tingulfsen:
Leaving this open until we've discussed logging with Christina and Erin - that might inform the approach.

jmartin-sul:
work on this implementation temporarily suspended, pending meeting tomorrow about possible future approaches to logging and error capture for the department in general

ndushay:
See https://docs.google.com/document/d/1Gxi73V2TfZviOIc6yr4cEpcUWN55Jowcgu93VxoxrIY

tingulfsen:
Today we agreed that this is probably the wrong approach for logging - we'll start off with Rails instrumentation instead per suggestion from @cmh2166.

PR #93, opened 2017-09-15
add postgres to rake commands

SaravShah:
Adding ```postgres``` to ```psql -f db/scripts/pcc_setup.sql postgres``` and ```psql -f db/scripts/preservation_core_catalog_setup.sql postgres```

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,8 @@ and that it's accepting connections:
 
 Using the `psql` utility, run these two setup scripts from the command line, like so:
 ```sh
-psql -f db/scripts/pcc_setup.sql
-psql -f db/scripts/preservation_core_catalog_setup.sql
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,8 +52,8 @@ and that it's accepting connections:
 
 Using the `psql` utility, run these two setup scripts from the command line, like so:
 ```sh
-psql -f db/scripts/pcc_setup.sql
-psql -f db/scripts/preservation_core_catalog_setup.sql
+psql -f db/scripts/pcc_setup.sql postgres
+psql -f db/scripts/preservation_core_catalog_setup.sql postgres
```
ndushay: hrmm - should it be presented as an either/or?

ndushay: sounds like this is good to go as is.

Issue #94, opened 2017-09-15
do we need dev and prod db separate?  user too?


ndushay:
Our database names do not reflect the functionality of said databases:

pcc -->  dev db user / tables

preservation_core_catalog --> test db user / tables

suggested:  pcc and pcc-test

research:  can pcc be used for dev and prod user / database table names

ndushay:
postgres examples might be avail: check puppet for quimby, kurma (geo) machines, cidr machines ... sul-reports ...

tallenaz:
The current db names are pcc for dev and pcc_test for test. 

see: https://github.com/sul-dlss/preservation_core_catalog/commit/cd625d2cf63639a5455112b8a5c30c8f235606b2

Should be no problem for the prod db to be names pcc

ndushay:
no, DEV and PROD.  :-)

tallenaz:
yes, DEV and PROD :)

PR #97, opened 2017-09-15
Change preservation_core_catalog db to pcc_test db

SaravShah:
- Only changed preservation_core_catalog to pcc_test
- haven't done anything with dev or prod

connects #94


ndushay:
Looks like this PR also needs to tweak Rakefile  as looking at .travis.yml, it calls a rake task to setup the database, and that is the refactored one ...

PR #117, opened 2017-09-25
Add timestamps to preservation_copies for checking db, storage roots …

### LINKED ISSUE #106: ensure PCC db has timestamp fields for ...
- storage_root to db 1:1 existence check 
- db to storage_root 1:1 existence check
- checksum validation
- any other specific validation checks as they come into play

tallenaz:
…and checksum validations

  Open to shortening the migration name to fewer than 90 characters...

  Closes #106

### DIFF
--- a/db/migrate/20170925210519_add_last_checked_on_storage_and_last_checksum_validation_to_preservation_copies.rb
+++ b/db/migrate/20170925210519_add_last_checked_on_storage_and_last_checksum_validation_to_preservation_copies.rb
@@ -1,0 +1,6 @@
+class AddLastCheckedOnStorageAndLastChecksumValidationToPreservationCopies < ActiveRecord::Migration[5.1]
+  def change
+    add_column :preservation_copies, :last_checked_on_storage, :datetime
+    add_column :preservation_copies, :last_checksum_validation, :datetime
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170908214835) do
+ActiveRecord::Schema.define(version: 20170925210519) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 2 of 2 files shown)


db/migrate/20170925210519_add_last_existed_in_storage_root_and_last_existed_in_db_and_last_checksum_validation_to_preservation_copies.rb
```diff
@@ -0,0 +1,7 @@
+class AddLastExistedInStorageRootAndLastExistedInDbAndLastChecksumValidationToPreservationCopies < ActiveRecord::Migration[5.1]
+  def change
+    add_column :preservation_copies, :last_existed_in_storage_root, :datetime
+    add_column :preservation_copies, :last_existed_in_db, :datetime
```
jmartin-sul: i would suggest maybe `last_checked_on_storage` and `last_checked_in_catalog`, respectively.  the bit of nuance being, in an error case, i.e. if the thing we expected isn't on disk at the version we expect, or it isn't already in the catalog, that column will indicate the last time we updated our knowledge of that state, but it won't necessarily indicate that the thing was in the place we expected at that time (because maybe it wasn't).

unless we stop updating those fields when we notice something's amiss, in which case those would be accurate names.

one more stray thought:  maybe we don't actually need `last_existed_in_db`/`last_checked_in_catalog`/whatever.  maybe that's implied by the `updated_at` timestamp.

and yet another stray thought, though it sort of ties into all of the above:  maybe `last_existed_in_db` is inherently non-sensical, because if the row did somehow disappear from the DB after being there for a time, we've already lost that information.  `last_checked_in_catalog` seems slightly less semantically sticky in that regard, but i'm also not sure it's any more useful than `updated_at` in light of that scenario.

so i dunno, maybe i'm actually just arguing to change the name of `last_existed_in_storage_root` and to drop `last_existed_in_db`.  i'm not entirely sure what the changes i'm looking for here are, obviously 😄.

db/migrate/20170925210519_add_last_checked_on_storage_and_last_checksum_validation_to_preservation_copies.rb
```diff
@@ -0,0 +1,6 @@
+class AddLastCheckedOnStorageAndLastChecksumValidationToPreservationCopies < ActiveRecord::Migration[5.1]
+  def change
+    add_column :preservation_copies, :last_checkeded_on_storage, :datetime
```
tingulfsen: Spelling: last_checkedED_on_storage

Issue #75, opened 2017-09-08
Maybe update spec tests  to factory girl


SaravShah:
-  run by team
- have  a lot of duplicated code for seeding/spec testing

ndushay:
we will still want actual moabs for "inventory functionality" code but perhaps factory girl would be useful in some of the other contexts.

ndushay:
I am closing this per PR #128  If this need comes up again, we'll probably open a new ticket (or reopen this one.)

PR #128, opened 2017-09-28
seed database with requisite foreign key referents

### LINKED ISSUE #112: seed database with default endpoints, endpoint_types, pres policies
need these things to actually create preservation_copies, need to enter pres copies to actually do inventory (as per schema design)

see also #75

jmartin-sul:
this adds code to create default `endpoints`, `endpoint_types`, `preservation_policies`, and `statuses` entries, by using class methods in the relevant models.  the definition for the `rake db:seed` task then just calls those methods in a transaction.

relates to #131: `preservation_copies` has a FK to endpoints, which has a FK to `endpoint_types`, which has relationships with `preservation_policies` via a join table; and `preservation_copies` also has a required FK to `statuses`.  so those things need to be there to create `preservation_copies` entries, per #131.

i didn't test the actual `rake db:seed` task definition.  happy to look into that if people want.  i couldn't find any info on preferred approaches after a cursory search, and the methods that do the heavy lifting are pretty well tested, so i figured that might be enough to start.  another option would be to ticket investigation of better testing for that.  i imagine some people must test their database seeding?

in a separate test branch (not pushed to GH), i rebased on @ndushay's branch for PR #130, and it was clean and tests passed.  but IIRC, that PR creates some pres policies inline, and that should probably be switched to call the new default getter (`PreservationPolicy.default_preservation_policy`) from this PR?  happy to wait till that PR is merged, rebase this, and make that switch in this PR in a new commit.  or happy to do it in a follow on ticket and PR.

closes #112


ndushay:
should this PR also affect #75?   Should #75 be closed??

config/settings.yml
```diff
@@ -1,3 +1,33 @@
...
+endpoint_types:
+  local_online_nfs:
+    endpoint_class: online
+    endpoint_defaults:
+      endpoint_node: 'localhost'
+      recovery_cost: 1
+
+# TODO: more illustrative fixture setup: create two storage roots, 'spec/fixtures/storage_root_1' and
+# 'spec/fixtures/storage_root_2', and split the moab fixtures between them.  also, rename 'moab_storage_root'
+# to 'moab_storage_trunk'.
```
jmartin-sul: created a [ticket](https://github.com/sul-dlss/preservation_core_catalog/issues/140), will get rid of this comment

jmartin-sul:
@ndushay: i'd be happy to close the factory girl issue for now and reopen it if it looks like it'd be handy.  or leave it open and push it down the backlog.  whatever.  can't think of anything off-hand at the moment that'd switch to it.

jmartin-sul:
ok, in the spirit of "save controversial commits for last", i removed a method that would've been more appropriately removed under then #115.  but having the method that commit removes makes me hesitant to wire up the actual storage roots on dev, because we don't want to glob any or all storage roots on demand from network requests.  we probably don't even want to use the more efficient method from #115 to return that to a caller.  callers should probably get lists from db queries, with the db being updated regularly through pcc inventory, and messages from ingest robots.

jmartin-sul:
rebased on master after #130 was merged.  then updated to use the default `PreservationPolicy` instead of creating a new one each time a `PreservedObject` is created.

config/settings.yml
```diff
@@ -1,3 +1,30 @@
...
+    endpoint_defaults:
+      endpoint_node: 'localhost'
+      recovery_cost: 1
+
 moab:
-  storage_roots: 'spec/fixtures'
+  # this is a list of named storage roots.  the names are used to seed the endpoints table.
+  storage_roots:
+    # name: '/location/on/disk'  # e.g.
+    fixtures: 'spec/fixtures'
```
SaravShah: what was the reason for adding ```fixtures``` under ```storage_roots```? Is it because we have multiple service disks, so when running ```storage_roots(Settings.moab.storage_roots.map { |_storage_root_name, storage_root_location| storage_root_location })``` you can iterate through all of the service disks?

config/settings.yml
```diff
@@ -1,3 +1,30 @@
...
   storage_trunk: 'moab_storage_root'
+
+# the first preservation policy listed will be considered the "default"
+# policy, to be used when a policy is otherwise left unspecified
+preservation_policies:
+  default:
+    archive_ttl: 604800 # _ttl values are in seconds
+    fixity_ttl: 604800
```
SaravShah: this is probably nitpicking .... but later in the code you put 604_800 because I'm guessing rubocop like that.

config/settings.yml
jmartin-sul: so, the idea was just to give each storage root a name, and `fixtures` was the name i chose for the `fixtures` directory.  but in practice, it'd probably be more like:
```yml
storage_roots:
  services-disk08: '/services-disk08'
  services-disk09: '/services-disk09'
```

i'd thought about doing something more clever, like using the path or parsing something from the path, but this felt more flexible, in case the paths were ever not just a single directory level.

config/settings.yml
jmartin-sul: i actually just wasn't sure whether you could do the underscore separator thing in yml files, and never bothered to try it out.  but happy to give it a shot if we'd prefer to be more consistent about that.

tallenaz: FWIW, just tested this out in `irb`, yml does support this:

```2.4.1 :003 > YAML.load("600_000")
 => 600000```

I'd prefer consistency, but in whatever direction -- if we want to disable that rubocop rule, I'd be fine with that, too.

**Comment by ndushay** [comment]
Isn't there a more "railsy" way to seed the db?  For example, I'm looking at

- http://guides.rubyonrails.org/active_record_migrations.html#migrations-and-seed-data
- https://davidmles.com/seeding-database-rails/ 

would our fixed data be appropriate in db/seeds.db  file?

File: config/settings.yml (hunk 3 above)
Comment (jmartin-sul) [reply config/settings.yml]: will fix in the yml.  or, fine to disable the rubocop rule, if people would prefer that.  no strong feelings either way from me.  i might have a slight preference for getting rid of the rubocop rule, but fine with whatever.

**Comment by jmartin-sul** [comment]
@ndushay:  this uses the `db/seeds.rb` file, but leverages methods in the relevant models.  that felt like the more testable and modular way to write the meat of the code, but it still lets you do the `rake db:seed` thing.

my feeling was that a lot of the time, we put the heavy lifting for tasks directly in rake task definitions, instead of calling neatly packaged methods from rake tasks.  that always felt like an anti-pattern to me, and this was just an attempt to get away from that.

FWIW, i actually had started out by just putting all this code inline in `seeds.db`, but this felt like a natural way to clean that up as i tried to make things more testable.

i would, however, like to find some way to neatly test that `db/seeds.rb` does the right thing, and it also feels like the existing model code has some repetition that could be DRYed up a bit.

File: app/services/preserved_object_handler.rb [hunk 4]
```diff
@@ -53,9 +53,7 @@ def update_or_create
       db_object = PreservedObject.find_by(druid: druid)
       results << update_per_version_comparison(db_object)
     else
-      pp_default = PreservationPolicy.create(preservation_policy_name: 'default',
-                                             archive_ttl: 604_800,
-                                             fixity_ttl: 604_800)
+      pp_default = PreservationPolicy.default_preservation_policy
```
ndushay: 👍

app/controllers/moab_storage_controller.rb
```diff
@@ -24,14 +16,4 @@ def show
...
-  private
-
-  def druids_from_storage_root
-    @druids ||= begin
-      @storage_root ||= "#{Moab::Config.storage_roots}/#{Moab::Config.storage_trunk}"
-      Dir.glob("#{@storage_root}/**/[a-z][a-z]*[0-9]").map { |d| d.split("/").last }
-    end
-  end
-
 end
```
ndushay: This might be better as a separate PR.  I  opened a ticket for "figure out what controller(s) we actually want" since the team isn't yet on the same page there.  

We're removing this code because we don't want to allow such a request against all the production storage_root data, yes?   Maybe comment out for now instead?   Or get rid of the routes?   I agree this code is likely to go away, but it's slightly possible we'll want some form of it.

app/models/endpoint.rb
```diff
@@ -11,4 +11,24 @@ class Endpoint < ApplicationRecord
   validates :endpoint_node, presence: true
   validates :storage_location, presence: true
   validates :recovery_cost, presence: true
+
+  # rubocop:disable Metrics/AbcSize
```
ndushay: this inline disables aren't my favorite, but it's certainly not worth more effort to simply shut up rubocop.  Shameless green is fine.

However, I did find it a bit more work than I would have liked to puzzle out the method below, and I'm wondering if my suggestions will help; and if they do, will they appease rubocop.

app/models/endpoint_type.rb
```diff
@@ -5,4 +5,24 @@ class EndpointType < ApplicationRecord
...
+      end
+    end
+  end
+
+  # the first endpoint type defined in the settings is considered the default endpoint type.
+  # NOTE: the assumption is that it will be a local "online" endpoint, i.e. on the same box as this
+  # app code, with exploded moab directories (not, e.g., a zipped archive on a remote machine).
+  def self.default_local_online_endpoint_type
+    default_endpoint_type_name = Settings.endpoint_types.first[0].to_s
+    find_by!(type_name: default_endpoint_type_name)
```
ndushay: I'm thinking this should be memoized  (`@@default ||= `)

app/models/endpoint.rb
```diff
@@ -11,4 +11,24 @@ class Endpoint < ApplicationRecord
...
+  # TODO: find a clean way to satisfy this style check, everything that silenced the complaint for me was
+  # less readable to my eyes due to what felt like needless indirection.
+  def self.seed_local_online_endpoints_from_config(preservation_policies = nil)
```
ndushay: decide to use "default" or "local_online" and use it consistently in the method name and var names  (or "default_local_online" or whatever).  I keep having to check what we're talking about.

app/models/endpoint.rb
```diff
@@ -11,4 +11,24 @@ class Endpoint < ApplicationRecord
...
+    default_endpoint_type_config = Settings.endpoint_types.first[1]
+    endpoint_defaults = default_endpoint_type_config.endpoint_defaults
+    endpoint_type = EndpointType.default_local_online_endpoint_type
+    preservation_policies = [PreservationPolicy.default_preservation_policy]
+    Settings.moab.storage_roots.map do |storage_root_name, storage_root_location|
+      find_or_create_by!(endpoint_name: storage_root_name.to_s) do |endpoint|
+        endpoint.endpoint_type = endpoint_type
```
ndushay: any reason not to put the right side of line 21  on the right side of line 25?   I would hope there is an optimizing compiler that would only do the lookup once, and i think it might make the code easier to read.  I'm having trouble tracking all the variables set above.

app/models/endpoint.rb
```diff
@@ -11,4 +11,24 @@ class Endpoint < ApplicationRecord
...
+        endpoint.endpoint_node = endpoint_defaults.endpoint_node
+        endpoint.storage_location = File.join(storage_root_location, Settings.moab.storage_trunk)
+        endpoint.recovery_cost = endpoint_defaults.recovery_cost
+        endpoint.preservation_policies = preservation_policies
```
ndushay: ditto here:  left side of line 22 could be put on the left side of line 29 ?

app/models/endpoint.rb
```diff
@@ -11,4 +11,24 @@ class Endpoint < ApplicationRecord
   validates :endpoint_node, presence: true
   validates :storage_location, presence: true
   validates :recovery_cost, presence: true
+
+  # rubocop:disable Metrics/AbcSize
+  # TODO: find a clean way to satisfy this style check, everything that silenced the complaint for me was
+  # less readable to my eyes due to what felt like needless indirection.
+  def self.seed_local_online_endpoints_from_config(preservation_policies = nil)
+    default_endpoint_type_config = Settings.endpoint_types.first[1]
+    endpoint_defaults = default_endpoint_type_config.endpoint_defaults
```
ndushay: I think lines 19-20 could be their own method `endpoint_defaults` ?   and that method should be memoized, no?

config/settings.yml
```diff
@@ -1,3 +1,30 @@
...
+      endpoint_node: 'localhost'
+      recovery_cost: 1
+
 moab:
-  storage_roots: 'spec/fixtures'
+  # this is a list of named storage roots.  the names are used to seed the endpoints table.
+  storage_roots:
+    # name: '/location/on/disk'  # e.g.
+    fixtures: 'spec/fixtures'
   storage_trunk: 'moab_storage_root'
```
ndushay: it might be nice to add a comment about storage_trunk, such as "storage_trunk" is the parent directory of all the storage roots".

I would also like to see the `path_method` explicitly set to `:druid_tree` here (see https://github.com/sul-dlss/moab-versioning/blob/master/spec/unit_tests/stanford/storage_repository_spec.rb and https://github.com/sul-dlss/moab-versioning/blob/master/spec/spec_config.rb and

config/settings.yml
ndushay: I think it would be good to indicate that 604_800 is a week's worth of seconds (if that's the correct answer).

db/seeds.rb
```diff
@@ -1,7 +1,15 @@
...
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+
+ApplicationRecord.transaction do
+  PreservationPolicy.seed_from_config
+  EndpointType.seed_from_config
+  Endpoint.seed_local_online_endpoints_from_config
+  Status.seed_from_config
+end
```
ndushay: so is the idea that we will call this via rake db:xxx task for all/any of the environments?  Should there be a comment that it can be safely run repeatedly?

spec/controllers/moab_storage_controller_spec.rb
```diff
@@ -2,44 +2,6 @@
...
-    it 'returns json when requested' do
-      get :index, format: :json
-      expect(response.content_type).to eq "application/json"
-    end
-    it 'returns xml when requested' do
-      get :index, format: :xml
-      expect(response.content_type).to eq "application/xml"
-    end
-  end
-
```
ndushay: See above per making this a separate PR, and possibly just commenting out the route for now.

ndushay:
Agree that this shouldn't be in a migration;  am open to code living in seeds.db or in models, depending on what rails experts say.  Dunno if you can access settings from seeds.db or not, dunno if it's a good idea or not.

As for tests - might it be possible for a test to run the rake task and then check the database state afterwards?  Or is that dumb?

jmartin-sul:
> Agree that this shouldn't be in a migration; am open to code living in seeds.db or in models, depending on what rails experts say. Dunno if you can access settings from seeds.db or not, dunno if it's a good idea or not.

you can access settings from seeds.db, i did it during my testing, since my implementation started out inline in seeds.db.  i am just super-hesitant to turn discreet model methods into a monolithic script, unless this approach is completely non-idiomatic and confusing for rails experts, or unless there's an architectural problem with it that i'm not seeing.  anyway, should be getting a once-over on this from someone more facile with rails patterns this afternoon (jessie said he'd take a look, though i can ask around and see if someone else is free if he can't get to it soon enough).

> As for tests - might it be possible for a test to run the rake task and then check the database state afterwards?

i would like to do that as an additional test, but i'd be hesitant to take out the existing unit tests.  it seems nice to have the seed code be well decomposed and testable if possible, and i'd be very hesitant to give that up now that the test coverage is there.  one of my initial concerns when i was implementing this inline in seeds.db was how to get good test coverage over the code there, without just doing one integration-y "did everything get seeded?" sort of test.

will remove the commit that removes the moab_storage_controller method and turn that into a separate PR

db/seeds.rb
jmartin-sul: yup, `rake db:seed` should be callable from where-ever, and it should be safe to run repeatedly.  good call on adding that as a note, will do...

app/models/endpoint.rb
jmartin-sul: ditto, `PreservationPolicy.default_preservation_policy` does a db lookup underneath

app/models/endpoint.rb
jmartin-sul: well, `EndpointType.default_local_online_endpoint_type` does a database lookup, and so it's something i'd like to not do every loop iteration, but also something i don't feel comfortable caching for more than the duration of this method call.  what about a comment indicating the reasoning for this approach?

app/models/endpoint.rb
jmartin-sul: ditto, `PreservationPolicy.default_preservation_policy` does a db lookup underneath

app/models/endpoint.rb
ndushay: ah - yeah, a comment.

app/models/endpoint.rb
ndushay: in both cases, you're looking up the single record for the default;   the default shouldn't change very often at all;  i think it could be a memoized class method in the appropriate model, and we could have methods in THIS model `default_preservation_policies`  and `default_endpoint_type` that are also memoized class methods here, for convenience.  Then we have nice brief, readable names.

app/controllers/moab_storage_controller.rb
tingulfsen: I'm happy to go with this commit and nuke the methods - just FYI. But I agree that it would be better as a separate PR.

app/models/endpoint.rb
jmartin-sul: well, i don't know that the names should line up so nicely, since they aren't referring to quite the same thing...

i don't really think we have the concept of a "default" endpoint or a "default" endpoint type in general (right?).  so it doesn't really make sense to me to name the `EndpointType` method `EndpointType.default_endpoint_type`.  there's a default configuration for local online endpoints, but not endpoint types in general.  so that's what `EndpointType.default_local_online_endpoint_type` gets.

similarly, with `Endpoint.seed_local_online_endpoints_from_config`...  i could've just done `Endpoint.seed_from_config`, as with the other classes, but i guess i wanted to make it clear that the method only does local endpoints at the moment, since there will be other endpoint types that need configuring as we do more development.  and since there didn't seem to me to be a concept of "default" endpoints, `.seed_default_endpoints_from_config` felt wrong.  though i'm not sure i made the right choice, and would totally be open to just calling it `Endpoint.seed_from_config`, with the assumption that the underlying code will be refactored, and `seed_from_config` will do all the seeding, for all types of endpoints, as we continue to develop them.

i dunno... `Endpoint.seed_local_online_endpoints_from_config` seeds local online endpoints from configuration, using the default configuration for online local endpoints, obtained via `EndpointType.default_local_online_endpoint_type`.  i know this is verbose, and the names seem like they don't quite fit the same pattern, but i'd find `Endpoint.seed_default_endpoints_from_config` (or any version of that method name with `default` in it) to be misleading.

um... how about `EndpointType.default_local_online_endpoint_type` and `Endpoint.seed_from_config`?  less verbose, less seeming mismatch in terminology, still entirely truthful about what its doing (assuming my understanding of endpoints and endpoint types is correct).

ndushay: if you can grab tommy before he leaves and agree on terms, go for it.  I may have confused the "default" prez. policy with the endpoint type stuff (which is telling in my ability to read the code).   I think i am also slightly confused by "default configuration for local online endpoints" - is it likely to change, except for the diff between local dev (with fixtures) and deployed?   

Which configuration data in settings expects certain things first?  that feels "default"-y and I agree with tommy - it could be clearer to call it out explicitly in the config as default rather than making it position dependent.

basically, my overall sentiments are:
- the code in this method, as written, wasn't skimmably clear to me -- as someone less intimately familiar with the database 
- use explicit default rather than positional if simple to do
- oy, super long names.
- eventually, we'll seed all the endpoints from config, no?  not just online local?   how might that impact naming choices?

jmartin-sul:
i believe i've addressed all the concerns that have been raised, and pushed up a cleaned up version of the branch.  in particular:
* fixed `settings.yml` to use the `_` style to separate thousands, note about what default `_ttl` is in weeks
* got rid of `MoabController` changes
* refactoring got rid of the inline rubocop disable, because AbcSize complaint disappeared
* `Endpoint` is hopefully clearer, with better naming, and less convolution
  * still uses `seed_storage_root_endpoints_from_config` as a name, but when we generalize, `seed_from_config` could just call that?
  * names are still kinda long, admittedly
* `Endpoint` seed method now always takes params, which get passed in by `seed.rb` and specs, which gets rid of some of the complexity naomi pointed out
* many of the things that were suggested for memoization or moving inline naturally evaporated in the refactoring
* usage comments on `seed_` class methods and in `seed.rb`
* got rid of positional defaults, all defaults should be named as such in settings, and settings has been regrouped (hopefully) more sanely
* since params always get passed to `Endpoint.seed_storage_root_endpoints_from_config`, no longer need to worry about checking default for nil
* jessie gives the thumbs up to this approach in general
  * hasn't seen this specific pattern, but thinks it's sane, and that it'd be hard to test thoroughly if everything lived inline in `seeds.rb`.
    * easier to test if the methods live with their models.  this seems like a good approach if testing seeding is important.
    * triggering seeding and checking results in db after startup sounds like only way to test if all's inline in `seeds.rb`, and that feels messy
    * referencing `Settings` for instance specific configs feels more idiomatic than e.g. including `seeds-dev.db` conditionally from `seeds.rb` depending on env
    * feels like the class methods that only know about their own respective class are fine
    * feels like maybe Endpoint knowing about other classes is iffy?
      * maybe that's better in a service that knows about all the relevant objects, but prob fine to wait and see if it needs to be broken out as we get more endpoints and endpoint types


normally, i'd have put the `storage_trunk` and `path_method` related changes in a separate PR, but i put them as the last two commits on this PR, just because naomi had requested them in the context of this PR.  but happy to lop those off and turn into a separate PR if people would prefer.

app/controllers/moab_storage_controller.rb
```diff
@@ -29,7 +29,7 @@ def show
 
   def druids_from_storage_root
     @druids ||= begin
-      @storage_root ||= "#{Moab::Config.storage_roots}/#{Moab::Config.storage_trunk}"
+      @storage_root ||= "#{Moab::Config.storage_roots.first}/#{Moab::Config.storage_trunk}"
```
ndushay: 👍 - elegant quick solution to the problem.

jmartin-sul: thanks!

PR #146, opened 2017-10-02
validate that ttl fields are only integers greater than zero

### LINKED ISSUE #135: schema/model polish:  add positive int validation on PreservationPolicy.archive_ttl and .fixity_ttl
Believe this is a change to the model code and maybe fixing up some existing code to accommodate.

if you spend more than 60 min on it, ask for help.

tallenaz:
Fixes #135 
  Arguably, existing specs cover the additional validations -- there's one to check that
  PreservationPolicy.new is not valid, for example.

### DIFF
--- a/app/models/preservation_policy.rb
+++ b/app/models/preservation_policy.rb
@@ -6,6 +6,6 @@ class PreservationPolicy < ApplicationRecord
   has_and_belongs_to_many :endpoints
 
   validates :preservation_policy_name, presence: true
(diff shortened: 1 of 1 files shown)


ndushay:
I'm ok without specs for this, as this is straight Rails validation code and if it doesn't work, then there's a lot of code out in the world that doesn't work.  :-)   Basically, this is a configuration.

👍  to merge.

Issue #108, opened 2017-09-20
(C2M)  catalog to moab


ndushay:
traversing druid in db for online Moabs

(possibly by storage root, or by oldest timestamp or ...), check existence of object at the storage root.

See https://docs.google.com/document/d/17QsWeb6au9fDBXmsJavb3iRyRql-stwuyDic1_ij_4M

ndushay:
![img_2233](https://user-images.githubusercontent.com/96775/30662630-9741d65e-9dfc-11e7-81e2-963a496c6101.JPG)

ndushay:
See also Julian's google doc (linked to from MVP "audit functionality" or some such -- there are checks for particular files in particular subdirs of a moab object to comprise "existence"

ndushay:
see `db-to-moab-existence` branch

PR #166, opened 2017-10-06
helper to load moab fixtures into catalog; always seeds test db for specs 

### LINKED ISSUE #149: test fixtures / seeded test db to enable catalog-to-moab existence check
In order to write code for the existence check (from db to storage_root), we need a seeded test db that corresponds with data in our test fixture objects.

Possibly use code from #107 to populate test db??

ndushay:
Wrote an rspec "shared context" https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-context to allow easy loading of our fixture moabs into the test db for specs.

Also:
- ensures that tests are run with test db seeded (added to rails_helper)
  - adjusts model specs affected by ^^
- bare beginnings of catalog to moab existence check code (useful for "testing" the load_fixtures_helper "shared context" ... and next work to do!)

Closes #149 (fixture moabs put in test db)
Connected to #108 (catalog to moab existence check)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -77,6 +77,9 @@ Rails/SkipsModelValidations:
     - 'app/services/preserved_object_handler.rb' # we're using touch on purpose
 
 # --- RSpec ---
+RSpec/BeforeAfterAll:
+  Exclude:
+    - 'spec/load_fixtures_helper.rb' # being careful to avoid state leakage
 
 RSpec/ExampleLength:
   Max: 50
--- a/lib/audit/catalog_to_moab_existence.rb
+++ b/lib/audit/catalog_to_moab_existence.rb
@@ -7,7 +7,7 @@ class CatalogToMoabExistence
   # FIXME:  temporarily turning off rubocop
   # rubocop:disable all
   def self.whole_db
-    PreservationCopy.all.each do |pc|
+    PreservationCopy.find_each do |pc|
       id = pc.preserved_object.druid
       version = pc.current_version
       storage_location = pc.endpoint.storage_location
--- a/lib/audit/catalog_to_moab_existence.rb
+++ b/lib/audit/catalog_to_moab_existence.rb
@@ -1,0 +1,30 @@
+require 'druid-tools'
+
+# Catalog to Moab existence check code
+class CatalogToMoabExistence
+
+  # NOTE: this is a bad thing -- HUGE list of objects;  just for now to populate test db
+  # FIXME:  temporarily turning off rubocop
+  # rubocop:disable all
+  def self.whole_db
+    PreservationCopy.all.each do |pc|
+      id = pc.preserved_object.druid
+      version = pc.current_version
+      storage_location = pc.endpoint.storage_location
+      druid = DruidTools::Druid.new(id)
+      object_dir = "#{storage_location}/#{druid.tree.join('/')}"
+
+      moab = Moab::StorageObject.new(id, object_dir)
+      moab_version = moab.current_version_id
+      if version == moab_version
+        p "hurray - #{id} versions match: #{version}"
+      else
+        p "boo - #{id} catalog has #{version} but moab has #{moab_version}"
+      end
+
+      # TODO: update status, timestamps, report errors, log, etc.
+    end
+  end
+  # rubocop:enable all
+
+end
--- a/lib/audit/catalog_to_moab_existence.rb
+++ b/lib/audit/catalog_to_moab_existence.rb
@@ -3,7 +3,9 @@ require 'druid-tools'
 # Catalog to Moab existence check code
 class CatalogToMoabExistence
 
-  # NOTE: this is a bad thing -- HUGE list of objects;  just for now to populate test db
+  # NOTE: this is at the chicken scratches stage
+  #  I'm not convinced we'll _ever_ do the whole db this way;  it was a way to start
+  #    step 1: ensure load_fixtures_helper works with very rough code here
   # FIXME:  temporarily turning off rubocop
   # rubocop:disable all
   def self.whole_db
--- a/spec/lib/audit/catalog_to_moab_existence_spec.rb
+++ b/spec/lib/audit/catalog_to_moab_existence_spec.rb
@@ -1,0 +1,11 @@
+require_relative '../../../lib/audit/catalog_to_moab_existence.rb'
+require_relative '../../load_fixtures_helper.rb'
+
+RSpec.describe CatalogToMoabExistence do
+  include_context 'fixture moabs in db'
+  describe '.whole_db' do
+    it 'test_of_load_fixtures_helper (eventually will be testing code in .whole_db)' do
+      described_class.whole_db
+    end
+  end
+end
--- a/spec/lib/audit/catalog_to_moab_existence_spec.rb
+++ b/spec/lib/audit/catalog_to_moab_existence_spec.rb
@@ -4,7 +4,9 @@ require_relative '../../load_fixtures_helper.rb'
 RSpec.describe CatalogToMoabExistence do
   include_context 'fixture moabs in db'
   describe '.whole_db' do
-    it 'test_of_load_fixtures_helper (eventually will be testing code in .whole_db)' do
+    # NOTE: this is at the chicken scratches stage
+    #  step 1: ensure load_fixtures_helper works
+    it 'test_of_load_fixtures_helper (eventually will be testing code for .whole_db)' do
       described_class.whole_db
     end
   end
--- a/spec/load_fixtures_helper.rb
+++ b/spec/load_fixtures_helper.rb
@@ -1,0 +1,59 @@
+require 'rails_helper.rb'
+
+RSpec.configure do |rspec|
+  # This config option will be enabled by default on RSpec 4,
+  # but for reasons of backwards compatibility, you have to
+  # set it on RSpec 3.
+  #
+  # It causes the host group and examples to inherit metadata
+  # from the shared context.
+  rspec.shared_context_metadata_behavior = :apply_to_host_groups
+end
+
+RSpec.shared_context "fixture moabs in db" do
+  before(:all) do
+    setup
+    load_fixture_moabs
+  end
+  after(:all) do
+    # TODO: danger - if there are objects in the test db we want to keep
+    Settings.moab.storage_roots.each do |_name, storage_root|
+      storage_dir = File.join(storage_root, Settings.moab.storage_trunk)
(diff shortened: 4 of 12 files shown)


ndushay:
@tingulfsen is especially requested as a reviewer because he hasn't seen this at all yet.

lib/audit/catalog_to_moab_existence.rb
```diff
@@ -0,0 +1,30 @@
+require 'druid-tools'
+
+# Catalog to Moab existence check code
+class CatalogToMoabExistence
+
+  # NOTE: this is a bad thing -- HUGE list of objects;  just for now to populate test db
+  # FIXME:  temporarily turning off rubocop
+  # rubocop:disable all
+  def self.whole_db
+    PreservationCopy.all.each do |pc|
```
jmartin-sul: you may want to use `find_each` for this:
http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each
> Looping through a collection of records from the database (using the ActiveRecord::Scoping::Named::ClassMethods#all method, for example) is very inefficient since it will try to instantiate all the objects at once.
> 
> In that case, batch processing methods allow you to work with the records in batches, thereby greatly reducing memory consumption.
> 
> The find_each method uses find_in_batches with a batch size of 1000 (or as specified by the :batch_size option).

had vague memories of something like that, did some googling, the stack exchange posts suggested `find_each` but weren't very well written, but thankfully the rails docs seem pretty helpful/thorough in this case.

seems like you could just sub `find_each` for `all.each` and keep the contents of the block unchanged, since you still get the object as the block parameter.

spec/load_fixtures_helper.rb
```diff
@@ -0,0 +1,59 @@
...
+RSpec.shared_context "fixture moabs in db" do
+  before(:all) do
+    setup
+    load_fixture_moabs
+  end
+  after(:all) do
+    # TODO: danger - if there are objects in the test db we want to keep
+    Settings.moab.storage_roots.each do |_name, storage_root|
+      storage_dir = File.join(storage_root, Settings.moab.storage_trunk)
+      PreservationCopy.where(endpoint_id: Endpoint.find_by(storage_location: storage_dir).id).each do |pc|
```
jmartin-sul: prob not worth it in this PR, but as a quick tech debt thing ticket, we might want to centralize this `storage_dir = File.join(storage_root, Settings.moab.storage_trunk)` thing into some little helper method somewhere, since we seem to be using it in at least a couple places (happy to create a ticket if that seems like a good idea).

spec/load_fixtures_helper.rb
```diff
@@ -0,0 +1,59 @@
...
+end
+
+RSpec.shared_context "fixture moabs in db" do
+  before(:all) do
+    setup
+    load_fixture_moabs
+  end
+  after(:all) do
+    # TODO: danger - if there are objects in the test db we want to keep
+    Settings.moab.storage_roots.each do |_name, storage_root|
```
jmartin-sul: or actually, since `setup` has been called, could you just use `@moab_storage_dirs` here instead of looping over `Settings.moab.storage_roots` again?

spec/load_fixtures_helper.rb
jmartin-sul: likewise with the `@storage_dir_to_endpoint_id` hash instead of `Endpoint.find_by`?

SaravShah:
looks good, i agree with john on the ```find_each``` method. 
When the time comes, we should also have consistent method names for the existence check :)

spec/rails_helper.rb
```diff
@@ -7,6 +7,11 @@
 require 'rspec/rails'
 # Add additional requires below this line. Rails is not loaded until this point!
 
+require 'rake'
+Rails.application.load_tasks
+Rake::Task["db:migrate"].invoke
+Rake::Task["db:seed"].invoke
+
```
tingulfsen: I like this - it's definitely an improvement. What about adding a statement to explicitly remove all contents from the database just to be 110% sure we start off with the expected test data?

spec/models/endpoint_spec.rb
```diff
@@ -93,11 +86,12 @@
...
     it 'returns the default endpoint type object for the storage root' do
-      EndpointType.seed_from_config
+      # db already seeded
       expect(Endpoint.default_storage_root_endpoint_type).to be_a_kind_of EndpointType
     end
 
     it "raises RecordNotFound if the default endpoint type doesn't exist in the db" do
+      skip('database seeded before running tests; not super-trivial to destroy relevant objects')
       expect { Endpoint.default_storage_root_endpoint_type }.to raise_error(ActiveRecord::RecordNotFound)
     end
```
tingulfsen: If we're not going to use this test, why not just delete it.

spec/models/preservation_policy_spec.rb
```diff
@@ -44,11 +44,12 @@
...
     it 'returns the default preservation policy object' do
-      PreservationPolicy.seed_from_config
+      # db already seeded
       expect(PreservationPolicy.default_preservation_policy).to be_a_kind_of PreservationPolicy
     end
 
     it "raises RecordNotFound if the default policy doesn't exist in the db" do
+      skip('database seeded before running tests; not super-trivial to destroy relevant objects')
       expect { PreservationPolicy.default_preservation_policy }.to raise_error(ActiveRecord::RecordNotFound)
     end
```
tingulfsen: Same as above - I agree we don't have to keep this test around. So let's just delete it instead of leaving skipped tests around.

lib/audit/catalog_to_moab_existence.rb
```diff
@@ -0,0 +1,30 @@
...
+      moab = Moab::StorageObject.new(id, object_dir)
+      moab_version = moab.current_version_id
+      if version == moab_version
+        p "hurray - #{id} versions match: #{version}"
+      else
+        p "boo - #{id} catalog has #{version} but moab has #{moab_version}"
+      end
+
+      # TODO: update status, timestamps, report errors, log, etc.
+    end
```
tingulfsen: Can you write some tickets for these TODOs? Or is that captured in another ticket?

spec/lib/audit/catalog_to_moab_existence_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+require_relative '../../load_fixtures_helper.rb'
+
+RSpec.describe CatalogToMoabExistence do
+  include_context 'fixture moabs in db'
+  describe '.whole_db' do
+    it 'test_of_load_fixtures_helper (eventually will be testing code in .whole_db)' do
+      described_class.whole_db
+    end
+  end
+end
```
tingulfsen: I'm perfectly happy to leave this as a bare-bones PR, but could you at least write down some comments here on what we should eventually be testing?

.rubocop.yml
```diff
@@ -77,6 +77,9 @@ Rails/SkipsModelValidations:
     - 'app/services/preserved_object_handler.rb' # we're using touch on purpose
 
 # --- RSpec ---
+RSpec/BeforeAfterAll:
+  Exclude:
+    - 'spec/load_fixtures_helper.rb' # being careful to avoid state leakage
```
tingulfsen: What is it about load_fixtures_helper.rb that makes it so RuboCop-antagonistic?

lib/audit/catalog_to_moab_existence.rb
ndushay: thanks for this.  I will change this PR.

I'm guessing that in the general case we're going to want to do "the least recently checked X moabs (? on storage_root 5 ?)" or some such.

spec/load_fixtures_helper.rb
ndushay: no, because the instance vars aren't available to before/after(:all) blocks

spec/load_fixtures_helper.rb
ndushay: great idea!  Please do make a ticket.

ndushay: no access to instance vars in before/after(:all)

spec/models/endpoint_spec.rb
ndushay: because @jmartin-sul intends to fix it and he also didn't want to lose the test.

spec/models/preservation_policy_spec.rb
ndushay: ditto to above comments.

lib/audit/catalog_to_moab_existence.rb
ndushay: I believe the work described by this comment is ticket #108;  I only have this much of a stub here because I needed some actual code for an actual test to determine if the load_fixture stuff worked.

spec/lib/audit/catalog_to_moab_existence_spec.rb
ndushay: again - part of ticket #108.  This is strictly meant here as "does my test fixture loading work?"

.rubocop.yml
ndushay: rubocop doesn't like before(:all) or after(:all) as they are not transactional fixtures that get rolled back.  I did them this way because it seemed silly to populate the test db with the fixture data for EACH test.  The objection was all about "is the database clean after your tests?" and I say: "yes, I was careful to clean the db".

spec/rails_helper.rb
ndushay: interesting idea.

ndushay: Did some research;  I think we'd have to code "unseed" methods and/or drop and re-create the whole test db.  That seems a bit heavy handed.  If you think this is worth the trouble, I suggest you create a separate ticket for it;  I'm not feeling passionate about it.

ndushay:
@jmartin-sul @tingulfsen  I believe I have changed the code where merited and have responded to your comments.  Please re-review.

spec/rails_helper.rb
tingulfsen: 👍  Created #169

spec/load_fixtures_helper.rb
jmartin-sul: ah, right, sorry

spec/models/preservation_policy_spec.rb
tingulfsen: OK, spoke to @jmartin-sul and created #170.

spec/load_fixtures_helper.rb
jmartin-sul: filed https://github.com/sul-dlss/preservation_core_catalog/issues/171

Issue #169, opened 2017-10-09
Clear database contents upon test start


tingulfsen:
spec/rails_helper.rb seeds the test database. Consider using one of the Rake methods to drop the test database before running `rake db:migrate`, or maybe just write a super basic script to remove the table rows.

PR #172, opened 2017-10-09
Add db:reset rake task

SaravShah:
Remove 
```
Rake::Task["db:migrate"].invoke
Rake::Task["db:seed"].invoke
```

and add
```
Rake::Task["db:reset"].invoke
```

``` rake db:reset``` drops, creates, migrates and seeds the db.

connects #169

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -9,8 +9,7 @@ require 'rspec/rails'
 
 require 'rake'
 Rails.application.load_tasks
-Rake::Task["db:migrate"].invoke
-Rake::Task["db:seed"].invoke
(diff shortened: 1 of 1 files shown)


ndushay:
I wonder if it's a bit heavy handed to wipe the test db when you run the tests.  Just a comment.  If it becomes a problem later, we'll know.

👍  to merge.

jmartin-sul:
yeah, seems like it could become heavy handed as we accumulate more data, but also seems like a fine way to get us to a known state for now, and i like the way it reads.  and agreed that we'll know if/when it becomes problematic (in which case, seems easy enough to move to a more selective setup).

Issue #183, opened 2017-10-10
(PC) add separate methods for create and update to PreservedObjectHandler


SaravShah:
Split into the Rails update and create methods for nice and easy REST calls :)

PR #187, opened 2017-10-12
[WIP] 183 split update or create

### LINKED ISSUE #131: (PC) PreservedObjectHandler needs to create PreservationCopy object
#180 creates an entry in the PreservedObject table.  It should be revised:

- [ ] also create an entry in the PreservationCopies table
   - [ ] which requires using data from the Endpoints table ...
- [ ] use existing default PreservationPolicy, not create a new one
  - [ ] ditto for specs ...

Note that the root_to_db existence check code will PASS IN a storage_directory per ticket #157.  Therefore, co-dependent with #157 -- maybe develop both in a feature branch???  Or do both tickets simultaneously.

### LINKED ISSUE #157: (M2C) needs to pass storage_dir to PreservedObjectHandler
... in order to allow preservation_copy table entry to include correct endpoint.

co-dependent with #131 -- maybe develop both in a feature branch???  Or do both tickets simultaneously

SaravShah:
connects #183
closes #157
closes #131


jmartin-sul:
superseded by #190

PR #190, opened 2017-10-13
split `PreservedObjectHandler#update_or_create` into separate `#update` and `#create` methods

### LINKED ISSUE #131: (PC) PreservedObjectHandler needs to create PreservationCopy object

### LINKED ISSUE #157: (M2C) needs to pass storage_dir to PreservedObjectHandler

### LINKED ISSUE #180: (M2C) discovered version less than Object version
If the discovered version of the druid is less than the version recorded in PRESERVED_OBJECT, alert noisily but attempt no remediation. Later development calls for this event to trigger a Recovery action (an attempt to recover this endpoint’s druid to the most recent version from another endpoint).

### LINKED ISSUE #183: (PC) add separate methods for create and update to PreservedObjectHandler

jmartin-sul:
* remaining TODO
  * [x] rebase on current master
  * [x] make sure all status codes are being used
  * [x] make sure status code messages say the right thing
  * [x] make sure status codes correspond to right logger severity
  * [x] fix `MoabToCatalog.check_moab_to_catalog_existence` to do the right po_handler calls
  * [ ] some new uses of `%{addl}` should maybe be `%{obj_class_name}` (actually, this prob requires assuming `addl` will be a hash w/ named params, since var name `addl` is expected by helper methods `result_code_msg` and `result_hash`) ticket #196 
  * [x] rubocop cleanup - 2 rubocop errors left, change ```to received``` to  ```to have_received```
  * [x] refactoring?  DRY things up?
  * [x] unit tests for `if invalid?` branch in `#update` and `#create` methods of PreservedObjectHandler
  * [ ] `#create` should prob trap `ActiveRecord::ActiveRecordError` for status code return like `#update` ticket #198
    * [ ] and corresponding unit test ticket #198 
  * [ ] add corresponding test for ```druid expected to exist in catalog but was not found``` in ```moab_to_catalog.rb```  ticket #197
  * [ ] test for `MoabToCatalog.check_moab_to_catalog_existence` `expect_to_create` flag

closes #183
closes #157
closes #131
closes #180

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,7 @@ group :development do
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
   gem 'ruby-prof'
+  gem 'hirb' # for db table display via rails console
 end
 
 group :test do
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -15,16 +15,21 @@ class PreservedObjectHandler
   UPDATED_DB_OBJECT_TIMESTAMP_ONLY = 6
   CREATED_NEW_OBJECT = 7
   DB_UPDATE_FAILED = 8
+  OBJECT_ALREADY_EXISTS = 9
+  OBJECT_DOES_NOT_EXIST = 10
 
   RESPONSE_CODE_TO_MESSAGES = {
     INVALID_ARGUMENTS => "encountered validation error(s): %{addl}",
-    VERSION_MATCHES => "incoming version (%{incoming_version}) matches db version",
-    ARG_VERSION_GREATER_THAN_DB_OBJECT => "incoming version (%{incoming_version}) greater than db version",
-    ARG_VERSION_LESS_THAN_DB_OBJECT => "incoming version (%{incoming_version}) less than db version; ERROR!",
-    UPDATED_DB_OBJECT => "db object updated",
-    UPDATED_DB_OBJECT_TIMESTAMP_ONLY => "updated db timestamp only",
+    VERSION_MATCHES => "incoming version (%{incoming_version}) matches %{addl} db version",
+    ARG_VERSION_GREATER_THAN_DB_OBJECT => "incoming version (%{incoming_version}) greater than %{addl} db version",
+    ARG_VERSION_LESS_THAN_DB_OBJECT => "incoming version (%{incoming_version}) less than %{addl} db version; ERROR!",
+    UPDATED_DB_OBJECT => "%{addl} db object updated",
+    UPDATED_DB_OBJECT_TIMESTAMP_ONLY => "%{addl} updated db timestamp only",
     CREATED_NEW_OBJECT => "added object to db as it did not exist",
-    DB_UPDATE_FAILED => "db update failed: %{addl}"
+    DB_UPDATE_FAILED => "db update failed: %{addl}",
+    OBJECT_ALREADY_EXISTS => "%{addl} db object already exists",
+    OBJECT_DOES_NOT_EXIST => "%{addl} db object does not exist"
+
   }.freeze
 
   include ActiveModel::Validations
@@ -33,32 +38,58 @@ class PreservedObjectHandler
   validates :druid, presence: true, format: { with: DruidTools::Druid.pattern }
   validates :incoming_version, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :incoming_size, numericality: { only_integer: true, greater_than: 0 }
+  validates :endpoint, presence: true
 
-  attr_reader :druid, :incoming_version, :incoming_size
+  attr_reader :druid, :incoming_version, :incoming_size, :storage_dir, :endpoint
 
-  def initialize(druid, incoming_version, incoming_size)
+  def initialize(druid, incoming_version, incoming_size, storage_dir)
     @druid = druid
     @incoming_version = version_string_to_int(incoming_version)
     @incoming_size = string_to_int(incoming_size)
+    @storage_dir = storage_dir
+    @endpoint = Endpoint.find_by(storage_location: storage_dir)
   end
 
-  def update_or_create
+  def create
     results = []
-
     if invalid?
       results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
     elsif PreservedObject.exists?(druid: druid)
-      Rails.logger.debug "update #{druid} called and object exists"
-
-      db_object = PreservedObject.find_by(druid: druid)
-      results << update_per_version_comparison(db_object)
+      results << result_hash(OBJECT_ALREADY_EXISTS, 'PreservedObject')
     else
       pp_default = PreservationPolicy.default_preservation_policy
-      PreservedObject.create(druid: druid,
-                             current_version: incoming_version,
-                             size: incoming_size,
-                             preservation_policy: pp_default)
+      po = PreservedObject.create!(druid: druid,
+                                   current_version: incoming_version,
+                                   size: incoming_size,
+                                   preservation_policy: pp_default)
+      status = Status.default_status
+      PreservationCopy.create(preserved_object: po,
+                              current_version: incoming_version,
+                              endpoint: endpoint,
+                              status: status)
       results << result_hash(CREATED_NEW_OBJECT)
+    end
+    results.flatten!
+    log_results(results)
+    results
+  end
+
+  def update
+    results = []
+    if invalid?
+      results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
+    elsif !PreservedObject.exists?(druid: druid)
+      results << result_hash(OBJECT_DOES_NOT_EXIST, 'PreservedObject')
+    else
+      Rails.logger.debug "update #{druid} called and object exists"
+      begin
+        po_db_object = PreservedObject.find_by(druid: druid)
+        results << update_per_version_comparison(po_db_object)
+        pc_db_object = PreservationCopy.find_by(preserved_object: po_db_object, endpoint: endpoint)
+        results << update_per_version_comparison(pc_db_object)
+      rescue ActiveRecord::ActiveRecordError => e
+        results << result_hash(DB_UPDATE_FAILED, "#{e.inspect} #{e.message} #{e.backtrace.inspect}")
+      end
     end
     results.flatten!
     log_results(results)
@@ -73,16 +104,16 @@ class PreservedObjectHandler
     version_comparison = db_object.current_version <=> incoming_version
     results = []
     if version_comparison.zero?
-      results << result_hash(VERSION_MATCHES)
+      results << result_hash(VERSION_MATCHES, db_object.class.name)
     elsif version_comparison == 1
       # TODO: needs manual intervention until automatic recovery services implemented
-      results << result_hash(ARG_VERSION_LESS_THAN_DB_OBJECT)
+      # TODO: we should also probably update status here?
+      results << result_hash(ARG_VERSION_LESS_THAN_DB_OBJECT, db_object.class.name)
     elsif version_comparison == -1
       db_object.current_version = incoming_version
-      db_object.size = incoming_size if incoming_size
-      results << result_hash(ARG_VERSION_GREATER_THAN_DB_OBJECT)
+      db_object.size = incoming_size if db_object.instance_of?(PreservedObject) && incoming_size
+      results << result_hash(ARG_VERSION_GREATER_THAN_DB_OBJECT, db_object.class.name)
     end
-
     update_db_object(db_object, results)
     results.flatten
   end
@@ -92,14 +123,12 @@ class PreservedObjectHandler
   def update_db_object(db_object, results)
     if db_object.changed?
       db_object.save
-      results << result_hash(UPDATED_DB_OBJECT)
+      results << result_hash(UPDATED_DB_OBJECT, db_object.class.name)
     else
       # FIXME: we may not want to do this, but instead to update specific timestamp for check
       db_object.touch
-      results << result_hash(UPDATED_DB_OBJECT_TIMESTAMP_ONLY)
+      results << result_hash(UPDATED_DB_OBJECT_TIMESTAMP_ONLY, db_object.class.name)
     end
-  rescue ActiveRecord::ActiveRecordError => e
-    results << result_hash(DB_UPDATE_FAILED, "#{e.inspect} #{e.message} #{e.backtrace.inspect}")
   end
 
   def result_hash(response_code, addl=nil)
@@ -111,7 +140,7 @@ class PreservedObjectHandler
   end
 
   def result_msg_prefix
-    @msg_prefix ||= "PreservedObjectHandler(#{druid}, #{incoming_version}, #{incoming_size})"
+    @msg_prefix ||= "PreservedObjectHandler(#{druid}, #{incoming_version}, #{incoming_size}, #{storage_dir})"
   end
 
   # results = [result1, result2]
@@ -121,7 +150,6 @@ class PreservedObjectHandler
     results.each do |r|
       severity = logger_severity_level(r.keys.first)
       msg = r.values.first
-      # Rails.logger.log(severity, msg, 'PreservedObjectHandler')
       Rails.logger.log(severity, msg)
     end
   end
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -43,6 +43,7 @@ group :development do
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
   gem 'ruby-prof'
+  gem 'hirb'
```
ndushay: please indicate what hirb does / is for?

app/services/preserved_object_handler.rb
```diff
@@ -33,73 +39,115 @@ class PreservedObjectHandler
...
+      # TODO: needs manual intervention until automatic recovery services implemented
+      results << result_hash(ARG_VERSION_LESS_THAN_DB_OBJECT, db_object.class.name)
+    elsif version_comparison == -1
+      db_object.current_version = incoming_version
+      results << result_hash(ARG_VERSION_GREATER_THAN_DB_OBJECT, db_object.class.name)
+    end
     update_db_object(db_object, results)
-    results.flatten
+    results.flatten!
+    results
```
ndushay: why not 
```
results.flatten
```

vs
```
results.flatten!
results
```

?

app/services/preserved_object_handler.rb
```diff
@@ -134,8 +182,10 @@ def logger_severity_level(result_code)
     when ARG_VERSION_LESS_THAN_DB_OBJECT then Logger::ERROR
     when UPDATED_DB_OBJECT then Logger::INFO
     when UPDATED_DB_OBJECT_TIMESTAMP_ONLY then Logger::INFO
-    when CREATED_NEW_OBJECT then Logger::WARN
+    when CREATED_NEW_OBJECT then Logger::INFO
```
ndushay: b/c the spec changed?  or should this line be dependent on `expect_to_create`?

lib/audit/moab_to_catalog.rb
```diff
@@ -1,14 +1,19 @@
 ##
 # Method that will check the a single moab service disk for the existence of moabs in postgres database
 class MoabToCatalog
-  def self.check_moab_to_catalog_existence(storage_dir)
+  def self.check_moab_to_catalog_existence(storage_dir, expect_to_create=false)
```
ndushay: method takes expect_to_create but does nothing with it yet?  (not even pass it thru?)

Gemfile
SaravShah: it allows us to display the sql database in the rails console as a table for easier reading. I used it during the demo. 
This is what it looks like
```>> Tag.last
+-----+-------------------------+-------------+---------------+-----------+-----------+-------+
| id  | created_at              | description | name          | namespace | predicate | value |
+-----+-------------------------+-------------+---------------+-----------+-----------+-------+
| 907 | 2009-03-06 21:10:41 UTC |             | gem:tags=yaml | gem       | tags      | yaml  |
+-----+-------------------------+-------------+---------------+-----------+-----------+-------+
1 row in set```

Comment (ndushay) [reply Gemfile]: that's cool!  I meant there should be a comment in the Gemfile like `for db table display via rails console` or something.

File: app/services/preserved_object_handler.rb (hunk 3 above)
Comment (SaravShah) [reply app/services/preserved_object_handler.rb]: IIRC we changed it to INFO because when the object is created, we don't want to log that as a warning as the create method was called on purpose....Then if the method was accidentally called on an object thats not in the db  we log an error in the audit check? @jmartin-sul

File: lib/audit/moab_to_catalog.rb (hunk 4 above)
Comment (SaravShah) [reply lib/audit/moab_to_catalog.rb]: Yeah I need to add the use case for that in the rspec test - not sure if this is the best way to do this though, we can probably come up with a better way so we don't have to set expect_to_create=true every time we want to create an object.. @jmartin-sul ?

File: app/services/preserved_object_handler.rb [hunk 5]
```diff
@@ -33,56 +38,99 @@ class PreservedObjectHandler
...
       pp_default = PreservationPolicy.default_preservation_policy
-      PreservedObject.create(druid: druid,
-                             current_version: incoming_version,
-                             size: incoming_size,
-                             preservation_policy: pp_default)
+      po = PreservedObject.create!(druid: druid,
+                                   current_version: incoming_version,
+                                   size: incoming_size,
+                                   preservation_policy: pp_default)
+      status = Status.find_by(status_text: "ok") # TODO: Replace the status_text to get a default status_text
```
ndushay: this is fine;  just want to make sure this is ticketed, as default status is now in master.

Or, you can just do `Status.default_status` and you're done already.

app/services/preserved_object_handler.rb
```diff
@@ -33,56 +38,99 @@ class PreservedObjectHandler
...
       db_object.current_version = incoming_version
       db_object.size = incoming_size if incoming_size
-      results << result_hash(ARG_VERSION_GREATER_THAN_DB_OBJECT)
+      results << result_hash(ARG_VERSION_GREATER_THAN_DB_OBJECT, db_object.class.name)
     end
+    update_db_object(db_object, results)
+    results.flatten
+  end
 
+  def update_per_version_comparison_pc(db_object)
```
ndushay: am I missing something obvious?  it seems like update_per_version_comparison_xx take the same argument, have the same comparison, do the same thing, and call the same method ... so why do we have separate methods?  line 115 could be in an "if is_instance_of?" clause.   

If there's a reason to keep them distinct, would it make sense to distinguish the param name (preserved_object vs preservation_copy) ?

app/services/preserved_object_handler.rb
```diff
@@ -92,14 +140,12 @@ def update_per_version_comparison(db_object)
...
-      results << result_hash(UPDATED_DB_OBJECT)
+      results << result_hash(UPDATED_DB_OBJECT, db_object.class.name)
     else
       # FIXME: we may not want to do this, but instead to update specific timestamp for check
       db_object.touch
-      results << result_hash(UPDATED_DB_OBJECT_TIMESTAMP_ONLY)
+      results << result_hash(UPDATED_DB_OBJECT_TIMESTAMP_ONLY, db_object.class.name)
     end
-  rescue ActiveRecord::ActiveRecordError => e
-    results << result_hash(DB_UPDATE_FAILED, "#{e.inspect} #{e.message} #{e.backtrace.inspect}")
```
ndushay: I see that rescue ActiveRecordError was moved up the method chain -- I'm fine with it but wonder if there's a principle or a "teach naomi" moment here?

app/services/preserved_object_handler.rb
ndushay: ah - maybe you're right about the spec changing.  I'm kinda at the "let's do it, fire it up, and let the users see if they like it" stage.

app/services/preserved_object_handler.rb
```diff
@@ -111,17 +157,13 @@ def result_code_msg(response_code, addl=nil)
...
 
-  # results = [result1, result2]
-  # result1 = {response_code => msg}
-  # result2 = {response_code => msg}
   def log_results(results)
     results.each do |r|
       severity = logger_severity_level(r.keys.first)
       msg = r.values.first
-      # Rails.logger.log(severity, msg, 'PreservedObjectHandler')
       Rails.logger.log(severity, msg)
```
ndushay: so we don't want info about the calling class?  or it's already in the logs?  or christina's logs are capturing it?

app/services/preserved_object_handler.rb
```diff
@@ -111,17 +157,13 @@ def result_code_msg(response_code, addl=nil)
   end
 
   def result_msg_prefix
-    @msg_prefix ||= "PreservedObjectHandler(#{druid}, #{incoming_version}, #{incoming_size})"
+    @msg_prefix ||= "PreservedObjectHandler(#{druid}, #{incoming_version}, #{incoming_size}, #{storage_dir})"
   end
 
-  # results = [result1, result2]
-  # result1 = {response_code => msg}
-  # result2 = {response_code => msg}
```
ndushay: i think the format of the results argument isn't inherently obvious, which is why I put in these comments.  Any reaction?

lib/audit/moab_to_catalog.rb
```diff
@@ -1,14 +1,19 @@
...
       moab_current_version = moab.current_version_id
       moab_size = Stanford::StorageServices.object_size(druid)
-      po_handler = PreservedObjectHandler.new(druid, moab_current_version, moab_size)
-      results << po_handler.update_or_create
+      po_handler = PreservedObjectHandler.new(druid, moab_current_version, moab_size, storage_dir)
+      if PreservedObject.exists?(druid: druid)
+        results << po_handler.update
+      else
+        Rails.logger.error "druid: #{druid} expected to exist in catalog but was not found" unless expect_to_create
+        results << po_handler.create
```
ndushay: er ... don't we only create if "expect_to_create" is true?   o.w. we don't?

I guess I'm thinking when 
- we're seeding the db for the first time -- we're calling create
- we're auditing the moabs against the catalog -- we're calling update and we might be "expect_to_create" ... or we might not. ??

maybe we just need to leave the code like this and have a ticket/question for "will we always "expect_to_create" for moab_to_catalog?"   and a separate ticket (if we don't already have one) for "seed the catalog" that calls create.

lib/benchmarking/moab_to_catalog_testing.rb
```diff
@@ -26,13 +29,16 @@
...
-# storage_dir = 'spec/fixtures/moab_storage_root'
-storage_dir = '/services-disk12/sdr2objects'
-time_to_check_existence = Benchmark.realtime do
-  RubyProf.start
-  logger.info "Check for output"
-  StorageRootToDB.check_moab_to_catalog_existence(storage_dir)
-  @prof = RubyProf.stop
+# storage_dir = 'spec/fixtures/storage_root01/moab_storage_trunk'
+# storage_dir = '/services-disk12/sdr2objects'
+Stanford::StorageServices.storage_roots.each do |storage_root|
```
ndushay: this is reading from whatever is in Settings and thus in config initializer, yes?  Might be good to indicate with a comment.

spec/lib/audit/moab_to_catalog_spec.rb
```diff
@@ -13,28 +13,51 @@
...
-      expected_argument_list.each do |arg_hash|
-        po_handler = instance_double('PreservedObjectHandler')
-        arg_hash[:po_handler] = po_handler
-        allow(PreservedObjectHandler).to receive(:new).with(
-          arg_hash[:druid],
-          arg_hash[:storage_root_current_version],
-          instance_of(Integer)
-        ).and_return(po_handler)
-        allow(po_handler).to receive(:update_or_create)
+    context "determine create or update method with expected values" do
```
ndushay: I may not be getting this because I'm not looking the whole spec file ... but ...   

- seeding job always calls create, no?
- audit check always calls update first, and MAY call create if "expect_to_create" is true?

Gemfile
```diff
@@ -43,6 +43,7 @@ group :development do
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
   gem 'ruby-prof'
+  gem 'hirb' # for db table display via rails console
```
ndushay: 👍  thank you for the comment.

app/services/preserved_object_handler.rb
SaravShah: 👍

app/services/preserved_object_handler.rb
SaravShah: No reason. We just wanted to get the code working and refactor later - i'll update this :)

app/services/preserved_object_handler.rb
SaravShah: Just ran a test and ``` Rails.logger.log(severity, msg, 'PreservedObjectHandler')``` logs the same thing as ``` Rails.logger.log(severity, msg)```

app/services/preserved_object_handler.rb
SaravShah: Yeah I think it is a good idea to add the format of results back. Will do that

lib/benchmarking/moab_to_catalog_testing.rb
SaravShah: will do!

app/services/preserved_object_handler.rb
jmartin-sul: yup, agreed w/ @SaravShah, i think by splitting `update_or_create` into two methods, we're pushing the responsibility for warning about an unexpected create out to the caller, which decides whether to create if the object isn't already there.

(which, from my point of view, is less an overall spec change, and more an implementation/decomposition change, which is to say we should still be logging a warning, which i think we're doing, but now it's here, AFAICT:  https://github.com/sul-dlss/preservation_core_catalog/pull/190/files#diff-7d6cab3132273db5a784afb09ac3b5e0R14)

lib/audit/moab_to_catalog.rb
jmartin-sul: i think the idea is that `expect_to_create` is usually `false` (because it's being used by the regular inventory process), but that it's `true` when we're populating the catalog for the first time (in which case, it'd be nice to not get warned 1.5 million times when we expect to be creating all the catalog entries on that particular invocation).

app/services/preserved_object_handler.rb
jmartin-sul: i think we wanted the catch to apply to the entire operation of creating the `PreservedObject` and the `PreservationCopy`.  if the former fails, we don't want to move on to the latter, but as it was, an error in pres copy creation would've been trapped, and would've allowed processing to move onto something that depends on a thing we just failed to create.

we may also want to wrap that whole thing in an explicit transaction... i think declaring transactions isn't something we've thought about much so far, but we might at least want to file a tech debt ticket to figure out where (if anyplace) we might need them.

lib/audit/moab_to_catalog.rb
jmartin-sul: i think the idea is that we always create, but when we expect the object to be there, we log a warning.

so, in your first example case (initially seeding db), the flag would be true, and we don't log anything, because we expect to be creating everything from scratch.  in your second example (auditing against the catalog), we don't expect to create, because we expect ingest robots to have created the entry already, so we log a warning that we just found an unexpected inconsistency between the catalog and what's on disk.

i guess my current thinking is that the code for regular inventory audit and initial spin up is mostly the same, other than the expectation about whether the object is already in the catalog.  though, now that i think about it, if ingest messages about updates too, then maybe any inconsistency (object not in catalog OR object already in catalog but at the wrong version) results in a warning?

lib/audit/moab_to_catalog.rb
ndushay: I actually think we will want to create specific code to create the catalog for the first time, and use the "create" method.  This is partly b/c we may add to this code to populate the TCR, since looping through each moab takes some CPU effort.

app/services/preserved_object_handler.rb
ndushay: I'm clueless about how to do transactions, but I think this makes sense.

lib/audit/moab_to_catalog.rb
ndushay: F2F group discussion -- we agreed that there will be a "seed db" entry point into code that calls "create" and a separate "M2C existence check" entry point into code.  @jmartin-sul created appropriate tickets for that work, so this is now 👍 in the context of this PR.

lib/audit/moab_to_catalog.rb
ndushay: F2F group discussion -- we agreed that there will be a "seed db" entry point into code that calls "create" and a separate "M2C existence check" entry point into code.  @jmartin-sul created appropriate tickets for that work, so this is now 👍 in the context of this PR.

spec/lib/audit/moab_to_catalog_spec.rb
ndushay: F2F group discussion -- we agreed that there will be a "seed db" entry point into code that calls "create" and a separate "M2C existence check" entry point into code.  @jmartin-sul created appropriate tickets for that work, so this is now 👍 in the context of this PR.

PR #192, opened 2017-10-13
make preservation_policy_name unique and indexed

### LINKED ISSUE #137: schema/model polish:  preservation_policies.preservation_policy_name should be unique and indexed
Believe this is a small db migration to be written and maybe updating some code to accommodate.

if you spend more than 60-90 min on it, ask for help.

unique may be a validation only??

tallenaz:
Using `default` as the same `preservation_policy_name` across different specs was making rspec try to create more than one `preservation_policy` object whose name was `default`, triggering the uniqueness validation, so my workaround was to make the name unique to each spec.

Resolve #137

### DIFF
--- a/spec/models/preservation_copy_spec.rb
+++ b/spec/models/preservation_copy_spec.rb
@@ -11,11 +11,7 @@ RSpec.describe PreservationCopy, type: :model do
       recovery_cost: '1'
     )
   end
-  let!(:preservation_policy) do
-    PreservationPolicy.create!(preservation_policy_name: 'default',
-                               archive_ttl: 604_800,
-                               fixity_ttl: 604_800)
-  end
+  let!(:preservation_policy) { PreservationPolicy.find_by(preservation_policy_name: 'default') }
 
   let!(:preserved_object) do
     PreservedObject.create!(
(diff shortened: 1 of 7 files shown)


spec/models/preservation_copy_spec.rb
```diff
@@ -12,7 +12,7 @@
     )
   end
   let!(:preservation_policy) do
-    PreservationPolicy.create!(preservation_policy_name: 'default',
+    PreservationPolicy.create!(preservation_policy_name: 'pc_spec_default',
                                archive_ttl: 604_800,
                                fixity_ttl: 604_800)
```
ndushay: I imagine we should use the default policy from the db, not create one ... will that fix it?

tallenaz:
@ndushay I've replaced inline unit test fixtures with references to seeded db objects

ndushay:
does this resolve #137?  If so, put 'Resolve #137' in the first comment/description

tallenaz:
It resolves #137. I've added the github keyword to the initial comment

spec/models/preservation_copy_spec.rb
```diff
@@ -11,11 +11,7 @@
       recovery_cost: '1'
     )
   end
-  let!(:preservation_policy) do
-    PreservationPolicy.create!(preservation_policy_name: 'default',
-                               archive_ttl: 604_800,
-                               fixity_ttl: 604_800)
-  end
+  let!(:preservation_policy) { PreservationPolicy.find_by(preservation_policy_name: 'default') }
```
ndushay: 👍

Issue #133, opened 2017-09-29
druid regex should come from druid-tools


ndushay:
perhaps in application_controller?  or PreservedObject?  or in a lib/contstants folder?

See #130

PR #195, opened 2017-10-16
change druid regex to use druid-tools

SaravShah:
instead of using ``` /\A[a-z]{2}\d{3}[a-z]{2}\d{4}\z/``` we can use ```DruidTools::Druid.pattern```
connects #133


jmartin-sul:
d'oh, i retract that comment, can't use the druid tools regex as-is in the place i mentioned, because it assumes the druid is the whole line.

merging as-is and ticketing follow up work.

jmartin-sul:
filed https://github.com/sul-dlss/druid-tools/issues/4 related to my comment above.

PR #208, opened 2017-10-20
[WIP] Update PreservationCopy status when DB version > Moab/local disk version

### LINKED ISSUE #181: (M2C)  discovered version less than Copy version
If the discovered version of the druid is less than the version recorded for this endpoint in PRESERVATION_COPY, alert noisily but attempt no remediation - this implies lost data or a failed attempt by IngestWF to update that requires investigation - again, this would be a trigger for a Recovery action (out of scope for this sprint).

tingulfsen:
Fixes #181 

This PR only deals with updating the database status. I believe the rest of what #181 asks for is already covered by the error that is being logged.

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -100,6 +100,7 @@ class PreservedObjectHandler
 
   # expects @incoming_version to be numeric
   # TODO: update existence check timestamps/status per each flavor of comparison?
+  # rubocop:disable Metrics/PerceivedComplexity
   def update_per_version_comparison(db_object)
     version_comparison = db_object.current_version <=> incoming_version
     results = []
@@ -107,8 +108,8 @@ class PreservedObjectHandler
       results << result_hash(VERSION_MATCHES, db_object.class.name)
     elsif version_comparison == 1
       # TODO: needs manual intervention until automatic recovery services implemented
-      # TODO: we should also probably update status here?
       results << result_hash(ARG_VERSION_LESS_THAN_DB_OBJECT, db_object.class.name)
+      db_object.status = unexpected_version_status if db_object.instance_of?(PreservationCopy)
     elsif version_comparison == -1
       db_object.current_version = incoming_version
       db_object.size = incoming_size if db_object.instance_of?(PreservedObject) && incoming_size
@@ -182,5 +183,9 @@ class PreservedObjectHandler
     return val if val.instance_of?(Integer) # NOTE: negative integers caught with validation
     return val.to_i if val.instance_of?(String) && val.scan(/\D/).empty?
     val
+  end
+
+  def unexpected_version_status
+    @unexpected_version_status ||= Status.find_by!(status_text: Settings.statuses.detect { |s| s == 'expected_version_not_found_on_disk' })
   end
 end
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -100,7 +100,7 @@ class PreservedObjectHandler
 
   # expects @incoming_version to be numeric
   # TODO: update existence check timestamps/status per each flavor of comparison?
-  # rubocop:disable Metrics/PerceivedComplexity
+  # rubocop:disable Metrics/PerceivedComplexity  -- temporary until we fix #211
   def update_per_version_comparison(db_object)
     version_comparison = db_object.current_version <=> incoming_version
     results = []
--- a/spec/services/preserved_object_handler_spec.rb
+++ b/spec/services/preserved_object_handler_spec.rb
@@ -354,6 +354,7 @@ RSpec.describe PreservedObjectHandler do
         let(:version_less_than_pc_msg) { "#{exp_msg_prefix} incoming version (1) less than PreservationCopy db version; ERROR!" }
         let(:updated_po_db_timestamp_msg) { "#{exp_msg_prefix} PreservedObject updated db timestamp only" }
         let(:updated_pc_db_timestamp_msg) { "#{exp_msg_prefix} PreservationCopy updated db timestamp only" }
+        let(:updated_pc_db_msg)           { "#{exp_msg_prefix} PreservationCopy db object updated" }
 
         it "entry version stays the same" do
           expect(po.current_version).to eq 2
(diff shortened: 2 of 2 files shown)


tingulfsen:
It looks like my change was the final drop that pushed the method over the edge - RuboCop now regards it as too complex. I'd like some feedback from reviewers on what to do - refactor the method or leave that for later or ignore?

app/services/preserved_object_handler.rb
```diff
@@ -183,4 +183,8 @@ def string_to_int(val)
     return val.to_i if val.instance_of?(String) && val.scan(/\D/).empty?
     val
   end
+
+  def unexpected_version_status
+    @unexpected_version_status ||= Status.find_by(status_text: Settings.statuses.detect { |s| s == 'expected_version_not_found_on_disk' })
```
jmartin-sul: i'd be inclined to do a `find_by!` here so it fails fast, but it'd fail soon enough w/ a `null` status being set if nothing gets found here.

SaravShah:
yeah i could be in favor for having a new ticket to refactor

tingulfsen:
I think it's a good idea to use find_by!() in this case as suggested, so I did that and created #211 - this means deferring refactoring the offending method until patch party day.

app/services/preserved_object_handler.rb
```diff
@@ -183,4 +184,8 @@ def string_to_int(val)
     return val.to_i if val.instance_of?(String) && val.scan(/\D/).empty?
     val
   end
```
jmartin-sul: i think you want a corresponding `# rubocop:enable Metrics/PerceivedComplexity` here

ndushay: ew - i'm only okay with that if it's temporary and so indicated.

spec/services/preserved_object_handler_spec.rb
```diff
@@ -371,13 +372,13 @@
...
           allow(Rails.logger).to receive(:log).with(Logger::INFO, updated_po_db_timestamp_msg)
-          allow(Rails.logger).to receive(:log).with(Logger::INFO, updated_pc_db_timestamp_msg)
+          allow(Rails.logger).to receive(:log).with(Logger::INFO, updated_pc_db_msg)
 
           po_handler.update
           expect(Rails.logger).to have_received(:log).with(Logger::ERROR, version_less_than_po_msg)
           expect(Rails.logger).to have_received(:log).with(Logger::ERROR, version_less_than_pc_msg)
           expect(Rails.logger).to have_received(:log).with(Logger::INFO, updated_po_db_timestamp_msg)
-          expect(Rails.logger).to have_received(:log).with(Logger::INFO, updated_pc_db_timestamp_msg)
+          expect(Rails.logger).to have_received(:log).with(Logger::INFO, updated_pc_db_msg)
```
ndushay: we don't also update the timestamp when we update the status?  Or does Rails do it automagically ?

spec/services/preserved_object_handler_spec.rb
```diff
@@ -403,9 +404,9 @@
             result_msg = results.select { |r| r[PreservedObjectHandler::UPDATED_DB_OBJECT_TIMESTAMP_ONLY] }.first.values.first
             expect(result_msg).to match(Regexp.escape(updated_po_db_timestamp_msg))
           end
-          it "PreservationCopy UPDATED_DB_OBJECT_TIMESTAMP_ONLY result" do
-            result_msg = results.select { |r| r[PreservedObjectHandler::UPDATED_DB_OBJECT_TIMESTAMP_ONLY] }.second.values.first
-            expect(result_msg).to match(Regexp.escape(updated_pc_db_timestamp_msg))
```
ndushay: I'm confused:  I'm looking for a test when we find what we expect and just update the timestamp, and another test for when we don't find the version we expect and we update the status and the timestamp.  Is this my failure to see the full spec in github GUI, or is the spec and/or code incomplete?

app/services/preserved_object_handler.rb
```diff
@@ -100,15 +100,16 @@ def update
 
   # expects @incoming_version to be numeric
   # TODO: update existence check timestamps/status per each flavor of comparison?
+  # rubocop:disable Metrics/PerceivedComplexity
```
ndushay: can we indicate this is temporary and what ticket corresponds to its removal?   I hate this noise in the code.

ndushay: alternately, can we just raise the default allowed?

spec/services/preserved_object_handler_spec.rb
jmartin-sul: my understanding is that this should happen automagically on object save (e.g.`#save` and `#touch` calls).

ndushay: but Logger doesn't get a message?  Or does it?  Are our tests consistent?

jmartin-sul: ah, i think i see what you're getting at... sorry, wasn't thinking about logging specifically when responding to your question even though that was the context.

i think the log msg is `UPDATED_DB_OBJECT_TIMESTAMP_ONLY => "%{addl} updated db timestamp only"`.  the var names for the tests lose that bit of nuance, and make it seem like the test is about whether we log a message anytime the timestamp is updated.  but i think that message should only get logged when *only* the timestamp is updated, which isn't the case here for the pres copy anymore (since its status field is getting changed).  the associated `PreservedObject` has no status of its own, and so it is only getting its timestamp updated.

so i think this is ok?  i think i actually had a similar moment of confusion when i looked at this, so maybe a comment or a clarifying var rename is in order here?  maybe even just as a refactor/tech-debt ticket?

spec/services/preserved_object_handler_spec.rb
jmartin-sul: d'oh, yeah... we should prob actively set `db_object.status` back to `ok` when `version_comparison.zero?` is true.  maybe under a separate ticket?  though that's a bug due to our incomplete thinking about this ticket, so maybe it does belong in this one.

maybe we only actively update `status` to `ok` if it isn't already `ok`?  this would then necessitate both of the test cases you describe here ("a test when we find what we expect and just update the timestamp, and another test for when we don't find the version we expect and we update the status and the timestamp", though i'd add the clarification that finding what we would expect once again after things had been wrong probably warrants a result code and log message about status change).

jmartin-sul: yeah, we assume that version match always results in only a timestamp update:
https://github.com/sul-dlss/preservation_core_catalog/blob/718d3e01dafa0dcf3b39ececcc6768c98de40ecb/spec/services/preserved_object_handler_spec.rb#L222-L281

which is the current behavior, but shouldn't be, i don't think.

jmartin-sul: i really like this auto-embedding when you link a section of code

tingulfsen:
It seems like there are multiple things to talk about here, and I feel like it's best to have this discussion in person on Monday. All the points that have been raised are valid, I just think it's going to be more efficient to figure this out face-to-face.

jmartin-sul:
@tingulfsen totally reasonable, sounds like a plan to me

jmartin-sul:
likely to be closed in favor of work on newer branch

tingulfsen:
Closing this PR, since we're taking a different approach now. We'll keep this branch around until we're done resolving the issue.

Issue #68, opened 2017-09-08
(M2C) (SPC) rake tasks to fire off PC seed and M2C check


ndushay:
Allows us to:
- seed the (prod) db
- run it from cron
...

- [ ] actually run the task on storage_root
   - if errors, write separate tickets

ndushay:
consider ways to run it per storage root, and for combos?   Just via settings.yml???

jmartin-sul:
blocked by #183
blocked by #158
blocked by #201

jmartin-sul:
per @ndushay's comment above ("consider ways to run it per storage root, and for combos? Just via settings.yml???"), this might not strictly be blocked by #158, in that we might not always want the rake task to do all the storage roots.  but completing this issue requires at least deciding what we want to do on that front.

all the roots sequentially seems like a fine first pass to me, but we should discuss as a team when it comes time to implement this.

PR #213, opened 2017-10-21
moab_to_catalog split out check_existence vs seed_catalog

### LINKED ISSUE #201: (M2C, PC) split out MoabToCatalog.check_existence into separate existence check and seed methods
we want two methods that will be called by rake tasks:  one to walk the storage roots to seed the empty catalog, one to walk the storage roots on a regular basis and check the moab objects on disk against the catalog database.

currently, this is conflated into one method, `MoabToCatalog.check_moab_to_catalog_existence`.  the code for the two new methods might be largely similar, but it's a short method, and you know, shameless green.

### LINKED ISSUE #204: (M2C) Test that logging happens when existence check calls create()
The `expect_to_create` argument for `lib/audit/moab_to_catalog:check_moab_to_catalog_existence()` affects logging. There is no test that utilizes that argument. We should have such a test. I imagine it would go in `moab_to_catalog_spec.rb`.

ndushay:
besides creating the seed_catalog method, I adjusted the logic in the check_existence method to only create the object if expect_to_create is true (and object doesn't already exist).

Resolves #201
Resolves #204

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -114,6 +114,7 @@ RSpec/NamedSubject:
 
 RSpec/NestedGroups:
   Exclude:
+    - 'spec/lib/audit/moab_to_catalog_spec.rb'
     - 'spec/services/preserved_object_handler_spec.rb'
 
 # --- Style ---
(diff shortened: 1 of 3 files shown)


SaravShah:
Is there still a TODO of creating the rake task to run these methods?

ndushay:
yes, and/or create the script to run with the seed action vs existence check action ...

SaravShah:
ah right i just saw the ticket for that. #68

Issue #218, opened 2017-10-23
(M2C) (UC) pohandler.update_version method


ndushay:
When the ingest robot has a new version of a digital object, it expects to update the version of said object in the pres catalog.  This would be via a ReST call that will ultimate call poh.update_version.

The same need can arise if the M2C existence check finds a newer version on disk than in the catalog -- it would notify that it (M2C) hit an unexpected situation, and then call update_version.

__update_version__ method:
- expects catalog entries to exist
  - if not ... call create????
- expects incoming version > catalog version
- updates the approp __pres copy version, status and timestamp(s)__
- updates the approp __pres object version, size (if avail), timestamp__
- __logs__ appropriately

ndushay:
version comparison cases:
- if incoming = (catalog + 1) 
     - set version, etc.
     - status ok
- if incoming > (catalog + 1)
     - notify skipping version
     - set version, etc.
     - status ok

Note cases below this could just be considered a single error case for now, and a separate ticket created in the backlog.

- if incoming = catalog
     - notify version equal, expected diff
     - update timestamp(s)
     - status ____ ?
- if incoming < catalog
     - notify noisily
     - status ____?
     - update timestamp(s)?

PR #222, opened 2017-10-24
pohandler_spec: use shared_examples to check for attribute validation

ndushay:
This is preparatory work for writing update_version  method (which will use these shared examples).

connected to #218 

This isn't rocket science, so don't need all of you to review.

One question is "am I checking that rails ActiveRecord validation works?" or "am I checking that the error messages are what we want, informative, and that we actually call validation appropriately early in the methods?"  Dunno.

### DIFF
--- a/spec/services/preserved_object_handler_spec.rb
+++ b/spec/services/preserved_object_handler_spec.rb
@@ -57,6 +57,77 @@ RSpec.describe PreservedObjectHandler do
     end
   end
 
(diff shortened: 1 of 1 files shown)


SaravShah:
I think we should be checking that the error messages are what we want and informative. I believe we are already testing the ActiveRecord validation in the model already....i think.

ndushay:
this is checking the validation calls used in PreservedObjectHandler service, not in the models ... :-P

PR #224, opened 2017-10-25
Get settings for fixture storage roots out of the way of shared_configs

### LINKED ISSUE #217: (PC) cap  deploy should not get fixture storage roots in settings in addition to those from shared_config
from #200, @tallenaz says:

```
The remaining problem is that config/settings.yml appends the fixture storage roots to settings in shared_configs.
```

tallenaz:
Fixes #217 

This seems like a bit of a hack but it does get the fixtures out of the settings object -- I'm open to discussing better approaches.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -12,10 +12,13 @@ endpoints:
 
 moab:
   # this is a list of named storage roots.  the names are used to seed the endpoints table.
-  storage_roots:
+  # any named storage roots added in the root settings.yml will always be inherited by env specific configs, even
+  # if they aren't listed directly in the settings/*.yml file for an environment. so list storage roots directly in
+  # the relevant settings file for the env, including for vanilla unit test env.
+  # storage_roots:
     # name: '/location/on/disk'  # e.g.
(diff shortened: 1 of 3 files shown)


tallenaz:
Yes, that makes total sense 👍

jmartin-sul:
agreed w/ @ndushay, overall approach looks good to me, but i'm also in favor of trimming the duplication.

config/settings.yml
```diff
@@ -14,8 +14,8 @@ moab:
   # this is a list of named storage roots.  the names are used to seed the endpoints table.
   storage_roots:
     # name: '/location/on/disk'  # e.g.
```
jmartin-sul: sorry, just thought of one other idea:  what about a short explanatory comment like:
```ruby
# any named storage roots added in the root settings.yml will always be inherited by env specific configs, even
# if they aren't listed directly in the settings/*.yml file for an environment. so list storage roots directly in
# the relevant settings file for the env, including for unit vanilla env.
```

jmartin-sul:
@ndushay says 👍 now that the dev and test settings files have been trimmed of duplicate info

PR #227, opened 2017-10-25
Status model: fix ok status retrieval, add tests for .ok and .unexpected_version

jmartin-sul:


app/models/status.rb
```diff
@@ -24,6 +24,6 @@ def self.unexpected_version
   end
 
   def self.ok
-    find_by!(status_text: Settings.statuses.detect { |s| s == 'status_ok' })
+    find_by!(status_text: Settings.statuses.detect { |s| s == 'ok' })
   end
```
ndushay: so these are simply convenience methods?

jmartin-sul: yeah, just convenience methods.  and i agree that this feels a little odd.  maybe the `Settings.statuses.detect` thing is overkill?

i think we might still want a "default" status, even if it's just an alias to "ok".  though i'm not sure about that.

i'd prob be ok with hardcoded status seeds instead of pulling them from settings, considering where this code seems to be going, though i'd slightly prefer to do that refactoring under a separate ticket, or later on in this `refactor-poh-update-methods` branch work, so that first we can knock out the harder parts of the logic you and tommy and i are trying to wrangle.

but yeah, `Status` is starting to feel more and more like a Rails `enum` than a model of its own, if we don't think it'll ever be anything more than `status_text` in the table.  but that refactoring def feels like a separate ticket/branch to me, if we decide to go that way.

ndushay: ok with sep ticket if you want me to merge this for now.  Yay messes!  Yay clean up later!

jmartin-sul: @ndushay: filed #228 to address this properly in the longer term

PR #244, opened 2017-11-01
Add method to loop through all the storage roots when seeding.

### LINKED ISSUE #158: (M2C)   method to use all  storage roots in the Settings/config file
check_online_to_db_existence method is set up to check a single storage_dir. 

We want to be able to loop through all 13 service disks

tingulfsen:
Fixes #158


tingulfsen:
Note that this method ignores the results of the computation - logging already happens elsewhere.

PR #249, opened 2017-11-01
MoabStorageDirectory class extracted to moab-versioning gem

### LINKED ISSUE #147: (PC) move app/models/moab_storage_directory.rb to ...
moab-versioning gem `lib/stanford` directory?

esp. if TCR rails app (or other non PCC code) uses moab_storage_directory to populate TCR with checksums

jmartin-sul:
this shouldn't get merged till ~~https://github.com/sul-dlss/moab-versioning/pull/76 gets merged.  once that's merged,~~ a new version of moab-versioning is released.  once that happens, we can fix the Gemfile to use a proper moab-versioning release, and then we can take this out of WIP.

closes #147


jmartin-sul:
@ndushay: released v2.3.0 of moab-versioning, and updated this PR accordingly.  should be ready for merge.

PR #253, opened 2017-11-01
(PC) PreservedObjectHandler result handling

### LINKED ISSUE #225: (PC) (M2C) (UC) pohandler.update_db_object should return results, not have them as arg
(ready once there is a feature branch for #218 and #181)

All the other code that adds to the results returns the results array for-that-method.  We should be consistent and do the same for `update_db_object` (or be consistent and always pass the results object or make it an instance var)

### LINKED ISSUE #240: (PC) pohandler: use result.concat instead of result.flatten
and avoid calls to result.flatten

jmartin-sul:
issue #225: PreservedObjectHandler#update_db_object should just return a result list, instead of taking one as a param and appending to it.

issue #240:  switch to using Array#concat instead of Array#<< for joining result lists from helper methods to running result lists in the caller.

see https://ruby-doc.org/core-2.4.0/Array.html#method-i-3C-3C and https://ruby-doc.org/core-2.4.0/Array.html#method-i-concat

closes #225
closes #240

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -119,11 +119,11 @@ class PreservedObjectHandler
           results << result_hash(ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_copy.class.name)
           update_preserved_copy(pres_copy, incoming_version, incoming_size)
           results << update_status(pres_copy, Status.default_status)
-          update_db_object(pres_copy, results)
+          results.concat(update_db_object(pres_copy))
           if incoming_version > pres_object.current_version # FIXME: need code/test for when it's NOT
             results << result_hash(ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_object.class.name)
             update_preserved_object(pres_object, incoming_version)
-            update_db_object(pres_object, results)
+            results.concat(update_db_object(pres_object))
           end
         else
           results << result_hash(UNEXPECTED_VERSION, 'PreservedCopy')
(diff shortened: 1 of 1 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -78,7 +78,7 @@ def create
         results << result_hash(DB_UPDATE_FAILED, "#{e.inspect} #{e.message} #{e.backtrace.inspect}")
```
ndushay: the above doesn't need a concat?

app/services/preserved_object_handler.rb
```diff
@@ -165,7 +165,7 @@ def version_comparison_results(db_object, version_symbol)
     elsif incoming_version > db_object.send(version_symbol)
       results << result_hash(ARG_VERSION_GREATER_THAN_DB_OBJECT, db_object.class.name)
```
ndushay: don't we want results.concat here?

jmartin-sul: i think we still want to use `<<` for single elements, and `concat` for lists

jmartin-sul: indeed:
```
m2dl-suntzu-01-mbpx:preservation_core_catalog suntzu$ bundle exec rails c
Running via Spring preloader in process 59610
Loading development environment (Rails 5.1.4)
[1] pry(main)> x = ['a']
=> ["a"]
[2] pry(main)> x.concat 'abc'
TypeError: no implicit conversion of String into Array
from (pry):2:in `concat'
[3] pry(main)> 
```

PR #266, opened 2017-11-03
A basic rake task to seed the catalog from all the storage roots.

### LINKED ISSUE #68: (M2C) (SPC) rake tasks to fire off PC seed and M2C check

tingulfsen:
Fixes #68

The new rake task is only for seeding the catalog, so maybe break out the rest of the ticket into a new ticket or two?

If anyone knows how to avoid adding the line to application.rb then feel free to amend this branch as you please.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,3 +13,11 @@ task default: [:spec, :rubocop]
 task :travis_setup_postgres do
   sh("psql -U postgres -f db/scripts/pres_test_setup.sql")
 end
+
+require_relative 'lib/audit/moab_to_catalog'
+task seed_catalog: :environment do
+  m2c = MoabToCatalog.new
+  puts "Seeding the database from all storage roots..."
+  m2c.seed_from_disk
+  puts "Done"
+end


ndushay:
hmm.  what if we load the approp lib/audit  file  in the rake task?  The rails app per se doesn't need them, does it?

I'm not confident of this so I didn't make the code change.

ndushay:
I just created tickets #267 and #268 for M2C existence check versions of these.

ndushay:
I think the benefit of not loading up unnec files is that startup time for the rails app is slower.  I think the lib/audit code is meant to support the rake tasks, not the controller.   The logic supporting the rake tasks and the controller (i.e. per-druid logic) should arguably all be in a services class?   

So I am slightly in favor of requiring / loading lib/audit code in the rakefile.

tingulfsen:
@ndushay I think the latest change/rebase should address your concerns?

PR #293, opened 2017-11-08
[WIP] Misc poh refactoring

jmartin-sul:


jmartin-sul:
will break this into multiple PRs...

PR #294, opened 2017-11-08
centralize common ActiveRecord rescue behavior in PreservedObjectHandler

jmartin-sul:
looks like more methods will be repeating this pattern, so i took a shot at centralizing it to a method that takes a block and does the error handling.  that part is in the third commit (and required no spec changes).

also, a couple other bits of minor housekeeping (which did require a bit of spec touchup).

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -218,7 +218,7 @@ class PreservedObjectHandler
   def update_db_object(db_object)
     results = []
     if db_object.changed?
-      db_object.save
+      db_object.save!
       results << result_hash(UPDATED_DB_OBJECT, db_object.class.name)
     else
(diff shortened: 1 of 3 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -142,6 +133,27 @@ def update_version
 
   private
 
+  # runs the provided block.  if any of the expected exceptions are raised, returns a list
```
SaravShah: I don't think we need these comments - can tell this by looking at the code and the caller/ where its used

SaravShah:
Suggest the new private method goes in the shared examples file

jmartin-sul:
@SaravShah and @ndushay:  pulled the comment that was seen as unnecessary, rebased on master.

Issue #264, opened 2017-11-03
(PC) update_version needs to be endpoint aware (Moab vs Archived)


ndushay:
possibly covered by storage_dir argument to pohandler.initialize().   

@jmartin-sul -- does that mean this ticket is already done?

ndushay:
I believe we are now set up to finish implementing #262 (or to close it, if we don't want to worry about archival preserved_copy objects yet) ??  @jmartin-sul

jmartin-sul:
@ndushay: yup, that sounds right to me

PR #296, opened 2017-11-08
PreservedObjectHandler takes endpoint

jmartin-sul:
~~we should merge #294 first, since this builds on that, and then rebase this.~~

there's no unique constraint on the `storage_location` field.  as such, we won't always know that we can lookup `endpoint` solely by `storage_dir` (aka `storage_location`).

this PR lets the callers to POH pass in `endpoint` instead of `storage_dir`, under the assumption that the caller knows how to look up the endpoint it cares about.  e.g., `MoabToCatalog` knows it's dealing with local online storage roots, and so can look up endpoints solely by `storage_dir` (for the moment, though this lookup may require more criteria if and when `storage_location` is no longer unique across endpoints).

efficiency bonus:  since `MoabToCatalog.check_existence` takes a single storage dir, and is always passing the same storage dir to the POHs it instantiates, this lets us look up the endpoint once for each `.check_existence` call, instead of once for each iteration of its loop.  (i also find this to be totally sensible semantically)

connects to #264


jmartin-sul:
rebased on master and re-pushed, travis appears to be happy.  ready for review, i think.

jmartin-sul:
`delegate` was a rubocop suggestion (though i then remembered seeing it before, e.g. we used it in argo a bit).

Issue #179, opened 2017-10-09
(M2C)  verify druid in manifest matches directory tree


SaravShah:
identify the most recent version, and positively identify the druid (read most recent manifestInventory.xml for objectId; match against druidtree path - i.e. does the path /aa/123/bb/1234/aa123bb1234 really contain a Moab object of the druid aa123bb1234? ).

connects #177

jmartin-sul:
this might ultimately belong in the moab-versioning gem, because it's about checking characteristics of a moab directory on disk

jmartin-sul:
this ticket gets closed when the call is in place in preservation catalog.

for the moab-versioning part, spawn a separate ticket and PR in the moab-versioning gem.  this ticket will require upgrading to the version of moab-versioning that includes the desired call, and making the call in PC.

tallenaz:
Related to sul-dlss/moab-versioning#71

tallenaz:
`moab-versioning` v2.5.0, which includes D2D check, is released. Once https://github.com/sul-dlss/preservation_catalog/pull/291 is merged, we can make the call in PC.

Issue #178, opened 2017-10-09
(MV)  verify each manifest subdir contains manifestInventory.xml and signatureCatalog.xml


SaravShah:
Each /version/manifest subdir contains at least a manifestInventory.xml file and a signatureCatalog.xml file (without these, it is impossible to perform checksum validation).

connects #177

ndushay:
It's possible moab-versioning  moab/storage_object_version #verify_manifest_inventory would help with this???

jmartin-sul:
this might ultimately belong in the moab-versioning gem, because it's about checking characteristics of a moab directory on disk

Issue #177, opened 2017-10-09
(MV) #create_with_validation


SaravShah:
That it does not contain any ‘nested’ Moabs (a complete Moab object dir placed inside the dir of another Moab, a result of bad initial file copy/move).

SaravShah:
Something to check for ```bj/102/hs/9687/bj102hs9687/v0004/bj/102/hs/9687/ba12fhs1447/```

SaravShah:
the easiest check is to just confirm there are only /manifests and /data directories under each version dir, and that there are only /content and /metadata dirs under /data. No other directories anywhere else.

jmartin-sul:
this might ultimately belong in the moab-versioning gem, because it's about checking characteristics of a moab directory on disk

jmartin-sul:
this ticket gets closed when the call is in place in preservation catalog.

for the moab-versioning part, spawn a separate ticket and PR in the moab-versioning gem.  this ticket will require upgrading to the version of moab-versioning that includes the desired call, and making the call in PC.

tallenaz:
Related to sul-dlss/moab-versioning/issues/63

ndushay:
See #279 -- please put this "hook up MV/D2D validation" code into pohandler.create_with_validation

PR #298, opened 2017-11-08
 Add #create_with_validation to PO_handler

SaravShah:
~~WIP because Stanford::StorageObjectValidator fails fast on invalid moabs and it shouldn't~~ - Fixed ✔️ 

connects #177 
connects to #179
connects to #178

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -56,6 +56,29 @@ class PreservedObjectHandler
     @endpoint = endpoint
   end
 
+  def create_with_validation
+    results = []
+    if invalid?
+      results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
+    elsif PreservedObject.exists?(druid: druid)
+      results << result_hash(OBJECT_ALREADY_EXISTS, 'PreservedObject')
+    else
+      dt = DruidTools::Druid.new(druid)
+      object_dir = "#{storage_location}/#{dt.tree.join('/')}"
+      moab = Moab::StorageObject.new(druid, object_dir)
+      object_validator = Stanford::StorageObjectValidator.new(moab)
+      validation_errors = object_validator.validation_errors
+      if validation_errors.empty?
+        results.concat(create_db_objects(Status.default_status))
+      else
+        results.concat(create_db_objects(Status.invalid))
+      end
+    end
+
+    log_results(results)
+    results
+  end
+
   def create
     results = []
     if invalid?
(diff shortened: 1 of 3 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -56,26 +56,35 @@ def initialize(druid, incoming_version, incoming_size, endpoint)
...
+    if invalid?
+      results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
+    else
+      dt = DruidTools::Druid.new(druid)
+      object_dir = "#{storage_location}/#{dt.tree.join('/')}"
+      moab = Moab::StorageObject.new(druid, object_dir)
+      object_validator = Stanford::StorageObjectValidator.new(moab)
+      validation_errors = object_validator.validation_errors
+      if validation_errors.empty?
+        create
```
jmartin-sul: i'm not sure i understand the call to `create` here, as opposed to a call to `results.concat(create_db_objects(Status.default_status))`... if it's to get the check for an extant `PreservedObject`, wouldn't you want to do that in either case?  entirely possible i'm misunderstanding things, but i'd have expected the two branches of this if to parallel each other a bit more closely.

spec/services/preserved_object_handler_create_spec.rb
```diff
@@ -87,4 +86,126 @@
...
+            it "exception's message" do
+              expect(results).to include(a_hash_including(result_code => a_string_matching('foo')))
+            end
+          end
+        end
+      end
+    end
+
+    context 'returns' do
+      let!(:result) { po_handler.create }
```
jmartin-sul: was the intention to test `create` here?  i'd have expected `.create_with_validation`?  (since previously written tests should cover the refactoring of `.create`, right?)

app/services/preserved_object_handler.rb
jmartin-sul: also, i see where result codes and log statements are tested for `create_with_validation` when it encounters validation errors, but not when there are no errors and the creation happens cleanly.

spec/services/preserved_object_handler_create_spec.rb
```diff
@@ -87,4 +86,126 @@
       end
     end
   end
+
+  describe '#create_with_validation' do
+    let(:valid_druid) { 'bp628nk4868' }
+    let(:storage_dir) { 'spec/fixtures/storage_root02/moab_storage_trunk' }
+
+    it_behaves_like 'attributes validated', :create
```
jmartin-sul: is this supposed to be `:create_with_validation`?

spec/services/preserved_object_handler_create_spec.rb
SaravShah: oops ya copy and paste left over lol

app/services/preserved_object_handler.rb
SaravShah: good point...:)

SaravShah: Also there was a test ``` 'calls #create when there are no validation errors' ```

SaravShah: ^ but i'm changing it to ` results.concat(create_db_objects(Status.default_status)) `

PR #299, opened 2017-11-09
add a description to the seed_catalog rake task

jmartin-sul:
so it shows up when you do a `rake -T` from console


jmartin-sul:
@tallenaz: i wasn't trying to modify the task itself, but good call, that would be useful, i filed an issue:  https://github.com/sul-dlss/preservation_catalog/issues/300

PR #306, opened 2017-11-10
Add profiler and version of seed_from_disk using it

### LINKED ISSUE #290: (PC) what should we do for performance testing?
the easiest thing would probably be to just use newrelic.

other options, should we want more detail or more target detail than newrelic provides:
* with argo reindexing, when we needed to tune performance we added some configurable performance logging on its indexing code:
  * https://github.com/sul-dlss/argo/blob/master/config/settings.yml#L42
  * https://github.com/sul-dlss/argo/blob/master/config/settings.yml#L55-L69
  * https://github.com/sul-dlss/argo/blob/master/app/jobs/local_indexing_job.rb
  * https://github.com/sul-dlss/argo/blob/master/lib/profiler.rb
  * https://github.com/sul-dlss/argo/blob/master/lib/indexer.rb
* a quick and dirty performance testing branch that ports some of our old PC perf testing code, without the intention of ever using that branch in prod
* there will likely be some basic DB stats in the logs per operation, but i think we’d need to do some log parsing for them to be useful at all, which is also work (or maybe someone in ops who’s fluent in whatever log analysis tool we have could make sense of them for us, or show us how to use that tool ourselves)

we're not sure how much of new relic’s magic `just works` for rake tasks (which is what the seeding is called from, as opposed to being called from a controller, which is what newrelic seemed to specialize in profiling, last i used newrelic).

tallenaz:
- lifted `ruby-prof` out of development gem group, to deploy it
- auto-loaded paths in lib, which made it possible to invoke the profiler in a console
- added a profiler that outputs `FlatPrinterWithLineNumbers` (other output options are available)
- added a profiled version of `seed_from_disk` and specs for it
- should probably add specs for the Profiler class

closes #290

### DIFF
--- a/lib/profiler.rb
+++ b/lib/profiler.rb
@@ -1,0 +1,23 @@
+# adapted from argo
+# takes a block of code, starts the profiler, runs the code, then
+# stops the profiler and returns the results of the profiling.
+# example usage:
+#  profiler = Profiler.new
+#  profiler.prof { MoabToCatalog.seed_catalog(storage_dir) }
+#  profiler.print_results_flat(out_file_id)
+class Profiler
+
+  attr_accessor :results
+
+  def prof
+    RubyProf.start
+    yield
+    @results = RubyProf.stop
+  end
+
+  def print_results_flat(out_file_id)
+    File.open "log/#{out_file_id}#{Time.current.localtime.strftime('%FT%T')}-flat.txt", 'w' do |file|
+      RubyProf::FlatPrinterWithLineNumbers.new(@results).print(file)
+    end
+  end
+end
--- a/spec/lib/audit/moab_to_catalog_spec.rb
+++ b/spec/lib/audit/moab_to_catalog_spec.rb
@@ -41,6 +41,20 @@ RSpec.describe MoabToCatalog do
     end
   end
 
+  describe "#seed_from_disk_with_profiling" do
+    let(:m2c) { described_class.new }
+
+    it "spins up a profiler, calling profiling and printing methods on it" do
+      mock_profiler = instance_double(Profiler)
+
+      expect(Profiler).to receive(:new).and_return(mock_profiler)
+      expect(mock_profiler).to receive(:prof)
+      expect(mock_profiler).to receive(:print_results_flat)
+
+      m2c.seed_from_disk_with_profiling
+    end
+  end
+
   describe ".check_existence" do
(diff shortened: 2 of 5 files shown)


config/application.rb
```diff
@@ -13,6 +13,7 @@ module PreservationCatalog
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 5.1
+    config.autoload_paths += %W[#{config.root}/lib #{config.root}/lib/audit]
```
ndushay: if this autoload stays here, it can probably be removed from the Rakefile

tallenaz:
- Removed the autoload, as it was just a convenience for developing
- Added specs for the Profiler class from pairing with @jmartin-sul

lib/profiler.rb
```diff
@@ -0,0 +1,23 @@
...
+  attr_accessor :results
+
+  def prof
+    RubyProf.start
+    yield
+    @results = RubyProf.stop
+  end
+
+  def print_results_flat(out_file_id)
+    File.open "log/#{out_file_id}-flat.txt", 'w' do |file|
```
jmartin-sul: what about a timestamp in the filename, just so profiling runs don't step on each other?  could also append to the file, but i think i'd have a slight pref for timestamp, because then you wouldn't have one indefinitely growing file.

spec/lib/audit/moab_to_catalog_spec.rb
```diff
@@ -41,6 +41,20 @@
     end
   end
 
+  describe "#seed_from_disk_with_profiling" do
+    let(:m2c) { described_class.new }
+
+    it "logs what it's about to do, uses the profiler, and prints profiling results" do
```
jmartin-sul: seems like the logging part got left off?  or i'm just missing it.  but if it did get left off, it'd be nice to either add it, or fix the test description.  logging would be nice, i'd be in favor of adding it, but also doesn't feel super-necessary, so happy to have it omitted if it doesn't feel worth the effort.

tallenaz:
- added a timestamp to the log generated by `print_results_flat`
- changed the test descriptions to omit suggestions that we're checking that the profiler actually logs, agreeing that a test like that would be nice to have, and it doesn't seem super-necessary to spec out.

tallenaz:
I'll add a comment that the expectation involving timestamps on the profiler log might need to be loosened.

PR #307, opened 2017-11-10
moab_to_catalog.seed calls create_with_validation

### LINKED ISSUE #279: (SPC) seeding the catalog method calls create_with_validation
MV/D2D should be called before create when seeding the catalog, so seed method in lib/audit/moab_to_catalog should call pohandler.create_with_validation()

ndushay:
Resolves #279

### DIFF
--- a/spec/lib/audit/moab_to_catalog_spec.rb
+++ b/spec/lib/audit/moab_to_catalog_spec.rb
@@ -193,10 +193,10 @@ RSpec.describe MoabToCatalog do
           ).and_return(po_handler)
         end
       end
-      it "call #create if object does not exist" do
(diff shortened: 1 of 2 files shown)


spec/lib/audit/moab_to_catalog_spec.rb
```diff
@@ -196,7 +196,7 @@
       it "call #create if object does not exist" do
```
jmartin-sul: super nitpicky, but would you mind updating this to `#create_with_validation`?

ndushay: I would be happy to - missed that, and also strongly prefer it to match the actual test code.

PR #308, opened 2017-11-10
add timestamps to log and puts statements for seeding and M2C

### LINKED ISSUE #300: add timestamp to puts that prints when seed_catalog rake task finishes
see https://github.com/sul-dlss/preservation_catalog/blob/master/Rakefile#L23

ndushay:
Resolves #300

I didn't feel it was worth it to write tests to prove that these messages were in the logs or written to to STDOUT ...

### DIFF
--- a/lib/audit/moab_to_catalog.rb
+++ b/lib/audit/moab_to_catalog.rb
@@ -39,14 +39,26 @@ class MoabToCatalog
   # Shameless green. In order to run several seed "jobs" in parallel, we would have to refactor.
   def seed_from_disk
     Settings.moab.storage_roots.each do |storage_root|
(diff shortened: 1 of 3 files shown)


lib/audit/moab_to_catalog.rb
```diff
@@ -39,14 +39,26 @@ def self.seed_catalog(storage_dir)
   # Shameless green. In order to run several seed "jobs" in parallel, we would have to refactor.
   def seed_from_disk
     Settings.moab.storage_roots.each do |storage_root|
+      start_msg = "#{Time.now.utc.iso8601} Seeding starting for #{storage_root}"
```
ndushay: it was a tossup whether to put this here, or in .seed_catalog.   Happy to hear a good reason to choose one vs. the other.

Issue #288, opened 2017-11-07
(PC) (UC)  pohandler.create (w, w/o validation) timestamps to update


ndushay:
create_with_validation:

PC:
  last_checked_on_storage 
  last_audited

create  (no validation)

PC:
  null for above timestamps

PR #309, opened 2017-11-10
Update pohandler.create/validations timestamps

SaravShah:
- Added a flag in the `create_db_objects()` method to run the timestamp update for ` create_with_validation` and not for `create` 
- we want to update PC.last_checked_on_storage and PC.last_audited for `create_with_validation`

- I figured that this was a good idea because I have access to the `PreservedCopy` in the `create_db_objects()` method and wouldn't have to use a `.find_by` call. 

connects #288

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -69,9 +69,9 @@ class PreservedObjectHandler
       object_validator = Stanford::StorageObjectValidator.new(moab)
       validation_errors = object_validator.validation_errors
       if validation_errors.empty?
-        results.concat(create_db_objects(Status.default_status))
+        results.concat(create_db_objects(Status.default_status, true))
       else
-        results.concat(create_db_objects(Status.invalid))
+        results.concat(create_db_objects(Status.invalid, true))
       end
     end
 
@@ -133,18 +133,25 @@ class PreservedObjectHandler
 
   private
 
-  def create_db_objects(status)
+  def create_db_objects(status, validate=false)
     results = []
     pp_default = PreservationPolicy.default_preservation_policy
     create_results = with_active_record_rescue do
       po = PreservedObject.create!(druid: druid,
                                    current_version: incoming_version,
                                    preservation_policy: pp_default)
-      PreservedCopy.create!(preserved_object: po,
-                            version: incoming_version,
-                            size: incoming_size,
-                            endpoint: endpoint,
-                            status: status)
+      pc = PreservedCopy.create!(preserved_object: po,
+                                 version: incoming_version,
+                                 size: incoming_size,
+                                 endpoint: endpoint,
+                                 status: status)
+      if validate
+        t = Time.current
+        # Returns the value of time as an integer number of seconds since the Epoch.
+        pc.last_audited = t.to_i
+        pc.last_checked_on_storage = t
+        pc.save
+      end
       results << result_hash(CREATED_NEW_OBJECT)
     end
     results.concat(create_results)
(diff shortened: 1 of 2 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -133,18 +133,25 @@ def update_version
 
   private
 
-  def create_db_objects(status)
+  def create_db_objects(status, validate=false)
```
ndushay: would it be more descriptive to say "validated" instead of "validate"?  It looks like it refers to whether or not validation has been performed already.

spec/services/preserved_object_handler_create_spec.rb
```diff
@@ -93,6 +93,27 @@
 
     it_behaves_like 'attributes validated', :create_with_validation
 
+    context 'updates timestamp' do
```
ndushay: would it be more correct to say "updates validation timestamps"?

app/services/preserved_object_handler.rb
```diff
@@ -133,18 +133,25 @@ def update_version
...
+                                 size: incoming_size,
+                                 endpoint: endpoint,
+                                 status: status)
+      if validate
+        t = Time.current
+        # Returns the value of time as an integer number of seconds since the Epoch.
+        pc.last_audited = t.to_i
+        pc.last_checked_on_storage = t
+        pc.save
+      end
```
jmartin-sul: what about something like:
```ruby
pc_attrs = {
  preserved_object: po,
  version: incoming_version,
  size: incoming_size,
  endpoint: endpoint,
  status: status
}

if validate
  t = Time.current
  # Returns the value of time as an integer number of seconds since the Epoch.
  pc_attrs.merge({
    last_audited: t.to_i,
    last_checked_on_storage: t
  })
end

```

and then you save yourself a possible extra trip to the database, since there's no optional update, just a create all at once, with some optional attributes.

app/services/preserved_object_handler.rb
jmartin-sul: agreed w/ @ndushay on verb tense

SaravShah:
In the `preserved_object_handler_create_spec.rb` I used `an_instance_of(Integer)` and `an_instance_of(ActiveSupport::TimeWithZone)` instead of using `Time.current.to_i` and `Time.current` respectively. Is everyone ok with the way I did this or should i do something that will account for the 1 second difference between code n test.

app/services/preserved_object_handler.rb
```diff
@@ -133,18 +133,28 @@ def update_version
...
+        version: incoming_version,
+        size: incoming_size,
+        endpoint: endpoint,
+        status: status
+      }
+
+      if validates
+        t = Time.current
+        # Returns the value of time as an integer number of seconds since the Epoch.
+        pc_attrs[:last_audited] = t.to_i
```
ndushay: I'm guessing `last_audited` is meant to be the integer format for easy sorting and comparisons and so on.

app/services/preserved_object_handler.rb
```diff
@@ -133,18 +133,28 @@ def update_version
...
+        pc_attrs[:last_checked_on_storage] = t
```
ndushay: for the datetime fields:  going the same route as ActiveRecord uses for created_at and updated_at may make sense.

There's a discussion brewing about UTC vs local time  for logging timestamps;  perhaps we should also agree on these datetime fields.   Not a bad storytime topic.  I would say let's do this for now and make a separate ticket for ensuring all timestamp fields are written to consistently?

Issue #285, opened 2017-11-07
(M2C) (UC) update_version (w, w/o validation) calls create_with_validation


ndushay:
So regardless of whether  originating caller may have done:

update_version_with_validation()
or
update_version()

we want "if object not in catalog" to call create_with_validation()

we're pretty sure, anyway.
until we're not.

let's wait for complaints.

ndushay:
the assumption for this ticket was that update_version actually called *create* when the object wasn't there.  But it seems the code says that if the catalog didn't have the object, then we should NOT create it.

Perhaps we decided that update version SHOULD error if the object didn't already exist -- I  forgot.  We should perhaps revisit which behavior we actually want update_version to do:  to flat out fail if the object doesn't exist, or to create it.

I'm excellent at forgetting decisions.  We should take notes at storytime.

ndushay:
From storytime notes from Nov 7, 2017:

```
dev question:   Will #update_version ALWAYS call #create if object does not exist? If not, please give specific example. 

Julian:  OK, so this is interesting. Presumably you’ve got some logic cooking here that tries an update, realizes that there’s nothing to update and does a create instead. If that’s the case, this would be a good place to do a MV/D2D before create. And also to alert like crazy.
``

ndushay:
Closing this ticket;  our current plan is to simply fail if update_version or update_version_after_validation  finds that the PO (or Moab PC) doesn't already exist.

Issue #287, opened 2017-11-07
(M2C) (UC) pohandler.update_version (w, w/o validation)  timestamps to update


ndushay:
(See also #278 and #286)

update_version (with or without validation) updates timestamps:

#### happy path:  version in PC updated:
PC:
- version
- updated_at (via ActiveRecord)
- if update_version_with_validation:
    - last_audited field
    - last_checked_on_storage field

PO:
- current_version
- updated_at  (via ActiveRecord)

#### if incoming version <= catalog version
ERROR (bad args) to caller
- this is not a status change but a "hey caller you have bad data" error:
    - if equal version to incoming, then caller has a problem version
    - if catalog higher than incoming, then caller also has a problem version

PC
- if update_version_with_validation:
   - last_audited field
   - last_checked_on_storage field
     
PO
- do nothing

#### if object does not exist:  it's an error.

Issue #283, opened 2017-11-07
(M2C) (UC) (PC)  update_version_after_validation method in pohandler


ndushay:
pohandler needs a public method to call MV/D2D before calling (plain) update_version.

check_existence will use update_version_with_validation.  

Ingest robot for new version will have a feature flag to use update_version_with_validation or just plain update_version

ndushay:
blocked by #279 and #177

ndushay:
Discussed at storytime.  Decisions:
- update PC version (and size) to match whatever is on disk (= incoming, no?)
- PC status set to "invalid_moab"
- update PO current_version as needed

```ruby
        # PC
        #  last_audited_field
        #  last_checked_on_storage_field
        #  status invalid
        #  version and size field
        # PO:  current_version if needed;  timestamps if needed (current_version)
```

PR #314, opened 2017-11-14
[WIP] Call create_with_validation in with_active_record_rescue with tests

tallenaz:
- Had to changed the RSpec Shared Example po_handle to a valid druid that could be created w/ create_with_validation otherwise tests weren't passing.

connects #285

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -112,28 +112,29 @@ class PreservedObjectHandler
     results
   end
 
-
-  # If validation doesn't pass then create with Status.invalid - so need another flag to tell create_db_object - 
+  # If validation doesn't pass then create with Status.invalid - so need another flag to tell create_db_object -
   # can jiust call that right?
-  
+
   def update_version_with_validation
     dt = DruidTools::Druid.new(druid)
     object_dir = "#{storage_location}/#{dt.tree.join('/')}"
     moab = Moab::StorageObject.new(druid, object_dir)
     object_validator = Stanford::StorageObjectValidator.new(moab)
     validation_errors = object_validator.validation_errors
-    results = []
-    if invalid?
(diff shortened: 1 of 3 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -112,13 +112,42 @@ def confirm_version
     results
   end
 
+  # If validation doesn't pass then create with Status.invalid - so need another flag to tell create_db_object -
+  # can jiust call that right?
+
+  def update_version_with_validation
```
ndushay: Sarav - we need to confer about this because this is EXACTLY what I was working on.  I thought ticket #285 was a very simple "we discover the object doesn't yet exist, so we called `create`" and substitute `create_after_validation` ... but this is essentially doing ticket #283, which I had started.

Maybe my assumption about "update_version" actually calling create when the druid wasn't in the catalog was mistaken, which would have made #285 very confusing ...

ndushay:
I  was thinking update_version looked something remotely like this:

```
  def update_version
    results = []
    if invalid?
      results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
    else
      Rails.logger.debug "update_version #{druid} called and druid in Catalog"
      if catalog has object
        upd_results = with_active_record_rescue do
          if endpoint.endpoint_type.endpoint_class == 'online'
            results.concat update_online_version
          elsif endpoint.endpoint_type.endpoint_class == 'archive'
            # TODO: create a new PC record for the new archived version on the endpoint
          end
        end
        results.concat(upd_results)
      else
        create_after_validation(object)
      end
    end

    log_results(results)
    results
  end
```

ndushay:
closing this;  ticket #285 was spurious (we fail if the object to be updated doesn't already exist), and #287 is the better ticket for what this branch was starting to do

PR #315, opened 2017-11-14
Document seeding the catalog

### LINKED ISSUE #275: (SPC) Document seeding the catalog
including how to invoke the rake task, use feature flag for performance data, etc.

tallenaz:
Adds an argument to rake seed_catalog
  to distinguish profiling runs

  Fixes #275

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,26 @@ For more info on postgres commands, see https://www.postgresql.org/docs/
 
 ## Usage Instructions
 
-TODO: write these
+### Seed the catalog
+
+Seeding the catalog presumes an empty or nearly empty database -- otherwise running the seed task will throw `druid NOT expected to exist in catalog but was found` errors for each found object. You can monitor the progress of the seed task by tailing `log/production.log`. The seed task can take a while -- check [the repo wiki for stats](https://github.com/sul-dlss/preservation_catalog/wiki/Stats) on the timing of past runs, and you can expect profiling runs will add overhead. The following rake tasks should be run from the root directory of the project, with whatever `RAILS_ENVIRONMENT` is appropriate. Because the task can take days when run over each storage root, consider running it in a [screen session](http://thingsilearned.com/2009/05/26/gnu-screen-super-basic-tutorial/).
+
+Without profiling:
+```ruby
+RAILS_ENVIRONMENT=production bundle exec rake seed_catalog
+```
+
+With profiling:
+```ruby
+RAILS_ENVIRONMENT=production bundle exec rake seed_catalog[profile]
+```
+As an alternative to `screen`, you can also run the task, with or without profiling, in the background under `nohup`. For example:
+
+```ruby
+RAILS_ENVIRONMENT=production bundle exec nohup rake seed_catalog &
+```
+
+this will generate a log at, for example, `log/seed_from_disk2017-11-13T13:57:01.log`
 
 ## Development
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,7 +45,18 @@ For more info on postgres commands, see https://www.postgresql.org/docs/
...
+### Seed the catalog
+
+Without profiling:
+```ruby
+RAILS_ENVIRONMENT=production bundle exec rake seed_catalog
+```
+
+With profiling:
+```ruby
+RAILS_ENVIRONMENT=production bundle exec rake seed_catalog[profile]
```
ndushay: is this actually what we type?  `seed_catalog[profile]`  ?  or do we do some string substitution for "profile" ?

Rakefile
```diff
@@ -16,9 +16,17 @@ end
...
+    p "Usage: rake seed_catalog || rake seed_catalog[profile]"
+    exit
+  end
   m2c = MoabToCatalog.new
   puts "#{Time.now.utc.iso8601} Seeding the database from all storage roots..."
-  m2c.seed_from_disk
+  if args[:profile] == 'profile'
+    m2c.seed_from_disk
+  elsif args[:profile].nil?
+    m2c.seed_from_disk_with_profiling
```
ndushay: uh - do you have these method calls backwards?  When arg is 'profile' then ... call method that does profiling?

README.md
tallenaz: Yes, you literally type `seed_catalog[profile]`, no substitution.

Rakefile
tallenaz: D'oh!

tallenaz:
Thanks for the questions @ndushay -- I'll add more documentation addressing them.

Rakefile
```diff
@@ -16,9 +16,17 @@ end
...
 require_relative 'lib/audit/moab_to_catalog'
 desc 'populate the catalog with the contents of the online storage roots'
-task seed_catalog: :environment do
+task :seed_catalog, [:profile] => [:environment] do |_t, args|
+  unless args[:profile] == 'profile' || args[:profile].nil?
+    p "Usage: rake seed_catalog || rake seed_catalog[profile]"
+    exit
+  end
   m2c = MoabToCatalog.new
   puts "#{Time.now.utc.iso8601} Seeding the database from all storage roots..."
```
jmartin-sul: might be nice to indicate in the `puts` msg that the run is being done with profiling, but not a big deal, and happy to ticket that as follow up if you do agree it'd be a good idea.

README.md
```diff
@@ -45,7 +45,20 @@ For more info on postgres commands, see https://www.postgresql.org/docs/
...
+```
```
jmartin-sul: since this could be a job that potentially runs for days, if running it over all prod storage roots, i'd maybe make two other practical suggestions for users:
* run it from a screen session (random cheatsheet/tutoral that i've found helpful for screen:  http://thingsilearned.com/2009/05/26/gnu-screen-super-basic-tutorial/)
* run it in the background under nohup:  `RAILS_ENV=production nohup bundle exec rake seed_catalog &`

Rakefile
ndushay: I think indicating profiling or not in the puts message is a great idea.

README.md
ndushay: I agree w John.

Rakefile
tallenaz: Thanks @jmartin-sul I'm agreed they're good suggestions, and I took a stab at both.

Issue #281, opened 2017-11-07
(PC) create a README.md under db/


jmartin-sul:
some concise guidelines to database development, e.g. use transactions for groups of operations that should succeed or fail atomically, and follow the AR guidelines for exception handling with transactions (see #203 for AR and PG docs to link to).

jmartin-sul:
notes here have links to useful supplemental docs:  https://github.com/sul-dlss/preservation_catalog/issues/203#issuecomment-340633610

PR #322, opened 2017-11-15
transactionality in POH: rename with_active_record_rescue to with_active_record_transaction_and_rescue, have it wrap its yield in a transaction

### LINKED ISSUE #203: (PC) use explicit DB transactions (at least for PO and PC object updates in PreservedObjectHandler)
what active record calls need to be grouped into transactions?  e.g., the calls in `PreservedObjectHandler#create`?  `PreservedObjectHandler#update`?  probably both of those... anywhere else?

jmartin-sul:
since that method is already wrapping high-level groups of DB operations which go together to form what should be a single atomic update

test transactions by forcing one part of the database updating to fail, after the prior step has succeeded, and check that the prior step was rolled back.

* the first commit is just a bit of log message clean up that i couldn't resist
* the second adds transactionality to the initial database seeding of endpoints etc
* the third is a bit of refactoring that sets up
* the fourth, which just adds transactionality to the method that already wraps high-level sets of DB operations

my plan is to put up a follow-on PR that maybe fleshes out the tests for this a bit more (#324), as well as adding a `db/README.md` (#281) which gives a bit of basic guidance about when and how to wrap sets of operations in transactions, etc.

closes #203

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -91,7 +91,7 @@ class PreservedObjectHandler
     if invalid?
       results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
     else
-      Rails.logger.debug "confirm_version #{druid} called and object exists"
+      Rails.logger.debug "confirm_version #{druid} called"
       confirm_results = with_active_record_rescue do
         po_db_object = PreservedObject.find_by!(druid: druid)
         results.concat(confirm_version_on_db_object(po_db_object, :current_version))
@@ -110,7 +110,7 @@ class PreservedObjectHandler
     if invalid?
       results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
     else
-      Rails.logger.debug "update_version #{druid} called and druid in Catalog"
+      Rails.logger.debug "update_version #{druid} called"
       upd_results = with_active_record_rescue do
         if endpoint.endpoint_type.endpoint_class == 'online'
           results.concat update_online_version
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -92,13 +92,7 @@ class PreservedObjectHandler
       results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
     else
       Rails.logger.debug "confirm_version #{druid} called"
-      confirm_results = with_active_record_rescue do
-        po_db_object = PreservedObject.find_by!(druid: druid)
-        results.concat(confirm_version_on_db_object(po_db_object, :current_version))
-        pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint)
-        results.concat(confirm_version_on_db_object(pc_db_object, :version))
-      end
-      results.concat(confirm_results)
+      results.concat(confirm_version_in_catalog)
     end
 
     log_results(results)
@@ -111,13 +105,10 @@ class PreservedObjectHandler
       results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
     else
       Rails.logger.debug "update_version #{druid} called"
-      upd_results = with_active_record_rescue do
-        if endpoint.endpoint_type.endpoint_class == 'online'
-          results.concat update_online_version
-        elsif endpoint.endpoint_type.endpoint_class == 'archive'
-        end
+      if endpoint.endpoint_type.endpoint_class == 'online'
+        results.concat(update_online_version)
+      elsif endpoint.endpoint_type.endpoint_class == 'archive'
       end
-      results.concat(upd_results)
     end
 
     log_results(results)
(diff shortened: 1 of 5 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -162,24 +153,37 @@ def create_db_objects(status, validated=false)
...
+        results.concat(version_comparison_results(pres_copy, :version))
+        results.concat(version_comparison_results(pres_object, :current_version))
+        # FIXME: TODO: should it update existence check timestamps/status?
+      end
     end
-    results
+    results.concat(upd_results)
+  end
+
+  def confirm_version_on_db_objects
```
jmartin-sul: i am not very happy with this name, since it's so similar to `confirm_version_on_db_object`, which it calls.  i'm totally open to suggestions for better names.

ndushay: how about `confirm_version_in_db` or `confirm_version_in_catalog` ?

app/services/preserved_object_handler.rb
```diff
@@ -91,14 +91,8 @@ def confirm_version
...
     else
-      Rails.logger.debug "confirm_version #{druid} called and object exists"
-      confirm_results = with_active_record_rescue do
-        po_db_object = PreservedObject.find_by!(druid: druid)
-        results.concat(confirm_version_on_db_object(po_db_object, :current_version))
-        pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint)
-        results.concat(confirm_version_on_db_object(pc_db_object, :version))
-      end
-      results.concat(confirm_results)
+      Rails.logger.debug "confirm_version #{druid} called"
```
ndushay: why remove the specificity of the debug logging messages when they are still in the same different conditions?

app/services/preserved_object_handler.rb
```diff
@@ -110,14 +104,11 @@ def update_version
     if invalid?
       results << result_hash(INVALID_ARGUMENTS, errors.full_messages)
     else
-      Rails.logger.debug "update_version #{druid} called and druid in Catalog"
-      upd_results = with_active_record_rescue do
-        if endpoint.endpoint_type.endpoint_class == 'online'
-          results.concat update_online_version
-        elsif endpoint.endpoint_type.endpoint_class == 'archive'
-        end
+      Rails.logger.debug "update_version #{druid} called"
```
ndushay: why remove the specificity of the debug logging messages when they are still in the same different conditions?

spec/services/preserved_object_handler_spec.rb
```diff
@@ -335,4 +335,42 @@
...
     it_behaves_like 'PreservedCopy does not exist', :confirm_version
   end
+
+  describe '#with_active_record_transaction_and_rescue' do
+    it '#create rolls back the preserved object if the preserved copy cannot be created' do
+      # so that pres copy creation fails with a constraint violation, since status can't be nil
+      allow(Status).to receive(:default_status).and_return(nil)
+      po_handler.create
+      expect(PreservedObject.where(druid: druid)).not_to exist
+    end
```
ndushay: should this "create" test go in the preserved_object_handler_create_spec  file?

spec/services/preserved_object_handler_spec.rb
```diff
@@ -335,4 +335,42 @@
 
     it_behaves_like 'PreservedCopy does not exist', :confirm_version
   end
+
+  describe '#with_active_record_transaction_and_rescue' do
```
ndushay: should there be a test for each of:

- create
- create_after_validation
- update_version
- update_version_after_validation (doesn't exist yet)
- confirm_version

proving that they roll back when they are supposed to?  Or maybe proving that they use the method?

app/services/preserved_object_handler.rb
jmartin-sul: at the time of the message, it's not clear that the object actually exists, since we haven't checked for that.  though i believe we were checking for existence when that log message was initially added.

spec/services/preserved_object_handler_spec.rb
jmartin-sul: i would actually like to do that, though i was having a bit of trouble coming up with a test for `update_version` that didn't feel too contrived.  but certainly possible.  at worst, we could force the pres obj update call to blow up, and confirm that the pres copy changes got rolled back.

i was also thinking it might be nice to test the private method specifically/explicitly, but maybe that's overkill?  see #324.

if we did add tests for the methods you suggested, would you prefer to do that as part of this PR, or as part of a follow on ticket (e.g. slightly change or expand the scope of #324)?

ndushay: I think you could either expand the scope of #324 or make a new ticket.

jmartin-sul: expanded scope of #324

jmartin-sul:
@ndushay, i think i've addressed all of your feedback (implemented the suggestions for method naming and test re-location, ticketed the suggested test improvements).  ready for another round of evaluation.

PR #329, opened 2017-11-16
pohandler_update_vers_spec: beef up specs and code pending specs

ndushay:
- implement the pending specs about update_version timestamp and status changes

- beef up the specs a bit in preparation for work on #283 

Connects to #287 -- it has the specs for timestamp and status updates expected for plain update_version

Connects to #283 -- it lays groundwork for the tests we will be changing for update_version_with_validation

### DIFF
--- a/spec/services/preserved_object_handler_update_version_spec.rb
+++ b/spec/services/preserved_object_handler_update_version_spec.rb
@@ -35,19 +35,24 @@ RSpec.describe PreservedObjectHandler do
         let(:updated_pc_db_msg) { "#{exp_msg_prefix} PreservedCopy db object updated" }
         let(:updated_po_db_msg) { "#{exp_msg_prefix} PreservedObject db object updated" }
 
-        it "updates entries with incoming version" do
+        it "updates PreservedCopy with incoming version" do
           expect(pc.version).to eq 2
+          po_handler.update_version
+          expect(pc.reload.version).to be > 2
+          expect(pc.reload.version).to eq incoming_version
+        end
+        it "updates PreservedObject with incoming version" do
           expect(po.current_version).to eq 2
           po_handler.update_version
-          expect(pc.reload.version).to eq incoming_version
+          expect(po.reload.current_version).to be > 2
           expect(po.reload.current_version).to eq incoming_version
         end
-        it 'updates entries with size if included' do
+        it 'updates PreservedCopy size if supplied' do
           expect(pc.size).to eq 1
           po_handler.update_version
           expect(pc.reload.size).to eq incoming_size
         end
-        it 'retains old size if incoming size is nil' do
+        it 'PreservedCopy retains old size if incoming size is nil' do
           expect(pc.size).to eq 1
           po_handler = described_class.new(druid, incoming_version, nil, ep)
           po_handler.update_version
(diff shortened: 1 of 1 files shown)


spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -35,19 +35,24 @@
...
           expect(pc.version).to eq 2
-          expect(po.current_version).to eq 2
           po_handler.update_version
+          expect(pc.reload.version).to be > 2
           expect(pc.reload.version).to eq incoming_version
+        end
+        it "updates PreservedObject with incoming version" do
+          expect(po.current_version).to eq 2
+          po_handler.update_version
+          expect(po.reload.current_version).to be > 2
```
jmartin-sul: ditto above comment about this feeling unnecessary in light of the following line and me not feeling strongly about changing it

spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -101,30 +116,41 @@
...
           po_handler.update_version
           expect(pc.reload.size).to eq 1
         end
-        it 'updates status of PreservedCopy to "ok"' do
-          skip("should it update status of PreservedCopy?")
-          expect(pc.status).to eq Status.unexpected_version
+        it 'does not update PreservedCopy last_audited field' do
+          orig_timestamp = pc.last_audited
           po_handler.update_version
-          expect(pc.reload.status).to eq Status.ok
```
jmartin-sul: seems like it maybe it should update to `Status.unexpected_version`?  but i'm not sure about that, and also not sure if i'm forgetting whether we already agreed on what the behavior should be in story time (or whether this really is still an open question.

i do see a TODO in the POH app code about whether timestamp/status should get updated in the case of unexpected version, so wouldn't expect a test for that to pass anyway.  but might be worth leaving in the skipped test?  though i guess a test would naturally get added if/when that TODO gets addressed.

https://github.com/sul-dlss/preservation_catalog/blob/master/app/services/preserved_object_handler.rb#L176-L181

spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -35,19 +35,24 @@
         let(:updated_pc_db_msg) { "#{exp_msg_prefix} PreservedCopy db object updated" }
         let(:updated_po_db_msg) { "#{exp_msg_prefix} PreservedObject db object updated" }
 
-        it "updates entries with incoming version" do
+        it "updates PreservedCopy with incoming version" do
           expect(pc.version).to eq 2
-          expect(po.current_version).to eq 2
           po_handler.update_version
+          expect(pc.reload.version).to be > 2
```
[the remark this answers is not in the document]
ndushay: basically, i got confused as to what "incoming_version" was and it seemed easier to just code both of these lines.

spec/services/preserved_object_handler_update_version_spec.rb
ndushay: Based on what is in #287, this was already changed to:

```
        it 'does not update status of PreservedCopy' do
          orig_status = pc.status
          po_handler.update_version
          expect(pc.reload.status).to eq orig_status
        end
```

PR #336, opened 2017-11-18
Log moab validation errors

### LINKED ISSUE #325: (PC) POH should log the MV validation errors it discovers
the moab-versioning validation code returns a nice list of error codes and messages, but it appears that nothing actually writes these error messages to the logs.  we should log that info.

SaravShah:
- Add moab validation errors to the log 
paired w/ @tingulfsen 

Note: Just saw in files changed that there are files from one of the druids I used for a test. Do you think I should remove those contents and add fake contents...so it takes up less space?

closes #325


ndushay:
It would be great to use teeny files, yeah.

app/services/preserved_object_handler.rb
```diff
@@ -118,10 +120,20 @@ def update_version
   private
 
   def moab_validation_errors
+    results = []
     object_dir = "#{storage_location}/#{DruidTools::Druid.new(druid).tree.join('/')}"
     moab = Moab::StorageObject.new(druid, object_dir)
     object_validator = Stanford::StorageObjectValidator.new(moab)
-    object_validator.validation_errors
+    errors = object_validator.validation_errors
+    unless errors.empty?
```
ndushay: this could be `if errors.any?`.   I find unless just slightly harder congnitively.  Okay if you disagree.

SaravShah: Ah yeah. I like that. I first did ``` if errors.size > 0 ``` then rubocop told me to change it to ``` unless errors.empty? ``` , this is a better solution ```` if errors.any? ``` !

SaravShah:
Just pushed up the new changes. Also, the information I used was from an existing moab on storage_root02, which is public. So i'm guessing the moabs on storage_root01 and storage_root02 do not have copyright or other permission errors?

ndushay:
correct:  our fixture objects on storage_roots 1 and 2 are public.

Issue #228, opened 2017-10-25
(PC) consider turning the `statuses` table and `Status` model into a Rails enum on `PreservationCopy`


jmartin-sul:
in our first pass at the schema, the `status` field on `PreservationCopy` was a text field.  @julianmorley was concerned that searching on this column wouldn't be performant, since text searches are more expensive than integer searches, even on indexed columns.  as such, we made the `status` col on pres copy a foreign key to a `statuses` table that holds the status text.  currently, status text is defined in the settings file, from which entries in the `statuses` table are seeded (using a method in the `Status` model).

this is getting unwieldy:  the code is aware of various statuses, and refers to them by name.  it seems unlikely that we'll add more columns to the `statuses` table, and it seems unlikely that users will add entries to the `status` settings without the code being made aware of those entries (unlike, say, endpoints, or endpoint types, or other things that might feel sort of like enums, but where there are multiple columns associated with each entry, and where the application code isn't necessarily aware of each specific entry by name in a way that's hardcoded like we're doing with status stuff).

as such, i'd propose that `status` should actually be a rails `enum`:
http://api.rubyonrails.org/v5.1/classes/ActiveRecord/Enum.html

underneath (in the DB), this will actually be an integer column (as it is right now since it's a FK to an ID in another table), and so it should still obviate julian's performance worry.  active record will do the translation that lets us refer to status by a human readable ruby symbol, as well as providing some other magical rails conveniences.

broadly, the reasons i could see to not make this change would be:
* there are actually more columns that are going to go in the `statuses` table/on the `Status` model.
  * can anyone actually think of any?
* we'll have user defined statuses that the code can work with, but that the code isn't aware of by name.
  * that sounds much more extensible and state machine-y than what we're planning on doing.
* there's some performance problem i'm not considering
  * though i'd actually think a plain int col would be more performant than a FK int col that needs to do a join.

so yeah... i vote that we do what this ticket suggests.  happy to be told i'm missing something.  i don't think the migration will be terrible:  we're still at a point where we can just build up the catalog from scratch again (and i think we're expecting to do that another time or two).  so i don't think we need to write a data migration, just a schema migration.  i think that migration is roughly:
* `preservation_copies.status_id` col gets dropped, replaced by `preservation_copies.status`, which is a plain `integer`, and not a `bigint`.
* `statuses` table gets dropped.
* `Status` model gets removed.
[...]

PR #351, opened 2017-11-27
Store statuses and policies that rarely change.

### LINKED ISSUE #347: (PC) pohandler constant for default PreservationPolicy; memoize endpoint_type lookups ...
Don't do a db table look up for each public method call in PreservedObjectHandler:

`PreservationPolicy.default` (looked up for every new catalog entry) could become a constant.

`endpoint.endpoint_type.endpoint_class` could be memoized in an instance method.

(note:  status lookups part of #228)

From logs, this could save maybe 1-2ms per #create call?

```
D, [2017-11-23T12:15:19.487354 #19661] DEBUG -- :   Status Load (0.8ms)  SELECT  "statuses".* FROM "statuses" WHERE "statuses"."status_text" = $1 LIMIT $2  [["status_text", "ok"], ["LIMIT", 1]]
D, [2017-11-23T12:15:19.488962 #19661] DEBUG -- :   PreservationPolicy Load (0.4ms)  SELECT  "preservation_policies".* FROM "preservation_policies" WHERE "preservation_policies"."preservation_policy_name" = $1 LIMIT $2  [["preservation_policy_name", "default"], ["LIMIT", 1]]
...
D, [2017-11-23T12:15:30.117925 #19661] DEBUG -- :   Status Load (1.1ms)  SELECT  "statuses".* FROM "statuses" WHERE "statuses"."status_text" = $1 LIMIT $2  [["status_text", "ok"], ["LIMIT", 1]]
D, [2017-11-23T12:15:30.119664 #19661] DEBUG -- :   PreservationPolicy Load (0.4ms)  SELECT  "preservation_policies".* FROM "preservation_policies" WHERE "preservation_policies"."preservation_policy_name" = $1 LIMIT $2  [["preservation_policy_name", "default"], ["LIMIT", 1]]
...
D, [2017-11-23T12:26:39.501242 #19661] DEBUG -- :   Status Load (0.7ms)  SELECT  "statuses".* FROM "statuses" WHERE "statuses"."status_text" = $1 LIMIT $2  [["status_text", "ok"], ["LIMIT", 1]]
D, [2017-11-23T12:26:39.503219 #19661] DEBUG -- :   PreservationPolicy Load (0.6ms)  SELECT  "preservation_policies".* FROM "preservation_policies" WHERE "preservation_policies"."preservation_policy_name" = $1 LIMIT $2  [["preservation_policy_name", "default"], ["LIMIT", 1]]

```

tingulfsen:
Closes #347 

I was in doubt about how to implement this. Another obvious way would be to do standard Ruby memoization. If anyone has any opinions on when to choose an approach over another I would be happy to hear about it.

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -37,6 +37,11 @@ class PreservedObjectHandler
     INVALID_MOAB => "Invalid moab, validation errors: %{addl}"
   }.freeze
 
+  DEFAULT_PRESERVATION_POLICY = PreservationPolicy.default_preservation_policy
+  DEFAULT_STATUS = Status.default_status
+  OK_STATUS = Status.ok
+  INVALID_STATUS = Status.invalid
+
   include ActiveModel::Validations
 
   # Note: supplying validations here to allow validation before use, e.g. incoming_version in numeric logic
@@ -65,9 +70,9 @@ class PreservedObjectHandler
     elsif PreservedObject.exists?(druid: druid)
       results << result_hash(OBJECT_ALREADY_EXISTS, 'PreservedObject')
     elsif moab_validation_errors.empty?
-      results.concat(create_db_objects(Status.default_status, true))
+      results.concat(create_db_objects(DEFAULT_STATUS, true))
     else
-      results.concat(create_db_objects(Status.invalid, true))
+      results.concat(create_db_objects(INVALID_STATUS, true))
     end
 
     log_results(results)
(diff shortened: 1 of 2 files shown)


jmartin-sul:
as per the comment i just left on #228, i would be in favor of the general idea behind this PR, but it also seems like the consensus is that we should do #228, which would obviate the need for the specific memoizations implemented here.  it might also be worth exploring @SaravShah's suggestion, though i'd also be happy with this approach if we did decide we wanted to memoize these values.

jmartin-sul:
the way things are in this PR, i believe those are class constants, and will be initialized once (when the class is defined)

tingulfsen:
Well this was a lot of fun. We'll try something altogether different with enumerated database types.

PR #352, opened 2017-11-28
PreservedObjectHandler.update_version_after_validation implemented

ndushay:
Implements `#update_version_after_validation` method in PreservedObjectHandler.

Also added FIXME notes in #confirm_version implementation methods.

closes #283 
closes #287

PR #354, opened 2017-11-28
Pohandler refactor

### LINKED ISSUE #360: (PC) ensure pohandler successfully completes action before results are logged
e.g.:

```
  def create_db_objects(status, validated=false)
    results = []
    create_results = with_active_record_transaction_and_rescue do
      po = PreservedObject.create!(druid: druid,
                                   current_version: incoming_version,
                                   preservation_policy: pp_default)
      pc_attrs = {
        preserved_object: po,
        version: incoming_version,
        size: incoming_size,
        endpoint: endpoint,
        status: status
      }

      if validated
        t = Time.current
        # Returns the value of time as an integer number of seconds since the Epoch.
        pc_attrs[:last_audited] = t.to_i
        pc_attrs[:last_checked_on_storage] = t
      end
    end
    results << result_hash(CREATED_NEW_OBJECT)
    results.concat(create_results)
  end
```

note that results gets CREATED_DB_OBJECT regardless of whether the previous code worked (?)

is this a problem anywhere else in the code?

ndushay:
Some small stuff I noticed while working on #352

Also, closes #360, which was created due to review comments.

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -285,7 +285,6 @@ class PreservedObjectHandler
   # TODO: revisit naming
   def confirm_version_on_db_object(db_object, version_symbol)
     results = []
-
     if incoming_version == db_object.send(version_symbol)
       results.concat(update_status(db_object, Status.ok)) if db_object.is_a?(PreservedCopy)
       results << result_hash(VERSION_MATCHES, db_object.class.name)
@@ -297,9 +296,7 @@ class PreservedObjectHandler
       results.concat(update_status(db_object, Status.unexpected_version)) if db_object.is_a?(PreservedCopy)
       results << result_hash(ARG_VERSION_LESS_THAN_DB_OBJECT, db_object.class.name)
     end
-
     results.concat(update_db_object(db_object))
-
     results
   end
 
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -220,8 +220,9 @@ class PreservedObjectHandler
     results = []
     confirm_results = with_active_record_transaction_and_rescue do
       po_db_object = PreservedObject.find_by!(druid: druid)
+      pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint) if po_db_object
+      break unless pc_db_object
       results.concat(confirm_version_on_db_object(po_db_object, :current_version))
-      pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint)
       results.concat(confirm_version_on_db_object(pc_db_object, :version))
     end
     results.concat(confirm_results)
--- a/spec/services/preserved_object_handler_spec.rb
+++ b/spec/services/preserved_object_handler_spec.rb
@@ -337,33 +337,27 @@ RSpec.describe PreservedObjectHandler do
   end
 
   describe '#with_active_record_transaction_and_rescue' do
-    context 'bogus endpoint' do
-      let(:wrong_ep) do
-        Endpoint.create!(
-          endpoint_name: 'wrong_endpoint',
-          endpoint_type: Endpoint.default_storage_root_endpoint_type,
-          endpoint_node: 'localhost',
-          storage_location: 'blah',
-          recovery_cost: 1
-        )
-      end
-      let(:bad_po_handler) { described_class.new(druid, 6, incoming_size, wrong_ep) }
-
-      before do
-        po = PreservedObject.create!(druid: druid, current_version: 2, preservation_policy: default_prez_policy)
-        PreservedCopy.create!(
-          preserved_object: po, # TODO: see if we got the preserved object that we expected
-          version: po.current_version,
-          size: 1,
-          endpoint: ep,
-          status: Status.default_status
-        )
-      end
-
-      it '#confirm_version rolls back preserved object if the preserved copy cannot be found' do
-        bad_po_handler.confirm_version
-        expect(PreservedObject.find_by(druid: druid).current_version).to eq 2
-      end
+    it '#confirm_version rolls back preserved object if there is a problem updating preserved copy' do
+      wrong_ep = Endpoint.create!(
+        endpoint_name: 'wrong_endpoint',
+        endpoint_type: Endpoint.default_storage_root_endpoint_type,
+        endpoint_node: 'localhost',
+        storage_location: 'blah',
+        recovery_cost: 1
+      )
+      po = PreservedObject.create!(druid: druid, current_version: 2, preservation_policy: default_prez_policy)
+      PreservedCopy.create!(
+        preserved_object: po,
+        version: po.current_version,
+        size: 1,
+        endpoint: wrong_ep,
+        status: Status.default_status
+      )
+      bad_po_handler = described_class.new(druid, 6, incoming_size, wrong_ep)
+      # NOTE: #increase_version checks class of object (rspec double != PreservedCopy)
+      allow_any_instance_of(PreservedCopy).to receive(:save!).and_raise(ActiveRecord::ActiveRecordError)
+      bad_po_handler.confirm_version
+      expect(PreservedObject.find_by(druid: druid).current_version).to eq 2
     end
   end
 end
--- a/spec/services/preserved_object_handler_spec.rb
+++ b/spec/services/preserved_object_handler_spec.rb
@@ -81,6 +81,22 @@ RSpec.describe PreservedObjectHandler do
         )
       end
 
+      it 'stops processing if there is no PreservedCopy' do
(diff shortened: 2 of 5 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -158,8 +158,8 @@ def create_db_objects(status, validated=false)
         pc_attrs[:last_checked_on_storage] = t
       end
       PreservedCopy.create!(pc_attrs)
-      results << result_hash(CREATED_NEW_OBJECT)
     end
+    results << result_hash(CREATED_NEW_OBJECT)
```
jmartin-sul: hmm... though the tests pass, this worries me just a bit, because if the `.create!` call fails with a db error, that'll get trapped by the surrounding `with_active_record_transaction_and_rescue` block, and the result code list would then rightfully have a db failure related result code, in addition to this success code.

that's all based on running this in my head, happy to take a shot at writing a test case for the worry i'm describing if you think i'm off-base.

app/services/preserved_object_handler.rb
```diff
@@ -191,8 +191,9 @@ def confirm_version_in_catalog
     results = []
     confirm_results = with_active_record_transaction_and_rescue do
       po_db_object = PreservedObject.find_by!(druid: druid)
+      pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint) if po_db_object
+      return unless pc_db_object
```
jmartin-sul: i am very slightly surprised this didn't require more edits to the tests, but i think i like this general approach better than what was there before.

however, i think you'd want to maybe append to the `results` list and return that?  or, if you just want to get out of the `with_active_record_transaction_and_rescue` block, you could call `break`.  but i'm assuming you want to return.

now that i think about it more, i'm also slightly unsure what happens when you call `return` in the middle of a transaction.  i'd hope it either implicitly aborts the transaction, or implicitly commits it, or something similarly intuitive (though i'd find the former approach much more immediately intuitive than the latter).  i'm happy to look into that if you did want to keep the `return`, as i think i'd like to know the expected behavior in that situation.

spec/services/preserved_object_handler_spec.rb
```diff
@@ -337,33 +337,27 @@
...
+        endpoint_node: 'localhost',
+        storage_location: 'blah',
+        recovery_cost: 1
+      )
+      po = PreservedObject.create!(druid: druid, current_version: 2, preservation_policy: default_prez_policy)
+      PreservedCopy.create!(
+        preserved_object: po,
+        version: po.current_version,
+        size: 1,
+        endpoint: wrong_ep,
```
jmartin-sul: i think if you make this just `ep`, then there will be a mismatch between where the copy lives and the endpoint you feed in when creating `bad_po_handler` (where i'd keep `wrong_ep`)... which will result in an error when calling `#confirm_version`, and then i think you could get rid of the explicit `and_raise`.

alternatively, if you'd prefer the explicit `and_raise` on `save!`, i think you could probably get rid of the `wrong_ep` definition and just use `ep` in both places?   i think i have a slight preference for the `wrong_ep` approach, but would certainly be ok with switching to the `and_raise` approach instead.  but pretty sure you don't need both (i'd initially implemented the `wrong_ep` thing to get around doing an `and_raise`).

but i am all for the general thing happening here, with the removal of some unnecessary `context` and `let` stuff.

app/services/preserved_object_handler.rb
ndushay: I don't *think* this is the only spot where we add result messages without checking if everything worked;  I would be in favor of it being a separate ticket (patch party?)

app/services/preserved_object_handler.rb
ndushay: break is a better choice; I will fix.

spec/services/preserved_object_handler_spec.rb
ndushay: My thoughts were that this ensured that there wasn't already a pres_copy in previous tests, and that raising an error on save! was a way to have crystal clarity in the test how we we triggered the desired situation.   The previous code a) no longer worked with other modifications and b) was not as crystal clear in the test code itself as to what was happening.

app/services/preserved_object_handler.rb
ndushay: plus, putting result writing within a transaction doesn't mean the logs don't get written until the transaction is done, does it?

tingulfsen: I feel like I'm missing something...why move the results message outside the transaction?

spec/services/preserved_object_handler_spec.rb
tingulfsen: I actually like the explicit raise here.

app/services/preserved_object_handler.rb
```diff
@@ -191,8 +191,9 @@ def confirm_version_in_catalog
     results = []
     confirm_results = with_active_record_transaction_and_rescue do
       po_db_object = PreservedObject.find_by!(druid: druid)
+      pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint) if po_db_object
+      break unless pc_db_object
```
tingulfsen: I like this very much. Should we have a test for this?

app/services/preserved_object_handler.rb
ndushay: because I'm foolish??

app/services/preserved_object_handler.rb
ndushay: good idea - I just added a commit with one.

ndushay:
@tingulfsen @jmartin-sul I think I addressed your concerns, either with follow on commits or rebased commits, or with comments.

The open question:  should I put the CREATED_NEW_OBJECT message back inside the transaction loop?  I think my ignorance about transactions has been exposed, and hopefully somewhat ameliorated, via this PR.

app/services/preserved_object_handler.rb
```diff
@@ -158,8 +158,8 @@ def create_db_objects(status, validated=false)
         pc_attrs[:last_checked_on_storage] = t
       end
       PreservedCopy.create!(pc_attrs)
-      results << result_hash(CREATED_NEW_OBJECT)
     end
+    results << result_hash(CREATED_NEW_OBJECT) unless create_results.include?(DB_UPDATE_FAILED)
```
SaravShah: nice!

app/services/preserved_object_handler.rb
jmartin-sul: to @ndushay's questions in this thread:

> putting result writing within a transaction doesn't mean the logs don't get written until the transaction is done, does it?

nope, i don't think so.  i believe they'll get written when the calls is made, as usual.

my understanding is that everything inside the transaction happens in real time as it normally would, but the database operations don't get committed (and thus don't become visible to other DB connections) unless the transaction succeeds in its entirety (if it doesn't, then the DB updates are rolled back).

> I don't think this is the only spot where we add result messages without checking if everything worked; I would be in favor of it being a separate ticket (patch party?)

good point... off-hand, i can't think of a spot where this happens, but does seem like a good patch party ticket (at least making sure we don't do that, and cleaning up instances where we do).

app/services/preserved_object_handler.rb
jmartin-sul: 👍

ndushay:
@jmartin-sul I added an explicit test for each of our "actions" re: writing the "yay, I xxx-ed for druid blah" message if the db update didn't happen.

tingulfsen:
If it's not too much of a PITA it would be nice to get that CREATED_NEW_OBJECT log message back inside the transaction, I won't hold this up over that though.

ndushay:
@jmartin-sul I added an explicit test for each of our "actions" re: writing the "yay, I xxx-ed for druid blah" message if the db update didn't happen.

ndushay:
I put the logging back inside the transaction, rebased on master, and then rebased these commits to be more atomic.

jmartin-sul:
like @tingulfsen, my preference was initially to put the create message back inside the transaction, but actually, i think @ndushay's initial instinct for placing it outside was on the right track, because it lets us check to see whether the transaction was successful before adding the creation success result code.  see the third commit on #364 (https://github.com/sul-dlss/preservation_catalog/pull/364/commits/d3357824e79123dabb64ec4b3477b30606624e27).

to @ndushay's other point ("The somewhat open question: should I use a diff druid with "ep" instead of "wrong_ep" for the transaction test?"), i have a commit that gets rid of the `wrong_ep` thing, since it felt unnecessary in light of the specific `raise`, which others seemed to have a preference for (and i'd agree now it is probably clearer).

jmartin-sul:
if the changes suggested in #364 are agreeable, i have no preference as to how they get incorporated here (merge into this and merge this to master, merge into this and cleanup commits before merging to master, cherry pick from that branch to this, etc).

PR #355, opened 2017-11-28
turn off provenance logging for development

ndushay:
we don't need it clogging up space (and grep results) on our local implementation, IMHO.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -45,4 +45,4 @@ statuses:
 default_status: ok
 
 provlog:
-  enable: true
+  enable: false


ndushay:
coveralls is ... confused.

jmartin-sul:
i'm a bit surprised it's so much, but my guess is just that the prov log code was getting tested implicitly since it was running all the time when other AR stuff was getting tested, but now it's not running, and it's still part of the code base, so percentage covered decreases (provlog doesn't appear to have its own unit tests).

PR #363, opened 2017-11-29
228 enums

### LINKED ISSUE #228: (PC) consider turning the `statuses` table and `Status` model into a Rails enum on `PreservationCopy`

tingulfsen:
Fixes #228 

Paired with @SaravShah 

We've removed the default status and are using 'ok' instead - a default status is awkward to implement with an enum approach.

RuboCop was complaining about a multiline expect block, wanting us to use do-end instead of curly braces. We think this looks ugly, so we turned that off for all the tests.

Note that we're now referring to statuses by strings such as 'ok'. We could potentially change this to the integers instead, or something like PreservedCopy.statuses[:ok].

MOVED TO 228-status-enums


app/models/preserved_copy.rb
```diff
@@ -3,13 +3,19 @@
...
-
+  enum status: {
+    ok: 0,
+    invalid_moab: 1,
+    invalid_checksum: 2,
+    not_found_on_disk: 3,
+    expected_version_not_found_on_disk: 4,
+    fixity_check_failed: 5
+  }
+  validates :status, inclusion: { in: statuses.keys }
```
ndushay: I am a bit uncomfy using plain strings in the code.  Could we turn those strings into, say:  STATUS_OK here?

will something like this fly?
```
STATUS_OK = status.ok
```

Then we could have 
```
STATUS_DEFAULT = STATUS_OK
```

app/services/preserved_object_handler.rb
```diff
@@ -65,9 +65,9 @@ def create_after_validation
     elsif PreservedObject.exists?(druid: druid)
       results << result_hash(OBJECT_ALREADY_EXISTS, 'PreservedObject')
     elsif moab_validation_errors.empty?
-      results.concat(create_db_objects(Status.default_status, true))
+      results.concat(create_db_objects("ok", true))
```
ndushay: yeah, these strings here make me nervous.  Can we get to something similar to `PreservedCopy.STATUS_OK` ?

The reasoning behind the "default" status is that we're trying to be more explicit that we're not really claiming it's okay -- we're just instantiating the damn thing and don't know its status.  This particular line was in error in its original form, as we were explicitly wanting "OK".

app/services/preserved_object_handler.rb
```diff
@@ -81,7 +81,7 @@ def create
     elsif PreservedObject.exists?(druid: druid)
       results << result_hash(OBJECT_ALREADY_EXISTS, 'PreservedObject')
     else
-      results.concat(create_db_objects(Status.default_status))
+      results.concat(create_db_objects("ok"))
```
ndushay: this is a correct place for `PreservedCopy.STATUS_DEFAULT`

app/models/preserved_copy.rb
```diff
@@ -3,13 +3,19 @@
...
   validates :version, presence: true
-  validates :status, presence: true
   # NOTE: size here is approximate and not used for fixity checking
   validates :size, numericality: { only_integer: true, greater_than: 0 }, allow_nil: true
-
+  enum status: {
+    ok: 0,
+    invalid_moab: 1,
+    invalid_checksum: 2,
+    not_found_on_disk: 3,
```
ndushay: while we're in here ... do we want to say "disk"?  Or maybe `online_moab_not_found` and `expected_version_not_found_in_moab` or some such?

spec/load_fixtures_helper.rb
```diff
@@ -40,7 +40,7 @@ def load_fixture_moabs
       PreservedCopy.create(preserved_object_id: po.id,
                            endpoint_id: @storage_dir_to_endpoint_id[storage_dir],
                            current_version: version,
-                           status: Status.default_status)
+                           status: "ok")
```
ndushay: DEFAULT_STATUS would very much be more correct here ...

spec/services/preserved_object_handler_spec.rb
```diff
@@ -210,10 +210,12 @@
...
         it "logs at error level" do
           expect(Rails.logger).to receive(:log).with(Logger::ERROR, version_less_than_po_msg)
-          expect(Rails.logger).to receive(:log).with(Logger::ERROR, version_less_than_pc_msg)
           expect(Rails.logger).to receive(:log).with(Logger::INFO, updated_po_db_timestamp_msg)
-          expect(Rails.logger).to receive(:log).with(Logger::INFO, updated_pc_db_obj_msg)
           expect(Rails.logger).to receive(:log).with(Logger::INFO, updated_pc_db_status_msg)
+          expect(Rails.logger).to receive(:log).with(Logger::ERROR, version_less_than_pc_msg)
+          expect(Rails.logger).to receive(:log).with(Logger::INFO, updated_pc_db_obj_msg)
+          # require 'byebug'
+          # byebug
```
ndushay: oops!  you meant to remove these comments, I bet.

ndushay:
So should this PR be closed???

SaravShah:
refer to #368

PR #364, opened 2017-12-01
Pohandler refactor touchups

jmartin-sul:
here are some suggested fixes to #354 that felt easier to implement with commit comments than to suggest in comments on that PR

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -220,8 +220,7 @@ class PreservedObjectHandler
     results = []
     confirm_results = with_active_record_transaction_and_rescue do
       po_db_object = PreservedObject.find_by!(druid: druid)
-      pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint) if po_db_object
-      break unless pc_db_object
+      pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint)
       results.concat(confirm_version_on_db_object(po_db_object, :current_version))
       results.concat(confirm_version_on_db_object(pc_db_object, :version))
     end
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -164,7 +164,7 @@ class PreservedObjectHandler
   def create_db_objects(status, validated=false)
     results = []
     pp_default = PreservationPolicy.default_preservation_policy
-    create_results = with_active_record_transaction_and_rescue do
+    transaction_results = with_active_record_transaction_and_rescue do
       po = PreservedObject.create!(druid: druid,
                                    current_version: incoming_version,
                                    preservation_policy: pp_default)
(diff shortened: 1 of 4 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -183,37 +183,60 @@ def create_db_objects(status, validated=false)
         pc_attrs[:last_checked_on_storage] = t
       end
       PreservedCopy.create!(pc_attrs)
+    end
+
+    if transaction_results.empty?
       results << result_hash(CREATED_NEW_OBJECT)
+    else
+      results.concat(transaction_results)
```
ndushay: meh - I don't think this makes the code clearer, tbh.

app/services/preserved_object_handler.rb
```diff
@@ -183,37 +183,60 @@ def create_db_objects(status, validated=false)
...
-    results.concat(upd_results)
+
+    # ok, now we're out of the woods:  if we're here, the transaction is over.  and if it produced no results
+    # of its own, it completed and committed successfully.  so if there were no error codes produced from the
+    # transaction running and committing, return the update results, otherwise, return the transaction failure code(s).
+    if transaction_results.empty?
+      full_results.concat(db_upd_results)
+    else
+      full_results.concat(transaction_results)
+    end
```
ndushay: again, I'm not convinced all of this makes the code clearer.  it certainly seems to make it more complex.  And longer.

I think some of the intent of this work might be better served with better naming -- being clear about updating ActiveRecord object vs updating the database, and being clear about what the transaction code does (I think it is fairly clear -- i'm just a noob to transactions.)

spec/services/preserved_object_handler_create_spec.rb
```diff
@@ -70,7 +70,7 @@
             expect(results).to include(a_hash_including(result_code => a_string_matching('foo')))
           end
           it 'does NOT get CREATED_NEW_OBJECT message' do
-            expect(results).to include(hash_not_including(PreservedObjectHandler::CREATED_NEW_OBJECT))
+            expect(results).not_to include(hash_including(PreservedObjectHandler::CREATED_NEW_OBJECT))
```
ndushay: this is a slightly interesting twist on how best to express that we don't expect something in the results.  I think I prefer your way.

jmartin-sul: well, the thing is, the app code changes above are a result of this change:  this change without those caused test failures, because i think this is expressing a slightly different expectation, which is unmet by the previous version of the app code.

i agree that the app code changes above are less readable, but i think they're more correct.

would you agree that this test change is correct?  and that it expresses something different than the prior version (expecting not to include a specific code, instead of expecting to include a code which is not the code in question, even if the code in question is also in the list).

this test change feels like the crux of this commit to me, and the other things feel like necessary follow ons.

app/services/preserved_object_handler.rb
jmartin-sul: yeah, i'm not excited about the added complexity and decreased readability, but i do think it's slightly more correct behaviorally.

jmartin-sul:
> Why is this so fricking hard?

`the joys of validation combined` * `the joys of transaction handling` = `too much complexity` 😆

app/services/preserved_object_handler.rb
tingulfsen: This new code is more complex but seems more readable to me.

jmartin-sul: > again, I'm not convinced all of this makes the code clearer

me neither 😄 

>  it certainly seems to make it more complex. And longer.

i would agree

app/services/preserved_object_handler.rb
```diff
@@ -164,7 +164,7 @@ def moab_validation_errors
   def create_db_objects(status, validated=false)
     results = []
     pp_default = PreservationPolicy.default_preservation_policy
-    create_results = with_active_record_transaction_and_rescue do
+    transaction_results = with_active_record_transaction_and_rescue do
```
SaravShah: So `transaction_results` will only have objects in it, if the transaction failed right? And the objects will be error codes if `OBJECT_DOES_NOT_EXIST` or `DB_UPDATE_FAILED` ?

SaravShah: If the above is the case then what if we included the word `errors` or something?

ndushay: I think transaction_results are result codes.  note that they get concatenated to results.

PR #365, opened 2017-12-02
 rake task to drop specific storage root

### LINKED ISSUE #342: (SPC) rake task  to remove db data for a particular storage root
to aid in seeding re-runs

SaravShah:
- I'm deleting all PreservedCopies that have an endpoint_name of whatever you specify
- Then i'm deleting the rest of the orphan PreservedObejcts

- Had to do this weird syntax in `moab_to_catalog_spec.rb` lines 259-264
```
begin
  subject
rescue
end
```
If anyone knows how to get around this, I would like to know!

- Also, should I add a `puts` statement that says something like `Congrats you deleted all the data off #{endpoint_name}` ?

closes #342

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -34,3 +34,19 @@ task :seed_catalog, [:profile] => [:environment] do |_t, args|
   puts "#{Time.now.utc.iso8601} Done"
   $stdout.flush
 end
+
+desc "Delete single endpoint db data"
+task :drop, [:storage_root] => [:environment] do |_t, args|
+  if args[:storage_root]
+    root = args[:storage_root]
+    puts "You're about to erase all the data for #{root}. Are you sure you want to continue? [y/N]"
+    input = STDIN.gets.chomp
+    if input.casecmp("y").zero? # rubocop prefers casecmp because it is faster than '.downcase =='
+      MoabToCatalog.drop_endpoint(root)
+    else
+      puts "You canceled erasing data from #{root}"
+    end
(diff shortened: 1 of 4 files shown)


lib/audit/moab_to_catalog.rb
```diff
@@ -65,4 +65,11 @@ def self.check_existence_for_all_storage_roots
       Rails.logger.info end_msg
     end
   end
+
+  def self.drop_storage_root(endpoint_name)
```
ndushay: this should probably be named drop_endpoint ??  (as we could conceivably use it for archival endpoints too?)

Rakefile
```diff
@@ -34,3 +34,14 @@ task :seed_catalog, [:profile] => [:environment] do |_t, args|
   puts "#{Time.now.utc.iso8601} Done"
   $stdout.flush
 end
+
+
+task :drop, [:storage_root] => [:environment] do |t,args|
+  if args[:storage_root]
+    p "you're in the if condition"
```
ndushay: It would be awesome if this rake task had a "you're about to erase all data for xxx - are you sure? (Y, N)" sort of prompt.  

This would be more complex than your puts statement above (which I assume is debug speak), but I think it's the right thing to do.

tallenaz:
@SaravShah this looks good to me. I would like to see a `puts` statement for successfully deleting a root, but I wouldn't hold up merging this for it.

PR #367, opened 2017-12-04
228 status enums

### LINKED ISSUE #228: (PC) consider turning the `statuses` table and `Status` model into a Rails enum on `PreservationCopy`

SaravShah:
Fixes #228 

paired with @tingulfsen


SaravShah:
Something got messed up and deleted a whole bunch of tests. Refer to https://github.com/sul-dlss/preservation_catalog/pull/368

PR #368, opened 2017-12-04
228 final enums

### LINKED ISSUE #228: (PC) consider turning the `statuses` table and `Status` model into a Rails enum on `PreservationCopy`

SaravShah:
Fixes #228

paired with @tingulfsen


app/models/preserved_copy.rb
```diff
@@ -3,13 +3,21 @@
...
-
+  enum status: {
+    ok: 0,
+    invalid_moab: 1,
+    invalid_checksum: 2,
+    online_moab_not_found: 3,
+    expected_version_not_found_online: 4,
+    fixity_check_failed: 5
+  }
+  DEFAULT_STATUS = statuses[:ok].freeze
```
ndushay: lovin' this 'DEFAULT_STATUS' constant

did the Settings.yml.  change to remove status stuff?   And/or github issues to update shared_configs?

spec/models/preserved_copy_spec.rb
```diff
@@ -6,7 +6,7 @@
     policy_id = PreservationPolicy.default_preservation_policy.id
     PreservedObject.create!(druid: 'ab123cd4567', current_version: 1, preservation_policy_id: policy_id)
   end
-  let!(:status) { Status.default_status }
+  let!(:status) { described_class.statuses[:ok] }
```
ndushay: `described_class::DEFAULT_STATUS` I think.

spec/services/preserved_object_handler_create_spec.rb
```diff
@@ -76,8 +76,8 @@
       end
 
       it "rolls back pres object creation if the pres copy can't be created (e.g. due to DB constraint violation)" do
-        # so that pres copy creation fails with a constraint violation, since status can't be nil
-        allow(Status).to receive(:default_status).and_return(nil)
+        # so that pres copy creation fails, thus forcing the transaction to be rolled back
+        allow(PreservedCopy).to receive(:create!).and_raise(ActiveRecord::RecordInvalid)
```
ndushay: yup - I like the explicit raise here better than the implicit one.

spec/services/preserved_object_handler_spec.rb
```diff
@@ -226,10 +226,10 @@
         end
         it "logs at error level" do
           expect(Rails.logger).to receive(:log).with(Logger::ERROR, version_less_than_po_msg)
-          expect(Rails.logger).to receive(:log).with(Logger::ERROR, version_less_than_pc_msg)
           expect(Rails.logger).to receive(:log).with(Logger::INFO, updated_po_db_timestamp_msg)
-          expect(Rails.logger).to receive(:log).with(Logger::INFO, updated_pc_db_obj_msg)
           expect(Rails.logger).to receive(:log).with(Logger::INFO, updated_pc_db_status_msg)
+          expect(Rails.logger).to receive(:log).with(Logger::ERROR, version_less_than_pc_msg)
+          expect(Rails.logger).to receive(:log).with(Logger::INFO, updated_pc_db_obj_msg)
```
ndushay: I have no problem with this change but I'm curious as to whether the order mattered.

spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -536,10 +536,10 @@
             expect(pc.reload.last_checked_on_storage).to be > orig_timestamp
           end
           it 'ensures status of PreservedCopy is invalid' do
-            pc.status = Status.ok
+            pc.status = PreservedCopy.statuses[:ok]
             pc.save!
             po_handler.update_version_after_validation
-            expect(pc.reload.status).to eq Status.invalid_moab
+            expect(PreservedCopy.statuses[pc.reload.status]).to eq PreservedCopy.statuses[:invalid_moab]
```
ndushay: `PreservedCopy.statuses[pc.reload.status])` is gnarly syntax.  plain old `pc.reload.status` didn't work, eh?

app/models/preserved_copy.rb
```diff
@@ -3,13 +3,21 @@
...
+
   # NOTE: size here is approximate and not used for fixity checking
   validates :size, numericality: { only_integer: true, greater_than: 0 }, allow_nil: true
-
+  enum status: {
+    ok: 0,
+    invalid_moab: 1,
+    invalid_checksum: 2,
+    online_moab_not_found: 3,
+    expected_version_not_found_online: 4,
```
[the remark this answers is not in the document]
SaravShah: oh, yeah but i think the name `expected_version_not_found_in_moab` sounds better? Thoughts?

app/models/preserved_copy.rb
SaravShah: shared_configs doesn't have status stuff...when i checked our two branches

spec/services/preserved_object_handler_spec.rb
SaravShah: Yes the order did matter.

spec/services/preserved_object_handler_update_version_spec.rb
SaravShah: Yeah the `pc.reload.status` wasn't working 🤔

app/models/preserved_copy.rb
```diff
@@ -3,13 +3,21 @@
...
 
   validates :preserved_object, presence: true
   validates :endpoint, presence: true
   validates :version, presence: true
-  validates :status, presence: true
+
   # NOTE: size here is approximate and not used for fixity checking
   validates :size, numericality: { only_integer: true, greater_than: 0 }, allow_nil: true
-
+  enum status: {
```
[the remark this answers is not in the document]
jmartin-sul: see #371

spec/services/preserved_object_handler_update_version_spec.rb
jmartin-sul: i think the simpler way to do this would be:
```ruby
expect(pc.reload.status).to eq 'invalid_moab'`
```

because enum fields in active record objects return a string (though they accept the string, representation, the equivalent symbol, or the underlying integer).

jmartin-sul:
merging, since @ndushay seemed fine to not hold up merging based on her remaining concerns.

PR #369, opened 2017-12-04
PreservedObjectHandler: refactored results and logging into separate class

### LINKED ISSUE #327: (PC) pohandler refactor out result codes / messages / logging into separate file
In thinking about how to reduce the size of the preserved_object_handler class, the first thing that looked plausible was to extract out the response codes, their messages, the results methods and the logging methods into a separate file.  Not sure if file should be its own class, or what ... thinking maybe `app/services/pohandler_errors.rb` or some such required by pohandler.

I would try to include the constant declarations, and probably the methods
- result_hash
- result_code_msg
- result_msg_prefix
- log_results
- logger_severity_level

ndushay:
This shortens PreservedObjectHandler by about 100 lines (from about 400 to about 300).

Closes #327

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -188,7 +188,7 @@ class PreservedObjectHandler
       end
     end
 
-    handler_results.remove_db_updated_result unless transaction_ok
+    handler_results.remove_db_updated_results unless transaction_ok
   end
 
   def confirm_version_in_catalog
@@ -198,7 +198,7 @@ class PreservedObjectHandler
       confirm_version_on_db_object(po_db_object, :current_version)
       confirm_version_on_db_object(pc_db_object, :version)
     end
-    handler_results.remove_db_updated_result unless transaction_ok
+    handler_results.remove_db_updated_results unless transaction_ok
   end
 
   # performs passed code wrapped in ActiveRecord transaction via yield
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -149,16 +149,16 @@ class PreservedObjectHandler
     object_dir = "#{storage_location}/#{DruidTools::Druid.new(druid).tree.join('/')}"
     moab = Moab::StorageObject.new(druid, object_dir)
     object_validator = Stanford::StorageObjectValidator.new(moab)
-    errors = object_validator.validation_errors(Settings.moab.allow_content_subdirs)
-    if errors.any?
+    moab_errors = object_validator.validation_errors(Settings.moab.allow_content_subdirs)
+    if moab_errors.any?
       moab_error_msg_list = []
-      errors.each do |error_hash|
+      moab_errors.each do |error_hash|
         error_hash.each_value { |moab_error_msgs| moab_error_msg_list << moab_error_msgs }
       end
       results << result_hash(INVALID_MOAB, moab_error_msg_list)
       log_results(results)
     end
-    errors
+    moab_errors
   end
 
   def create_db_objects(status, validated=false)
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -42,7 +42,7 @@ class PreservedObjectHandler
       create_db_objects(PreservedCopy.statuses[:invalid_moab], true)
     end
 
-    handler_results.log_my_results
+    handler_results.log_results
     handler_results.result_array
   end
 
@@ -55,7 +55,7 @@ class PreservedObjectHandler
       create_db_objects(PreservedCopy::DEFAULT_STATUS)
     end
 
-    handler_results.log_my_results
+    handler_results.log_results
     handler_results.result_array
   end
 
@@ -67,7 +67,7 @@ class PreservedObjectHandler
       confirm_version_in_catalog
     end
 
-    handler_results.log_my_results
+    handler_results.log_results
     handler_results.result_array
   end
 
@@ -89,7 +89,7 @@ class PreservedObjectHandler
       end
     end
 
-    handler_results.log_my_results
+    handler_results.log_results
     handler_results.result_array
   end
 
@@ -106,7 +106,7 @@ class PreservedObjectHandler
       end
     end
 
-    handler_results.log_my_results
+    handler_results.log_results
     handler_results.result_array
   end
 
@@ -122,9 +122,7 @@ class PreservedObjectHandler
       moab_errors.each do |error_hash|
         error_hash.each_value { |msg| moab_error_msgs << msg }
       end
-      PreservedObjectHandlerResults.log_results(
-        [ handler_results.result_hash(PreservedObjectHandlerResults::INVALID_MOAB, moab_error_msgs) ]
-      )
+      handler_results.add_result(PreservedObjectHandlerResults::INVALID_MOAB, moab_error_msgs)
     end
     moab_errors
   end
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -237,7 +237,7 @@ class PreservedObjectHandler
     elsif incoming_version < db_object.send(version_symbol)
       handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, db_object.class.name)
     elsif incoming_version > db_object.send(version_symbol)
-      handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, db_object.class.name)
+      handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, db_object.class.name)
     end
   end
 
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -7,35 +7,7 @@
 # inspired by http://www.thegreatcodeadventure.com/smarter-rails-services-with-active-record-modules/
 class PreservedObjectHandler
 
-  INVALID_ARGUMENTS = 1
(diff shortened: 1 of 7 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -56,115 +28,110 @@ def initialize(druid, incoming_version, incoming_size, endpoint)
...
-      results.concat(create_db_objects(Status.default_status, true))
+      create_db_objects(Status.default_status, true)
     else
-      results.concat(create_db_objects(Status.invalid_moab, true))
+      create_db_objects(Status.invalid_moab, true)
     end
 
-    log_results(results)
-    results
+    handler_results.log_my_results
```
ndushay: `log_my_results` is a hokey name;  i wanted to distinguish it from the class `log_results` method ... suggestions welcome.

app/services/preserved_object_handler.rb
```diff
@@ -56,115 +28,110 @@ def initialize(druid, incoming_version, incoming_size, endpoint)
...
+        error_hash.each_value { |msg| moab_error_msgs << msg }
       end
-      results << result_hash(INVALID_MOAB, moab_error_msg_list)
-      log_results(results)
+      PreservedObjectHandlerResults.log_results(
+        [ handler_results.result_hash(PreservedObjectHandlerResults::INVALID_MOAB, moab_error_msgs) ]
+      )
     end
-    errors
+    moab_errors
```
ndushay: I refactored moab_validation_errors method in addition to the logging changes -- I think it's clearer now.  Happy to entertain dissent.

app/services/preserved_object_handler.rb
```diff
@@ -185,81 +152,69 @@ def create_db_objects(status, validated=false)
       PreservedCopy.create!(pc_attrs)
     end
 
-    if transaction_results.empty?
-      results << result_hash(CREATED_NEW_OBJECT)
-    else
-      results.concat(transaction_results)
-    end
+    handler_results.add_result(PreservedObjectHandlerResults::CREATED_NEW_OBJECT) if transaction_ok
```
ndushay: here, having the transaction wrapper method return a boolean works nicely, IMO.

app/services/preserved_object_handler.rb
```diff
@@ -185,81 +152,69 @@ def create_db_objects(status, validated=false)
...
 
-    # ok, now we're out of the woods:  if we're here, the transaction is over.  and if it produced no results
-    # of its own, it completed and committed successfully.  so if there were no error codes produced from the
-    # transaction running and committing, return the update results, otherwise, return the transaction failure code(s).
-    if transaction_results.empty?
-      full_results.concat(db_upd_results)
-    else
-      full_results.concat(transaction_results)
-    end
+    handler_results.remove_db_updated_result unless transaction_ok
```
ndushay: here, having the transaction wrapper method return a boolean, and then doing this "remove results" is a little weird ... BUT I think it makes things clearer to the point that I felt John's block comments were no longer needed.

app/services/preserved_object_handler.rb
```diff
@@ -185,81 +152,69 @@ def create_db_objects(status, validated=false)
...
+    transaction_ok = with_active_record_transaction_and_rescue do
       po_db_object = PreservedObject.find_by!(druid: druid)
       pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint)
-      results.concat(confirm_version_on_db_object(po_db_object, :current_version))
-      results.concat(confirm_version_on_db_object(pc_db_object, :version))
+      confirm_version_on_db_object(po_db_object, :current_version)
+      confirm_version_on_db_object(pc_db_object, :version)
     end
-    results.concat(confirm_results)
+    handler_results.remove_db_updated_result unless transaction_ok
```
ndushay: adding some actual logic here dependent on transaction results ...

spec/services/preserved_object_handler_create_spec.rb
```diff
@@ -40,13 +40,14 @@
       it 'logs an error' do
         po_handler.create
         expect(Rails.logger).to receive(:log).with(Logger::ERROR, exp_msg)
-        po_handler.create
+        new_po_handler = described_class.new(druid, incoming_version, incoming_size, ep)
+        new_po_handler.create
```
ndushay: this is the one test I had to change -- it's abnormal to re-use a pohandler object so I changed this test so it no longer did.

app/services/preserved_object_handler.rb
jmartin-sul: i think it'd be reasonable to collapse the two into an instance method called `log_results`, something like:
```ruby
def log_results
  result_array.each do |r|
    severity = logger_severity_level(r.keys.first)
    msg = r.values.first
  end
end
```

spec/services/preserved_object_handler_create_spec.rb
jmartin-sul: as a stretch-goal/patch-party type thing, maybe we could be nice to the caller and set a flag internally that'd let us raise an error when an instance has been used and a new one has to be instantiated?

app/services/preserved_object_handler.rb
```diff
@@ -56,115 +28,110 @@ def initialize(druid, incoming_version, incoming_size, endpoint)
...
+      moab_error_msgs = []
       moab_errors.each do |error_hash|
-        error_hash.each_value { |moab_error_msgs| moab_error_msg_list << moab_error_msgs }
+        error_hash.each_value { |msg| moab_error_msgs << msg }
       end
-      results << result_hash(INVALID_MOAB, moab_error_msg_list)
-      log_results(results)
+      PreservedObjectHandlerResults.log_results(
+        [ handler_results.result_hash(PreservedObjectHandlerResults::INVALID_MOAB, moab_error_msgs) ]
+      )
```
jmartin-sul: maybe:
```ruby
handler_results.add_result(PreservedObjectHandlerResults::INVALID_MOAB, moab_error_msgs)
```

app/services/preserved_object_handler_results.rb
```diff
@@ -0,0 +1,112 @@
...
+    @msg_prefix = "PreservedObjectHandler(#{druid}, #{incoming_version}, #{incoming_size}, #{endpoint})"
+    @result_array = []
+  end
+
+  def add_result(code, msg_args=nil)
+    result_array << result_hash(code, msg_args)
+  end
+
+  # used when updates wrapped in transaction fail, and there is a need to ensure there is no db updated result
+  def remove_db_updated_result
```
jmartin-sul: nitpicking, but i would prefer pluralizing the name since it's possible multiple results will get removed.

app/services/preserved_object_handler.rb
```diff
@@ -277,124 +232,66 @@ def update_pc_validation_timestamps(pres_copy)
...
-    results = []
     if incoming_version == db_object.send(version_symbol)
-      results << result_hash(VERSION_MATCHES, db_object.class.name)
+      handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, db_object.class.name)
     elsif incoming_version < db_object.send(version_symbol)
-      results << result_hash(ARG_VERSION_LESS_THAN_DB_OBJECT, db_object.class.name)
+      handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, db_object.class.name)
     elsif incoming_version > db_object.send(version_symbol)
-      results << result_hash(ARG_VERSION_GREATER_THAN_DB_OBJECT, db_object.class.name)
+      handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, db_object.class.name)
```
jmartin-sul: was changing the result code here intentional?  seems like this should stay as it was?  but also this didn't seem to break the tests, so i guess either i'm missing something or this is a test case we missed?

app/services/preserved_object_handler.rb
jmartin-sul: slightly surprised/worried this didn't require any tests to be fixed.

spec/services/preserved_object_handler_create_spec.rb
ndushay: sure, could be a patch party ticket.  Make it so.  Also make a case for it in the ticket??

app/services/preserved_object_handler.rb
ndushay: probably a typo - will fix.

app/services/preserved_object_handler.rb
ndushay: confirm_version was written early on;  i think our mental models grew over time blah blah.

app/services/preserved_object_handler_results.rb
ndushay: agree - will fix.

app/services/preserved_object_handler.rb
ndushay: Actually, it's a test case I fudged, so I'm delighted you caught it.

https://github.com/sul-dlss/preservation_catalog/blob/7120f7411f70ef97f9d479373475d614340a85d5/spec/services/preserved_object_handler_update_version_spec.rb#L169-L179

app/services/preserved_object_handler.rb
ndushay: actually - this is where the separate logging was, and you were right about this being part of the fix.

ndushay:
@jmartin-sul I think I addressed your concerns with 3 follow-on commits;  I have a slight preference to squash commits if you are good with the resulting code.

jmartin-sul:
thanks, @ndushay!  all looks fine to me now.

spec/services/preserved_object_handler_create_spec.rb
jmartin-sul: see #388

PR #374, opened 2017-12-06
Add rake task and method to populate a single endpoint

tallenaz:


tallenaz:
The WIP was just me being cautious, but I think updating the README is a good idea. I'll make that change, and then erase the WIP designation.

PR #377, opened 2017-12-06
Refactor migration to allow it to run when the table is populated

### LINKED ISSUE #376: Fix cap deployment fail on AddStatusToPreservedCopies migration
see https://www.viget.com/articles/adding-a-not-null-column-to-an-existing-table

tallenaz:
The problem and solution this addresses is documented here: https://www.viget.com/articles/adding-a-not-null-column-to-an-existing-table

The migration would run locally (because of an empty table), but wouldn't run on prod during a capistrano deployment, because the table was populated.

Fixes #376

### DIFF
--- a/db/migrate/20171204215532_add_status_to_preserved_copies.rb
+++ b/db/migrate/20171204215532_add_status_to_preserved_copies.rb
@@ -1,6 +1,11 @@
 class AddStatusToPreservedCopies < ActiveRecord::Migration[5.1]
-  def change
-    add_column :preserved_copies, :status, :int, null: false
+  def up
+    add_column :preserved_copies, :status, :int
+    change_column_null :preserved_copies, :status, false
     add_index :preserved_copies, :status
+  end
+
+  def down
+    remove_column :preserved_copies, :status, :int
   end
 end


db/migrate/20171204215532_add_status_to_preserved_copies.rb
```diff
@@ -1,6 +1,11 @@
 class AddStatusToPreservedCopies < ActiveRecord::Migration[5.1]
-  def change
-    add_column :preserved_copies, :status, :int, null: false
+  def up
+    add_column :preserved_copies, :status, :int
+    change_column_null :preserved_copies, :status, false
```
SaravShah: so this adds the constraint which tells the database not to accept NULL values?

tallenaz: Yes. I just tested this out via console in test with:
```
pc = PreservedCopy.first
pc.status = nil
```
and was not allowed to save:
```
[8] pry(main)> pc.save
   (0.3ms)  BEGIN
  PreservedObject Load (0.3ms)  SELECT  "preserved_objects".* FROM "preserved_objects" WHERE "preserved_objects"."id" = $1 LIMIT $2  [["id", 197], ["LIMIT", 1]]
  Endpoint Load (0.3ms)  SELECT  "endpoints".* FROM "endpoints" WHERE "endpoints"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
   (0.2ms)  ROLLBACK
=> false
```

PR #375, opened 2017-12-06
enums touchups

jmartin-sul:
~~NOTE: this is WIP, because we should merge #369 first, and then rebase this on master.  though i have rebased this on the branch used by #369, under the assumption that it'll get merged in something pretty similar to its current form.~~

* a bit of re-arrangement so that like things are grouped contiguously
* a slightly more conventional approach to enum usage
* db/README.md: fix for accuracy now that PreservedCopy.status is an enum
* slightly more thorough #status validation tests

closes #373

PR #379, opened 2017-12-07
[WIP] refactor confirm version

### LINKED ISSUE #353: (M2C) confirm_version needs to use update_version_after_validation
POHandler.confirm_version, when it discovers a later version, is supposed to (effectively) call update_version_after_validation.   The current method does NOT do this -- there is an #increase_version method ONLY called by confirm_version, and it needs to go away in favor of the correct methods.  

Likely confirm_version won't really call update_version_after_validation but instead will call some inner methods used by it.

jmartin-sul:
~~NOTE: before this is merged, we should get #375 merged, and then we should rebase this on master (as this branch builds off of the branch for #375, which itself builds off the branch for #369).~~

paired w/ @SaravShah

closes #353

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -127,7 +127,7 @@ Rails/SkipsModelValidations:
 
 RSpec/AnyInstance:
   Exclude:
-    - spec/services/preserved_object_handler_spec.rb # #increase_version checks class of object (rspec double != PreservedCopy)
+    - spec/services/preserved_object_handler_spec.rb # Mock object doesn't get returned so need to stub it
 
 RSpec/BeforeAfterAll:
   Exclude:
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -64,7 +64,11 @@ class PreservedObjectHandler
       handler_results.add_result(PreservedObjectHandlerResults::INVALID_ARGUMENTS, errors.full_messages)
     else
       Rails.logger.debug "confirm_version #{druid} called"
-      confirm_version_in_catalog
+      if endpoint.endpoint_type.endpoint_class == 'online'
+        confirm_online_version
+      elsif endpoint.endpoint_type.endpoint_class == 'archive'
+        # TODO: do some confirmation with archive version
+      end
     end
 
     handler_results.log_results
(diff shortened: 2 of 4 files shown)


ndushay:
I'm a bit loathe to review this until it is rebased ...

jmartin-sul:
@ndushay: totally reasonable

tallenaz:
The commits

https://github.com/sul-dlss/preservation_catalog/pull/379/commits/aa63c2c681799b3a7bbef17690c136b771cd796f
https://github.com/sul-dlss/preservation_catalog/pull/379/commits/cb6fbe95594cab87e650a1009faf138a43e457b0

look good to me.

app/services/preserved_object_handler.rb
```diff
@@ -191,12 +195,31 @@ def update_online_version(validated=false, status=nil)
...
 
-  def confirm_version_in_catalog
+  def confirm_online_version
     transaction_ok = with_active_record_transaction_and_rescue do
       po_db_object = PreservedObject.find_by!(druid: druid)
       pc_db_object = PreservedCopy.find_by!(preserved_object: po_db_object, endpoint: endpoint)
-      confirm_version_on_db_object(po_db_object, :current_version)
-      confirm_version_on_db_object(pc_db_object, :version)
+
+      if incoming_version > po_db_object.current_version && po_db_object.current_version == pc_db_object.version
```
ndushay: for this method, I think the first case should be the happy path in which po and pc agree on version and incoming version matches.

ndushay: I think if the incoming version is higher, it is supposed to call update_version_after_validation.

app/services/preserved_object_handler.rb
```diff
@@ -191,12 +195,31 @@ def update_online_version(validated=false, status=nil)
...
+        if moab_validation_errors.empty?
+          update_status(pc_db_object, 'ok')
+        else
+          update_status(pc_db_object, 'invalid')
+        end
+        update_preserved_copy_version_etc(pc_db_object, incoming_version, incoming_size, true)
+      elsif incoming_version == po_db_object.current_version && po_db_object.current_version == pc_db_object.version
+        handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, po_db_object.class.name)
+        handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pc_db_object.class.name)
+      else
```
ndushay: this logic is flawed:  you could get here if po and pc versions don't match, also.    I'm not sure if we do the same thing in either case, but certainly you only output arg_version_less_than if it's true ...

Happy to pair on this;  it might be more efficient.

lib/audit/moab_to_catalog.rb
```diff
@@ -16,6 +16,7 @@ def self.check_existence_for_dir(storage_dir, expect_to_create=false)
         results << po_handler.confirm_version
       else
         Rails.logger.error "druid: #{druid} expected to exist in catalog but was not found"
+        # FIXME: might need to call create_after_validation - See Audit Requirement google doc
```
ndushay: I think this might be true inside pohandler.check_existence, not here ...

.rubocop.yml
```diff
@@ -127,7 +127,7 @@ Rails/SkipsModelValidations:
 
 RSpec/AnyInstance:
   Exclude:
-    - spec/services/preserved_object_handler_spec.rb # #increase_version checks class of object (rspec double != PreservedCopy)
+    - spec/services/preserved_object_handler_spec.rb # Mock object doesn't get returned so need to stub it
```
ndushay: do we still even need this rubocop exception here?  if increase_version goes away?

spec/services/preserved_object_handler_spec.rb
```diff
@@ -382,7 +382,6 @@
       )
       bad_po_handler = described_class.new(druid, 6, incoming_size, ep)
       allow(bad_po_handler).to receive(:moab_validation_errors).and_return([])
-      # NOTE: #increase_version checks class of object (rspec double != PreservedCopy)
       allow_any_instance_of(PreservedCopy).to receive(:save!).and_raise(ActiveRecord::ActiveRecordError)
```
ndushay: if you are removing the comment, to me it implies we could look to receive save on a specific object.  If that is NOT true, then what's changed is the method in question is no longer #increase_version and thus the comment should stay but the method name should be updated.

jmartin-sul:
this is now rebased on current master, but will remain WIP till the above concerns are addressed.

also, it appears that @ndushay's github comment about the removal of a code comment was inadvertently collapsed by the rebasing.

app/services/preserved_object_handler.rb
ndushay: hrmm ... actually, i'm no longer sure.  ugh.

jmartin-sul: seems to me like we should call some of the helper methods called by `update_version_after_validation`, but not the method itself, since that'd re-run e.g. `if invalid?`, the transaction wrapping, the endpoint type check, etc.

ndushay:
I *think* this is going to be closed in favor of #401

jmartin-sul:
@ndushay that's my understanding also

jmartin-sul:
since #401 seems close to approval, i'm going to go ahead and close this one out.  easy enough to refer back to the closed PR if need be.

PR #380, opened 2017-12-07
default PreservationPolicy caching

### LINKED ISSUE #347: (PC) pohandler constant for default PreservationPolicy; memoize endpoint_type lookups ...

jmartin-sul:
i went with caching the id for the default preservation policy, instead of caching the whole active record object.  this felt safer to me -- my gut feeling left me slightly leery of caching a whole AR object, i know ActiveJob advises users to pass around job IDs instead of job objects (which are often activerecord objects underneath if e.g. delayed_job is used), and users on stack exchange etc advised against caching active record objects.  we only need the id anyway, since either the id or the object can be used to set the association.  as a bonus, this is a slightly lighter thing to cache, but i'd guess that's pretty negligible.

the cache eviction is naive, as pointed out in the comments, but i think it'll be performant enough, with as infrequently as we'd expect to add pres policies.

#347 suggested also caching `endpoint_type` lookups, but this seems unnecessary in practice for `PreservedObjectHandler`:  right now, in our usage of `PreservedObjectHandler`, all its instances get passed the same `endpoint` instance, and active record does a lazy lookup to get the associated `endpoint_type` the first time, and then caches the result on the object.  example from pry, notice only one `SELECT` on `endpoint_types`:
```ruby
[1] pry(main)> 
[2] pry(main)> ep = Endpoint.first
  Endpoint Load (2.0ms)  SELECT  "endpoints".* FROM "endpoints" ORDER BY "endpoints"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> #<Endpoint:0x00007fd7cdc3dc08
 id: 1,
 endpoint_name: "fixture_sr1",
 created_at: Tue, 05 Dec 2017 21:26:47 UTC +00:00,
 updated_at: Tue, 05 Dec 2017 21:26:47 UTC +00:00,
 endpoint_node: "localhost",
 storage_location: "spec/fixtures/storage_root01/moab_storage_trunk",
 recovery_cost: 1,
 access_key: nil,
 endpoint_type_id: 1>
[3] pry(main)> 
[4] pry(main)> ep.endpoint_type.endpoint_class
  EndpointType Load (1.0ms)  SELECT  "endpoint_types".* FROM "endpoint_types" WHERE "endpoint_types"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
=> "online"
[5] pry(main)> 
[6] pry(main)> ep.endpoint_type.endpoint_class
=> "online"
[7] pry(main)> 
[8] pry(main)> ep.endpoint_type               
=> #<EndpointType:0x00007fd7d4110568
 id: 1,
 type_name: "online_nfs",
 endpoint_class: "online",
 created_at: Tue, 05 Dec 2017 21:26:47 UTC +00:00,
 updated_at: Tue, 05 Dec 2017 21:26:47 UTC +00:00>
[9] pry(main)> 
```

as such, i felt like we were best off punting on that caching for now, since we're getting that for free in our current usage.

closes #347

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -128,7 +128,9 @@ RSpec/AnyInstance:
 
 RSpec/BeforeAfterAll:
   Exclude:
-    - 'spec/load_fixtures_helper.rb' # being careful to avoid state leakage
+    # we use after(:all) specifically to prevent state leakage
+    - 'spec/load_fixtures_helper.rb'
+    - 'spec/models/preservation_policy_spec.rb'
 
 RSpec/ExampleLength:
   Max: 50
--- a/app/models/preservation_policy.rb
+++ b/app/models/preservation_policy.rb
@@ -9,6 +9,11 @@ class PreservationPolicy < ApplicationRecord
   validates :archive_ttl, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :fixity_ttl, presence: true, numericality: { only_integer: true, greater_than: 0 }
 
+  # this is a *very* naive cache eviction algorithm: if any pres policy changes, clear
+  # the cache.  we expect pres policies to change very infrequently, so probably no big deal.
+  # we can get smarter if/when we need to.
+  after_save { |_record| self.class.send(:clear_id_cache) }
+
   # iterates over the preservation policies enumerated in the settings, creating any that don't already exist.
   # returns an array with the result of the ActiveRecord find_or_create_by! call for each settings entry (i.e.,
   # the PreservationPolicy rows defined in the config, whether newly created by this call, or previously created).
@@ -25,5 +30,17 @@ class PreservationPolicy < ApplicationRecord
 
   def self.default_preservation_policy
     find_by!(preservation_policy_name: Settings.preservation_policies.default_policy_name)
+  end
+
+  def self.cached_default_preservation_policy_id
+    id_cache[:default_preservation_policy_id] ||= default_preservation_policy.id
+  end
+
+  private_class_method def self.id_cache
+    @id_cache ||= {}
+  end
+
+  private_class_method def self.clear_id_cache
+    @id_cache = {}
   end
 end
--- a/app/models/preservation_policy.rb
+++ b/app/models/preservation_policy.rb
@@ -28,12 +28,15 @@ class PreservationPolicy < ApplicationRecord
     end
   end
 
-  def self.default_preservation_policy
+  def self.default_policy
     find_by!(preservation_policy_name: Settings.preservation_policies.default_policy_name)
   end
 
-  def self.cached_default_preservation_policy_id
-    id_cache[:default_preservation_policy_id] ||= default_preservation_policy.id
+  # note that this is cached.  prefer calling this method over .default_policy if an id
+  # will suffice (e.g., you only need the id, not the whole object, to set an association
+  # to a pres policy on another ActiveRecord object).
+  def self.default_policy_id
+    id_cache[:default_policy_id] ||= default_policy.id
   end
 
   private_class_method def self.id_cache
--- a/spec/models/preservation_policy_spec.rb
+++ b/spec/models/preservation_policy_spec.rb
@@ -56,21 +56,21 @@ RSpec.describe PreservationPolicy, type: :model do
     end
   end
 
-  describe '.default_preservation_policy' do
+  describe '.default_policy' do
     it 'returns the default preservation policy object' do
       # db already seeded
-      expect(PreservationPolicy.default_preservation_policy).to be_a_kind_of PreservationPolicy
+      expect(PreservationPolicy.default_policy).to be_a_kind_of PreservationPolicy
     end
 
     it "raises RecordNotFound if the default policy doesn't exist in the db" do
       # a bit contrived, but just want to test that lack of default PreservationPolicy causes lookup to
       # fail fast.  since db is already seeded, we just make it look up something that we know isn't there.
       allow(Settings.preservation_policies).to receive(:default_policy_name).and_return('nonexistent')
-      expect { PreservationPolicy.default_preservation_policy }.to raise_error(ActiveRecord::RecordNotFound)
+      expect { PreservationPolicy.default_policy }.to raise_error(ActiveRecord::RecordNotFound)
     end
   end
 
-  describe '.cached_default_preservation_policy_id' do
+  describe '.default_policy_id' do
     # clear the cache before each test to reset
     before { PreservationPolicy.send(:clear_id_cache) }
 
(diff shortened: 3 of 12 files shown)


jmartin-sul:
this felt pretty orthogonal to #369, #375, and #379, so this is just based on master as of this PR.

.rubocop.yml
```diff
@@ -129,6 +129,7 @@ RSpec/AnyInstance:
 RSpec/BeforeAfterAll:
   Exclude:
     - 'spec/load_fixtures_helper.rb' # being careful to avoid state leakage
+    - 'spec/models/preservation_policy_spec.rb' # the after(:all) here specifically prevents state leakage
```
ndushay: both of the comments here say the same thing ... is the shorter one sufficient??

spec/models/preservation_policy_spec.rb
```diff
@@ -69,4 +69,46 @@
...
+    it 'returns the default preservation policy id' do
+      default_pres_policy_id = PreservationPolicy.default_preservation_policy.id
+      expect(PreservationPolicy.cached_default_preservation_policy_id).to eq default_pres_policy_id
+    end
+
+    it "doesn't re-run the query if a cached value is available" do
+      expect(PreservationPolicy).to receive(:find_by!).once.and_call_original
+      PreservationPolicy.cached_default_preservation_policy_id
+      PreservationPolicy.cached_default_preservation_policy_id
+    end
```
ndushay: I love this test.

app/models/preservation_policy.rb
```diff
@@ -26,4 +31,16 @@ def self.seed_from_config
   def self.default_preservation_policy
     find_by!(preservation_policy_name: Settings.preservation_policies.default_policy_name)
   end
+
+  def self.cached_default_preservation_policy_id
```
ndushay: question:  do the callers need to know it's cached?  would we be okay with naming this method `default_pp_id`, for example, or `default_policy_id` or `default_id`?  

I would also be in favor of changing `default_preservation_policy` method name to `default_policy` ...

SaravShah: +1 to changing it to `default_policy`

jmartin-sul: sure, `default_policy_id` and `default_policy` it is.  i sorta like making it explicit that it's cached, but i don't feel strongly about that, and am happy to leave it out of the name if others would prefer.  could also leave a comment.

tallenaz: I'm in favor of leaving a comment about caching if, as it looks likely, we call this `default_policy`, if only b/c `id_cache` seems like a really terse method name to me, but I wouldn't demand a comment.

jmartin-sul:
i believe i have implemented both of @ndushay's suggestions, and have just re-pushed that as follow on commits.

app/models/preservation_policy.rb
jmartin-sul: i left a comment in my follow-on commits

app/models/preservation_policy.rb
```diff
@@ -9,6 +9,11 @@ class PreservationPolicy < ApplicationRecord
   validates :archive_ttl, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :fixity_ttl, presence: true, numericality: { only_integer: true, greater_than: 0 }
 
+  # this is a *very* naive cache eviction algorithm: if any pres policy changes, clear
+  # the cache.  we expect pres policies to change very infrequently, so probably no big deal.
```
tingulfsen: why did you decide to roll your own rather than use Rails built-in caching? is there something about it that won't work for us?

jmartin-sul: @tingulfsen:  my understanding from everything i found online and from @ndushay's assessment is that the rails caching is very focused on caching http responses.  the one library i found that leveraged the rails caching infrastructure for general caching didn't seem very well maintained.

i'm not sure that the rails caching machinery would result in less code, or more readable code, though i do think that if we find ourselves doing more caching, we should more thoroughly explore off the shelf patterns and/or gems, or at least run our code by someone w/ more ActiveRecord experience, to tell us if we're walking into some anti-pattern.

also, if i totally misunderstood rails caching, or missed a library that we could've used, i'm totally open to suggestions.

but i basically took the general approach you used in #351, and added cache eviction, and just cached the id instead of the whole object.  the discussion on that one seemed to indicate that we weren't necessarily itching for a heavier caching solution (or maybe i was misinterpreting that).

jmartin-sul: my feeling was that since the solution is short and relatively readable and hopefully correct and performant for our usage, it felt like "good enough"/"shameless green".

.rubocop.yml
```diff
@@ -128,8 +128,9 @@ RSpec/AnyInstance:
 
 RSpec/BeforeAfterAll:
   Exclude:
-    - 'spec/load_fixtures_helper.rb' # being careful to avoid state leakage
-    - 'spec/models/preservation_policy_spec.rb' # the after(:all) here specifically prevents state leakage
+    # we use after(:all) specifically to prevent state leakage
```
ndushay: haha - this is fine;  I was thinking of keeping the comments after the file name and then repeating it.

jmartin-sul:
@ndushay: squashed into what felt like 3 logical commits, happy to squash into one if you'd prefer (or feel free to do that w/ the github button).  also, rebased on current master.

PR #382, opened 2017-12-07
Update readme

### LINKED ISSUE #381: (PC) README clarity
https://github.com/sul-dlss/preservation_catalog/pull/375

Remove `(note that the argument is not a string):` on line 87

line 93: Naomi says : 
" I would prefer to have the "what will it do" in the same position for each of these rake task explanations - preferably before. I don't want to read it; I want to skim it. I might even have a ##### heading for each task? "

SaravShah:
Closes #381

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,15 +82,17 @@ WARNING! this will erase the catalog, and thus require re-seeding from scratch. 
   Basically an especially inconvenient confirmation dialogue.  For safety's sake, the full command that skips that warning can be constructed by the user as needed, so as to prevent unintentional copy/paste dismissal when the user might be administering multiple deployment environments simultaneously.  Inadvertent database wipes are no fun.
   * `db:reset` will make sure db is migrated and seeded.  If you want to be extra sure: `RAILS_ENV=[environment] bundle exec rake db:migrate db:seed`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -82,15 +82,14 @@ WARNING! this will erase the catalog, and thus require re-seeding from scratch.
   Basically an especially inconvenient confirmation dialogue.  For safety's sake, the full command that skips that warning can be constructed by the user as needed, so as to prevent unintentional copy/paste dismissal when the user might be administering multiple deployment environments simultaneously.  Inadvertent database wipes are no fun.
   * `db:reset` will make sure db is migrated and seeded.  If you want to be extra sure: `RAILS_ENV=[environment] bundle exec rake db:migrate db:seed`
 
-#### Drop or populate the catalog for a single endpoint
 
-There are rake tasks for clearing out all database entries for a single endpoint. To run either of the rake tasks, invoke them as follows, prepending the `RAILS_ENV` and giving the name of the endpoint as an argument (note that the argument is not a string):
+#### Drop all database entries for a single endpoint:
```
ndushay: do we need to include the bit "giving the name of the endpoint as an argument (note that the argument is not a string)" or some such?

SaravShah: good point

Issue #286, opened 2017-11-07
(M2C)  check_existence timestamps to be updated (blocked by / simult w   #284)


ndushay:
PC timestamps:
if check_existence clean:
- [ ] last_checked_on_storage AND 
- [ ] last_audited AND 
- [ ] updated_at (but ActiveRecord does this, right?)

PO timestamp(s) changes only when check_existence:
- calls create_with_validation (already done in plain create method, timestamp info in ticket #288)
- calls update_version_with_validation (timestamp info already in ticket #287)

See #284

ndushay:
It is believed that these timestamps will be updated in confirm_version happy path;  when confirm_version is NOT in happy path ... should timestamps be updated by confirm_version?

When check_existence calls confirm version or update_version_after_vdn or create_after_vdn, these timestamp updates will be done within those methods, no?

So again, it's the same question for the UNHAPPY path for both confirm_version and check_existence -- do we update the timestamps?

ndushay:
Believe this was addressed with PR for #284  cloing

PR #401, opened 2017-12-12
pohandler.confirm_version re-implemented

### LINKED ISSUE #353: (M2C) confirm_version needs to use update_version_after_validation

### LINKED ISSUE #384: (PC) #confirm_version timestamp update
PC and PO update timestamps...they are done with
-  create_with_validation (already done in plain create method, timestamp info in ticket #288)
-  calls update_version_with_validation (timestamp info already in ticket #287)

ndushay:
Shameless green implementation of pohandler.confirm_version, which is supposed to:

- update timestamps if the versions match
- error if the versions don't match (done via the result_array returned)

Closes #353 
Closes #384
(and may also be the implementation of #286)


spec/services/preserved_object_handler_confirm_version_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+        allow(pc).to receive(:status).and_return(status)
+        allow(pc).to receive(:last_audited=)
+        allow(pc).to receive(:last_checked_on_storage=)
+        allow(pc).to receive(:changed?).and_return(true)
+        allow(pc).to receive(:save!)
+        po_handler.confirm_version
+        expect(po).not_to have_received(:save!)
+        expect(pc).to have_received(:save!)
+      end
+      it 'calls PreservedObject.touch and PreservedCopy.touch if the existing record is NOT altered' do
```
jmartin-sul: looks like title should actually be `'calls PreservedCopy.touch (but not PreservedObject.touch) if the existing record is NOT altered'` (or similar)

spec/services/preserved_object_handler_confirm_version_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+          context 'changed' do
+            it 'last_audited' do
+              orig = pc.last_audited
+              po_handler.confirm_version
+              expect(pc.reload.last_audited).not_to eq orig
+            end
+            it 'last_checked_on_storage' do
+              orig = pc.last_checked_on_storage
+              po_handler.confirm_version
+              expect(pc.reload.last_checked_on_storage).not_to eq orig
```
jmartin-sul: i think i'd have expected `.to be > orig`, as in the subsequent tests (ditto the test before this one)

spec/services/preserved_object_handler_confirm_version_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+            end
+            it 'last_audited' do
+              orig = pc.last_audited
+              po_handler.confirm_version
+              expect(pc.reload.last_audited).not_to eq orig
+            end
+            it 'last_checked_on_storage' do
+              orig = pc.last_checked_on_storage
+              po_handler.confirm_version
+              expect(pc.reload.last_checked_on_storage).not_to eq orig
```
jmartin-sul: i think i'd have expected `.to be > orig`, as in the subsequent tests (ditto the test before this one)

spec/services/preserved_object_handler_confirm_version_spec.rb
ndushay: this would be a good change, except in this test, the orig value is nil.

spec/services/preserved_object_handler_confirm_version_spec.rb
ndushay: good catch - meant to change that test description.

ndushay:
@jmartin-sul I believe I have addressed your three concerns above.  Thank you for your careful review.

spec/services/preserved_object_handler_spec.rb
```diff
@@ -363,16 +76,15 @@
...
         preserved_object: po,
         version: po.current_version,
         size: 1,
         endpoint: ep,
         status: PreservedCopy::DEFAULT_STATUS
       )
       bad_po_handler = described_class.new(druid, 6, incoming_size, ep)
-      # NOTE: #increase_version checks class of object (rspec double != PreservedCopy)
-      allow_any_instance_of(PreservedCopy).to receive(:save!).and_raise(ActiveRecord::ActiveRecordError)
+      allow(pc).to receive(:save!).and_raise(ActiveRecord::ActiveRecordError)
```
SaravShah: 👍

spec/services/preserved_object_handler_confirm_version_spec.rb
jmartin-sul: yup, figured i was missing something!

PR #407, opened 2017-12-13
README tweaks

ndushay:


README.md
```diff
@@ -84,44 +88,53 @@ WARNING! this will erase the catalog, and thus require re-seeding from scratch.
...
 ```
 
 ### Run Moab to Catalog existence check for a single root and for all storage roots
 
+- To run rake tasks below, give the name of the moab storage_root (e.g. from settings/development.yml) as an argument.
+
+- You can monitor the progress of the rake task by tailing `log/production.log`.  The rake task may take many hours.
+
+- The following rake tasks should be run from the root directory of the project, with whatever `RAILS_ENV` is appropriate. Because the task can take hours when run over all storage roots, consider running it in a [screen session](http://thingsilearned.com/2009/05/26/gnu-screen-super-basic-tutorial/), or with "nohup" (inserted just before bundle exec).
+
```
tingulfsen: Is there a way to not repeat entire paragraphs here? This README is already long...

PR #408, opened 2017-12-13
Add check_existence to PreservedObjectHandler

### LINKED ISSUE #284: (M2C) (PC) check_existence method in pohandler (blocked by #353, #384)
check_existence method will:

- [ ] call confirm_version
- [ ] if object not found error, calls create_after_validation
- [ ] if Moab version > catalog, calls update_version_after_validation
- [ ] else error

(timestamp info also in #286)

PC timestamp changes:
if check_existence clean:
- [ ] last_checked_on_storage AND 
- [ ] last_audited AND 
- [ ] updated_at (but ActiveRecord does this, right?)

PO timestamp(s) changes only when check_existence:
- calls create_after_validation (already done in plain create method, timestamp info in ticket #288)
- calls update_version_after_validation (timestamp info already in ticket #287)

- [ ] lib/audit code will call check_existence method instead of confirm_version

### LINKED ISSUE #394: (M2C) lib/audit/m2c single root method calls pohandler.check_existence
Currently lib/audit/m2c, when checking existence, has some logic.  It should become stupider, more like seed from single disk, and the logic should be part of check_existence method (see #284)

- [ ] do for single storage root
- [ ] remove boolean argument for create_if_does_not_exist
- [x] ensure call for all storage roots follows this approach. (already done)

tallenaz:
Fixes #284 
Fixes #394 

Connects to #286 

Question for reviewers: on the "unhappy path", specifically, where the po and pc versions don't match, and we don't even look at the moab on disk, do we want to update timestamps? Currently, we don't, though it seems like on all other "unhappy paths", we do.


app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,19 @@ def create
   end
 
   def check_existence
-    # this is just a stub
+    if invalid?
+      handler_results.add_result(PreservedObjectHandlerResults::INVALID_ARGUMENTS, errors.full_messages)
+    elsif PreservedObject.exists?(druid: druid)
+      confirm_version
+      if handler_results.contains_result_code? PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT
+        update_version_after_validation
```
SaravShah: see ticket #410

tingulfsen: Should we just make confirm_version return a value for you to use so we don't have to search our own results? I realize we talked about this, but now that I see the code, it just seems a little weird.

jmartin-sul: i sorta like that we can just examine the results we've already collected, so i think i have a slight preference for this, but i'd be totally fine with switching if you have a strong preference for using a return value, i'd be open to switching.

actually, one other thing in favor of this, is that `confirm_version` already returns the result array, which is structurally consistent with other similar methods, which i like.  we could look at that result array, but that's actually what tony and i had started out doing, and it looked a little messy, and so we ended up at this new `contains_result_code?` method.

ndushay: `confirm_version` and all the similar public methods must return the results array so we can turn them into ReST calls with meaningful HTTP response codes for errors.  That's what started us down this whole results array approach in the first place.

spec/lib/audit/moab_to_catalog_spec.rb
```diff
@@ -123,38 +123,18 @@
...
+      it 'calls check_existence' do
+        fake_codes = %w[fake_code1 fake_code2]
         expected_argument_list.each do |arg_hash|
-          expect(PreservedObject).to receive(:exists?).with(druid: arg_hash[:druid]).and_return(true)
-          expect(arg_hash[:po_handler]).to receive(:confirm_version)
+          expect(arg_hash[:po_handler]).to receive(:check_existence).and_return(fake_codes)
         end
-        subject
+        # * 3 will magically give us a flat, 6 element array
+        expect(subject).to eq(fake_codes * 3)
```
SaravShah: seems like this and line 137 are kind of doing the same thing?
ok to keep it.. Why does subject give us ` ["fake_code1", "fake_code2", "fake_code1", "fake_code2", "fake_code1", "fake_code2"] `? maybe im not understanding this test?

app/services/preserved_object_handler.rb
ndushay: calling the public method 'update_version_after_validation' here is likely contributing to double logging and double db lookups.

spec/lib/audit/moab_to_catalog_spec.rb
SaravShah: Is it to show that `check_existence` returns an array of codes?

app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,19 @@ def create
   end
 
   def check_existence
-    # this is just a stub
+    if invalid?
+      handler_results.add_result(PreservedObjectHandlerResults::INVALID_ARGUMENTS, errors.full_messages)
+    elsif PreservedObject.exists?(druid: druid)
+      confirm_version
```
ndushay: If we call confirm_version immediately, it will do the argument validation, and it avoids the extra call to "PreservedObject.exists?" as well. Another approach would be to call confirm_version itself but to call the private methods it uses.

app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,19 @@ def create
...
+      if handler_results.contains_result_code? PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT
+        update_version_after_validation
+      end
+    else
+      handler_results.add_result(PreservedObjectHandlerResults::OBJECT_DOES_NOT_EXIST, 'PreservedObject')
+      create_after_validation
```
ndushay: again, calling the public method 'create_after_validation' here is likely contributing to double logging and double lookups.

spec/services/preserved_object_handler_spec.rb
```diff
@@ -67,9 +67,78 @@
   end
 
   describe '#check_existence' do
-    # FIXME: move the actual specs to a separate file?
```
ndushay: I believe we would like the check_existence specs to live in a separate file, as indicated by the comment?

spec/services/preserved_object_handler_spec.rb
```diff
@@ -67,9 +67,78 @@
...
-    it 'stubbed spec for now' do
-      skip
+    context "(calls create or confirm_version)" do
+      it 'calls #create_after_validation when the object does not exit' do
```
ndushay: as stated above, I don't think we want to test that it calls the method, as much as we want to test that the outcome is what we want, with the correct db changes and the correct result_array.  This sort of test is how we missed duplicate stuff.

spec/services/preserved_object_handler_spec.rb
```diff
@@ -67,9 +67,78 @@
...
+        expect(PreservedObject).to receive(:exists?).with(druid: druid).and_return(false)
+        no_exist_msg = "#{exp_msg_prefix} PreservedObject db object does not exist"
+        expect(Rails.logger).to receive(:log).with(Logger::ERROR, no_exist_msg)
+        # because create_after_validation isn't called, we don't check
+        # for the usual result codes it would return (doesn't exist and add to the db)
+        expect(po_handler).to receive(:create_after_validation)
+        po_handler.check_existence
+      end
+
+      it 'calls confirm_version when the object exists' do
```
ndushay: as stated above, I don't think we want to test that it calls the method, as much as we want to test that the outcome is what we want, with the correct db changes and the correct result_array. This sort of test is how we missed duplicate stuff.

spec/services/preserved_object_handler_spec.rb
```diff
@@ -67,9 +67,78 @@
...
+        expect(PreservedObject).to receive(:exists?).with(druid: druid).and_return(true)
+        expect(po_handler).to receive(:confirm_version)
+        po_handler.check_existence
+      end
+
+      it 'calls update_version_after_validation when confirm_version returns ARG_VERSION_GREATER_THAN_DB_OBJECT' do
```
ndushay: as stated above, I don't think we want to test that it calls the method, as much as we want to test that the outcome is what we want, with the correct db changes and the correct result_array. This sort of test is how we missed duplicate stuff.

spec/services/preserved_object_handler_spec.rb
```diff
@@ -67,9 +67,78 @@
...
+      let(:updated_pc_db_msg) { "#{exp_msg_prefix} PreservedCopy db object updated" }
+
+      before do
+        allow(po_handler).to receive(:moab_validation_errors).and_return([])
+        allow(PreservedObject).to receive(:exists?).with(druid: druid).and_return(true)
+        allow(po_handler).to receive(:confirm_version) do
+          po_handler.handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, 'PreservedObject')
+          po_handler.handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, 'PreservedCopy')
+          po_handler.handler_results.add_result(PreservedObjectHandlerResults::UPDATED_DB_OBJECT, 'PreservedCopy')
+        end
```
ndushay: I think mocking this is NOT an appropriate test, because we really do want to check the logic -- did executive return the correct results.  I also think we need to test results for each of the cases.

ndushay:
for the record, in discussion with @tallenaz and @SaravShah; I am embarking on a follow-on commit with new tests in a separate file, to replace the existing tests.

PR #411, opened 2017-12-13
quick fix up

SaravShah:


README.md
```diff
@@ -109,8 +109,7 @@ bundle exec rake m2c_exist_single_root[fixture_sr1]
 ```ruby
 bundle exec rake m2c_exist_single_root[fixture_sr1,profile]
 ```
-this will generate a log at, for example, ```log/profiler_flat_check_existence_for_dir2017-12-11T14:34:06-flat.txt
-```
+this will generate a log at, for example, ```log/profiler_flat_check_existence_for_dir2017-12-11T14:34:06.txt```
```
ndushay: I believe, besides fixing the README, you will need to fix the actual code generating this file name

Rakefile
```diff
@@ -88,7 +88,7 @@ task :m2c_exist_single_root, [:storage_root, :profile] => [:environment] do |_t,
   elsif args[:profile].nil?
     MoabToCatalog.check_existence_for_dir(storage_dir)
   end
-  puts "#{Time.now.utc.iso8601} Moab to Catalog Existence Check for a single storage root is done"
+  puts "#{Time.now.utc.iso8601} Moab to Catalog Existence Check for #{storage_dir} is done"
```
ndushay: this looks awesome - can you do it for the single root seeding, too?

README.md
jmartin-sul: on that note, i think i'd prefer that we get rid of the first `_flat` and leave the last `-flat`, since the latter is put there by the profiler code (which seems like the more logical bit of code to note the type of profile data in the logfile name, esp since its also the code responsible for constructing the timestamp and file extension parts of the file name).

SaravShah: i removed the `_flat`  and left the `-flat` in.

Issue #403, opened 2017-12-12
(PC) PO and PC version mismatch error should indicate versions in error message


ndushay:
https://github.com/sul-dlss/preservation_catalog/blob/master/app/services/preserved_object_handler_results.rb#L38

should be explicit about the versions it finds

- [ ] `confirm_version`
- [ ] `check_existence`

PR #423, opened 2017-12-15
Add check_existence to PreservedObjectHandler

### LINKED ISSUE #284: (M2C) (PC) check_existence method in pohandler (blocked by #353, #384)

### LINKED ISSUE #394: (M2C) lib/audit/m2c single root method calls pohandler.check_existence

tallenaz:
note that `PreservedObjectHandler#check_existence` errs very much on the side of shameless green:  it passes the tests, and while there's clearly some similarity with other POH code, we didn't see immediately obvious/easy refactorings (the sorta-repetition is a result of extracting code from existing methods that did almost what we wanted, and then altering the extracted code to get the actual desired behavior).

our hope is that follow on tickets can find new small shared behaviors to extract into helper methods, or something like that.

@jmartin-sul paired w/ @tallenaz on the initial implementation, and then @ndushay and @SaravShah paired on the tests that showed us what was wrong with that implementation.  @tallenaz and @jmartin-sul then fixed the original implementation to conform to the expected behavior the new tests defined.  then everything was squashed into a cleaner commit history.

so TLDR, we definitely want @tingulfsen to review since he's the freshest set of eyes on this.

Fixes #284 
Fixes #394 
Connects to #286


app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,66 @@ def create
...
+          if pres_copy.version != pres_object.current_version
+            handler_results.add_result(PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH, pres_copy.class.name)
+            raise ActiveRecord::Rollback, 'PO current_version != PC version'
+          end
+
+          if incoming_version == pres_copy.version
+            handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_copy.class.name)
+            handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_object.class.name)
+            update_pc_validation_timestamps(pres_copy)
+            update_db_object(pres_copy)
```
SaravShah: Do we want to update `pres_obj` as well? Right now `pres_obj` is only updated when the incoming version is less than the PC or PO . @ndushay @tingulfsen

app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,66 @@ def create
...
+                update_pc_validation_timestamps(pres_copy)
+                update_db_object(pres_copy)
+              end
+            elsif endpoint.endpoint_type.endpoint_class == 'archive'
+              # TODO: perform archive object validation; then create a new PC record for the new
+              #  archived version on the endpoint
+            end
+          end
+        end
+        handler_results.remove_db_updated_results unless transaction_ok
```
jmartin-sul: ok, now something jumps out at me as a possible bit of refactoring:  lines 68-109 could be extracted into `check_online_existence`, just to make `check_existence` a bit more readable, and to follow a pattern we use for the other public methods in the class.  not sure if that'd be helpful in practice, but seemed worth throwing out there.

spec/services/shared_examples_preserved_object_handler.rb
```diff
@@ -71,17 +71,20 @@
 
 RSpec.shared_examples 'druid not in catalog' do |method_sym|
   let(:druid) { 'rr111rr1111' }
-  let(:exp_msg) { "#{exp_msg_prefix} #<ActiveRecord::RecordNotFound: Couldn't find PreservedObject> db object does not exist" }
+  # let(:exp_msg) { "#{exp_msg_prefix} #<ActiveRecord::RecordNotFound: Couldn't find PreservedObject> db object does not exist" }
```
jmartin-sul: whoops, looks like we left in a bit of commented code for an expectation that we loosened slightly (since sometimes this error message comes as a result of a trapped exception, and sometimes as a result of a pro-active check).  happy to either fix this PR or remove in a tiny follow-on PR.  ditto the next two lines of commented code in this file.

app/services/preserved_object_handler.rb
tingulfsen: Why don't you make a ticket for that?

app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,66 @@ def create
   end
 
   def check_existence
```
ndushay: it's probably work adding a comment such as
`# this is a long, complex method (shameless green);  if it is refactored, revisit the exceptions in rubocop.yml`

app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,66 @@ def create
...
+    elsif PreservedObject.exists?(druid: druid)
+      Rails.logger.debug "check_existence #{druid} called"
+      if endpoint.endpoint_type.endpoint_class == 'online'
+        transaction_ok = with_active_record_transaction_and_rescue do
+          pres_object = PreservedObject.find_by!(druid: druid)
+          # FIXME: what if there is more than one associated pres_copy?
+          pres_copy = PreservedCopy.find_by!(preserved_object: pres_object, endpoint: endpoint) if pres_object
+
+          if pres_copy.version != pres_object.current_version
+            handler_results.add_result(PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH, pres_copy.class.name)
```
ndushay: added info to #403  to indicate the actual versions in this message (and maybe rollback message) -- that should happen both here and in confirm_version (#403 was created stemming from that PR review)

app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,66 @@ def create
...
+            handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_copy.class.name)
+            handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_object.class.name)
+            update_pc_validation_timestamps(pres_copy)
+            update_db_object(pres_copy)
+          else
+            version_comp_result = if incoming_version > pres_copy.version
+                                    PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT
+                                  else
+                                    PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT
+                                  end
```
ndushay: unfortunately, I believe this is incorrect implementation.

I believe we need to proceed to do the changes to the catalog below when incoming > catalog, but we are to report the error ("loudly") if the incoming < catalog -- we are definitely NOT to update the version.

I suspect the specs written by Sarav and I were changed in a mistaken understanding of the expected behavior here.  

In the interests of forward progress, I have created a new ticket about this: #424.

spec/services/shared_examples_preserved_object_handler.rb
ndushay: Slight preference for you to fix it here.

app/services/preserved_object_handler.rb
ndushay: no - I don't think we want to update the version on either object (PO or PC) when incoming version < catalog version.

app/services/preserved_object_handler.rb
ndushay: yes - i think there may be some other wins along these lines ...

jmartin-sul:
here's a branch that has the commit history from before we squashed and cleaned things up:
https://github.com/sul-dlss/preservation_catalog/compare/284-check-exist-in-poh_pre-commit-squashing

the only difference between the code in this branch and the code in that branch is the removal of the `contains_result_code?` method from our first pass at this, which we didn't end up using in the final implementation.  for whatever reason, this is clearer when diffing from the command line.  the github diff, at least for me, displays differently, and shows code that's in both versions as being added between the two.  i'd be curious whether others saw this when they pulled the latest code and diffed from command line:
```shell
$ git checkout 284-check-exist-in-poh_final_cut 
Switched to branch '284-check-exist-in-poh_final_cut'
Your branch is up-to-date with 'origin/284-check-exist-in-poh_final_cut'.
$ 
$ git diff 284-check-exist-in-poh_pre-commit-squashing 
diff --git a/app/services/preserved_object_handler_results.rb b/app/services/preserved_object_handler_results.rb
index a2d70fe..cc0aa96 100644
--- a/app/services/preserved_object_handler_results.rb
+++ b/app/services/preserved_object_handler_results.rb
@@ -85,10 +85,6 @@ class PreservedObjectHandlerResults
     { code => result_code_msg(code, msg_args) }
   end
 
-  def contains_result_code?(code)
-    result_array.detect { |result_hash| result_hash.keys.include?(code) } != nil
-  end
-
diff --git a/spec/services/preserved_object_handler_results_spec.rb b/spec/services/preserved_object_handler_results_spec.rb
index 6e9c929..984762b 100644
--- a/spec/services/preserved_object_handler_results_spec.rb
+++ b/spec/services/preserved_object_handler_results_spec.rb
@@ -57,14 +57,4 @@ RSpec.describe PreservedObjectHandlerResults do
       skip
     end
   end
-  context '#contains_result_code?' do
-    it 'returns true if the result code is there, false if not' do
-      expect(pohr.result_array.size).to eq 0
-      added_code = PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH
-      other_code = PreservedObjectHandlerResults::VERSION_MATCHES
-      pohr.add_result(added_code)
-      expect(pohr.contains_result_code?(added_code)).to eq true
-      expect(pohr.contains_result_code?(other_code)).to eq false
-    end
-  end
 end
$ 
```

the commits showing the changes that @tallenaz and i made to @ndushay and @SaravShah's tests are:
* https://github.com/sul-dlss/preservation_catalog/commit/fb7d40dccd931b9f16c9a48b15c4d332632c355d#diff-a86addf91eaa9d589fbe0483306dc0cc
* https://github.com/sul-dlss/preservation_catalog/commit/256fe4ec2f22d12b2977359e51ccaa2975e335bc (actually, this is us getting rid of some of our original bad tests)
* https://github.com/sul-dlss/preservation_catalog/commit/8add38937452c909e35a385bebffa691ba467111#diff-a86addf91eaa9d589fbe0483306dc0cc
* https://github.com/sul-dlss/preservation_catalog/commit/8add38937452c909e35a385bebffa691ba467111#diff-d5f237be9d502ca2392dc5de77b125d4
* https://github.com/sul-dlss/preservation_catalog/commit/99ebf5cf82e15a5ce9b4bea7a68787e84c9a633e#diff-a86addf91eaa9d589fbe0483306dc0cc
* https://github.com/sul-dlss/preservation_catalog/commit/99ebf5cf82e15a5ce9b4bea7a68787e84c9a633e#diff-b94b0af9a8066f6ccc50d5cde120432b (us getting rid of the rest of our original bad tests)
* https://github.com/sul-dlss/preservation_catalog/commit/830215b81e7410073277c635eb980d0937c4fb27#diff-a86addf91eaa9d589fbe0483306dc0cc (turn explicit `fail` tests to `skip`)

app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,66 @@ def create
...
+            handler_results.add_result(version_comp_result, pres_copy.class.name)
+            handler_results.add_result(version_comp_result, pres_object.class.name)
+            if endpoint.endpoint_type.endpoint_class == 'online'
```
tingulfsen: You already checked the endpoint_class on L67...so why check it again here?

app/services/preserved_object_handler.rb
jmartin-sul: d'oh.  yeah, i think we got this case wrong, in that we always update version.  good catch.

app/services/preserved_object_handler.rb
tallenaz: See issue https://github.com/sul-dlss/preservation_catalog/issues/425 for refactoring 68-109

app/services/preserved_object_handler.rb
```diff
@@ -60,7 +60,66 @@ def create
...
+            update_db_object(pres_copy)
+          else
+            version_comp_result = if incoming_version > pres_copy.version
+                PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT
+              else
+                PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT
+              end
+            handler_results.add_result(version_comp_result, pres_copy.class.name)
+            handler_results.add_result(version_comp_result, pres_object.class.name)
+            if endpoint.endpoint_type.endpoint_class == 'online'
```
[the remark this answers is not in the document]
jmartin-sul: yup, this was copypasta, we built up `check_existence` v2 by extracting code from helper methods that almost did what we wanted, and then changing that code.  two of the methods we used had that check.

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -0,0 +1,756 @@
...
+              before do
+                allow(po_handler).to receive(:moab_validation_errors).and_return([])
+              end
+              it 'status if former status was ok' do
+                pc.status = PreservedCopy::OK_STATUS
+                pc.save!
+                po_handler.check_existence
+                expect(pc.reload.status).to eq PreservedCopy::OK_STATUS
+              end
+              it 'size if incoming size is nil' do
```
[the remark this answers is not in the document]
jmartin-sul: looked at this with @tallenaz, and we think this is as it should be.  it appears that we test non-nil size above:  https://github.com/sul-dlss/preservation_catalog/pull/423/files/0c173915c5ecc6045581a9fcc72104306344b640..361d44c0385f5b9e8539fa73fb7cf8031297e806#diff-a86addf91eaa9d589fbe0483306dc0ccR139

and that the tests are grouped by "situations where fields should change" and "situations where fields should remain unchanged", since the overall context is that the incoming version is greater than the catalog version and that the moab is valid, which indicates that the preserved copy should change in general (at least its `version` and `updated_at` fields, which are notably absent from the `context 'unchanged'` section).

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -0,0 +1,756 @@
...
+              expect(results).to include(a_hash_including(code => version_gt_pc_msg))
+              expect(results).to include(a_hash_including(code => version_gt_po_msg))
+            end
+            it "UPDATED_DB_OBJECT results" do
+              code = PreservedObjectHandlerResults::UPDATED_DB_OBJECT
+              expect(results).to include(a_hash_including(code => updated_pc_db_msg))
+              expect(results).to include(a_hash_including(code => updated_po_db_msg))
+            end
+            it 'what results/logging desired for incoming version > catalog' do
+              fail 'we need clarification of requirements on results/logging in this case'
```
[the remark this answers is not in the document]
jmartin-sul: @tingulfsen, @tallenaz and i aren't sure.

@ndushay and/or @SaravShah, did you have specific concerns about result codes that might be missing, or that might be extraneous in the currently returned values?  or is this maybe a holdover from when the tests were being sketched out?  it does appear that logging and return values are both tested for this scenario, and the tests appear internally consistent (expects four results, looks for four different result codes, checks for four different log statements and the absence of a fifth).

jmartin-sul:
@ndushay, @SaravShah, @tingulfsen:  i think @tallenaz and i have addressed all of the feedback so far, either in reply comments or follow on commits.  we left the commits as additions to what was already there, to make it easier to review.  happy to do some squashing if people want once something gets approved.

spec/services/preserved_object_handler_check_exist_spec.rb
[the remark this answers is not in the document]
ndushay: I think this is more about "we need to speak to Julian" and now we know basically, he would prefer a single log message for a druid.   I'm pondering whether that means we have a single Result code, or maybe just more logic around what we write to logs or something.

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -0,0 +1,756 @@
...
+            Endpoint.find_or_create_by!(endpoint_name: 'bad_fixture_dir') do |endpoint|
+              endpoint.endpoint_type = Endpoint.default_storage_root_endpoint_type
+              endpoint.endpoint_node = Settings.endpoints.storage_root_defaults.endpoint_node
+              endpoint.storage_location = invalid_storage_dir
+              endpoint.recovery_cost = Settings.endpoints.storage_root_defaults.recovery_cost
+            end
+            # these need to be in before loop so it happens before each context below
+            invalid_po = PreservedObject.create!(druid: invalid_druid, current_version: 2, preservation_policy: default_prez_policy)
+            PreservedCopy.create!(
+              preserved_object: invalid_po, # TODO: see if we got the preserved object that we expected
```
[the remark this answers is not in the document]
ndushay: I think there is a lot of copypasta here -- that's a comment I made a looooong time ago on one test and it has propagated everywhere.  Another thing to attack in refactoring, perhaps?

app/services/preserved_object_handler.rb
ndushay: To actually reply properly to Sarav's comment above:   no - we are only indicating that we have audited the pres copy;  nothing to change in the preservedObject in this happy path case.

PR #426, opened 2017-12-16
[WIP] sending test errors to argo

SaravShah:
Note: This is all WIP and i'm just testing everything out.

I'm hard coding the druid in just because I couldn't find our test fixtures in argo-stage. So i'm using `dh966tq9465`. 
This is all work in progress, but i've created a class called `ReportErrorsToArgo` and put the file in `lib/report_errors_to_argo.rb`. Right now I've just added the test code to `poh#moab_validation_errors` - i've been using the druid `jj925bx9565` and adding/renaming version directories to make sure moab validation errors are sent to argo. If you go to https://argo-stage.stanford.edu/view/druid:dh966tq9465 you'll see the most recent error. I also changed `create` to `create_after_validation` in the `moab_to_catalog.rb` just so we can get moab validation errors. What i've been doing in console is this:
```
input> load 'lib/report_errors_to_argo.rb'
input> load 'lib/audit/moab_to_catalog.rb'
input> MoabToCatalog.check_existence_for_dir('spec/fixtures/storage_root01/moab_storage_trunk', true)
```
And then depending on how you manipulate the druid(`jjj925bx9565`) you should see the corresponding validation error. Remember I couldn't find this druid in argo-stage so i'm just sending the validation error to `dh966tq9465`. 

pairing w/ @tingulfsen


tingulfsen:
Just pushed a few minor things up that I *think* takes the work in the right direction.

We might just nuke this PR and start it up from scratch again since the commit history is a bit ugly. But I would really like your feedback on the *basic approach*, especially with respect to code organization/design (forget about the naming etc).

app/services/report_errors_to_argo.rb
```diff
@@ -0,0 +1,24 @@
...
+  WORKFLOW = 'preservationWF'.freeze
+  WORKFLOW_DEF = <<-EOXML
+    <workflow id="preservationWF">
+         <process name="start" />
+         <process name="moab-valid"  />
+         <process name="M2C-check-exist" />
+         <process name="C2M-check-exist" />
+    </workflow>
+    EOXML
+  REPO = "dor".freeze
```
ndushay: I *think* the pattern is to put this stuff in config -- see https://github.com/sul-dlss/was_robot_suite/blob/9fa3d924c62d5b496a305d258ac15c1c9a9731a2/config/workflows/dor/wasCrawlDisseminationWF.xml for example ...

app/services/preserved_object_handler.rb
```diff
@@ -133,6 +133,10 @@ def moab_validation_errors
       end
       handler_results.add_result(PreservedObjectHandlerResults::INVALID_MOAB, moab_error_msgs)
     end
+    if moab_errors.any?
+      reta = ReportErrorsToArgo.new("dh966tq9465")
+      reta.send_errors("moab-valid", "#{moab_errors}")
+    end
```
ndushay: I guess I envisioned the code to actually send the stuff to workflows happening in PreservedObjectHandlerResults ... but perhaps we can discuss pros and cons in person

app/services/report_errors_to_argo.rb
```diff
@@ -0,0 +1,24 @@
+class ReportErrorsToArgo
```
ndushay: I'm inclined to substitute 'workflow' for 'argo' as it's perhaps more accurate and more descriptive.  But wonder if that makes sense when we get around to adding the code to do the solr indexing ... again - perhaps debate pros and cons in person.  `services/dor_workflows.rb` ??

app/services/report_errors_to_argo.rb
```diff
@@ -0,0 +1,24 @@
...
+    </workflow>
+    EOXML
+  REPO = "dor".freeze
+
+  def initialize(druid)
+    @druid = druid
+  end
+
+  def send_errors(workflow_process, error_msg)
+    unless Dor::WorkflowService.get_active_workflows(REPO,"druid:#{@druid}").include?(WORKFLOW)
```
ndushay: ugh - seems sound; wondering about performance.  Guess we'll find out!  Will we need newrelic to help us with knowing wait times for network call?

config/application.rb
```diff
@@ -17,5 +17,7 @@ class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+
+    Dor::WorkflowService.configure(Settings.workflow_url, { 'logger': Logger.new(Settings.workflow_services_log) } )
```
ndushay: wondering if the pattern is for this to live here, or elsewhere (initializer?  other?)

app/services/report_errors_to_argo.rb
tallenaz: Good point @ndushay -- we turned newrelic off because it didn't register seeding (and such) activity. I think the monitoring it offered us was pretty controller-centric, though, so it's not clear it would help us out in this case, but maybe benchmarking will?

app/services/preserved_object_handler.rb
jmartin-sul: same here... i was thinking there'd be something similar to POHR's `#log_results`, like `#report_errors_to_workflows`, that we could call alongside `#log_errors` at the end of each of the public POH methods.

app/services/report_errors_to_argo.rb
jmartin-sul: i think i agree with this... ultimately, the errors are being consumed by argo from WFS, but as far as pres catalog knows, it's just updating some workflow info (which, theoretically, could be consumed by things other than argo, even if that's the only service that cares to consume them for now).

config/application.rb
jmartin-sul: a `config/initializers` file sounds like the right home to me.

SaravShah:
close in favor of #435

PR #427, opened 2017-12-16
Version mismatch error shows object versions in message

### LINKED ISSUE #403: (PC) PO and PC version mismatch error should indicate versions in error message

tallenaz:
Fixes #403

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -72,8 +72,8 @@ class PreservedObjectHandler
           pres_copy = PreservedCopy.find_by!(preserved_object: pres_object, endpoint: endpoint) if pres_object
 
           if pres_copy.version != pres_object.current_version
-            handler_results.add_result(PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH, pres_copy.class.name)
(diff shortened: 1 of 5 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -72,7 +72,7 @@ def check_existence
           pres_copy = PreservedCopy.find_by!(preserved_object: pres_object, endpoint: endpoint) if pres_object
 
           if pres_copy.version != pres_object.current_version
-            handler_results.add_result(PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH, pres_copy.class.name)
+            handler_results.add_result(PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH, { pc_version: pres_copy.version, po_version: pres_object.current_version })
             raise ActiveRecord::Rollback, 'PO current_version != PC version'
```
ndushay: this looks great!  Can you stick the versions in the ActiveRecord rollback message, too?

PR #429, opened 2017-12-18
Validity unknown status

### LINKED ISSUE #371: add scary warning to PreservedCopy.status enum
a strongly worded note like the following might be helpful:
```ruby
  # DO NOT change the underlying constants for enum values that have been merged to
  # master/used in prod (or at least, consider the necessary migration)
```

### LINKED ISSUE #409: (M2C) when creating missing catalog entry, status should be ??
"OK" if it was validated
"DEFAULT" if it wasn't validated

only should get unexpected_version when incoming_version < catalog version ...

ndushay:
Since the "default" status in PreservedCopy was essentially a stand in for "validity unknown", I revised the code to have a distinct status for "validity unknown" and removed the "default" status.

Note that no db migration needed because a value was added to the status enum, but no existing values were changed.

- Resolves #409 (status when creating missing catalog entry)
- Resolves #371 (add scary warning about changes to PreservedCopy.status enum)
- bonus: gets rid of a copy pasta comment that annoyed @tingulfsen

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -37,7 +37,7 @@ class PreservedObjectHandler
     elsif PreservedObject.exists?(druid: druid)
       handler_results.add_result(PreservedObjectHandlerResults::OBJECT_ALREADY_EXISTS, 'PreservedObject')
     elsif moab_validation_errors.empty?
-      create_db_objects(PreservedCopy::DEFAULT_STATUS, true)
+      create_db_objects(PreservedCopy::OK_STATUS, true)
     else
       create_db_objects(PreservedCopy::INVALID_MOAB_STATUS, true)
     end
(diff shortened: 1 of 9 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -37,7 +37,7 @@ def create_after_validation
     elsif PreservedObject.exists?(druid: druid)
       handler_results.add_result(PreservedObjectHandlerResults::OBJECT_ALREADY_EXISTS, 'PreservedObject')
     elsif moab_validation_errors.empty?
-      create_db_objects(PreservedCopy::DEFAULT_STATUS, true)
+      create_db_objects(PreservedCopy::OK_STATUS, true)
```
SaravShah: ah, good catch

ndushay:
@jmartin-sul re: your concern:

If we hook up a ReST call for plain `create` (i.e.  NOT `create_after_validation`), then a person could create an object with  "validity unknown" status.  I think we should plan for that case to exist, but it is definitely an edge case, not critical path, and could be captured in a separate ticket and be addressed as time permits.

ndushay:
I created #433 for the concern John raised about PreservedCopy objects with `VALIDITY_UNKNOWN_STATUS` ever getting dealt with.

PR #432, opened 2017-12-19
fix M2C.check_exist bug when disk version < catalog version

### LINKED ISSUE #424: (M2C) check_exist: do not update catalog if incoming version < catalog
It is an error state if M2C check_exist has an incoming version < catalog -- this implies the catalog somehow got ahead of the actual Moab.  We are, I believe, intended to do something like (check with Julian)

- indicate UNEXPECTED VERSION on disk in results
- set status to UNEXPECTED VERSION
- error "loudly"
- complain endlessly

It is telling that this condition passed the tests -- or perhaps the tests were changed due to a misunderstanding of the spec.

I suspect it remains correct to do a moab validation in this circumstance.

ndushay:
At the moment, this is the ultimate in shameless green implementation.
Do we want to:
- merge it as is and make a separate ticket to refactor?
- refactor check_exist as part of this PR?
- move on to more pressing concerns?

Resolves #424

Review advice:   look at the commits separately

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -79,28 +79,28 @@ class PreservedObjectHandler
           if incoming_version == pres_copy.version
             handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_copy.class.name)
             handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_object.class.name)
-            update_pc_validation_timestamps(pres_copy)
-            update_db_object(pres_copy)
-          else
-            version_comp_result = if incoming_version > pres_copy.version
-                                    PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT
-                                  else
-                                    PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT
-                                  end
-            handler_results.add_result(version_comp_result, pres_copy.class.name)
-            handler_results.add_result(version_comp_result, pres_object.class.name)
+          elsif incoming_version > pres_copy.version
+            handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_copy.class.name)
+            handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_object.class.name)
             if moab_validation_errors.empty?
-              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, true)
+              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, false)
               update_status(pres_copy, PreservedCopy::OK_STATUS)
-              update_db_object(pres_copy)
               pres_object.current_version = incoming_version
               update_db_object(pres_object)
             else
               update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
-              update_pc_validation_timestamps(pres_copy)
-              update_db_object(pres_copy)
+            end
+          else # incoming_version < pres_copy.version
+            handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, pres_copy.class.name)
+            handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, pres_object.class.name)
+            if moab_validation_errors.empty?
+              update_status(pres_copy, PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS)
+            else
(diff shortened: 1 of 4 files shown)


spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -114,15 +114,15 @@
           @pc.save!
         end
 
-        it_behaves_like 'unexpected version', 8
+        it_behaves_like 'unexpected version', :update_version, 8
```
SaravShah: nice

app/services/preserved_object_handler.rb
```diff
@@ -79,28 +79,28 @@ def check_existence
...
-                                    PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT
-                                  end
-            handler_results.add_result(version_comp_result, pres_copy.class.name)
-            handler_results.add_result(version_comp_result, pres_object.class.name)
+          elsif incoming_version > pres_copy.version
+            handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_copy.class.name)
+            handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_object.class.name)
             if moab_validation_errors.empty?
-              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, true)
+              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, false)
```
jmartin-sul: should this stay `true` since we looked at the validation errors and determined there were none?

app/services/preserved_object_handler.rb
```diff
@@ -79,28 +79,28 @@ def check_existence
...
               update_status(pres_copy, PreservedCopy::OK_STATUS)
-              update_db_object(pres_copy)
               pres_object.current_version = incoming_version
               update_db_object(pres_object)
             else
               update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
```
jmartin-sul: would we want to `update_pc_validation_timestamps` here too, since we did check validity?  my recollection is that we update the timestamps if we did the check, regardless of the result?  i'm having a bit of trouble finding an issue/comment to cite, though that was my recollection from conversations (possibly incorrect).

app/services/preserved_object_handler.rb
```diff
@@ -79,28 +79,28 @@ def check_existence
...
+          else # incoming_version < pres_copy.version
+            handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, pres_copy.class.name)
+            handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, pres_object.class.name)
+            if moab_validation_errors.empty?
+              update_status(pres_copy, PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS)
+            else
+              update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
             end
           end
+          update_pc_validation_timestamps(pres_copy)
```
jmartin-sul: ah, ok, this ameliorates the concerns from the two prior comments... but won't this also be hit in the "happy path" (`incoming_version == pres_copy.version` evaluates to true)?  in which case it seems we didn't validate (which i believe is the behavior described by the audit doc)?

app/services/preserved_object_handler.rb
ndushay: I made it false because I moved the timestamp updates to outside the if block, and that's all that "true" as an arg is doing:  telling the method to update validation timestamps.  "true" is more correct, but "false" avoids double updates.

app/services/preserved_object_handler.rb
ndushay: ah yes - you are correct here.  Hrmm.

ndushay: oh wait - we're fine because in ALL cases, we have in fact done an audit and therefore update audit timestamps.

ndushay: perhaps we are running into more terminology woes.  The timestamps are:

last_audited
last_checked_on_storage

I think maybe we are now wondering about distinctions among:

- last-checksum-audit
- last-moab-validation-audit
- last-version-audit

etc.

oh joy!

ndushay:
@jmartin-sul if you are so inclined, I'd be happy to make a separate ticket re:  migrating data and code from `last_checked_on_storage`  to  two fields:  `last_version_audit` and `last_moab_validation`   and then maybe we could merge this and move on ...

jmartin-sul:
@ndushay ticket sounds fine to me.

and, uh, yeah, more timestamp fields, i guess!

app/services/preserved_object_handler.rb
jmartin-sul: filed issues #434 to address the follow on work.

Issue #404, opened 2017-12-12
(M2C) error notification by writing audit errors to DOR workflow


ndushay:
NOTE:  this is complicated SDR hoo-ha and probably should not be attempted without the participation of John, Tommy, Joe or some other workflow-aware person.

(see #402 for high level view)

#### Testing:

- https://github.com/sul-dlss/wfs_rails is a rails engine used by argo to TEST workflow services.   See especially https://github.com/sul-dlss/wfs_rails/blob/master/config/routes.rb.   @mejackreed knows the most about this.

(in addition?) these sul-dlss/argo spec files mock calls and responses:

- https://github.com/sul-dlss/argo/blob/60f26262a1026d3e831cc34ebbf50e4ad53fdce9/spec/controllers/items_controller_spec.rb
- https://github.com/sul-dlss/argo/blob/a2df7df51bc45b2a5a8f28d69ff30698330aa11a/spec/jobs/modsulator_job_spec.rb

#### Documentation
(probably outdated) 
- https://consul.stanford.edu/display/DOR/DOR+services#DORservices-workflow_updateThis

PR #435, opened 2017-12-20
[WIP] Update PCWF via ReST call

SaravShah:
paired w/ @tallenaz

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -43,7 +43,6 @@ class PreservedObjectHandler
     end
 
     handler_results.report_results
-    handler_results.result_array
   end
 
   def create
@@ -56,7 +55,6 @@ class PreservedObjectHandler
     end
 
     handler_results.report_results
-    handler_results.result_array
   end
 
   # this is a long, complex method (shameless green); if it is refactored, revisit the exceptions in rubocop.yml
@@ -112,7 +110,6 @@ class PreservedObjectHandler
       end
     end
     handler_results.report_results
-    handler_results.result_array
   end
 
   def confirm_version
@@ -129,7 +126,6 @@ class PreservedObjectHandler
     end
 
     handler_results.report_results
-    handler_results.result_array
   end
 
   def update_version_after_validation
@@ -151,7 +147,6 @@ class PreservedObjectHandler
     end
 
     handler_results.report_results
-    handler_results.result_array
   end
 
   def update_version
@@ -168,7 +163,6 @@ class PreservedObjectHandler
     end
 
     handler_results.report_results
-    handler_results.result_array
   end
 
   private
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -42,7 +42,7 @@ class PreservedObjectHandler
       create_db_objects(PreservedCopy::INVALID_MOAB_STATUS, true)
     end
 
-    handler_results.log_results
+    handler_results.report_results
     handler_results.result_array
   end
 
@@ -55,7 +55,7 @@ class PreservedObjectHandler
       create_db_objects(PreservedCopy::VALIDITY_UNKNOWN_STATUS)
     end
 
-    handler_results.log_results
+    handler_results.report_results
     handler_results.result_array
   end
 
@@ -111,7 +111,7 @@ class PreservedObjectHandler
         create_db_objects(PreservedCopy::INVALID_MOAB_STATUS, true)
       end
     end
-    handler_results.log_results
+    handler_results.report_results
     handler_results.result_array
   end
 
@@ -128,7 +128,7 @@ class PreservedObjectHandler
       end
     end
 
-    handler_results.log_results
+    handler_results.report_results
     handler_results.result_array
   end
 
@@ -150,7 +150,7 @@ class PreservedObjectHandler
       end
     end
 
-    handler_results.log_results
+    handler_results.report_results
     handler_results.result_array
   end
 
@@ -167,7 +167,7 @@ class PreservedObjectHandler
       end
     end
 
-    handler_results.log_results
+    handler_results.report_results
     handler_results.result_array
   end
 
--- a/app/services/preserved_object_handler_results.rb
+++ b/app/services/preserved_object_handler_results.rb
@@ -64,9 +64,10 @@ class PreservedObjectHandlerResults
     end
   end
 
-  attr_reader :result_array, :incoming_version, :msg_prefix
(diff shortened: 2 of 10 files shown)


app/services/preserved_object_handler_results.rb
```diff
@@ -88,7 +89,8 @@ def result_hash(code, msg_args=nil)
   # result_array = [result1, result2]
   # result1 = {response_code => msg}
   # result2 = {response_code => msg}
-  def log_results
+  def report_results
```
ndushay: if `report_results` method returned the `result_array`, then we could end pohandler public methods with:

```
handler_results.report_results
```

instead of 
```
handler_results.report_results
handler_results.result_array
```

app/services/preserved_object_handler.rb
```diff
@@ -42,7 +42,7 @@ def create_after_validation
       create_db_objects(PreservedCopy::INVALID_MOAB_STATUS, true)
     end
 
-    handler_results.log_results
+    handler_results.report_results
     handler_results.result_array
```
ndushay: if `report_results` method returned the `result_array`, then we could end pohandler public methods with:

```
handler_results.report_results
```

instead of 
```
handler_results.report_results
handler_results.result_array
```

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,31 @@
...
+class WorkflowErrorsReporter
+
+  def self.update_workflow(druid, error_message)
+    error_message.select { |error| request(druid, 'moab-valid', moab_error(error)) if error.key?(PreservedObjectHandlerResults::INVALID_MOAB) }
+  end
+
+  private_class_method def self.conn
+    if Settings.workflow_services.url.present?
+      @connection ||= Faraday.new(url: Settings.workflow_services.url)
+      Rails.logger.log(warn, 'no workflow hookup - assume you are in test or dev environment')
```
ndushay: I think we want the logic to say:

```ruby
private_class_method def self.conn
  @@connection || = begin
    if Settings.workflow_services.url.present?
       Faraday.new(url: Settings.workflow_services.url)
    else
      'no workflow hookup - assume you are in test or dev environment'
    end
  end
end
```

Would like the above to make the connection object ONE time, as a CLASS constant, and would like to only error ONE time if there is no url.  However, if we set it to nil when there is no url, then it will try each time to make the connection.  And if we don't set it to nil when we have none, then we have to check in the `request` method to see if it's a connection or not ...

Note: the logging can be simpler:

```ruby
Rails.logger.warn(msg_string)
```

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,31 @@
...
+    return unless conn
+    @connection.put do |request|
+      request.headers['content-type'] = "application/xml"
+      request.url  "/workflow/dor/objects/druid:#{druid}/workflows/preservationWF/#{process_name}"
+      request.body = "<process name='#{process_name}' status='error' errorMessage='#{error_message}'/>"
+    end
+  end
+
+  private_class_method def self.moab_error(error)
+    /^.*\)(.*)$/.match(error[PreservedObjectHandlerResults::INVALID_MOAB]).captures.first
```
ndushay: would it be clearer / appropriate to just split on `) ` and take the last piece?  Also, note the space after the paren.

And would that be generalized, so you could use it for any errors you wanted, not just moab errors?

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  private_class_method def self.conn
+    if Settings.workflow_services.url.present?
+      @connection ||= Faraday.new(url: Settings.workflow_services.url)
+      Rails.logger.log(warn, 'no workflow hookup - assume you are in test or dev environment')
+    end
+    @connection
+  end
+
+  private_class_method def self.request(druid, process_name, error_message)
```
ndushay: "request" is for the faraday http request, yes?  Would "send_request" be clearer or less clear as a name?  

"process_name" ain't bad, and I know in the context of workflows they use the term "process".  I wonder if we think "workflow_step" or "wf_step" is clearer or less clear.

and I think we may have to do this here:  (can't figure out how to only assign it once and also print warnings -- see previous comment)

```ruby
 private_class_method def self.request(druid, process_name, error_message)
   if conn.is_a?(String)
      logger.warn("{#druid} would write #{error_message} for wf step #{process_name} but #{conn}")
      return
  else
    ...
   conn.put do |request|
  ...
  end
end
```

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,31 @@
+require 'faraday'
+# This class will send a put ReST call to the preservationWF via faraday. This also parses
+# out a shorter error message, that only includes the moab validation error.
+class WorkflowErrorsReporter
+
+  def self.update_workflow(druid, error_message)
+    error_message.select { |error| request(druid, 'moab-valid', moab_error(error)) if error.key?(PreservedObjectHandlerResults::INVALID_MOAB) }
```
ndushay: I see that this is just for the INVALID_MOAB error, mapping to the 'moab-valid' step.   To generalize, would we want a constant in this class of the particular result codes triggering the write to workflow?  Roughly:

in workflow_errors_reporter.rb:
```ruby
ERRORS_TO_REPORT = [
  PreservedObjectHandlerResults::INVALID_MOAB,
  PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH  # or whatever
].freeze

def self.update_workflow(druid, pohandler_results)
  pohander_results.each do |result| 
     result_code = result.keys.first
     if ERRORS_TO_REPORT.include?(result_code)
        request(druid, map_to_workflow_step(result_code), result.values.first)
    end
end
```

in this case, workflow_errors_reporter is a bit savvy about result codes and result messages, but in a more generic way (not specific to a single error).


OR

you could have the mapping from result_codes to workflow steps in pohandler_result, and from there:

in pohandler_results:

```ruby
def report_results
  report_to_workflows
  ...
end

private

def report_to_workflows
  result_array.each do |result| 
     result_code = result.keys.first
     if MAP_TO_WORKFLOW_STEPS.keys.include?(result_code)
        WorkflowErrorsReporter.update_workflow(druid, MAP_TO_WORKFLOW_STEPS[result_code], workflow_err_msg(results.values.first))
     end
end

def workflow_err_msg(result_value)
  result_value.split(') ').last
end
```

And then WorkflowErrorsReporter could be stupider:  just "i have a workflow step, I have a message, I'm reporting an error."

ndushay:
Is this the implementation of #404 ?

SaravShah:
closed in favor of #454

PR #437, opened 2017-12-21
Refactoring happens

ndushay:
Some refactoring I did while working on #434 (improve audit timestamps).

Should be easy to review commit by commit.

PR #438, opened 2017-12-21
[WIP] Check existence skips bad status

### LINKED ISSUE #431: (M2C) modify check_existence to deal with pres copies that already have a version mismatch status
~~skip further checks for moabs that are still in a version mismatch status, but log a warning that they are still that way.~~

~~the assumption is that since remediation of the moab content is currently manual, manually correcting the status of the bad pres copy in the catalog should also be part of remediation.  this is not much more work for the remediator, and could still involve using a pre-defined method for setting status back to ok (including appropriate logging).~~

~~we may revisit this when more of the remediation/recovery functionality becomes automated, and/or when we start dealing with archive copies.~~

actually, per further discussion this past tuesday at story time, we want to re-validate pres copies that have a bad status, and update the status to reflect the current state on disk, then proceed with checks as before.

jmartin-sul:
we should merge #437, then #439, and then rebase this, which is one of the reasons for the WIP tag (the other is the open question below).

* standardize and refactor common version mismatch error code to helper method (see below)
* fix what seemed like copypasta in a shared spec invocation
* have `PreservedObjectHandler` `#check_existence` and `#confirm_version` deal with items that already have a bad status

includes has some tangentially related clean-up work.  might be easier to review by commit.  one of the bits of refactoring is duplicative of some work in #437 (and should just get dropped from this PR if that one's merged first).  though some work for this PR adopts the same pattern for a similar bit of functionality.
https://github.com/sul-dlss/preservation_catalog/pull/437/files#diff-db66cd3a51226f58d609af9d477eda0cR74
https://github.com/sul-dlss/preservation_catalog/commit/91793244594d751c2f456298723569b9e43bb43e#diff-db66cd3a51226f58d609af9d477eda0cR74
https://github.com/sul-dlss/preservation_catalog/commit/cb33e5ad8ac3834fcfe5294d343452f4c7226924#diff-db66cd3a51226f58d609af9d477eda0cR74

fine for the team to take over in my absence to get this finished, should be pretty straight-forward if this is generally in the right direction.

my question:  does any non-ok status stop further work from happening?  that's the approach i went with.  the conversation w/ @ndushay and @julianmorley that sparked this work was initially about version mismatch errors (between pres copy and pres obj).  but since what we ended up deciding was that M2C shouldn't detect when an object goes "good" (status should be set to OK as part of remediation, which is already manual for the time being), i just had things halt at anything that wasn't already ok.  should be easy enough to relax that a bit, hopefully.

edit 2017-12-22, a couple things we should probably do before closing this:
[ ] update the audit document to capture this?  something more permanent?  seems like the intent here should be captured somewhere?
[ ] file a ticket indicating that we may have to revisit this when dealing with archive copies, and/or when automating more of the remediation process.

closes #431

### DIFF
--- a/spec/services/preserved_object_handler_check_exist_spec.rb
+++ b/spec/services/preserved_object_handler_check_exist_spec.rb
@@ -162,12 +162,6 @@ RSpec.describe PreservedObjectHandler do
                 po_handler.check_existence
                 expect(pc.reload.updated_at).to be > orig
               end
-              it 'status becomes "ok" if it was invalid_moab (b/c after validation)' do
-                pc.status = PreservedCopy::INVALID_MOAB_STATUS
-                pc.save!
-                po_handler.check_existence
-                expect(pc.reload.status).to eq PreservedCopy::OK_STATUS
-              end
             end
             context 'unchanged' do
               before do
@@ -314,12 +308,6 @@ RSpec.describe PreservedObjectHandler do
               end
               it 'ensures status becomes invalid_moab from ok' do
                 invalid_pc.status = PreservedCopy::OK_STATUS
-                invalid_pc.save!
-                invalid_po_handler.check_existence
-                expect(invalid_pc.reload.status).to eq PreservedCopy::INVALID_MOAB_STATUS
-              end
-              it 'ensures status becomes invalid_moab from expected_vers_not_found_on_storage' do
-                invalid_pc.status = PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS
                 invalid_pc.save!
                 invalid_po_handler.check_existence
                 expect(invalid_pc.reload.status).to eq PreservedCopy::INVALID_MOAB_STATUS
@@ -477,6 +465,7 @@ RSpec.describe PreservedObjectHandler do
         allow(po).to receive(:current_version).and_return(1)
         allow(po).to receive(:touch)
         allow(PreservedCopy).to receive(:find_by).with(preserved_object: po, endpoint: ep).and_return(pc)
+        allow(pc).to receive(:status).and_return(PreservedCopy::OK_STATUS)
         allow(pc).to receive(:version).and_return(1)
         allow(pc).to receive(:last_audited=)
         allow(pc).to receive(:last_checked_on_storage=)
--- a/spec/services/preserved_object_handler_check_exist_spec.rb
+++ b/spec/services/preserved_object_handler_check_exist_spec.rb
@@ -180,15 +180,8 @@ RSpec.describe PreservedObjectHandler do
                 expect(pc.reload.size).to eq orig
               end
             end
-            it 'what about other statuses????' do
-              skip 'need to know what to do when status does NOT start as ok or invalid moab'
-              pc.status = PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS
-              pc.save!
-              po_handler.check_existence
-              expect(pc.reload.status).to eq PreservedCopy::OK_STATUS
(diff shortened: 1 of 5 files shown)


spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -162,12 +162,6 @@
                 po_handler.check_existence
                 expect(pc.reload.updated_at).to be > orig
               end
-              it 'status becomes "ok" if it was invalid_moab (b/c after validation)' do
-                pc.status = PreservedCopy::INVALID_MOAB_STATUS
-                pc.save!
-                po_handler.check_existence
-                expect(pc.reload.status).to eq PreservedCopy::OK_STATUS
-              end
```
ndushay: I think if we are performing a validation anyway, and it is staring from INVALID_MOAB, we should be able to set it to "ok"

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -318,12 +312,6 @@
                 invalid_po_handler.check_existence
                 expect(invalid_pc.reload.status).to eq PreservedCopy::INVALID_MOAB_STATUS
               end
-              it 'ensures status becomes invalid_moab from expected_vers_not_found_on_storage' do
-                invalid_pc.status = PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS
-                invalid_pc.save!
-                invalid_po_handler.check_existence
-                expect(invalid_pc.reload.status).to eq PreservedCopy::INVALID_MOAB_STATUS
-              end
```
ndushay: I think "invalid_moab" should trump most other statuses, if moab validation has already been run.

spec/services/preserved_object_handler_check_exist_spec.rb
jmartin-sul: i think that the way it's implemented in this PR, things are halting before moab validation is performed.  but i'd be totally open to something slightly more clever, where a status of `invalid_moab` triggered a moab validation, and marked things `ok` and proceeded if no errors were detected.

on the other hand, it did seem like the idea was to punt on any sort of clever remediation detection, which does make me slightly hesitant to add that bit of nuance.  similarly, if the only problem were an unexpected version error, we could see if the version on disk equal to or greater than the version expected by the catalog.  but my recollection is that punting on that was the genesis of this ticket.  and they feel like the same general category of error to me, i think (i.e., errors that need manual attention, where we don't expect our current processes to repair the automatically).  i guess some types of moab validation failure are not things that you'd want to block further checking... maybe there should be a threshold of moab validation error?  maybe that's getting too clever.

i'm not sure i'm convinced that "validate and set to ok if it is" the right way to go, but i'm also definitely not sure this PR's approach is the right way to go.  i'm also worried there will be unforseen pitfalls in the auto-mark-ok approach that will come out once we start testing and using the code, which is one of the things that made "skip and manually remediate status" so attractive to me in the first place.

maybe face to face discussion after break?  seems unlikely this will get merged this afternoon anyway?

spec/services/preserved_object_handler_check_exist_spec.rb
jmartin-sul: i'd be open to that logic

jmartin-sul:
rebased and re-pushed.  no idea why these `updated_at` fields are getting updated at all, let alone why they're getting updated and failing the travis build, while passing on my laptop.  strange.

i propose revisiting in january?

jmartin-sul:
created #443 to capture some possible follow-on work that i'd proposed in the discussion that spawned this ticket and PR.

jmartin-sul:
superseded by #456

PR #439, opened 2017-12-21
Improve audit timestamps

### LINKED ISSUE #434: (M2C) more audit timestamp granularity
we discovered that we might want to differentiate between `last_checked_on_storage` use cases to a couple more granular things like `last_moab_validation` and `last_version_audit`.

surfaced by PR #432.

ndushay:
~~WIP until #437 is merged;  then this will be rebased on master.~~

PreservedCopy datestamps:
- last_moab_validation
- last_version_audit

replace:
- last_audited
- last_check_on_storage

To allow us to distinguish between these two types of audit checks.  This was discussed here:  https://github.com/sul-dlss/preservation_catalog/pull/432#discussion_r157886007

I tried to break up the commits so it would be somewhat manageable for review.  The truth is after the db migration, the specs won't pass without all these commits.  :-P

Resolves #434

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -164,7 +164,7 @@ class PreservedObjectHandler
       Rails.logger.debug "update_version #{druid} called"
       if endpoint.endpoint_type.endpoint_class == 'online'
         # NOTE: we deal with active record transactions in update_online_version, not here
-        update_online_version
+        update_online_version(false, nil, true)
       elsif endpoint.endpoint_type.endpoint_class == 'archive'
         # TODO: create a new PC record for the new archived version on the endpoint
       end
@@ -217,7 +217,7 @@ class PreservedObjectHandler
 
   # TODO: this is "too complex" per rubocop: shameless green implementation
   # NOTE: if we can reduce complexity, remove Metrics/PerceivedComplexity exception in .rubocop.yml
-  def update_online_version(moab_validated=false, status=nil)
+  def update_online_version(moab_validated=false, status=nil, set_status_to_unexp_version=false)
     transaction_ok = with_active_record_transaction_and_rescue do
       pres_object = PreservedObject.find_by!(druid: druid)
       pres_copy = PreservedCopy.find_by!(preserved_object: pres_object, endpoint: endpoint) if pres_object
@@ -232,11 +232,14 @@ class PreservedObjectHandler
         handler_results.add_result(code, pres_object.class.name)
 
         update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, moab_validated)
-        update_status(pres_copy, status) if status
+        update_status(pres_copy, status) if status && moab_validated
         update_db_object(pres_copy)
         pres_object.current_version = incoming_version
         update_db_object(pres_object)
       else
+        if set_status_to_unexp_version
+          status = PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS
+        end
         update_pc_unexpected_version(pres_copy, pres_object, status, moab_validated)
       end
     end
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -80,24 +80,24 @@ class PreservedObjectHandler
           elsif incoming_version > pres_copy.version
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_copy.class.name)
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_object.class.name)
+            ran_moab_validation = true
             if moab_validation_errors.empty?
-              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, false)
+              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, ran_moab_validation)
               update_status(pres_copy, PreservedCopy::OK_STATUS)
               pres_object.current_version = incoming_version
               update_db_object(pres_object)
             else
               update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
             end
-            ran_moab_validation = true
           else # incoming_version < pres_copy.version
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, pres_copy.class.name)
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, pres_object.class.name)
+            ran_moab_validation = true
             if moab_validation_errors.empty?
               update_status(pres_copy, PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS)
             else
               update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
             end
-            ran_moab_validation = true
           end
           update_pc_audit_timestamps(pres_copy, ran_moab_validation, true)
           update_db_object(pres_copy)
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -145,7 +145,7 @@ class PreservedObjectHandler
           # NOTE: we deal with active record transactions in update_online_version, not here
           update_online_version(true, PreservedCopy::OK_STATUS)
         else
-          update_online_version(true, PreservedCopy::INVALID_MOAB_STATUS)
+          update_pc_invalid_moab
         end
       elsif endpoint.endpoint_type.endpoint_class == 'archive'
         # TODO: perform archive object validation; then create a new PC record for the new
@@ -253,6 +253,18 @@ class PreservedObjectHandler
       handler_results.add_result(res_code, { pc_version: pc_version, po_version: po_version })
       raise ActiveRecord::Rollback, "PreservedCopy version #{pc_version} != PreservedObject current_version #{po_version}"
     end
+  end
+
+  def update_pc_invalid_moab
+    transaction_ok = with_active_record_transaction_and_rescue do
+      pres_object = PreservedObject.find_by!(druid: druid)
+      pres_copy = PreservedCopy.find_by!(preserved_object: pres_object, endpoint: endpoint) if pres_object
+      # FIXME: what if there is more than one associated pres_copy?
+      update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
+      update_pc_audit_timestamps(pres_copy, true, false)
+      update_db_object(pres_copy)
+    end
+    handler_results.remove_db_updated_results unless transaction_ok
   end
 
   def update_pc_unexpected_version(pres_copy, pres_object, new_status, moab_validated)
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -73,7 +73,7 @@ class PreservedObjectHandler
 
           raise_rollback_if_pc_po_version_mismatch(pres_copy.version, pres_object.current_version)
 
-          moab_validated = false
+          ran_moab_validation = false
           if incoming_version == pres_copy.version
             handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_copy.class.name)
             handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_object.class.name)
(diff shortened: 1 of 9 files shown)


app/services/preserved_object_handler.rb
```diff
@@ -161,7 +164,7 @@ def update_version
       Rails.logger.debug "update_version #{druid} called"
       if endpoint.endpoint_type.endpoint_class == 'online'
         # NOTE: we deal with active record transactions in update_online_version, not here
-        update_online_version
+        update_online_version(false, nil, true)
```
SaravShah: why is `set_status_to_enexp_version` here true?

ndushay:
@SaravShah given that bigint lacked sub-second granularity, I made an executive decision to stick to DateTime stuff.   Pushback is welcome.

app/services/preserved_object_handler.rb
ndushay: i don't understand your question - typo?

SaravShah: `update_online_version` the method takes in three parameters: `moab_validated`, `status`, and `set_status_to_unexp_version`, here you set it to be ` true`. I was wondering why we are setting it to `true`.

ndushay: my thinking was:  

in the case of update_version, if we don't get  incoming > catalog, then we should set the status to "whoa!  this catalog entry isn't what I expected!"   But maybe that's wrong.  I sort of made it up as I went.

db/schema.rb
```diff
@@ -53,17 +53,19 @@
 
   create_table "preserved_copies", force: :cascade do |t|
     t.integer "version", null: false
-    t.bigint "last_audited"
     t.bigint "preserved_object_id", null: false
     t.bigint "endpoint_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.datetime "last_checked_on_storage"
+    t.datetime "last_moab_validation"
```
tingulfsen: I think it was Julian who suggested using bigint (seconds since the epoch) for this field. Maybe run this change by him to see if there was an important reason that we might have missed with the change to datetime? I do agree that it's preferably to use a proper date abstraction instead of counting seconds since some randomly chosen time in the past.

jmartin-sul: my recollection is that it was a performance concern.  if @julianmorley would be willing, i'd like to try just using dates and seeing how that goes.  i have to imagine the postgres date indexing is at least somewhat performant?  seems worth testing?  i agree, a proper date abstraction would be nice.  and while it didn't seem terrible to keep a shadow bigint column for one audit field, keeping a shadow column for multiple audit timestamps starts to feel (to me) like too much overhead (coginitive and storage), when we could just hopefully have our db do that for us.  and a single `last_audited` column that's the last of any of those feels less useful because presumably we want to query at the level of granularity we're tracking.

i was going to suggest that we need to start indexing those date columns, but looks like that's already part of #439!  maybe we should also index some/all of the `updated_at` fields in the schema?

jmartin-sul:
note:  the base branch here is still `refactoring-happens`.  should probably be switched to master since the current base has already been merged to master?

app/services/preserved_object_handler.rb
```diff
@@ -87,6 +88,7 @@ def check_existence
             else
               update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
             end
+            moab_validated = true
```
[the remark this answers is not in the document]
ndushay: I'm thinking more "moab validation has occurred" -- not that it PASSED validation.  Therefore, it is correct that last_moab_validation was now even if it was a failure.

app/services/preserved_object_handler.rb
```diff
@@ -232,11 +232,14 @@ def update_online_version(moab_validated=false, status=nil)
...
+        update_status(pres_copy, status) if status && moab_validated
         update_db_object(pres_copy)
         pres_object.current_version = incoming_version
         update_db_object(pres_object)
       else
+        if set_status_to_unexp_version
+          status = PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS
+        end
         update_pc_unexpected_version(pres_copy, pres_object, status, moab_validated)
       end
```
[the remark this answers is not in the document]
ndushay: this is the sort of thing that made me conclude that we need a wiki page here on:
- PreservedCopy.status   vs
- PohandlerResults result code  vs
- what we will be logging vs
- what errors we report to honeybadger.

I *think* we want PreservedCopy.status to reflect the known state of the object outside the catalog.  i'm not certain we have all the statuses we need or that we know what each of them means ...

bleah.

db/schema.rb
ndushay: I am glad you are calling this out here.

I suspect julian wanted bigint to save a byte or two per record.  However, the bigint approach only goes to the granularity of seconds and was making us have to "sleep" for tests involving that field. 

I have grown dubious of Julian's performance concerns and went in the direction of what made the most sense cognitively and writing code wise.

app/services/preserved_object_handler.rb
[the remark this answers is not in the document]
tallenaz: Is it worth considering tracking the distinction between a validation occurring and validation passing in the variable name? So, instead of `moab_validated`, we call the variable something like `moab_validation_occurred`?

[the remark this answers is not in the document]
ndushay: note the distinction between `moab_validateD` and `moab_validateS`.   I would use the former to say "I performed the moab validation checks" and I would us the latter to say "this object passes validation."   Is that too subtle?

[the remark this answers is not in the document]
tallenaz: The subtlety for me is that I would also naturally use the former to say "This object passed validation". I don't think it's too subtle with your distinction in mind, though.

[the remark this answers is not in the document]
ndushay: `ran_moab_validation`

app/services/preserved_object_handler.rb
ndushay: if incoming is NOT > catalog for an update_version call (NOT a check_existence call) then the issue is with the "incoming" argument, and the PC status should NOT be updated.

Example:  ingest robot has a new version, so it calls update_version with v3.  The catalog says it is already at v3.  Now what?   Ask Julian / pick something for now.

app/services/preserved_object_handler.rb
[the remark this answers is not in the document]
ndushay: See wiki;  perhaps this is ok?

db/schema.rb
ndushay: Ensure with Julian that performance is adequate.   This has more to do with date query performance than storage costs, I imagine.  We believe postgres can do date searching performantly.

app/services/preserved_object_handler.rb
ndushay: Per Ben:

Note:  if the call is coming from ingest robot, then _we can trust the incoming version_.  If
- incoming version == catalog version + 1:  perfect.  update catalog version.  Joy to the world.
- incoming version == catalog version + (more than 1):  report error to dor workflows.  log error.  But go ahead and update the catalog.
- incoming version == catalog version OR incoming version < catalog version:  report error to dor workflows.  log error.  Do not update.  Do not pass go.

db/schema.rb
ndushay: Julian has confirmed that datetime is fine by him.

spec/services/shared_examples_preserved_object_handler.rb
```diff
@@ -119,53 +119,81 @@
...
 
   context 'returns' do
     let!(:results) { po_handler.send(method_sym) }
+    let(:num_results) do
+      if method_sym == :update_version
+        5
+      else
+        4
+      end
+    end
```
jmartin-sul: does this actually get used in this context?  i can see where the context that starts on line 305 uses the `num_results` it defines, but this one looks unused to me.

spec/services/shared_examples_preserved_object_handler.rb
```diff
@@ -278,8 +335,10 @@
       ]
       obj_version_results = results.select { |r| codes.include?(r.keys.first) }
       msgs = obj_version_results.map { |r| r.values.first }
-      expect(msgs).to include(a_string_matching("PreservedObject"))
-      expect(msgs).to include(a_string_matching("PreservedCopy"))
+      unless results.find { |r| r.keys.first == PreservedObjectHandlerResults::INVALID_MOAB }
+        expect(msgs).to include(a_string_matching("PreservedObject"))
+        expect(msgs).to include(a_string_matching("PreservedCopy"))
+      end
```
jmartin-sul: random observation:  i'm starting to feel just a bit uncomfortable with the amount of conditional logic and complexity creeping into the tests, but i don't have a good suggestion for how to go about things differently without making our already voluminous test code even more voluminous.  e.g., i would not go back to not having shared examples.

app/services/preserved_object_handler.rb
ndushay: I created #451 to capture the above from Ben, rather than addressing it here.

app/services/preserved_object_handler.rb
```diff
@@ -219,8 +222,7 @@ def update_online_version(moab_validated=false, status=nil)
       pres_object = PreservedObject.find_by!(druid: druid)
       pres_copy = PreservedCopy.find_by!(preserved_object: pres_object, endpoint: endpoint) if pres_object
 
-      # TODO: add this later with audit timestamp updates
-      # raise_rollback_if_pc_po_version_mismatch(pres_copy.version, pres_object.current_version)
+      raise_rollback_if_pc_po_version_mismatch(pres_copy.version, pres_object.current_version)
```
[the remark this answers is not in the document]
ndushay: @jmartin-sul : I think it was because this PR was done in tandem with some refactoring in a separate PR, and I commented out the line in one context, waiting for the other.  Something like that, anyway.

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -393,41 +393,8 @@
...
-          it 'PC_PO_VERSION_MISMATCH result' do
-            code = PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH
-            expect(results).to include(hash_including(code => version_mismatch_msg))
-          end
-          it 'does NOT get UPDATED_DB_OBJECT message' do
-            expect(results).not_to include(hash_including(PreservedObjectHandlerResults::UPDATED_DB_OBJECT))
-            expect(results).not_to include(hash_including(PreservedObjectHandlerResults::UPDATED_DB_OBJECT_TIMESTAMP_ONLY))
-          end
-        end
+        it_behaves_like 'PreservedObject current_version does not match online PC version', :update_version, 3, 2, 8
```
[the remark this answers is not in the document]
ndushay: good 👀 , thanks!

spec/services/preserved_object_handler_confirm_version_spec.rb
```diff
@@ -221,41 +221,8 @@
...
-          it 'PC_PO_VERSION_MISMATCH result' do
-            code = PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH
-            expect(results).to include(hash_including(code => version_mismatch_msg))
-          end
-          it 'does NOT get UPDATED_DB_OBJECT message' do
-            expect(results).not_to include(hash_including(PreservedObjectHandlerResults::UPDATED_DB_OBJECT))
-            expect(results).not_to include(hash_including(PreservedObjectHandlerResults::UPDATED_DB_OBJECT_TIMESTAMP_ONLY))
-          end
-        end
+        it_behaves_like 'PreservedObject current_version does not match online PC version', :update_version, 3, 2, 8
```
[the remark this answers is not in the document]
ndushay: also good 👀 , thanks!

spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -460,54 +524,14 @@
...
-                allow(pc).to receive(:size=).with(incoming_size)
-                allow(pc).to receive(:last_audited=)
-                allow(pc).to receive(:last_checked_on_storage=)
-                allow(pc).to receive(:status).and_return(PreservedCopy::OK_STATUS)
-                allow(pc).to receive(:status=)
-                allow(pc).to receive(:changed?).and_return(true)
-                allow(pc).to receive(:save!)
-                po_handler.update_version_after_validation
-              end
-
```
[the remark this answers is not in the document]
ndushay: Adding a comment:  PreservedObject won't get updated if moab is invalid, that's why the test isn't there.

spec/services/shared_examples_preserved_object_handler.rb
ndushay: completely agree.  Also had no handy solution short of a big refactoring which feels non-critical path.

Eventually we could probably morph most of the logging tests to something appropriate in the results_spec, and perhaps test for pohandler results directly, which might help a bit.

I agree the tests (and the code) are ugly as they are now.

spec/services/shared_examples_preserved_object_handler.rb
ndushay: again, good :eyes:.  Removed.

ndushay:
@tingulfsen @jmartin-sul @SaravShah: I believe I have addressed all your concerns.  Would love to get this merged and move on.

PR #454, opened 2018-01-11
Implemented the workflow services rest call to send error messages to preservationWF.

SaravShah:
Incorporated it into the PreservedObjectHandler.report_results method. 
Wrote RSpec tests using WebMock. 

Only have two workflow steps, `'moab-valid' and 'pc-process'` The `'pc-process'` name is just all other errors other than `PreservedObjectHandlerResults::INVALID_MOAB` so if there is another name we should use, let me know.

connects to #404 

(Travis fails will be fixed in follow on PR)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -86,6 +86,7 @@ Metrics/LineLength:
   Exclude:
     - 'app/services/preserved_object_handler.rb' # line 269 is 121
     - 'app/services/preserved_object_handler_results.rb' # line 35 is 121
+    - 'app/services/workflow_errors_reporter.rb' # line 7 is 143
     - 'spec/lib/audit/moab_to_catalog_spec.rb' # 1 line 126
     - 'spec/models/status_spec.rb' # Line length of 132 for better readability
     - spec/services/preserved_object_handler_spec.rb
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -60,6 +60,7 @@ Lint/RescueWithoutErrorClass:
 Metrics/AbcSize:
   Exclude:
     - 'app/services/preserved_object_handler.rb' # methods still readable; shameless green
+    - 'app/services/preserved_object_handler_results.rb' # method still readable
     - 'lib/audit/moab_to_catalog.rb' # .xxx_for_all_storage_roots is decidedly readable
 
 Metrics/BlockLength:
@@ -74,6 +75,7 @@ Metrics/BlockNesting:
 Metrics/ClassLength:
   Exclude:
     - 'app/services/preserved_object_handler.rb'
+    - 'app/services/preserved_object_handler_results.rb'
 
 Metrics/CyclomaticComplexity:
   Exclude:
@@ -89,15 +91,13 @@ Metrics/LineLength:
     - 'app/services/workflow_errors_reporter.rb' # line 7 is 143
     - 'spec/lib/audit/moab_to_catalog_spec.rb' # 1 line 126
     - 'spec/models/status_spec.rb' # Line length of 132 for better readability
-    - spec/services/preserved_object_handler_spec.rb
     - 'spec/services/preserved_object_handler_check_exist_spec.rb' # 17 lines, but who's counting, officer?
     - 'spec/services/preserved_object_handler_confirm_version_spec.rb'
     - 'spec/services/preserved_object_handler_create_spec.rb'
+    - 'spec/services/preserved_object_handler_spec.rb' # line 94 is 121
     - 'spec/services/preserved_object_handler_update_version_spec.rb'
     - 'spec/services/shared_examples_preserved_object_handler.rb'
-    - 'spec/models/status_spec.rb' # Line length of 132 for better readability
-    - 'spec/services/preserved_object_handler_check_exist_spec.rb' # 17 lines, but who's counting, officer?
-    - 'spec/services/workflow_errors_reporter_spec.rb' # lines for body and error_msg kept long for readability
+    - 'spec/services/workflow_errors_reporter_spec.rb'
 
 Metrics/MethodLength:
   Max: 25
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -95,6 +95,9 @@ Metrics/LineLength:
     - 'spec/services/preserved_object_handler_create_spec.rb'
     - 'spec/services/preserved_object_handler_update_version_spec.rb'
     - 'spec/services/shared_examples_preserved_object_handler.rb'
+    - 'spec/models/status_spec.rb' # Line length of 132 for better readability
+    - 'spec/services/preserved_object_handler_check_exist_spec.rb' # 17 lines, but who's counting, officer?
+    - 'spec/services/workflow_errors_reporter_spec.rb' # lines for body and error_msg kept long for readability
 
 Metrics/MethodLength:
   Max: 25
--- a/app/services/preserved_object_handler_results.rb
+++ b/app/services/preserved_object_handler_results.rb
@@ -64,9 +64,10 @@ class PreservedObjectHandlerResults
     end
   end
 
-  attr_reader :result_array, :incoming_version, :msg_prefix
+  attr_reader :result_array, :incoming_version, :msg_prefix, :druid
 
   def initialize(druid, incoming_version, incoming_size, endpoint)
+    @druid = druid
     @incoming_version = incoming_version
     @msg_prefix = "PreservedObjectHandler(#{druid}, #{incoming_version}, #{incoming_size}, #{endpoint.endpoint_name if endpoint})"
     @result_array = []
@@ -88,7 +89,8 @@ class PreservedObjectHandlerResults
   # result_array = [result1, result2]
   # result1 = {response_code => msg}
   # result2 = {response_code => msg}
-  def log_results
+  def report_results
+    WorkflowErrorsReporter.update_workflow(druid, result_array)
     result_array.each do |r|
       severity = self.class.logger_severity_level(r.keys.first)
       msg = r.values.first
--- a/app/services/preserved_object_handler_results.rb
+++ b/app/services/preserved_object_handler_results.rb
@@ -96,6 +96,7 @@ class PreservedObjectHandlerResults
       msg = r.values.first
       Rails.logger.log(severity, msg)
     end
+    result_array
   end
 
   private
--- a/app/services/preserved_object_handler_results.rb
+++ b/app/services/preserved_object_handler_results.rb
@@ -38,6 +38,15 @@ class PreservedObjectHandlerResults
     PC_PO_VERSION_MISMATCH => "PreservedCopy online moab version %{pc_version} does not match PreservedObject current_version %{po_version}"
   }.freeze
 
+  WORKFLOW_REPORT_CODES = [
+    ARG_VERSION_LESS_THAN_DB_OBJECT,
+    DB_UPDATE_FAILED,
+    OBJECT_ALREADY_EXISTS,
+    OBJECT_DOES_NOT_EXIST,
+    UNEXPECTED_VERSION,
+    PC_PO_VERSION_MISMATCH
+  ].freeze
+
   DB_UPDATED_CODES = [
     UPDATED_DB_OBJECT,
     UPDATED_DB_OBJECT_TIMESTAMP_ONLY,
@@ -86,20 +95,41 @@ class PreservedObjectHandlerResults
     { code => result_code_msg(code, msg_args) }
   end
 
-  # result_array = [result1, result2]
-  # result1 = {response_code => msg}
-  # result2 = {response_code => msg}
+  # output results to Rails.logger and send errors to WorkflowErrorsReporter
+  # @return result_array
   def report_results
-    WorkflowErrorsReporter.update_workflow(druid, result_array)
+    candidate_workflow_results = []
     result_array.each do |r|
-      severity = self.class.logger_severity_level(r.keys.first)
-      msg = r.values.first
-      Rails.logger.log(severity, msg)
+      log_result(r)
+      if r.key?(PreservedObjectHandlerResults::INVALID_MOAB)
+        WorkflowErrorsReporter.update_workflow(druid, 'moab-valid', r.values.first)
+      elsif WORKFLOW_REPORT_CODES.include?(r.keys.first)
+        candidate_workflow_results << r
+      end
     end
+    stack_trace = caller(1..1).first[/.+?(?=:in)/]
+    report_errors_to_workflows(candidate_workflow_results, stack_trace)
     result_array
   end
 
+  def report_errors_to_workflows(candidate_workflow_results, stack_trace)
+    value_array = []
+    candidate_workflow_results.each do |x|
+      x.each_value do |val|
+        value_array << val
+      end
+    end
+    value_array << stack_trace
+    WorkflowErrorsReporter.update_workflow(druid, 'preservation-audit', value_array.join(" || "))
+  end
+
   private
+
+  def log_result(result)
+    severity = self.class.logger_severity_level(result.keys.first)
+    msg = result.values.first
+    Rails.logger.log(severity, msg)
+  end
 
   def result_code_msg(code, addl=nil)
     arg_hash = { incoming_version: incoming_version }
--- a/app/services/workflow_errors_reporter.rb
+++ b/app/services/workflow_errors_reporter.rb
@@ -1,4 +1,5 @@
 require 'faraday'
+require 'retries'
 # This class will send a put ReST call to the preservationWF via faraday. This also parses
 # out a shorter error message, that only includes the moab validation error.
 class WorkflowErrorsReporter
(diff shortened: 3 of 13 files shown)


ndushay:
I think `pc-process` is not going to work because `pc` isn't enough context.  Maybe `preservation-process` ?  `preservation-catalog` ?  `presvtn-service` ?

app/services/preserved_object_handler_results.rb
```diff
@@ -88,12 +98,32 @@ def result_hash(code, msg_args=nil)
...
+      elsif REPORT_ERROR_LIST.include?(r.keys.first)
+        combined_error_messages << r
+      end
+    end
+    stack_trace = caller(1..1).first[/.+?(?=:in)/]
+    display_message(combined_error_messages, stack_trace)
+    result_array
+  end
+
+  def display_message(combined_error_messages, stack_trace)
```
ndushay: how about `report_errs_to_workflows` ? or something along those lines?

SaravShah:
maybe it should have the word `error` in it?

app/services/preserved_object_handler_results.rb
```diff
@@ -88,12 +98,32 @@ def result_hash(code, msg_args=nil)
...
+  def report_results
+    combined_error_messages = []
     result_array.each do |r|
       severity = self.class.logger_severity_level(r.keys.first)
       msg = r.values.first
       Rails.logger.log(severity, msg)
+      if r.key?(PreservedObjectHandlerResults::INVALID_MOAB)
+        WorkflowErrorsReporter.update_workflow(druid, 'moab-valid', r.values.first)
+      elsif REPORT_ERROR_LIST.include?(r.keys.first)
+        combined_error_messages << r
```
ndushay: how about `candidate_workflow_results` instead of `combined_error_messages` ?   We are still dealing with "result" objects, after all.  Or  `candidate_workflow_result_array` ...

Or, don't do it this way.  

You could have a method 'report_errors_to_workflows`

```ruby
def report_errors_to_workflows
  # not quite the right syntax, just the idea
  moab-validation-err-result = result_array.find { |r| r.code == PreservedObjectHandlerResults::INVALID_MOAB }
  WorkflowErrorsReporter.update_workflow(druid, 'moab-valid', moab-validation-err-result.values.first) if moab-validation-err-result

  other_results = result_array.dup - moab-validation-err-result
  err_results = other_results.select_all { |r| r << r if REPORT_ERROR_LIST.include?(r.keys.first) }
  report_combined_errs_to_workflows(err_results, stacktrace)
  result_array
end

def report_combined_errs_to_workflows # new name for "display_message"

end
```

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,39 @@
+require 'faraday'
+require 'retries'
+# This class will send a put ReST call to the preservationWF via faraday. This also parses
+# out a shorter error message, that only includes the moab validation error.
```
ndushay: This looks like the comment from an earlier commit.   How about:

```ruby
# send errors to preservationWF workflow for an object via ReST calls.
```

app/services/preserved_object_handler_results.rb
```diff
@@ -88,12 +98,32 @@ def result_hash(code, msg_args=nil)
   # result_array = [result1, result2]
   # result1 = {response_code => msg}
   # result2 = {response_code => msg}
-  def log_results
+  def report_results
```
ndushay: this change should have some tests, but since I already wrote some of them, you can leave that to me.

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    let(:body) { "<process name='moab-valid' status='error' errorMessage='#{error_msg}'/>" }
+
+    before do
+      Settings.workflow_services_url = 'https://sul-lyberservices-test.stanford.edu/workflow/'
+      stub_request(:put, put)
+        .with(body: body,
+              headers: headers)
+        .to_return(status: 204, body: "", headers: {})
+    end
+    it 'updates preservationWF with invalid moab errors' do
```
ndushay: `returns Faraday::Response` ?   I think that's what it is showing.

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(described_class.update_workflow('jj925bx9565', 'moab-valid', error_msg)).to be_an_instance_of(Faraday::Response)
+    end
+
+    it 'rescues for Faraday::Error' do
+      allow(described_class).to receive(:update_workflow).and_raise(Faraday::Error)
+      expect { described_class.update_workflow('jj925bx9565', 'moab-valid', 'Foo() error') }.to raise_error(Faraday::Error)
```
ndushay: I don't understand this test.  The expect statement says you expect to raise a `Faraday::Error` ... so where does the rescue get tested?

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+
+RSpec.describe WorkflowErrorsReporter do
+  let(:put) { 'https://sul-lyberservices-test.stanford.edu/workflow/dor/objects/druid:jj925bx9565/workflows/preservationWF/moab-valid' }
+  let(:headers) { { 'Content-Type' => 'application/xml' } }
+
+  describe '.update_workflow' do
+    let(:error_msg) do
```
ndushay: isn't this actually a "result" per pohandler_results?

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        .with(body: body,
+              headers: headers)
+        .to_return(status: 200, body: "", headers: {})
+    end
+    it 'returns Rails warning' do
+      Settings.workflow_services_url = ''
+      expect(Rails.logger).to receive(:warn).with('no workflow hookup - assume you are in test or dev environment')
+      described_class.update_workflow('jj925bx9565', 'moab-valid', [{ 13 => "Foo() Invalid Moab" }])
+    end
+  end
```
ndushay: do you want a test for a bad process name?  Or is that silly?   What about when a druid isn't recognized by workflows (like we will have in stage) ?

ndushay:
More on the process name:  since "error" is going to be the status of the thing, and most often, it will NOT be in an "error" status, I'm not sure we want "error" as part of the name.  at this point, we're specifically saying that something during moab-to-catalog-existence-and-version-auditing process had an error, otherwise we're silent.  We know we will also have catalog-to-moab-existence-and-version-auditing, and checksum validations.   If we want to be generic, at the moment, they are all some sort of 'preservation-audit-process' ??

.rubocop.yml
```diff
@@ -94,6 +97,9 @@ Metrics/LineLength:
     - 'spec/services/preserved_object_handler_create_spec.rb'
     - 'spec/services/preserved_object_handler_update_version_spec.rb'
     - 'spec/services/shared_examples_preserved_object_handler.rb'
+    - 'spec/models/status_spec.rb' # Line length of 132 for better readability
```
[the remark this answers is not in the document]
SaravShah: oops i think i messed up during a merge conflict error

spec/services/workflow_errors_reporter_spec.rb
ndushay: Trying to find examples for you (not even sure you need webmock?):

- https://github.com/sul-dlss/sul-requests/blob/master/spec/models/library_hours_api_spec.rb#L18-L25
- https://github.com/sul-dlss/sul-requests/blob/master/spec/models/symphony_curr_loc_request_spec.rb#L48-L64
- https://github.com/sul-dlss/sul-requests/blob/master/spec/models/symphony_user_name_request_spec.rb#L27-L51
- https://github.com/sul-dlss/sul-requests/blob/master/spec/models/proxy_access_spec.rb#L64-L98

ndushay: - https://github.com/sul-dlss/argo/blob/master/spec/models/index_queue_spec.rb#L7-L12

- https://github.com/sul-dlss/dor-workflow-service/blob/master/spec/workflow_service_spec.rb#L76-L79 <-- might be the closest to the sort of test I *think* you were going for???

ndushay: Note also:  do we want to test the response code from Dor workflows?  Make sure it didn't get an error?  Log it if it did?

There can be a connection failure;  there can be bad http response codes ...

SaravShah: okay cool - i'll try to disable webmock and try these

SaravShah: if i change the test to : `      expect { described_class.update_workflow('jj925bx9565', 'moab-val', 'Foo() error') }.to raise_error(Faraday::Error)
` i think this satisfies the test?

SaravShah: (note its `moab-val` instead of `moab-valid`)

.rubocop.yml
```diff
@@ -86,9 +88,9 @@ Metrics/LineLength:
   Exclude:
     - 'app/services/preserved_object_handler.rb' # line 269 is 121
     - 'app/services/preserved_object_handler_results.rb' # line 35 is 121
+    - 'app/services/workflow_errors_reporter.rb' # line 7 is 143
     - 'spec/lib/audit/moab_to_catalog_spec.rb' # 1 line 126
     - 'spec/models/status_spec.rb' # Line length of 132 for better readability
-    - spec/services/preserved_object_handler_spec.rb
```
ndushay: always delighted when we remove exclusions for rubocop ...

app/services/preserved_object_handler_results.rb
```diff
@@ -85,15 +95,34 @@ def result_hash(code, msg_args=nil)
...
-  # result2 = {response_code => msg}
-  def log_results
+  # output results to Rails.logger and send errors to WorkflowErrorsReporter
+  # @return result_array
+  def report_results
+    candidate_workflow_results = []
     result_array.each do |r|
       severity = self.class.logger_severity_level(r.keys.first)
       msg = r.values.first
       Rails.logger.log(severity, msg)
```
ndushay: I wonder if the code would be clearer if the above 3 lines were in a separate method `log_result` (singular)

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,50 @@
+require 'faraday'
+require 'retries'
+# send errors to preservationWF workflow for an object via ReST calls.
+class WorkflowErrorsReporter
+
+  def self.update_workflow(druid, process_name, error_message)
+    response = http_workflow_request(druid, process_name, error_message)
+
+    # withouth this `if response` I was getting:
```
ndushay: `withouth` 😆 

I think you can leave out the comment about "if response"

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,50 @@
...
+    # withouth this `if response` I was getting: 
+    # got: (#<NameError: uninitialized class variable @@connection in WorkflowErrorsReporter>)
+    # had to add this condition to get around it. This is for the test:
+    # workflow_errors_reporter_spec.rb:41
+    if response 
+      if response.status == 204
+        Rails.logger.debug("Server responded with status 204, PUT request went through")
+      else
+        # There is not `response.status == 400` because double logging will occur
+        # with the rescue. The rescue will handle the status 400.
```
ndushay: how about this comment becomes something like "note that status == 400 will be handled by the rescue clause"

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'rails_helper'
+
+RSpec.describe WorkflowErrorsReporter do
+  let(:put) { 'https://sul-lyberservices-test.stanford.edu/workflow/dor/objects/druid:jj925bx9565/workflows/preservationWF/moab-valid' }
```
ndushay: "put" ?  really?  Not "full_url" or something like that?

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  let(:headers) { { 'Content-Type' => 'application/xml' } }
+  let(:result_array) do
```
ndushay: this is a single result, not an array

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    { 13 => "Invalid moab, validation error...ential version directories." }
+  end
+  let(:body) { "<process name='moab-valid status='error' errorMessage='{13=>\"Invalid moab, validation error...ential version directories.\"}'/>" }
```
ndushay: you can use previous "let" variables in other "let" definitions ... which could make this more readable (e.g. if you have a var "result_msg" ...)

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  describe '.update_workflow with 204' do
```
ndushay: I would tweak this slightly (and no, i do not love infinite nesting)

```ruby
context '.update_workflow' do
  it '204 response' do
     # code within the before
     # actual test code showing you expect message
  end

  it '400 response' do
   # code within THAT before
   # actual test code
  end
 
  it 'invalid blah blah url' do
    etc
  end
end
```

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,50 @@
...
+        Rails.logger.warn("PUT request did not go through")
```
ndushay: I think you want a little more context here:  what put request?  for what?  how about  "#{druid} - unable to update workflow: #{response.body}" ??

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,50 @@
...
+  def self.update_workflow(druid, process_name, error_message)
+    response = http_workflow_request(druid, process_name, error_message)
+
+    # withouth this `if response` I was getting: 
+    # got: (#<NameError: uninitialized class variable @@connection in WorkflowErrorsReporter>)
+    # had to add this condition to get around it. This is for the test:
+    # workflow_errors_reporter_spec.rb:41
+    if response 
+      if response.status == 204
+        Rails.logger.debug("Server responded with status 204, PUT request went through")
```
ndushay: which server?  How about "#{druid} - sent error to workflow service for preservationWF #{process_name}"

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,50 @@
...
+      else
+        # There is not `response.status == 400` because double logging will occur
+        # with the rescue. The rescue will handle the status 400.
+        Rails.logger.warn("PUT request did not go through")
+      end
+    end
+  rescue Faraday::Error, StandardError => e
+    Rails.logger.warn(e)
```
ndushay: again, maybe prefix with a little context:  "#{druid} unable to update workflow #{e.inspect}" ???

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      Settings.workflow_services_url = 'https://sul-lyberservices-test.stanford.edu/workflow/'
+      stub_request(:put, put)
+         .with(body: body,
+              headers: headers)
+         .to_return(status: 204, body: "", headers: {})
+    end
+
+    it 'returns a 204 response status with rails info' do
+      # used debug here because i couldn't get receive(:info) to work??
+      expect(Rails.logger).to receive(:debug).with("put request went through")
```
ndushay: i think this is debug level information, so we are ok.  I'm sure we could get 'info' to work but would rather move on.  Remove the comment of line 21?

ndushay:
... and the specs need to pass, too.  LOL.

ndushay:
last thought about workflow process name:  how about 'preservation-audit' ?  "process" is just ... noise.

spec/services/workflow_errors_reporter_spec.rb
SaravShah: i need each of the `it` blocks to have a `describe` block so i can use the `before do...` block

ndushay: you don't need a `before do` block if you just put that code inside the `it` block.  That is, if the setup is only for a single test, it can be part of a single test.

If there is setup applying to multiple tests, it can be a `before` block inside the context block.  `context` is basically equal to `describe` anyway.

SaravShah: oh right, i knew that..

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,44 @@
...
+class WorkflowErrorsReporter
+
+  def self.update_workflow(druid, process_name, error_message)
+    response = http_workflow_request(druid, process_name, error_message)
+    if response
+      if response.status == 204
+        Rails.logger.debug("#{druid} - sent error to workflow service for preservationWF #{process_name}")
+      else
+        # Note: status == 400 will be handled by the rescue clause
+        Rails.logger.warn("#{druid} - unable to update workflow: #{response.body}")
```
ndushay: include the process_name (and error_message?)?

app/services/workflow_errors_reporter.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+    end
+  rescue StandardError => e
+    Rails.logger.warn("#{druid} - unable to update workflow #{e.inspect}")
```
ndushay: include the process_name (and error_message?)?

Issue #451, opened 2018-01-09
update_version:  if skipping versions, report as its own  error (still update catalog)


ndushay:
@blalbrit says that he would like an error reported to dor-workflows when **update_version** (as called by **ingest robot**):
- skips a version (e.g.  incoming version is 4, but PC  version is 2).   We should go ahead and update the preserved copy object, but we should also report the skip to dor workflows and log it.
- incoming <= version in the catalog (this is an ERROR for an update_version call).  We should *not* change the version in the catalog, we should set an error status (I think) and we should report the error to dor workflows and log it.

The above behavior also applies to what we want when M2C existence check finds incoming version > catalog version:
- if incoming version <= version in catalog:  error condition;  status becomes error, error reported in dor workflows, logs, etc.
- incoming version skips version in catalog (online moab is version 4, but catalog is version 2):  update PC object, report skip to dor workflows and log it.

jmartin-sul:
we'd want confirmation from @andrewjbtw if we make changes on this front, but revisiting this ticket while doing some backlog grooming w/ @edsu, i think:
* "skips a version (e.g. incoming version is 4, but PC version is 2). We should go ahead and update the preserved copy object, but we should also report the skip to dor workflows and log it."  -- i'd think we probably want to e.g. log and honeybadger alert, but if this is a somewhat expected (and not very worrisome) situation, i'd say we probably don't want to report an error to workflow service, because we don't want to gate accessioning based on clearing this (non?) error?  we also have added the dor-services-app event service since this ticket was filed -- perhaps that'd be a more appropriate reporting destination?
* warrants looking for confirmation, but i think this is already the case?  "incoming <= version in the catalog (this is an ERROR for an update_version call). We should not change the version in the catalog, we should set an error status (I think) and we should report the error to dor workflows and log it."
  * if we do have to make a change on this front, might be better as a separate ticket?

ndushay:
@jmartin-sul this seems worth a discussion with @andrewjbtw.  Maybe this no longer happens.  Maybe it does happen, silently.  What does Andrew know?  What does he wish to happen?

Maybe the answers will allow us to just close the ticket.

andrewjbtw:
I've never seen this and must have missed the earlier mention of me while on vacation. I now wonder if there are moabs where this happened and what we should do about them. I don't think it's possible to get into this situation now.

andrewjbtw:
[...]

PR #456, opened 2018-01-12
Check existence revalidates on bad status

### LINKED ISSUE #431: (M2C) modify check_existence to deal with pres copies that already have a version mismatch status

### LINKED ISSUE #433: (C2M) pres_copy with VALIDITY_UNKNOWN_STATUS becomes OK_STATUS how?
How do we expect any VALIDITY_UNKNOWN_STATUS objects to get validated and set to OK_STATUS ?  

Is this part of C2M?   and/or update_version_after_validation when something gets a new version via a robot?

Objects are assigned VALIDITY_UNKNOWN_STATUS when they are created without moab validation.

jmartin-sul:
~~pretty close, added new tests, still need to:~~
- [x] move most of the new tests to a shared example, except for the ones that deal with status becoming `EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS`, since the conditions for that differ between `check_existence` and `confirm_version`.
- [x] re-implement the `EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS` tests appropriately for `confirm_version` (e.g. add a test where the version increases, which is fine by `check_existence`, but not `confirm_version`).
- [x] call the shared example on `check_existence` and `confirm_version` in the appropriate place in the respective spec files.

paired with @tingulfsen for most of this.

supersedes #438 

closes #431

closes #433

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -37,9 +37,9 @@ class PreservedObjectHandler
     elsif PreservedObject.exists?(druid: druid)
       handler_results.add_result(PreservedObjectHandlerResults::OBJECT_ALREADY_EXISTS, 'PreservedObject')
     elsif moab_validation_errors.empty?
-      create_db_objects(PreservedCopy::OK_STATUS, true)
+      create_db_objects(PreservedCopy::OK_STATUS)
     else
-      create_db_objects(PreservedCopy::INVALID_MOAB_STATUS, true)
+      create_db_objects(PreservedCopy::INVALID_MOAB_STATUS)
     end
 
     handler_results.report_results
@@ -76,16 +76,14 @@ class PreservedObjectHandler
           found_expected_version = (incoming_vers_eq_pc_vers || incoming_vers_gt_pc_vers)
           check_status_if_not_ok_or_unexpected_version(pres_copy, found_expected_version)
 
-          ran_moab_validation = false
           if incoming_vers_eq_pc_vers
             handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_copy.class.name)
             handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_object.class.name)
           elsif incoming_vers_gt_pc_vers
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_copy.class.name)
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_object.class.name)
-            ran_moab_validation = true
             if moab_validation_errors.empty?
-              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, ran_moab_validation)
+              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size)
               update_status(pres_copy, PreservedCopy::OK_STATUS)
               pres_object.current_version = incoming_version
               update_db_object(pres_object)
@@ -95,9 +93,8 @@ class PreservedObjectHandler
           else # incoming_version < pres_copy.version
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, pres_copy.class.name)
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_LESS_THAN_DB_OBJECT, pres_object.class.name)
-            ran_moab_validation = true # moab validation ran because check_status_if_not_ok_or_unexpected_version saw an unexpected incoming version == true flag
           end
-          update_pc_audit_timestamps(pres_copy, ran_moab_validation, true)
+          update_pc_audit_timestamps(pres_copy, true)
           update_db_object(pres_copy)
         end
         handler_results.remove_db_updated_results unless transaction_ok
@@ -107,9 +104,9 @@ class PreservedObjectHandler
     else
       handler_results.add_result(PreservedObjectHandlerResults::OBJECT_DOES_NOT_EXIST, 'PreservedObject')
       if moab_validation_errors.empty?
-        create_db_objects(PreservedCopy::OK_STATUS, true)
+        create_db_objects(PreservedCopy::OK_STATUS)
       else
-        create_db_objects(PreservedCopy::INVALID_MOAB_STATUS, true)
+        create_db_objects(PreservedCopy::INVALID_MOAB_STATUS)
       end
     end
     handler_results.report_results
@@ -139,7 +136,7 @@ class PreservedObjectHandler
       if endpoint.endpoint_type.endpoint_class == 'online'
         if moab_validation_errors.empty?
           # NOTE: we deal with active record transactions in update_online_version, not here
-          update_online_version(true, PreservedCopy::OK_STATUS)
+          update_online_version(PreservedCopy::OK_STATUS)
         else
           update_pc_invalid_moab
         end
@@ -159,7 +156,7 @@ class PreservedObjectHandler
       Rails.logger.debug "update_version #{druid} called"
       if endpoint.endpoint_type.endpoint_class == 'online'
         # NOTE: we deal with active record transactions in update_online_version, not here
-        update_online_version(false, nil, true)
+        update_online_version(nil, true)
       elsif endpoint.endpoint_type.endpoint_class == 'archive'
         # TODO: create a new PC record for the new archived version on the endpoint
       end
@@ -175,6 +172,7 @@ class PreservedObjectHandler
     moab = Moab::StorageObject.new(druid, object_dir)
     object_validator = Stanford::StorageObjectValidator.new(moab)
     moab_errors = object_validator.validation_errors(Settings.moab.allow_content_subdirs)
+    @ran_moab_validation = true
     if moab_errors.any?
       moab_error_msgs = []
       moab_errors.each do |error_hash|
@@ -185,7 +183,11 @@ class PreservedObjectHandler
     moab_errors
   end
 
-  def create_db_objects(status, ran_moab_validation=false)
+  def ran_moab_validation?
+    @ran_moab_validation
+  end
+
+  def create_db_objects(status)
     pp_default_id = PreservationPolicy.default_policy_id
     transaction_ok = with_active_record_transaction_and_rescue do
       po = PreservedObject.create!(druid: druid,
@@ -198,7 +200,7 @@ class PreservedObjectHandler
         endpoint: endpoint,
         status: status
       }
-      if ran_moab_validation
+      if ran_moab_validation?
         t = Time.current
         pc_attrs[:last_version_audit] = t
         pc_attrs[:last_moab_validation] = t
@@ -211,7 +213,7 @@ class PreservedObjectHandler
 
   # TODO: this is "too complex" per rubocop: shameless green implementation
   # NOTE: if we can reduce complexity, remove Metrics/PerceivedComplexity exception in .rubocop.yml
-  def update_online_version(ran_moab_validation=false, status=nil, set_status_to_unexp_version=false)
+  def update_online_version(status=nil, set_status_to_unexp_version=false)
     transaction_ok = with_active_record_transaction_and_rescue do
       pres_object = PreservedObject.find_by!(druid: druid)
       pres_copy = PreservedCopy.find_by!(preserved_object: pres_object, endpoint: endpoint) if pres_object
@@ -225,8 +227,8 @@ class PreservedObjectHandler
         handler_results.add_result(code, pres_copy.class.name)
         handler_results.add_result(code, pres_object.class.name)
 
-        update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, ran_moab_validation)
-        update_status(pres_copy, status) if status && ran_moab_validation
+        update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size)
+        update_status(pres_copy, status) if status && ran_moab_validation?
         update_db_object(pres_copy)
         pres_object.current_version = incoming_version
         update_db_object(pres_object)
@@ -234,7 +236,7 @@ class PreservedObjectHandler
         if set_status_to_unexp_version
           status = PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS
         end
-        update_pc_unexpected_version(pres_copy, pres_object, status, ran_moab_validation)
+        update_pc_unexpected_version(pres_copy, pres_object, status)
       end
     end
 
@@ -255,7 +257,7 @@ class PreservedObjectHandler
       pres_copy = PreservedCopy.find_by!(preserved_object: pres_object, endpoint: endpoint) if pres_object
       # FIXME: what if there is more than one associated pres_copy?
       update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
-      update_pc_audit_timestamps(pres_copy, true, false)
+      update_pc_audit_timestamps(pres_copy, false)
       update_db_object(pres_copy)
     end
     handler_results.remove_db_updated_results unless transaction_ok
@@ -296,13 +298,13 @@ class PreservedObjectHandler
     update_status(pres_copy, PreservedCopy::OK_STATUS)
   end
 
-  def update_pc_unexpected_version(pres_copy, pres_object, new_status, ran_moab_validation)
+  def update_pc_unexpected_version(pres_copy, pres_object, new_status)
     handler_results.add_result(PreservedObjectHandlerResults::UNEXPECTED_VERSION, 'PreservedCopy')
     version_comparison_results(pres_copy, :version)
     version_comparison_results(pres_object, :current_version)
 
     update_status(pres_copy, new_status) if new_status
-    update_pc_audit_timestamps(pres_copy, ran_moab_validation, true)
+    update_pc_audit_timestamps(pres_copy, true)
     update_db_object(pres_copy)
   end
 
@@ -324,7 +326,7 @@ class PreservedObjectHandler
       else
         handler_results.add_result(PreservedObjectHandlerResults::UNEXPECTED_VERSION, pres_copy.class.name)
       end
-      update_pc_audit_timestamps(pres_copy, false, true)
+      update_pc_audit_timestamps(pres_copy, true)
       update_db_object(pres_copy)
     end
     handler_results.remove_db_updated_results unless transaction_ok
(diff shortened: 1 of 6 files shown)


jmartin-sul:
i think this is ready for review.  in the process of trying to figure out why the tests i initially added as skipped in the second commit weren't working, i saw what seemed like a couple convenient/easy refactorings that'd help me out.

the first was to have the method that was validating the object in the case of an already bad status also do the validation and status setting if the incoming version is unexpected compared to the catalog.  with a bit of extra mocking (since it triggered slightly more frequent object validation), this allowed enabling the previously skipped tests.  it also allowed a bit of simplification in the status setting that `check_existence` does.

the second was to centralize the setting of the `ran_moab_validation` flag.  i think there was a moment where that felt directly related to the rest of this PR's work.  i'm not sure it actually is, but it did feel like an improvement in maintainability in the vein of some other refactoring we've done (e.g. turning result codes into a centrally managed list).

i tried to build this up with a logical ordering of commits, with the ones i thought might be more controversial saved for later.  but fixing the skipped tests seemed important, because they seemed to be testing reasonable expectations.  and hopefully i wasn't way off on the validation flag one.

probably easier to review commit by commit.

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -386,6 +376,38 @@
...
+            expect(pc.reload.status).to eq PreservedCopy::OK_STATUS
+          end
+          it 'had INVALID_MOAB_STATUS, was remediated, should now have OK_STATUS' do
+            pc.status = PreservedCopy::INVALID_MOAB_STATUS
+            pc.save!
+            allow(po_handler).to receive(:moab_validation_errors).and_return([])
+            po_handler.check_existence
+            expect(pc.reload.status).to eq PreservedCopy::OK_STATUS
+          end
+          it 'had EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS, seems to have an acceptable version now' do
```
ndushay: "seems to have" ?   hahaha - it's a test - I think we could be assertive here.  Mostly added this comment to amuse myself, not because I think changing the text is important.

spec/services/preserved_object_handler_confirm_version_spec.rb
```diff
@@ -124,6 +124,31 @@
         end
       end
```
ndushay: question:  for "confirm_version" do we now have the capacity to easily report when incoming version > catalog version + 1?    That is, if incoming is 4 and catalog is 2 ... does the code now support us detecting this case?  Asking because it pertains to ticket #451

spec/services/shared_examples_preserved_object_handler.rb
```diff
@@ -474,3 +474,106 @@
...
+    expect(pc.reload.status).to eq PreservedCopy::OK_STATUS
+  end
+  it 'had VALIDITY_UNKNOWN_STATUS, but is now INVALID_MOAB_STATUS' do
+    pc.status = PreservedCopy::VALIDITY_UNKNOWN_STATUS
+    pc.save!
+    allow(po_handler).to receive(:moab_validation_errors).and_return([{ Moab::StorageObjectValidator::MISSING_DIR => 'err msg' }])
+    po_handler.send(method_sym)
+    expect(pc.reload.status).to eq PreservedCopy::INVALID_MOAB_STATUS
+  end
+  it 'had EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS, seems to have an acceptable version now' do
```
ndushay: another "seems to"

app/services/preserved_object_handler.rb
```diff
@@ -76,7 +76,7 @@ def check_existence
           incoming_vers_eq_pc_vers = (incoming_version == pres_copy.version)
           incoming_vers_gt_pc_vers = (incoming_version > pres_copy.version)
           found_expected_version = (incoming_vers_eq_pc_vers || incoming_vers_gt_pc_vers)
-          check_status_if_not_ok(pres_copy, found_expected_version)
+          check_status_if_not_ok_or_unexpected_version(pres_copy, found_expected_version)
```
ndushay: I'm wondering if the code is clearer if you take each comparator relationship separately below and call `check_status_if_not_ok_or_unexpected_version(pres_copy, false)` or true as the case may be.  

The reason I suggest this is the comment with line 100 about ran_moab_validation setting -- it feels like there is some slight of hand going on now ...

But have yet to look at the whole -- still going commit by commit ...

app/services/preserved_object_handler.rb
```diff
@@ -272,14 +267,19 @@ def update_pc_invalid_moab
     handler_results.remove_db_updated_results unless transaction_ok
   end
 
-  def check_status_if_not_ok(pres_copy, found_expected_version)
-    if pres_copy.status != PreservedCopy::OK_STATUS
+  # TODO: what if this checked status in the case of version mismatcch also?  would that simplify
```
ndushay: mismatccccch

app/services/preserved_object_handler.rb
```diff
@@ -272,14 +267,19 @@ def update_pc_invalid_moab
...
+  # check_existence and/or confirm_version in their handling after calling this?  could they just
+  # deal with the result code wranging after calling this if it also checked status on vers mismatch?
```
ndushay: The question you raise here is worthy - I wonder if it should be a refactor ticket rather than a comment here?  Or in addition?  Or maybe worth no more than 1 hour's experimentation?

app/services/preserved_object_handler.rb
```diff
@@ -272,14 +267,19 @@ def update_pc_invalid_moab
...
+  def check_status_if_not_ok_or_unexpected_version(pres_copy, found_expected_version)
+    # if pres_copy.status != PreservedCopy::OK_STATUS
+    unless pres_copy.status == PreservedCopy::OK_STATUS && found_expected_version
       set_status_as_seen_on_disk(pres_copy, found_expected_version)
     end
   end
 
   # given a PreservedCopy instance and whether the caller found the expected version of it on disk, this will perform
   # other validations of what's on disk, and will update the status accordingly
+  # FIXME: this method needs to update validation timestamps, and tests need to get fixed to reflect that
```
ndushay: I hope this is addressed in a later commit ...

app/services/preserved_object_handler.rb
```diff
@@ -272,14 +267,19 @@ def update_pc_invalid_moab
     handler_results.remove_db_updated_results unless transaction_ok
   end
 
-  def check_status_if_not_ok(pres_copy, found_expected_version)
-    if pres_copy.status != PreservedCopy::OK_STATUS
+  # TODO: what if this checked status in the case of version mismatcch also?  would that simplify
+  # check_existence and/or confirm_version in their handling after calling this?  could they just
+  # deal with the result code wranging after calling this if it also checked status on vers mismatch?
+  def check_status_if_not_ok_or_unexpected_version(pres_copy, found_expected_version)
```
ndushay: such a long name.  How about `validate_for_status`?  or `validate_status` or `status_validation` or ...

app/services/preserved_object_handler.rb
```diff
@@ -174,19 +176,25 @@ def update_version
...
-    object_validator = Stanford::StorageObjectValidator.new(moab)
-    moab_errors = object_validator.validation_errors(Settings.moab.allow_content_subdirs)
-    @ran_moab_validation = true
-    if moab_errors.any?
-      moab_error_msgs = []
-      moab_errors.each do |error_hash|
-        error_hash.each_value { |msg| moab_error_msgs << msg }
+    # validation is a bit expensive, so memoize the result.  keeps us from having to worry about a
+    # performance hit if a caller wants to check whether there are validation errors from more than
+    # one conditional statement.
```
ndushay: i think you can leave out this comment - memoization is a common practice and devs know why it is used.

app/services/preserved_object_handler.rb
```diff
@@ -73,33 +73,32 @@ def check_existence
...
+          if incoming_vers_eq_pc_vers
             handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_copy.class.name)
             handler_results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, pres_object.class.name)
-          elsif incoming_version > pres_copy.version
+          elsif incoming_vers_gt_pc_vers
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_copy.class.name)
             handler_results.add_result(PreservedObjectHandlerResults::ARG_VERSION_GREATER_THAN_DB_OBJECT, pres_object.class.name)
-            ran_moab_validation = true
+            # we're not too worried if the version increased, but we normally expect it to be kept current
+            # by calls from ingest robots, so validate to make sure all's well.
```
ndushay: not sure this comment is giving us anything -- i would prefer to read the code to figure it out.  ("so if the incoming is greater then it validates ...")

app/services/preserved_object_handler.rb
```diff
@@ -73,33 +73,32 @@ def check_existence
 
           raise_rollback_if_pc_po_version_mismatch(pres_copy.version, pres_object.current_version)
 
-          ran_moab_validation = false
-          if incoming_version == pres_copy.version
+          incoming_vers_eq_pc_vers = (incoming_version == pres_copy.version)
+          incoming_vers_gt_pc_vers = (incoming_version > pres_copy.version)
+          found_expected_version = (incoming_vers_eq_pc_vers || incoming_vers_gt_pc_vers)
+          check_status_if_not_ok_or_unexpected_version(pres_copy, found_expected_version)
```
ndushay: what if you put `check_status_if_not_ok_or_unexpected_version(pres_copy, true)` (or false)  in each of the if clauses below?   I have a preference for actually seeing "incoming_version > pres_copy.version" over the variable.

app/services/preserved_object_handler.rb
```diff
@@ -73,33 +73,32 @@ def check_existence
...
-            ran_moab_validation = true
+            # we're not too worried if the version increased, but we normally expect it to be kept current
+            # by calls from ingest robots, so validate to make sure all's well.
             if moab_validation_errors.empty?
-              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size, ran_moab_validation)
-              update_status(pres_copy, PreservedCopy::OK_STATUS)
+              update_preserved_copy_version_etc(pres_copy, incoming_version, incoming_size)
               pres_object.current_version = incoming_version
               update_db_object(pres_object)
+              # don't have to set status here: check_status_if_not_ok_or_unexpected_version should've corrected a bad status
```
ndushay: putting the call to `check_status_if_not_ok_or_unexpected_version(pres_copy, true)` within the if-elsif-else statement makes this comment unnecessary.   As it stands, I feel like it's calling out some slight of hand.

Or maybe simply removing this comment works.

app/services/preserved_object_handler.rb
```diff
@@ -261,19 +271,51 @@ def update_pc_invalid_moab
...
-      update_pc_audit_timestamps(pres_copy, true, false)
+      update_pc_audit_timestamps(pres_copy, false)
       update_db_object(pres_copy)
     end
     handler_results.remove_db_updated_results unless transaction_ok
   end
 
-  def update_pc_unexpected_version(pres_copy, pres_object, new_status, ran_moab_validation)
+  def check_status_if_not_ok_or_unexpected_version(pres_copy, found_expected_version)
+    # if pres_copy.status != PreservedCopy::OK_STATUS
```
ndushay: I think this comment can go.  I think the unless reads clearly.

app/services/preserved_object_handler.rb
```diff
@@ -261,19 +271,51 @@ def update_pc_invalid_moab
...
+    unless pres_copy.status == PreservedCopy::OK_STATUS && found_expected_version
+      set_status_as_seen_on_disk(pres_copy, found_expected_version)
+    end
+  end
+
+  # given a PreservedCopy instance and whether the caller found the expected version of it on disk, this will perform
+  # other validations of what's on disk, and will update the status accordingly
+  # FIXME: this method probably needs to update validation timestamps?  and tests would need to get fixed to reflect that.
```
ndushay: i think this FIXME comment can now go away?

app/services/preserved_object_handler.rb
```diff
@@ -261,19 +271,51 @@ def update_pc_invalid_moab
...
+  def set_status_as_seen_on_disk(pres_copy, found_expected_version)
+    # TODO: do the check that'd set PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS
+    # TODO: nothing in the code seems to set ONLINE_MOAB_NOT_FOUND_STATUS atm, but everything we're checking right now is M2C and thus found by definition
```
ndushay: I think the above two comments could be combined.  Maybe "TODO: when we implement C2M, this is likely where we'll set status to ..."

app/services/preserved_object_handler.rb
```diff
@@ -261,19 +271,51 @@ def update_pc_invalid_moab
...
+    unless found_expected_version
+      update_status(pres_copy, PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS)
+      return
+    end
+
+    # TODO: do the check that'd set INVALID_CHECKSUM_STATUS
+
+    # TODO: do the check that'd set FIXITY_CHECK_FAILED_STATUS
+
+    # should automatically get things out of VALIDITY_UNKNOWN_STATUS, because if we fall through to here, we can set status to ok
```
ndushay: this comment could be removed as well.

app/services/preserved_object_handler.rb
```diff
@@ -286,14 +328,16 @@ def confirm_online_version
 
       raise_rollback_if_pc_po_version_mismatch(pres_copy.version, pres_object.current_version)
 
-      if incoming_version == pres_copy.version
+      found_expected_version = (incoming_version == pres_copy.version)
```
ndushay: again, I have a slight preference for `incoming_version == pres_copy.version` in the if statement.  Here, tho, the logic is simpler so my cognitive overload is lower.

app/services/preserved_object_handler.rb
```diff
@@ -73,11 +73,16 @@ def check_existence
 
           raise_rollback_if_pc_po_version_mismatch(pres_copy.version, pres_object.current_version)
 
+          incoming_vers_eq_pc_vers = (incoming_version == pres_copy.version)
```
SaravShah: i do have a slight preference for the conditions in the if statement, but i dont mind it here.

app/services/preserved_object_handler.rb
```diff
@@ -267,6 +272,36 @@ def update_pc_invalid_moab
     handler_results.remove_db_updated_results unless transaction_ok
   end
 
+  def check_status_if_not_ok(pres_copy, found_expected_version)
```
SaravShah: 👍

app/services/preserved_object_handler.rb
```diff
@@ -191,7 +189,11 @@ def moab_validation_errors
     moab_errors
   end
 
-  def create_db_objects(status, ran_moab_validation=false)
+  def ran_moab_validation?
+    @ran_moab_validation
```
SaravShah: 👍

app/services/preserved_object_handler.rb
jmartin-sul: yeah, that was a commented out line from a prior iteration, not an intentional comment, will remove

spec/services/preserved_object_handler_check_exist_spec.rb
jmartin-sul: "gee, sure looks like it has an acceptable version now, but really who are we to judge what's an acceptable version?"

spec/services/preserved_object_handler_confirm_version_spec.rb
jmartin-sul: i don't think this work makes it easier (or harder).  we'd still have to differentiate the case where the version is greater, since we only look for whether the versions are equal or not in `confirm_version`.

and while i don't think this PR makes that differentiation easier in `check_existence` either, that differentiation is already easier in `check_existence` than in `confirm_version`, since `check_existence` already differentiates the `incomnig_version > pres_copy.version` case from other version mismatches (in the `incoming_version > pres_copy.version` condition in `check_existence`, we could just see whether the difference is greater than one, and return different result codes accordingly).

app/services/preserved_object_handler.rb
jmartin-sul: ah, good question, hadn't thought about that explicitly... but i think what's happening is that if an instance var hasn't been set yet, and it gets read, `nil` will get returned, which is "falsey", so essentially yeah, it returns `false` for our use case.

but that's a little unclear, so i'll change the method to do `@ran_moab_validation ||= false`.  which'll explicitly init `@ran_moab_validation` to false if it's still unset, or leave it alone if it's already there.  hopefully things read a bit clearer with that change.

ndushay: good catch, @SaravShah and @jmartin-sul !

app/services/preserved_object_handler.rb
jmartin-sul: was about to say i think it's still needed, but now i agree it can go.  i think i left that comment before doing the `ran_moab_validation` refactoring.  now that that flag gets set on the POH instance whenever validation is run, i agree that this should be able to go away, because the top-level callers to `set_status_as_seen_on_disk` both also call `update_pc_audit_timestamps` (which now reads the centrally set `ran_moab_validation` flag).

sorry if you'd already worked all that out, i wrote that down as much for my own clarification as anything.

jmartin-sul:
i believe i've addressed all of the feedback from @ndushay and @SaravShah, should be ready for re-evaluation.

i just pushed follow on commits for the changes i made.  normally i'd be more of a stickler about a tidy commit history, but this isn't a terrible amount of churn, the squashing would be a bit of a pain, and this is probably a little easier for follow-on review anyway.  happy to cleanup if people would prefer that.

app/services/preserved_object_handler.rb
```diff
@@ -261,19 +263,42 @@ def update_pc_invalid_moab
...
     end
     handler_results.remove_db_updated_results unless transaction_ok
   end
 
-  def update_pc_unexpected_version(pres_copy, pres_object, new_status, ran_moab_validation)
+  # given a PreservedCopy instance and whether the caller found the expected version of it on disk, this will perform
+  # other validations of what's on disk, and will update the status accordingly
+  def set_status_as_seen_on_disk(pres_copy, found_expected_version)
+    # TODO: when we implement C2M, we should add a check here to look for the object on disk, and set PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS
+    # if needed.  or maybe the caller will have done that already, since it needs to try reading the moab to obtain incoming_version?
```
ndushay: this assumes C2M will be in pohandler.  That is ... unclear at this time.   Should this comment stay?

jmartin-sul: good point, i'd been assuming C2M would leverage POH methods that called this new method, but that seems less clear after the discussions this afternoon.  happy to remove this comment or make it appropriately vague.

ndushay: I dunno what to do;  confusion reigns in what some what call my "mind"

Issue #461, opened 2018-01-16
(C2M) rake tasks for C2M methods


ndushay:
(See PR #459 if it hasn't been merged yet)

We need rake tasks analogous to the ones for M2C check_existence.   

Note that C2M will take an additional parameter last_checked_version_b4_date.  

Format of the date needed is tied to #460

PR #465, opened 2018-01-17
rake tasks for c2m methods

SaravShah:
- Rake Task with and without profiling for `check_version_on_dir` and `check_version_all_dirs`

connects #461

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ For more info on postgres commands, see https://www.postgresql.org/docs/
 
 ### General Info About Running These Rake Tasks
 
+- Note: If the rake task takes multiple arguments, DO NOT put a space in between the commas.
+
 - You can monitor the progress of most rake tasks by tailing `log/production.log`, or by querying the database from Rails console using ActiveRecord. The tasks for large storage_roots can take a while -- check [the repo wiki for stats](https://github.com/sul-dlss/preservation_catalog/wiki) on the timing of past runs (and some suggested overview queries). Profiling will add some overhead.
 
 - Rake tasks must be run from the root directory of the project, with whatever `RAILS_ENV` is appropriate. Because the task can take days when run over all storage roots, consider running it in a [screen session](http://thingsilearned.com/2009/05/26/gnu-screen-super-basic-tutorial/) so you don't need to keep your connection open but can still see the output.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -131,6 +131,34 @@ RAILS_ENV=production bundle exec rake m2c_exist_all_storage_roots[profile]
 ```
 this will generate a log at, for example, ```log/profile_check_existence_for_all_storage_roots2017-12-11T14:25:31-flat.txt```
 
+### Run Catalog to Moab existence check for a single root and for all storage roots
```
ndushay: "for a single root or for all storage roots"  (or instead of and??)

Given that Julian and Vivian both expressed their confusion about what C2M is, perhaps either the heading or the first sentence after it should explain something like 
```
"given a catalog entry for an online moab, ensure that the online moab exists and that the catalog version matches the online moab version."
```

README.md
```diff
@@ -131,6 +131,34 @@ RAILS_ENV=production bundle exec rake m2c_exist_all_storage_roots[profile]
...
+
+- To run rake tasks below, give the name of the moab storage_root (e.g. from settings/development.yml) and the last checked before date as an argument.
+
+#### Single Root
+- Without profiling
+```ruby
+RAILS_ENV=production bundle exec rake c2m_check_version_on_dir[last_checked_b4_date, fixture_sr1]
```
ndushay: Use an exemplar date in the right format here, not 'last_checked_b4_date.  @tallenaz will know the correct date format.

README.md
```diff
@@ -131,6 +131,34 @@ RAILS_ENV=production bundle exec rake m2c_exist_all_storage_roots[profile]
 ```
 this will generate a log at, for example, ```log/profile_check_existence_for_all_storage_roots2017-12-11T14:25:31-flat.txt```
 
+### Run Catalog to Moab existence check for a single root and for all storage roots
+
+- To run rake tasks below, give the name of the moab storage_root (e.g. from settings/development.yml) and the last checked before date as an argument.
```
ndushay: we need a better explanation of "last checked before date" ... 

we're also going to want an exemplar date in the right format, not 'last_checked_b4_date. @tallenaz will know the correct date format.

How about:
```
To run rake tasks below, give a date (timestamp?  whatever tony says)** and the name of the moab storage_root (e.g. from settings/development.yml) as arguments.

The (date/timestamp) argument is a threshold:  it will run the check on all catalog entries which last had a version check BEFORE the argument.   It should be in the format xxxx.
```

(also ask @tallenaz if it is BEFORE or  BEFORE or EQUAL TO)

README.md
```diff
@@ -131,6 +131,34 @@ RAILS_ENV=production bundle exec rake m2c_exist_all_storage_roots[profile]
...
+
+#### Single Root
+- Without profiling
+```ruby
+RAILS_ENV=production bundle exec rake c2m_check_version_on_dir[last_checked_b4_date, fixture_sr1]
+```
+- With profiling
+```ruby
+RAILS_ENV=production bundle exec rake c2m_check_version_on_dir[last_checked_b4_date, fixture_sr1,profile]
```
ndushay: ditto on exemplar date, and do we want/need a space after the comma between `fixture_sr1` and `profile` or  must we NOT have one?

SaravShah: Good catch, NO spaces!

SaravShah:
made the changes requested! ready for review.

PR #468, opened 2018-01-18
Quick fix

SaravShah:
- adds single quote that was breaking workflow call
- add test to make sure this doesnt happen again


ndushay:
fix yer tests and I'll approve.

PR #469, opened 2018-01-18
Add query to C2M check_version_on_dir

### LINKED ISSUE #460: (C2M) SQL query for looping through catalog entries
(See PR #459 if it hasn't been merged yet)

We will be selecting PreservedCopy objects by storage_dir / endpoint and by last_version_audit date.   We want a SQL query to get us the correct PreservedCopy objects, but it must be done in a scalable way (i.e. chunked).

@jmartin-sul says: "re: the C2M query, i think you could use plain `find` if you just wanted a list of IDs to loop over, but passing a block to `find_each` was the more efficient approach, IIRC.  because then you don't have to go back and do a `find` on each object ID (i assume that `find_each` does some paging under the hood, and that it'll make multiple DB connections, but fewer connections than `find` to get the list of IDs and `find` on each returned ID).  one more random suggestion: if you only want the last N records, i'd have the query do the sort/limit (as opposed to querying broadly and filtering the list in ruby-land).  presumably we'll be querying on indexed cols, which should be pretty efficient."

tallenaz:
This method could use some further specs, and that work is ticketed (connects to #476)

Fixes #460

### DIFF
--- a/spec/lib/audit/catalog_to_moab_spec.rb
+++ b/spec/lib/audit/catalog_to_moab_spec.rb
@@ -17,6 +17,15 @@ RSpec.describe CatalogToMoab do
       expect(described_class).to receive(:check_catalog_version).at_least(3).times
       described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir)
     end
+    it 'will not check a PreservedCopy with a future last_version_audit date' do
+      expect(described_class).to receive(:check_catalog_version).exactly(6).times
+      described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir)
+      pc = PreservedCopy.first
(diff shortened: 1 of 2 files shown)


spec/lib/audit/catalog_to_moab_spec.rb
```diff
@@ -17,6 +17,15 @@
       expect(described_class).to receive(:check_catalog_version).at_least(3).times
       described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir)
     end
+    it 'will not check a pc with a future last_version_audit date' do
```
SaravShah: `PreservedCopy` instead of pc?

jmartin-sul:
see #476 for follow-on work to shore up testing

PR #473, opened 2018-01-18
add regex and return statement

SaravShah:


jmartin-sul:
more tests will be implemented by #474

PR #488, opened 2018-01-22
Test stack trace help

SaravShah:
- added new stack_trace
- moved `result_hash`,  `report_errors_to_workflows` to private
- rename variable from `x` to `results`

### DIFF
--- a/app/services/preserved_object_handler_results.rb
+++ b/app/services/preserved_object_handler_results.rb
@@ -91,10 +91,6 @@ class PreservedObjectHandlerResults
     result_array.delete_if { |res_hash| DB_UPDATED_CODES.include?(res_hash.keys.first) }
   end
 
-  def result_hash(code, msg_args=nil)
-    { code => result_code_msg(code, msg_args) }
-  end
-
   # output results to Rails.logger and send errors to WorkflowErrorsReporter
   # @return result_array
   def report_results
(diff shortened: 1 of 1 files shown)


app/services/preserved_object_handler_results.rb
```diff
@@ -107,25 +103,29 @@ def report_results
...
+    { code => result_code_msg(code, msg_args) }
+  end
+
   def report_errors_to_workflows(candidate_workflow_results, stack_trace)
     return if candidate_workflow_results.empty?
     value_array = []
-    candidate_workflow_results.each do |x|
-      x.each_value do |val|
+    candidate_workflow_results.each do |results|
+      results.each_value do |val|
```
ndushay: is this "results" or a single "result" ?

app/services/preserved_object_handler_results.rb
```diff
@@ -107,25 +103,29 @@ def report_results
         candidate_workflow_results << r
       end
     end
-    stack_trace = caller(1..1).first[/`\S+/].chop[1..-1]
+    stack_trace = caller(0..1).join("\n")
```
ndushay: my range stuff is a bit muzzy, and my caller knowledge more so  -- is this basically taking the first 2 lines from the caller stack?

SaravShah: yup first 2 lines

app/services/preserved_object_handler_results.rb
SaravShah: its an array of hashes, where it could be 1 result (if it a moab-validation error) or multiple ( if it is anything else in the preservation-audit step)

ndushay: isn't `candidate_workflow_results` the array of hashes?   and isn't the variable currently called `results` actually a single `result_hash`?

Names matter.

app/services/preserved_object_handler_results.rb
tallenaz: Yep, `caller` returns an Array of String lines, and this grabs the first two elements.

app/services/preserved_object_handler_results.rb
SaravShah: oh you're right, i had to print everything out. i changed it to `result_hash` makes it easier to understand

SaravShah: good catch

Issue #483, opened 2018-01-22
(C2M) make po/pc version matching method callable from M2C and C2M


ndushay:
Both M2C and C2M need to confirm that PO and PC versions match.

Possibly make a class method for pohandler -- this could be used by either check.

ndushay:
@tallenaz note that this ticket shouldn't be closed until your new `PreservedCopy.matches_po_current_version?` method is called by C2M code appropriately.

It should ALSO be called by M2C code appropriately, if that's simple.  :-). If it's not simple, make a refactor / patch party ticket to do it (may need a yield???)

PR #492, opened 2018-01-23
Add method to match pc version against its po current version

tallenaz:
- adds a method on preserved_copy to check the version agains its preserved_object current_version. I picked that location instead of the po_handler, since preserved copy is handed into `check_catalog_version`.

- implements the new method inside of `check_catalog_version` with an `if` statement that adds results to the handler (includes a new code for when versions match)

- specs for the new method

Connects to #483
(#483 is finished when the code in this PR is called from `C2M.check_catalog_version` )

### DIFF
--- a/lib/audit/catalog_to_moab.rb
+++ b/lib/audit/catalog_to_moab.rb
@@ -53,8 +53,12 @@ class CatalogToMoab
   def check_catalog_version
     druid = preserved_copy.preserved_object.druid
     results = PreservedObjectHandlerResults.new(druid, nil, nil, preserved_copy.endpoint)
-
-    # TODO: Pohandler.ensure_po_version_matches_this_pc_version (for non-archived, online moab) - see #483
+    unless preserved_copy.matches_po_current_version?
+      results.add_result(PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH,
+                         pc_version: preserved_copy.version,
+                         po_version: preserved_copy.preserved_object.current_version)
+      return
+    end
 
     # TODO: anything special if preserved_copy.status is not OK_STATUS? - see #480
 
--- a/spec/models/preserved_copy_spec.rb
+++ b/spec/models/preserved_copy_spec.rb
@@ -180,4 +180,18 @@ RSpec.describe PreservedCopy, type: :model do
       expect(pc.changed?).to eq true
     end
   end
+
+  context '#matches_po_current_version?' do
+    it 'returns true when its version matches its preserved objects current version' do
+      preserved_copy.version = 666
+      preserved_copy.preserved_object.current_version = 666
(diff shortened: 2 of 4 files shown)


tallenaz:
@ndushay I removed the extra code for when po and pc versions match, agreed that it would be too noisy. Tried to make it visible in the specs that they're not comparing nil values. This should probably lie fallow until 493 is merged, so that we have a spot to test out the `unless`.

spec/models/preserved_copy_spec.rb
```diff
@@ -180,4 +180,20 @@
       expect(pc.changed?).to eq true
     end
   end
+
+  context '#matches_po_current_version?' do
+    it 'returns true when its version matches its preserved objects current version' do
+      preserved_copy.version = 1
+      preserved_object.current_version = 1
```
ndushay: so the code is actually comparing `preserved_copy.preserved_object`'s version, not the plain `preserved_object` that is floating around in the spec.  I think this test could read:

```ruby
it 'returns true when its version matches its preserved objects current version' do
      preserved_copy.version = 1234
      preserved_copy.preserved_object.current_version = 1234
      expect(preserved_copy.matches_po_current_version?).to be true
    end
```

note that I picked version numbers that are highly unlikely to be occurring in our fixture objects ...

ndushay: Note that this PASSES:

```ruby
    it 'returns true when its version matches its preserved objects current version' do
      preserved_copy.version = 1
      preserved_object.current_version = 8
      expect(preserved_copy.matches_po_current_version?).to be true
    end
```

lib/audit/catalog_to_moab.rb
```diff
@@ -59,6 +59,10 @@ def self.check_version_all_dirs_profiled(last_checked_b4_date)
 
     # TODO: anything special if preserved_copy.status is not OK_STATUS? - see #480
 
+    unless preserved_copy.matches_po_current_version?
+      results.add_result(PreservedObjectHandlerResults::PC_PO_VERSION_MISMATCH, { pc_version: preserved_copy.version, po_version: preserved_copy.preserved_object.current_version })
```
ndushay: looking at the analogous code in M2C, this should stop processing, so needs a "return" line.

spec/models/preserved_copy_spec.rb
```diff
@@ -180,4 +180,18 @@
       expect(pc.changed?).to eq true
     end
   end
+
+  context '#matches_po_current_version?' do
+    it 'returns true when its version matches its preserved objects current version' do
+      preserved_copy.version = 666
```
SaravShah: 👿

ndushay:
again, looks great;  thinking AFTER #499 merge.

ndushay:
PR #500 does the method beginning reorg for you!

PR #495, opened 2018-01-23
[WIP] Moab exists on Catalog but does not exist online

### LINKED ISSUE #482: (C2M) handle error if catalog has a moab that is not found on disk
presumably, this means:

- add a PohandlerResult code
- ensure that error is logged
- ensure that error is reported to workflows

SaravShah:
### NOTE: IGNORE THIS PR, will close this PR once #508 is merged :)

- I would like 2 reviewers to approve this before it gets merged.

- Couple of thoughts on this PR:
1. In `catalog_to_moab_spec.rb:130`, I wanted to test:  `if 'moab is nil'`, we stop processing the rest of the code in `check_catalog_version`. I couldn't think of a straightforward way to do this and thought once we add more code in `check_catalog_version` we would have an easier way to test this.
2. Do we like `.moab_found_on_disk?` or `.moab_found_online`? 
3. Should I add a test that says something like: `if 'moab exists; code runs normally'`?

closes #482

### DIFF
--- a/lib/audit/catalog_to_moab.rb
+++ b/lib/audit/catalog_to_moab.rb
@@ -67,7 +67,8 @@ class CatalogToMoab
     object_dir = "#{storage_dir}/#{DruidTools::Druid.new(druid).tree.join('/')}"
     moab = Moab::StorageObject.new(druid, object_dir)
 
-    # TODO: report error if moab doesn't exist - see #482
+
+    return unless moab_found_on_disk?(moab, results)
 
     moab_version = moab.current_version_id
     catalog_version = preserved_copy.version
@@ -95,4 +96,14 @@ class CatalogToMoab
     # update_pc_audit_timestamps(preserved_copy, ran_moab_validation, true) - see #477
     # update_db_object(preserved_copy) - see #478
   end
+  
+  def moab_found_on_disk?(moab, results)
+    unless moab
+      results.add_result(PreservedObjectHandlerResults::ONLINE_MOAB_DOES_NOT_EXIST)
+      results.report_results
+      return false
+    end
+    true
+  end
+
 end
--- a/lib/audit/catalog_to_moab.rb
+++ b/lib/audit/catalog_to_moab.rb
@@ -61,16 +61,14 @@ class CatalogToMoab
                          po_version: preserved_copy.preserved_object.current_version)
       return
     end
-
     # TODO: anything special if preserved_copy.status is not OK_STATUS? - see #480
+    unless online_moab_found?(druid, storage_dir)
+      results.add_result(PreservedObjectHandlerResults::ONLINE_MOAB_DOES_NOT_EXIST)
+      results.report_results
+      return
+    end
 
-    object_dir = "#{storage_dir}/#{DruidTools::Druid.new(druid).tree.join('/')}"
-    moab = Moab::StorageObject.new(druid, object_dir)
-
-
-    return unless moab_found_on_disk?(moab, results)
-
-    moab_version = moab.current_version_id
+    moab_version = @moab.current_version_id
     catalog_version = preserved_copy.version
     if catalog_version == moab_version
       results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, preserved_copy.class.name)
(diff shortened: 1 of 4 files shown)


lib/audit/catalog_to_moab.rb
```diff
@@ -81,4 +81,13 @@ def self.check_version_all_dirs_profiled(last_checked_b4_date)
     # update_db_object(preserved_copy) - see #478
   end
 
+  private_class_method def self.moab_found_on_disk?(moab, results)
```
ndushay: I find myself slightly uncomfortable because this isn't a single purpose method and the name doesn't match what it's doing.   
- return true or false depending if moab is on disk  (`online_moab_found`) might be better as `online` is the language we use for the db to distinguish vs. archived??
- add error and report results if moab NOT found

How might we avoid this confusion?

(Note:  I am thinking we should use an INSTANCE of the class to actually do the work of check_catalog_version;  creating the instance inside the loop of `self.check_version_on_dir`)

we could have some methods:

```ruby
def online_moab(druid, storage_dir)
  # code here to instantiate Moab::Storage object from params passed to check_catalog_version
  # e.g.
  moab ||= begin
    object_dir = something_to_manipulate_storage_dir
    moab = Moab::StorageObject.new(druid, object_dir)
  end
end
```

so we then get something like:
```ruby
def check_catalog_version(preserved_copy, storage_dir)
    # TODO: Pohandler.ensure_po_version_matches_this_pc_version (for non-archived, online moab) - see #483

    results = PreservedObjectHandlerResults.new(druid, nil, nil, preserved_copy.endpoint)

    unless online_moab
        results.add_result(PreservedObjectHandlerResults::ONLINE_MOAB_DOES_NOT_EXIST)
        results.report_results
        return
     end
    
      ...

end
```

And further, for switching to an instance, I am thinking something vaguely like

```ruby
  def self.check_version_on_dir(last_checked_b4_date, storage_dir)
    # TODO: ensure last_checked_version_b4_date is in the right format for query - see #485
    pcs = PreservedCopy
          .where('last_version_audit < ? OR last_version_audit IS NULL', last_checked_b4_date)
          .order('last_version_audit IS NOT NULL, last_version_audit ASC')
    pcs.find_each do |pc|
      c2m = CatalogToMoab.new(pc, storage_dir)
      c2m.check_catalog_version
    end
  end
```

spec/lib/audit/catalog_to_moab_spec.rb
```diff
@@ -108,6 +108,30 @@
...
+        allow(Moab::StorageObject).to receive(:new).with(druid, instance_of(String)).and_return(nil)
+        pohandler_results = instance_double(PreservedObjectHandlerResults, report_results: nil)
+        allow(PreservedObjectHandlerResults).to receive(:new).and_return(pohandler_results)
+        expect(pohandler_results).to receive(:add_result).with(
+          PreservedObjectHandlerResults::ONLINE_MOAB_DOES_NOT_EXIST
+        )
+        described_class.send(:check_catalog_version, pres_copy, nil)
+      end
+      it 'stops processing .check_catalog_version' do
+        skip 'add this test once we add more code to .check_catalog_version???'
```
ndushay: for this, right now you could do  

```ruby
expect(moab_double_object).not_to receive(:current_version_id)
```

because that is the next statement, and it is executed regardless of which comparison case occurs.

spec/lib/audit/catalog_to_moab_spec.rb
```diff
@@ -108,6 +108,31 @@
...
+        pohandler_results = instance_double(PreservedObjectHandlerResults, report_results: nil)
+        allow(PreservedObjectHandlerResults).to receive(:new).and_return(pohandler_results)
+        expect(pohandler_results).to receive(:add_result).with(
+          PreservedObjectHandlerResults::ONLINE_MOAB_DOES_NOT_EXIST
+        )
+        described_class.send(:check_catalog_version, pres_copy, nil)
+      end
+      it 'stops processing .check_catalog_version' do
+        moab = instance_double(Moab::StorageObject)
+        expect(moab).not_to receive(:current_version_id)
```
jmartin-sul: seems like there should be a `described_class.send(:check_catalog_version, pres_copy, nil)` here?

ndushay: good catch, John!

SaravShah: d'oh!

spec/lib/audit/catalog_to_moab_spec.rb
```diff
@@ -108,6 +108,31 @@
       described_class.send(:check_catalog_version, pres_copy, nil)
     end
 
+    it 'calls .moab_found_on_disk?(moab,results)' do
```
tallenaz: I hate to be that guy, but could you put a space between the args? :)

lib/audit/catalog_to_moab.rb
tallenaz: I'm agreed on having a separation of concerns here.

SaravShah:
closed in favor #508

PR #501, opened 2018-01-24
HTML Escape text in error message sent to workflow

SaravShah:
When we sent this errorMessage:
`"PreservedObjectHandler(bp534qj1626, 7, 12877062629, services-disk02) Invalid moab, validation errors: [\"Version directory name not in 'v00xx' format: original-v1\"]"` 

- Ruby didn't escape the quotes, the errorMessage didn't have parsable structure when a PUT request was sent.
- WorkflowService responded with a 400 bad request.
- Wrote RSpec test to test this. 
- We discovered this in prod, but was reproducible in stage.

paired w/ @tallenaz @jmartin-sul

### DIFF
--- a/app/services/workflow_errors_reporter.rb
+++ b/app/services/workflow_errors_reporter.rb
@@ -32,7 +32,7 @@ class WorkflowErrorsReporter
   private_class_method def self.request_params(request, druid, process_name, error_message)
     request.headers['content-type'] = "application/xml"
     request.url  "/workflow/dor/objects/druid:#{druid}/workflows/preservationAuditWF/#{process_name}"
-    request.body = "<process name='#{process_name}' status='error' errorMessage='#{error_message}'/>"
+    request.body = "<process name='#{process_name}' status='error' errorMessage='#{CGI.escapeHTML(error_message)}'/>"
   end
 
   private_class_method def self.conn
--- a/spec/services/workflow_errors_reporter_spec.rb
+++ b/spec/services/workflow_errors_reporter_spec.rb
@@ -5,10 +5,8 @@ RSpec.describe WorkflowErrorsReporter do
     'https://sul-lyberservices-test.stanford.edu/workflow/dor/objects/druid:jj925bx9565/workflows/preservationAuditWF/moab-valid'
   end
   let(:headers) { { 'Content-Type' => 'application/xml' } }
-  let(:result) do
-    { 13 => "Invalid moab, validation error...ential version directories." }
-  end
-  let(:body) { "<process name='moab-valid' status='error' errorMessage='{13=>\"Invalid moab, validation error...ential version directories.\"}'/>" }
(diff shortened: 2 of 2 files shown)


app/services/workflow_errors_reporter.rb
```diff
@@ -32,7 +32,7 @@ def self.update_workflow(druid, process_name, error_message)
   private_class_method def self.request_params(request, druid, process_name, error_message)
     request.headers['content-type'] = "application/xml"
     request.url  "/workflow/dor/objects/druid:#{druid}/workflows/preservationAuditWF/#{process_name}"
-    request.body = "<process name='#{process_name}' status='error' errorMessage='#{error_message}'/>"
+    request.body = "<process name='#{process_name}' status='error' errorMessage='#{Rack::Utils.escape_html(error_message)}'/>"
```
ndushay: I checked the SearchWorks code, and they seem to use `CGI.escape_html` which is part of ruby stdlib:

http://ruby-doc.org/stdlib-2.5.0/libdoc/cgi/rdoc/CGI/Util.html#method-i-escape_html

https://github.com/sul-dlss/SearchWorks/blob/master/app/models/concerns/eds_subjects.rb#L43

SaravShah: Cool  - ill make this fix and get the tests working! thank you

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -28,7 +28,7 @@
         .with(body: body,
               headers: headers)
         .to_return(status: 400, body: "", headers: {})
-      expect(Rails.logger).to receive(:warn).with("#{druid} - unable to update workflow for preservationAuditWF moab-valid #<Faraday::ClientError response={:status=>400, :headers=>{}, :body=>\"\"}>. Error message: #{result}")
+      expect(Rails.logger).to receive(:warn).with("jj925bx9565 - unable to update workflow for preservationAuditWF moab-valid #<Faraday::ClientError response={:status=>400, :headers=>{}, :body=>\"\"}>. Error message: Invalid moab, validation error...ential version directories.")
```
ndushay: question:  why did you stop using the vars here?

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -6,9 +6,9 @@
   end
   let(:headers) { { 'Content-Type' => 'application/xml' } }
   let(:result) do
-    { 13 => "Invalid moab, validation error...ential version directories." }
+    "Invalid moab, validation error...ential version directories."
```
ndushay: This can go on one line now, yes?

ndushay:
Another comment:  if you look at your commit comment, and this PR comment, it doesn't have enough context, and it's slightly misleading.  Perhaps "html escape text sent to workflows" would be better understood a week or two, or a year or two from now?

PR #507, opened 2018-01-25
m2c uses matches_po_current_version? to compare po/pc versions

### LINKED ISSUE #483: (C2M) make po/pc version matching method callable from M2C and C2M

tallenaz:
Fixes #483 

paired with @ndushay 

Should need just one set of eyes.

The method to compare a preserved_copy version against its preserved_object current version is now used in m2c as well as c2m.

We also corrected a couple of cases where `receive(:find_by)`should have been `receive(find_by!)`


app/services/preserved_object_handler.rb
```diff
@@ -69,7 +69,7 @@ def check_existence
           # FIXME: what if there is more than one associated pres_copy?
           pres_copy = PreservedCopy.find_by!(preserved_object: pres_object, endpoint: endpoint) if pres_object
 
-          raise_rollback_if_pc_po_version_mismatch(pres_copy.version, pres_object.current_version)
+          raise_rollback_if_pc_po_version_mismatch(pres_copy)
```
SaravShah: ah i like this, now we don't have to pass `pres_obj` around

Issue #478, opened 2018-01-22
(M2C) (C2M) PreservedObjectHandler.update_db_object:  do we still want it? or just .save!


ndushay:
This is critical path if it impacts the writing of C2M;  otherwise it is refactoring.

I think with our new understanding of our timestamps, we now are served by doing a simple `PreservedCopy.save!` or `PreservedObject.save!` which will do the right thing, yes?  We can leave `updated_at` timestamp to its natural behavior?

Thus, the method below becomes superfluous (do we really need to have those particular handler results???).

Note that there are rubocop exceptions tied to this code (for the `touch` method, I believe, if nothing else) that should also go away ...

current code (https://github.com/sul-dlss/preservation_catalog/blob/master/app/services/preserved_object_handler.rb#L372L383)

```ruby
# TODO: this may need reworking if we need to distinguish db timestamp updates when
  #   version matched vs. incoming version less than db object
  def update_db_object(db_object)
    if db_object.changed?
      db_object.save!
      handler_results.add_result(PreservedObjectHandlerResults::UPDATED_DB_OBJECT, db_object.class.name)
    else
      # FIXME: we may not want to do this, but instead to update specific timestamp for check
      db_object.touch
      handler_results.add_result(PreservedObjectHandlerResults::UPDATED_DB_OBJECT_TIMESTAMP_ONLY, db_object.class.name)
    end
  end
```

PR #519, opened 2018-01-26
Refactor to spec for ordering in c2m SQL query

### LINKED ISSUE #476: (C2M) write check_version_on_dir specs to test ordering of sql result pres_copies
Connects to #460

Will become testable with further work on `check_catalog_version`s updating of timestamps

### LINKED ISSUE #537: (C2M) add order clause in find_each loop of SQL query
Copied from #519 

"one slightly annoying thing i did notice when playing around with this stuff in the console:
```ruby
[5] pry(main)> PreservedCopy.by_endpoint_name('fixture_sr1').fixity_check_expired.find_each { |x| p x }
Scoped order is ignored, it's forced to be batch order.
  PreservedCopy Load (1.2ms)  SELECT  "preserved_copies".* FROM "preserved_copies" INNER JOIN "endpoints" ON "endpoints"."id" = "preserved_copies"."endpoint_id" INNER JOIN preserved_objects ON preserved_objects.id = preserved_object_id INNER JOIN preservation_policies ON preservation_policies.id = preserved_objects.preservation_policy_id AND extract(day from (CURRENT_TIMESTAMP - last_checksum_validation)*86400) > fixity_ttl OR last_checksum_validation IS NULL WHERE "endpoints"."endpoint_name" = $1 ORDER BY "preserved_copies"."id" ASC LIMIT $2  [["endpoint_name", "fixture_sr1"], ["LIMIT", 1000]]
#<PreservedCopy id: 1, version: 3, preserved_object_id: 1, endpoint_id: 1, created_at: "2017-12-13 20:19:49", updated_at: "2018-01-29 20:50:44", last_moab_validation: "2017-12-13 20:19:48", last_checksum_validation: nil, size: 2012695, status: "ok", last_version_audit: nil>
#<PreservedCopy id: 2, version: 3, preserved_object_id: 2, endpoint_id: 1, created_at: "2017-12-13 20:19:49", updated_at: "2017-12-13 20:19:49", last_moab_validation: "2017-12-13 20:19:49", last_checksum_validation: nil, size: 277588, status: "ok", last_version_audit: nil>
#<PreservedCopy id: 3, version: 2, preserved_object_id: 3, endpoint_id: 1, created_at: "2017-12-13 20:19:49", updated_at: "2017-12-13 20:19:49", last_moab_validation: "2017-12-13 20:19:49", last_checksum_validation: nil, size: 6570668, status: "ok", last_version_audit: nil>
=> nil
```

which indicates to me that we might want to move the ordering clause to the `find_each`, or have it in both places since some callers of the scope might want it, or something.  would be happy to ticket that consideration as follow on work.  doesn't seem essential to getting this PR merged.

* an ActiveRecord relation represents a query that can be run, but constructing it doesn't automatically run the query -- the query only gets run when the result set needs to be retrieved, e.g. when invoking something like `#find_each`, `#first`, `#all`, or `#count` on the relation)."

tallenaz:
Extracts the SQL query from check_version_on_dir into PreservedCopy
  This made writing the spec more straightforward

  Adds a spec on c2m to test that sql orders objects differently
  depending on last_version_audit timestamp

  To get the spec running we saved the preserved_copy, though
  we haven't agreed on the best way of saving the pc

  The spec seems a bit fragile in the way it uses timestamps in
  method calls

 Paired with @jmartin-sul 

Closes #476 but shouldn't be merged until we get further along with how to save the preserved_copy in `check_catalog_version`, which connects this to #478 

Closes #537

### DIFF
--- a/app/models/preserved_copy.rb
+++ b/app/models/preserved_copy.rb
@@ -31,6 +31,13 @@ class PreservedCopy < ApplicationRecord
   validates :status, inclusion: { in: statuses.keys }
   validates :version, presence: true
 
+  scope :least_recent_version_audit, lambda { |last_checked_b4_date, storage_dir|
+    joins(:endpoint)
+      .where(endpoints: { storage_location: storage_dir })
+      .where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date)
+      .order('last_version_audit IS NOT NULL, last_version_audit ASC')
+  }
+
   def update_audit_timestamps(moab_validated, version_audited)
     t = Time.current
     self.last_moab_validation = t if moab_validated
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -39,3 +39,4 @@ provlog:
   enable: false
 
 workflow_services_url: ''
+c2m_sql_limit: 1000
--- a/lib/audit/catalog_to_moab.rb
+++ b/lib/audit/catalog_to_moab.rb
@@ -5,16 +5,18 @@ require 'profiler.rb'
 class CatalogToMoab
 
   # allows for sharding/parallelization by storage_dir
-  def self.check_version_on_dir(last_checked_b4_date, storage_dir)
+  def self.check_version_on_dir_of_batch(last_checked_b4_date, storage_dir, limit)
     # TODO: ensure last_checked_version_b4_date is in the right format for query - see #485
-    pcs = PreservedCopy
-          .joins(:endpoint)
-          .where(endpoints: { storage_location: storage_dir })
-          .where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date)
-          .order('last_version_audit IS NOT NULL, last_version_audit ASC')
-    pcs.find_each do |pc|
+    pcs = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).limit(limit)
+    pcs.each do |pc|
       c2m = CatalogToMoab.new(pc, storage_dir)
       c2m.check_catalog_version
+    end
+  end
+
+  def self.check_version_on_dir(last_checked_b4_date, storage_dir)
+    unless PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count.zero?
+      check_version_on_dir_of_batch(last_checked_b4_date, storage_dir, Settings.c2m_sql_limit)
     end
   end
 
--- a/spec/lib/audit/catalog_to_moab_class_spec.rb
+++ b/spec/lib/audit/catalog_to_moab_class_spec.rb
@@ -4,27 +4,33 @@ require_relative '../../load_fixtures_helper.rb'
 RSpec.describe CatalogToMoab do
   let(:last_checked_version_b4_date) { (Time.now.utc - 1.day).iso8601 }
   let(:storage_dir) { 'spec/fixtures/storage_root01/moab_storage_trunk' }
+  let(:limit) { Settings.c2m_sql_limit }
 
-  context '.check_version_on_dir' do
+  context '.check_version_on_dir_of_batch' do
     include_context 'fixture moabs in db'
-    let(:subject) { described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir) }
+    let(:subject) { described_class.check_version_on_dir_of_batch(last_checked_version_b4_date, storage_dir, limit) }
 
     it 'creates an instance and calls #check_catalog_version' do
       c2m_mock = instance_double(described_class)
       allow(described_class).to receive(:new).and_return(c2m_mock)
       expect(c2m_mock).to receive(:check_catalog_version).exactly(3).times
-      described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir)
+      described_class.check_version_on_dir_of_batch(last_checked_version_b4_date, storage_dir, limit)
     end
-    it 'will not check a PreservedCopy with a future last_version_audit date' do
-      c2m_mock = instance_double(described_class)
-      allow(described_class).to receive(:new).and_return(c2m_mock)
-      expect(c2m_mock).to receive(:check_catalog_version).exactly(3).times
-      described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir)
-      pc = PreservedCopy.first
-      pc.last_version_audit = (Time.now.utc + 1.day).iso8601
-      pc.save
-      expect(c2m_mock).to receive(:check_catalog_version).exactly(2).times
-      described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir)
+  end
+
+  context ".check_version_on_dir" do
+    include_context 'fixture moabs in db'
+    let(:subject) { described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir) }
+
+    it "calls check_version_on_dir_of_batch when there are objects to audit" do
+      expect(described_class).to receive(:check_version_on_dir_of_batch).exactly(1).times
+      subject
+    end
+
+    it "skips calling check_version_on_dir when there are no objects to audit" do
+      expect(described_class).not_to receive(:check_version_on_dir_of_batch)
+      PreservedCopy.all.update(last_version_audit: (Time.now.utc + 2.days))
+      subject
     end
   end
 
--- a/spec/models/preserved_copy_spec.rb
+++ b/spec/models/preserved_copy_spec.rb
@@ -194,4 +194,46 @@ RSpec.describe PreservedCopy, type: :model do
       expect(preserved_copy.matches_po_current_version?).to be false
     end
   end
+
+  context '.least_recent_version_audit(last_checked_b4_date, storage_dir)' do
+    let!(:newer_timestamp_pc) do
+      PreservedCopy.create!(
+        preserved_object_id: preserved_object.id,
+        endpoint_id: endpoint.id,
+        version: 1,
+        status: status,
+        size: 1,
+        last_version_audit: (Time.now.utc - 1.day)
+      )
+    end
+    let!(:older_timestamp_pc) do
+      PreservedCopy.create!(
+        preserved_object_id: preserved_object.id,
+        endpoint_id: endpoint.id,
+        version: 1,
+        status: status,
+        size: 1,
+        last_version_audit: (Time.now.utc - 2.days)
+      )
+    end
+    let!(:future_timestamp_pc) do
+      PreservedCopy.create!(
+        preserved_object_id: preserved_object.id,
+        endpoint_id: endpoint.id,
+        version: 1,
+        status: status,
+        size: 1,
+        last_version_audit: (Time.now.utc + 1.day)
+      )
+    end
+    let!(:nil_timestamp_pc) { preserved_copy }
+    let!(:pcs_ordered_by_query) { PreservedCopy.least_recent_version_audit(Time.now.utc, endpoint.storage_location) }
+
+    it 'returns PreservedCopies with nils first, then old to new timestamps' do
+      expect(pcs_ordered_by_query).to eq [nil_timestamp_pc, older_timestamp_pc, newer_timestamp_pc]
+    end
+    it 'returns no PreservedCopies with future timestamps' do
+      expect(pcs_ordered_by_query).not_to include future_timestamp_pc
+    end
+  end
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -193,6 +193,7 @@ Style/FormatStringToken:
 Style/GuardClause:
   Exclude:
     - app/services/preserved_object_handler.rb # update_status method easy to read as is
+    - lib/audit/catalog_to_moab.rb # wrapping method on line 17 is more readable
 
 Style/PercentLiteralDelimiters:
   Exclude:


spec/lib/audit/catalog_to_moab_class_spec.rb
```diff
@@ -26,6 +26,25 @@
       expect(c2m_mock).to receive(:check_catalog_version).exactly(2).times
       described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir)
     end
+    it 'checks a PreservedCopy previously audited before one that is not audited' do
```
ndushay: I think this is the opposite of what we want

app/models/preserved_copy.rb
```diff
@@ -31,6 +31,13 @@ class PreservedCopy < ApplicationRecord
   validates :status, inclusion: { in: statuses.keys }
   validates :version, presence: true
 
+  scope :least_recent_version_audit, lambda { |last_checked_b4_date, storage_dir|
+    joins(:endpoint)
+      .where(endpoints: { storage_location: storage_dir })
+      .where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date)
+      .order('last_version_audit IS NOT NULL, last_version_audit ASC')
+  }
```
ndushay: please pardon my ignorance   (I didn't understand this before this refactor, either):

For performance or other reasons, do we need to break up the results returned into "chunks"?

Does this approach already "chunk" results?

If it's straight up Rails-foo, no need for tests of this ... just wondering.

app/models/preserved_copy.rb
```diff
@@ -31,6 +31,13 @@ class PreservedCopy < ApplicationRecord
   validates :status, inclusion: { in: statuses.keys }
   validates :version, presence: true
 
+  scope :least_recent_version_audit, lambda { |last_checked_b4_date, storage_dir|
```
ndushay: also - this scope thing is very cool.  Neat-o!

spec/models/preserved_copy_spec.rb
```diff
@@ -194,4 +194,18 @@
...
+  context '.least_recent_version_audit(last_checked_b4_date, storage_dir)' do
+    let(:db_query) {
+      described_class.least_recent_version_audit(Time.now.utc, 'spec/fixtures/storage_root01/moab_storage_trunk')
+    }
+
+    it 'returns an ActiveRecord::Relation' do
+      expect(db_query).to be_a(ActiveRecord::Relation)
+    end
+
+    it 'returns the preserved_copy in the db' do
```
ndushay: does it return one preserved_copy, or an array of them?

app/models/preserved_copy.rb
jmartin-sul: this doesn't chunk, but that's just because it just constructs the query w/o actually running it.  the query execution is handled by the caller's `.find_each` invocation on this query, which does do the chunking.

app/models/preserved_copy.rb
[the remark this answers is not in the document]
jmartin-sul: 😆 _rimshot_

since it's a lambda, i think you could have it check at the top of the block and throw an exception upon encountering a bad date, since what the scope is returning is the result of the AR query method chain (which it would still return if the formatting guard clause on the block failed to raise an error).

but i have no idea whether that's an idiomatic/common/acceptable way to do that sort of validation.  so i'd be inclined to have the caller do the check.

spec/models/preserved_copy_spec.rb
jmartin-sul: confusingly, i think it's the test name that's wrong here.  the code currently uses this for the order by:
`'last_version_audit IS NOT NULL, last_version_audit ASC'`

what i think that does is sort nulls first, then non-nulls in ascending order.  the logic being:
* the first sort criterion is: 'date not null?'  all rows where the date is null will be false by that criterion, and i believe false results sort above true results (0 < 1).
* the second sort criterion just the date in ascending order, so all non-null dates will be sorted next by the actual date value.

in oracle, you could just do `'last_version_audit ASC NULLS FIRST'`, but when i went looking around on the internet to see if that was standard SQL, i saw stack exchange posts and whatnot jumping through hoops because they said it wasn't.  didn't do any testing of my own to confirm whether they were right.

spec/lib/audit/catalog_to_moab_class_spec.rb
jmartin-sul: confusingly, i think it's the test name that's wrong here.  the code currently uses this for the order by:
`'last_version_audit IS NOT NULL, last_version_audit ASC'`

what i think that does is sort nulls first, then non-nulls in ascending order.  the logic being:
* the first sort criterion is: 'date not null?'  all rows where the date is null will be false by that criterion, and i believe false results sort above true results (0 < 1).
* the second sort criterion just the date in ascending order, so all non-null dates will be sorted next by the actual date value.

in oracle, you could just do `'last_version_audit ASC NULLS FIRST'`, but when i went looking around on the internet to see if that was standard SQL, i saw stack exchange posts and whatnot jumping through hoops because they said it wasn't.  didn't do any testing of my own to confirm whether they were right.

jmartin-sul: just did some manual testing in rails console, the kind of `order by` statement used in this PR indeed sorted nulls first for me.

app/models/preserved_copy.rb
jmartin-sul: ok, just did a bit of manual testing.  at first, i was worried that i was incorrect, because i ran some scopes in the rails console and got results immediately.  but then i noticed the same thing happened when i did a `join` without using a scope, and i know that returns an ActiveRecord relation and not an immediate result*.  so, i think what was happening was that pry was trying to immediately print the result of invoking `inspect` on the scope, and that was causing the query to be run.  for example, when i play around with some scopes i was experimenting with on friday, which are similar to the one in the PR, i got:
```ruby
# prevent pry from calling `inspect` on the return value of the scope, by having the result of the line be something other than the return value of the scope
[68] pry(main)> scope_query = PreservedCopy.by_endpoint_name('fixture_sr1').fixity_check_expired; p 'var is set'
"var is set"
=> "var is set"
...
[81] pry(main)> scope_query.class                                                                               
=> PreservedCopy::ActiveRecord_Relation
[82] pry(main)> 
```

so, i think scope should play just find with `find_each`, and should do the chunking thing correctly.  seems like it should be as composable as using `join` and `where` directly, since the scope returns an AR `Relation` object just like those do.

one slightly annoying thing i did notice when playing around with this stuff in the console:
```ruby
[5] pry(main)> PreservedCopy.by_endpoint_name('fixture_sr1').fixity_check_expired.find_each { |x| p x }
Scoped order is ignored, it's forced to be batch order.
  PreservedCopy Load (1.2ms)  SELECT  "preserved_copies".* FROM "preserved_copies" INNER JOIN "endpoints" ON "endpoints"."id" = "preserved_copies"."endpoint_id" INNER JOIN preserved_objects ON preserved_objects.id = preserved_object_id INNER JOIN preservation_policies ON preservation_policies.id = preserved_objects.preservation_policy_id AND extract(day from (CURRENT_TIMESTAMP - last_checksum_validation)*86400) > fixity_ttl OR last_checksum_validation IS NULL WHERE "endpoints"."endpoint_name" = $1 ORDER BY "preserved_copies"."id" ASC LIMIT $2  [["endpoint_name", "fixture_sr1"], ["LIMIT", 1000]]
#<PreservedCopy id: 1, version: 3, preserved_object_id: 1, endpoint_id: 1, created_at: "2017-12-13 20:19:49", updated_at: "2018-01-29 20:50:44", last_moab_validation: "2017-12-13 20:19:48", last_checksum_validation: nil, size: 2012695, status: "ok", last_version_audit: nil>
#<PreservedCopy id: 2, version: 3, preserved_object_id: 2, endpoint_id: 1, created_at: "2017-12-13 20:19:49", updated_at: "2017-12-13 20:19:49", last_moab_validation: "2017-12-13 20:19:49", last_checksum_validation: nil, size: 277588, status: "ok", last_version_audit: nil>
#<PreservedCopy id: 3, version: 2, preserved_object_id: 3, endpoint_id: 1, created_at: "2017-12-13 20:19:49", updated_at: "2017-12-13 20:19:49", last_moab_validation: "2017-12-13 20:19:49", last_checksum_validation: nil, size: 6570668, status: "ok", last_version_audit: nil>
=> nil
```

which indicates to me that we might want to move the ordering clause to the `find_each`, or have it in both places since some callers of the scope might want it, or something.  would be happy to ticket that consideration as follow on work.  doesn't seem essential to getting this PR merged.

* an ActiveRecord relation represents a query that can be run, but constructing it doesn't automatically run the query -- the query only gets run when the result set needs to be retrieved, e.g. when invoking something like `#find_each, `#first`, `#all`, or `#count` on the relation).

jmartin-sul: the scopes referenced in the above comment can be seen here:
https://github.com/sul-dlss/preservation_catalog/compare/pres-copy-expired-fixity-scope?expand=1

they're pretty similar to the scope in this PR, i just happened to have that branch checked out already (because i was messing around for a bit with scopes we might want at the end of the day friday).

ndushay: cool - thanks for this super helpful detailed info.   I would say, yes please do create the ticket for adding ordering in for each loop.

lib/audit/catalog_to_moab.rb
```diff
@@ -93,7 +89,8 @@ def check_catalog_version
     end
 
     preserved_copy.update_audit_timestamps(ran_moab_validation?, true)
-
+    # This may not be the best way to save the preserved_copy!
+    preserved_copy.save!
```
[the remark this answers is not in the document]
jmartin-sul: i think so?

spec/lib/audit/catalog_to_moab_class_spec.rb
ndushay: again, thanks for the details.  I think we want tests that explicitly show

- null value for `last_version_audit` gets done before a dated value
- dates come in oldest -> newest order
- dates AFTER the date arg are not included

jmartin-sul: based on how completely un-intuitive this construction is, though, a comment might be _in order_ 😆

ndushay: ow ow ow

app/models/preserved_copy.rb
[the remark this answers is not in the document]
ndushay: @tallenaz i would add the suggestion to the "validate date arg" ticket, not include the work here ...

spec/models/preserved_copy_spec.rb
tallenaz: An array: ```#<ActiveRecord::Relation [#<PreservedCopy id: 1, version: 0, preserved_object_id: 1, endpoint_id: 1, created_at: "2018-01-29 22:04:52", updated_at: "2018-01-29 22:04:52", last_moab_validation: nil, last_checksum_validation: nil, size: 1, status: "validity_unknown", last_version_audit: nil>]>``` I can change the description to clear that up, if it's too ambiguous.

tallenaz: Well, it returns one preserved copy as an AR::Relation that you can call `to_a` on to generate an array from, so actually, I'm not sure what to say.

spec/lib/audit/catalog_to_moab_class_spec.rb
tallenaz: oh wow, I got that just now! slow burn!

spec/models/preserved_copy_spec.rb
```diff
@@ -194,4 +194,18 @@
...
+      expect(db_query).to include(preserved_copy)
```
ndushay: What I'm trying to express is ... can/should these tests be more explicit in how you get to preserved copies from ActiveRecord::Relation?

The calling code does:

```ruby
 pcs = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir)
```

so do we want to be explicit that we can view the ActiveRecord relation as an Array of PreservedCopy objects?

Maybe that would go with the preview test (line 204) or in a separate test ... or nowhere because I'm so ignorant about ActiveRecord relations.   (second cousin?  first cousin once removed?  🤔 )

spec/models/preserved_copy_spec.rb
jmartin-sul: yeah, i think technically the scope (like a `join` or a `where` or whatever used directly) returns an ActiveRecord relation.  and i *think* AR relations have some magic to so that they behave somewhat transparently like the object or objects you likely want if you start using that relation directly as if it were the query result itself (i.e. as if the AR relation is actually the AR model instance).  which would mean the result has the class `ActiveRecord::Relation`, but that it might behave like a list, or like a single result, depending on how you interact with it.  ✨

ndushay: 💀 🐔 💀 🐔 💀  <-- voodoo

🎩 🐰 <-- magic

spec/models/preserved_copy_spec.rb
tallenaz: I added one commit that tries to check that dates come in oldest -> newest order, and took it that we already have specs testing that -- null value for last_version_audit gets done before a dated value and that -- dates AFTER the date arg are not included. Since a couple of the tests are similar, I tried to wrap them up in a shared context block, and extract some `let` statements to cut down on the verbosity.

I also switched the preserved_copy spec to use rspec's ActiveRecord::Relation `match_array` matcher (https://relishapp.com/rspec/rspec-rails/docs/matchers/activerecord-relation-match-array), which hopefully addresses the concern about making it clear how you get to preserved copies from an ActiveRecord::Relation.

app/models/preserved_copy.rb
tallenaz: created https://github.com/sul-dlss/preservation_catalog/issues/537

spec/models/preserved_copy_spec.rb
```diff
@@ -194,4 +194,14 @@
...
     end
   end
+
+  context '.least_recent_version_audit(last_checked_b4_date, storage_dir)' do
+    let(:db_query) {
+      described_class.least_recent_version_audit(Time.now.utc, 'spec/fixtures/storage_root01/moab_storage_trunk')
+    }
+
+    it 'returns all PreservedCopy objects, in any order' do
+      expect(db_query).to match_array([preserved_copy])
```
SaravShah: `db_query` returns `#<ActiveRecord::Relation [#<PreservedCopy id: 1, version: 0, preserved_object_id: 1, endpoint_id: 1, created_at: "2018-01-31 00:33:37", updated_at: "2018-01-31 00:33:37", last_moab_validation: nil, last_checksum_validation: nil, size: 1, status: "validity_unknown", last_version_audit: nil>]>` this is only 1 object, so you can't really test to see if the PreservedCopy object is returned in any order...

tallenaz:
@ndushay @SaravShah this should be ready for the next round of review

spec/models/preserved_copy_spec.rb
```diff
@@ -194,4 +194,46 @@
...
+        status: status,
+        size: 1,
+        last_version_audit: (Time.now.utc + 1.day)
+      )
+    end
+    let!(:nil_timestamp_pc) { preserved_copy }
+    let!(:pcs_ordered_by_query) { PreservedCopy.least_recent_version_audit(Time.now.utc, endpoint.storage_location) }
+
+    it 'returns PreservedCopies with nils first, then old to new timestamps' do
+      expect(pcs_ordered_by_query).to eq [nil_timestamp_pc, older_timestamp_pc, newer_timestamp_pc]
```
SaravShah: 👍

config/settings.yml
```diff
@@ -39,3 +39,4 @@ provlog:
   enable: false
 
 workflow_services_url: ''
+c2m_sql_limit: 1000
```
ndushay: are we ok with this simply being set here, or should it also go in shared_configs?   (If you already have PRs for shared_configs, you need to paste the urls in the channel for us to see them.)

lib/audit/catalog_to_moab.rb
```diff
@@ -5,19 +5,21 @@
...
-    pcs.find_each do |pc|
+    pcs = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).limit(limit)
+    pcs.each do |pc|
       c2m = CatalogToMoab.new(pc, storage_dir)
       c2m.check_catalog_version
     end
   end
 
+  def self.check_version_on_dir(last_checked_b4_date, storage_dir)
+    unless PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count.zero?
```
ndushay: question:  is there a way to avoid doing this query twice per batch (once here in `check_version_on_dir` and once in `check_version_on_dir_of_batch`) ?   I'm fine with it being a second ticket.

spec/lib/audit/catalog_to_moab_class_spec.rb
```diff
@@ -4,27 +4,33 @@
...
+  end
+
+  context ".check_version_on_dir" do
+    include_context 'fixture moabs in db'
+    let(:subject) { described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir) }
+
+    it "calls check_version_on_dir_of_batch when there are objects to audit" do
+      expect(described_class).to receive(:check_version_on_dir_of_batch).exactly(1).times
+      subject
+    end
```
ndushay: for through-ness, it might be nice to prove that it works for multiple batches.  Maybe by using dir with 3 objects, and setting limit to 2 ?

config/settings.yml
tallenaz: It might be good to have them also in shared_configs, if we ever want to alter the limit between deployed environments, for example. I'll set that up.

spec/lib/audit/catalog_to_moab_class_spec.rb
tallenaz: Agreed.

lib/audit/catalog_to_moab.rb
jmartin-sul: yeah, we could probably do that without too much trouble.  though it's also worth noting that a count query like this should usually be much cheaper to run than the same query actually fetching the records`*`.  or so DB devs better than me have told me in the past.  i think the query optimizer has tricks for shortcutting unnecessary computations/retrievals, plus it's just an integer value getting sent back from the DB instead of 1000 rows, plus those 1000 rows don't have to get turned into 1000 ActiveRecord objects in memory.

`*` and so if you really only wanted the count, doing this with `#size` instead of `#count` wouldn't be advisable because you'd be getting the records and checking array length after a bunch of AR instantiations instead of doing a count query in PG.

jmartin-sul: you could also just do it without getting the count, i think?  if there aren't any rows that satisfy the query, the array is empty and nothing happens.

i think the way to use count would be to loop until there's nothing left that's older than the cut-off.  something like:
```ruby
while PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count > 0
  check_version_on_dir_of_batch(last_checked_b4_date, storage_dir, Settings.c2m_sql_limit)
end
```

alternatively, i just found out ruby has an `until`, so you could also do `until PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count.zero?`.

tallenaz:
Per discussion at this morning's standup, the additional changes documented https://github.com/sul-dlss/preservation_catalog/pull/519#pullrequestreview-93136679 aren't required for this PR, but are ticketed in #549

PR #525, opened 2018-01-26
Use .save! instead of custom method to save db objects

### LINKED ISSUE #478: (M2C) (C2M) PreservedObjectHandler.update_db_object:  do we still want it? or just .save!

tallenaz:
Closes #478 

Could probably use more than 1 set of eyes, since this touches a lot of code.

### DIFF
--- a/app/services/preserved_object_handler_results.rb
+++ b/app/services/preserved_object_handler_results.rb
@@ -10,25 +10,21 @@ class PreservedObjectHandlerResults
   VERSION_MATCHES = 2
   ARG_VERSION_GREATER_THAN_DB_OBJECT = 3
   ARG_VERSION_LESS_THAN_DB_OBJECT = 4
-  UPDATED_DB_OBJECT = 5
-  UPDATED_DB_OBJECT_TIMESTAMP_ONLY = 6
-  CREATED_NEW_OBJECT = 7
-  DB_UPDATE_FAILED = 8
-  OBJECT_ALREADY_EXISTS = 9
-  OBJECT_DOES_NOT_EXIST = 10
-  PC_STATUS_CHANGED = 11
-  UNEXPECTED_VERSION = 12
-  INVALID_MOAB = 13
-  PC_PO_VERSION_MISMATCH = 14
-  ONLINE_MOAB_DOES_NOT_EXIST = 15
+  CREATED_NEW_OBJECT = 5
+  DB_UPDATE_FAILED = 6
+  OBJECT_ALREADY_EXISTS = 7
+  OBJECT_DOES_NOT_EXIST = 8
+  PC_STATUS_CHANGED = 9
+  UNEXPECTED_VERSION = 10
+  INVALID_MOAB = 11
+  PC_PO_VERSION_MISMATCH = 12
+  ONLINE_MOAB_DOES_NOT_EXIST = 13
 
   RESPONSE_CODE_TO_MESSAGES = {
     INVALID_ARGUMENTS => "encountered validation error(s): %{addl}",
     VERSION_MATCHES => "incoming version (%{incoming_version}) matches %{addl} db version",
     ARG_VERSION_GREATER_THAN_DB_OBJECT => "incoming version (%{incoming_version}) greater than %{addl} db version",
     ARG_VERSION_LESS_THAN_DB_OBJECT => "incoming version (%{incoming_version}) less than %{addl} db version; ERROR!",
-    UPDATED_DB_OBJECT => "%{addl} db object updated",
-    UPDATED_DB_OBJECT_TIMESTAMP_ONLY => "%{addl} updated db timestamp only",
     CREATED_NEW_OBJECT => "added object to db as it did not exist",
     DB_UPDATE_FAILED => "db update failed: %{addl}",
     OBJECT_ALREADY_EXISTS => "%{addl} db object already exists",
@@ -51,8 +47,6 @@ class PreservedObjectHandlerResults
   ].freeze
 
   DB_UPDATED_CODES = [
-    UPDATED_DB_OBJECT,
-    UPDATED_DB_OBJECT_TIMESTAMP_ONLY,
     CREATED_NEW_OBJECT,
     PC_STATUS_CHANGED
   ].freeze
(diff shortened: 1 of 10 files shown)


spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -470,10 +450,10 @@
         allow(pc).to receive(:version).and_return(1)
         allow(pc).to receive(:update_audit_timestamps)
         allow(pc).to receive(:changed?).and_return(false)
-        allow(pc).to receive(:touch)
+        allow(pc).to receive(:save!)
```
jmartin-sul: looks like the test needs a rename now too

spec/services/preserved_object_handler_confirm_version_spec.rb
```diff
@@ -307,11 +295,11 @@
         allow(pc).to receive(:version).and_return(1)
         allow(pc).to receive(:update_audit_timestamps)
         allow(pc).to receive(:changed?).and_return(false)
-        allow(pc).to receive(:touch)
+        allow(pc).to receive(:save!)
```
jmartin-sul: ditto this test name

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -457,23 +437,21 @@
         end
       end
 
-      it 'calls PreservedCopy.touch (but not PreservedObject.touch) if the existing record is NOT altered' do
+      it 'calls PreservedCopy.save! if the existing record is NOT altered' do
```
ndushay: how about "calls PreservedCopy.save! (but note PreservedObject.save!) if the existing version is NOT altered" ?

The original test shows that PreservedCopy timestamps are updated, but PreservedObject was unchanged.

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -457,23 +437,21 @@
...
         po_handler = described_class.new(druid, 1, 1, ep)
         po = instance_double(PreservedObject)
         pc = instance_double(PreservedCopy)
         allow(PreservedObject).to receive(:find_by!).with(druid: druid).and_return(po)
         allow(po).to receive(:current_version).and_return(1)
-        allow(po).to receive(:touch)
```
ndushay: which would change this line to be included, but :save! instead of :touch

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -457,23 +437,21 @@
...
         allow(PreservedCopy).to receive(:find_by!).with(preserved_object: po, endpoint: ep).and_return(pc)
         allow(pc).to receive(:matches_po_current_version?).and_return(true)
         allow(pc).to receive(:status).and_return(PreservedCopy::OK_STATUS)
         allow(pc).to receive(:version).and_return(1)
         allow(pc).to receive(:update_audit_timestamps)
         allow(pc).to receive(:changed?).and_return(false)
-        allow(pc).to receive(:touch)
+        allow(pc).to receive(:save!)
         po_handler.check_existence
-        expect(po).not_to have_received(:touch)
```
ndushay: and would change this line to be included, but :save! instead of :touch

spec/services/preserved_object_handler_confirm_version_spec.rb
```diff
@@ -295,23 +283,21 @@
         expect(po).not_to have_received(:save!)
         expect(pc).to have_received(:save!)
       end
-      it 'calls PreservedCopy.touch (but not PreservedObject.touch) if the existing record is NOT altered' do
+      it 'calls PreservedCopy.save! if the existing record is NOT altered' do
```
ndushay: save as in check_existence spec:  this test shows that PreservedCopy timestamps were updated for the check, but PreservedObject remained unchanged.

spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -249,26 +230,6 @@
         expect(pc).to have_received(:save!)
       end
 
-      it 'does not call PreservedObject.touch when PreservedCopy only has timestamp updates' do
```
ndushay: again save as in check_existence spec:  this test shows that PreservedCopy timestamps were updated for the check, but PreservedObject remained unchanged.  So test should stay, but be modified

spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -458,27 +419,6 @@
           expect(Rails.logger).to have_received(:debug).with(msg)
         end
 
-        it 'does not call PreservedObject.touch when PreservedCopy only has timestamp updates' do
```
ndushay: ditto for this test

app/services/preserved_object_handler_results.rb
```diff
@@ -10,25 +10,21 @@ class PreservedObjectHandlerResults
...
+  PC_PO_VERSION_MISMATCH = 12
+  ONLINE_MOAB_DOES_NOT_EXIST = 13
 
   RESPONSE_CODE_TO_MESSAGES = {
     INVALID_ARGUMENTS => "encountered validation error(s): %{addl}",
     VERSION_MATCHES => "incoming version (%{incoming_version}) matches %{addl} db version",
     ARG_VERSION_GREATER_THAN_DB_OBJECT => "incoming version (%{incoming_version}) greater than %{addl} db version",
     ARG_VERSION_LESS_THAN_DB_OBJECT => "incoming version (%{incoming_version}) less than %{addl} db version; ERROR!",
-    UPDATED_DB_OBJECT => "%{addl} db object updated",
-    UPDATED_DB_OBJECT_TIMESTAMP_ONLY => "%{addl} updated db timestamp only",
```
ndushay: heh - funny how we're all comfortable calling this "noise" now.  Yay us!

Issue #372, opened 2017-12-05
PreservedCopy has duplicate statuses for checksum validation failure


jmartin-sul:
PreservedCopy status enum needs some love:

1.  'expected_vers_not_found_on_storage' is too long.  How about:
  `unexpected_version_in_catalog` ?   or `unexpected_version_on_storage` ?

2.  we have:

- INVALID_MOAB_STATUS = 'invalid_moab'.freeze  # <-- moab validation failed
- INVALID_CHECKSUM_STATUS = 'invalid_checksum'.freeze  # <-- checksum validation failed
- FIXITY_CHECK_FAILED_STATUS = 'fixity_check_failed'.freeze # <-- duplicate of the above?  before we wanted to get more specific?

Note that these changes could imply a migration is needed.

Note that a big migration of the data itself might be not worth it and we should just wipe, re-seed, and move on.

ndushay:
I prefer `INVALID_CHECKSUM_STATUS`

jmartin-sul:
* i'd agree about dropping `FIXITY_CHECK_FAILED_STATUS` as a dupe of `INVALID_CHECKSUM_STATUS`
* no objection to renaming `expected_vers_not_found_on_storage` to `unexpected_version_on_storage`.  not that much shorter but i think it reads a little better while communicating the same thing.
* > Note that a big migration of the data itself might be not worth it and we should just wipe, re-seed, and move on.
  * i think the only thing we'd have to migrate is `expected_vers_not_found_on_storage`, which we might not even have any entries for?  but it'd also be one of the simplest types of data migrations you can do, and so maybe good practice/exposure for that part of active record (but also very much *not* critical path at the moment).

jmartin-sul:
currently, nothing on prod has `EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS`:
```ruby
[12] pry(main)> PreservedCopy.joins(:preserved_object, :endpoint).where(status: PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS).count
=> 0
[15] pry(main)> PreservedCopy.joins(:preserved_object, :endpoint).where(status: PreservedCopy::OK_STATUS).count
=> 1192819
```

so, i think no migration needed.

[...]

PR #526, opened 2018-01-27
C2M#check_catalog_version: handle pres copy status status already bad

### LINKED ISSUE #480: (C2M) do something useful if pres_copy.status is not OK
probably analogous to what we do for M2C (there was a PR for this that @jmartin-sul spearheaded.)

jmartin-sul:
since this adds its own `preserved_copy#save!` call, which also happens in #519 and #525, so those should probably get merged first, and then this should get rebased.

assumed similar behavior to `PreservedObjectHandler#check_existence`.

paired w/ @SaravShah

closes #480

### DIFF
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -271,7 +271,7 @@ class PreservedObjectHandler
   # other validations of what's on disk, and will update the status accordingly
   # TODO: near duplicate of method in CatalogToMoab - extract superclass or moab wrapper class??
   def set_status_as_seen_on_disk(pres_copy, found_expected_version)
-    unless moab_validation_errors.empty?
+    if moab_validation_errors.any?
       update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
       return
     end
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -269,6 +269,7 @@ class PreservedObjectHandler
 
   # given a PreservedCopy instance and whether the caller found the expected version of it on disk, this will perform
   # other validations of what's on disk, and will update the status accordingly
+  # TODO: near duplicate of method in CatalogToMoab - extract superclass or moab wrapper class??
   def set_status_as_seen_on_disk(pres_copy, found_expected_version)
     unless moab_validation_errors.empty?
       update_status(pres_copy, PreservedCopy::INVALID_MOAB_STATUS)
--- a/lib/audit/catalog_to_moab.rb
+++ b/lib/audit/catalog_to_moab.rb
@@ -142,7 +142,7 @@ class CatalogToMoab
   # other validations of what's on disk, and will update the status accordingly.
   # TODO: near duplicate of method in POHandler - extract superclass or moab wrapper class??
   def set_status_as_seen_on_disk(found_expected_version)
-    unless moab_validation_errors.empty?
+    if moab_validation_errors.any?
       update_status(PreservedCopy::INVALID_MOAB_STATUS)
       return
     end
--- a/lib/audit/catalog_to_moab.rb
+++ b/lib/audit/catalog_to_moab.rb
@@ -63,11 +63,11 @@ class CatalogToMoab
       return
     end
 
-    # TODO: anything special if preserved_copy.status is not OK_STATUS? - see #480
-
     unless online_moab_found?(druid, storage_dir)
+      update_status(PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS)
       results.add_result(AuditResults::ONLINE_MOAB_DOES_NOT_EXIST)
       results.report_results
+      preserved_copy.save!
       return
     end
 
@@ -75,20 +75,18 @@ class CatalogToMoab
     results.actual_version = moab_version
     catalog_version = preserved_copy.version
     if catalog_version == moab_version
+      set_status_as_seen_on_disk(true) unless preserved_copy.status == PreservedCopy::OK_STATUS
       results.add_result(AuditResults::VERSION_MATCHES, preserved_copy.class.name)
       results.report_results
     elsif catalog_version < moab_version
+      set_status_as_seen_on_disk(true)
       results.add_result(AuditResults::UNEXPECTED_VERSION, preserved_copy.class.name)
       # TODO: avoid repetitious results ... (leave out line above??) - see #484
       pohandler = PreservedObjectHandler.new(druid, moab_version, moab.size, preserved_copy.endpoint)
       pohandler.update_version_after_validation # results reported by this call
     else # catalog_version > moab_version
+      set_status_as_seen_on_disk(false)
       results.add_result(AuditResults::UNEXPECTED_VERSION, preserved_copy.class.name)
-      if moab_validation_errors.empty?
-        update_status(PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS)
-      else
-        update_status(PreservedCopy::INVALID_MOAB_STATUS)
-      end
       results.report_results
     end
 
@@ -140,4 +138,24 @@ class CatalogToMoab
     false
   end
 
+  # given whether the caller found the expected version of preserved_copy on disk, this will perform
+  # other validations of what's on disk, and will update the status accordingly.
+  # TODO: near duplicate of method in POHandler - extract superclass or moab wrapper class??
+  def set_status_as_seen_on_disk(found_expected_version)
+    unless moab_validation_errors.empty?
+      update_status(PreservedCopy::INVALID_MOAB_STATUS)
+      return
+    end
+
+    unless found_expected_version
+      update_status(PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS)
+      return
+    end
+
+    # TODO: do the check that'd set INVALID_CHECKSUM_STATUS
+
+    # TODO: do the check that'd set FIXITY_CHECK_FAILED_STATUS
+
+    update_status(PreservedCopy::OK_STATUS)
+  end
 end
--- a/spec/lib/audit/catalog_to_moab_instance_spec.rb
+++ b/spec/lib/audit/catalog_to_moab_instance_spec.rb
@@ -78,7 +78,30 @@ RSpec.describe CatalogToMoab do
         expect(pohandler_results).to receive(:add_result).with(
           AuditResults::ONLINE_MOAB_DOES_NOT_EXIST
         )
+        expect(pohandler_results).to receive(:add_result).with(
+          AuditResults::PC_STATUS_CHANGED, old_status: "ok", new_status: "online_moab_not_found"
+        )
         c2m.check_catalog_version
+      end
+      context 'updates status correctly' do
+        before do
+          allow(Moab::StorageObject).to receive(:new).with(druid, instance_of(String)).and_return(nil)
+        end
+
+        [
+          PreservedCopy::VALIDITY_UNKNOWN_STATUS,
+          PreservedCopy::OK_STATUS,
+          PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS,
+          PreservedCopy::INVALID_MOAB_STATUS,
+          PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS
+        ].each do |orig_status|
+          it "had #{orig_status}, should now have ONLINE_MOAB_NOT_FOUND_STATUS" do
+            pres_copy.status = orig_status
+            pres_copy.save!
+            c2m.check_catalog_version
+            expect(pres_copy.reload.status).to eq PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS
+          end
+        end
       end
       it 'stops processing .check_catalog_version' do
         moab = instance_double(Moab::StorageObject)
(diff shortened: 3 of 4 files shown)


lib/audit/catalog_to_moab.rb
```diff
@@ -66,32 +66,33 @@ def check_catalog_version
...
     end
 
     moab_version = moab.current_version_id
     catalog_version = preserved_copy.version
     if catalog_version == moab_version
+      set_status_as_seen_on_disk(true) unless preserved_copy.status == PreservedCopy::OK_STATUS
       results.add_result(PreservedObjectHandlerResults::VERSION_MATCHES, preserved_copy.class.name)
       results.report_results
     elsif catalog_version < moab_version
+      set_status_as_seen_on_disk(true) unless preserved_copy.status == PreservedCopy::OK_STATUS
```
jmartin-sul: should we always just set status in this case, regardless of current catalog status?  see skipped test...

spec/lib/audit/catalog_to_moab_instance_spec.rb
```diff
@@ -133,6 +223,56 @@
...
+      context 'check whether PreservedCopy already has a status other than OK_STATUS, re-check status if so' do
+        it 'had OK_STATUS, version increased, should still have OK_STATUS' do
+          pres_copy.status = PreservedCopy::OK_STATUS
+          pres_copy.save!
+          allow(c2m).to receive(:moab_validation_errors).and_return([])
+          c2m.check_catalog_version
+          expect(pres_copy.reload.status).to eq PreservedCopy::OK_STATUS
+        end
+        it 'had OK_STATUS, but is now INVALID_MOAB_STATUS' do
+          skip "validate if version increases on disk? as in POH#check_existence, checks status iff pres copy not OK."
```
jmartin-sul: in `PreservedObjectHandler#check_existence`, we validate explicitly when the incoming version is greater than the catalog version.  that was already happening in that branch, and so `set_status_as_seen_on_disk` was only called if the object already had a non-OK status.  i'm inclined to say we should just call `set_status_as_seen_on_disk` regardless of the current catalog status when incoming version is greater than catalog version.  this seems simple and reasonable, and like it accomplishes similar behavior to the M2C counterpart with less code than what's in this PR now.

lib/audit/catalog_to_moab.rb
```diff
@@ -66,32 +66,33 @@ def check_catalog_version
     # TODO: anything special if preserved_copy.status is not OK_STATUS? - see #480
 
     unless online_moab_found?(druid, storage_dir)
+      update_status(PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS)
       results.add_result(PreservedObjectHandlerResults::ONLINE_MOAB_DOES_NOT_EXIST)
```
tallenaz: When I saw the response code and status together like this, I took a look at codes and statuses in general, and see that we have some codes and statuses that use `DOES_NOT_EXIST`, and some `NOT_FOUND`. My question is whether these are synonymous, and if so, do we want to decide on a unified convention? If so, I'd be for `NOT FOUND`.

lib/audit/catalog_to_moab.rb
```diff
@@ -66,32 +66,33 @@ def check_catalog_version
...
       results.report_results
+      preserved_copy.save!
       return
```
tallenaz: Do we want an inline `if` for these saves? I remember this topic coming up in standup this morning...

lib/audit/catalog_to_moab.rb
```diff
@@ -142,4 +143,23 @@ def online_moab_found?(druid, storage_dir)
...
+    unless moab_validation_errors.empty?
+      update_status(PreservedCopy::INVALID_MOAB_STATUS)
+      return
+    end
+
+    unless found_expected_version
+      update_status(PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS)
+      return
+    end
+
```
tallenaz: Just noticing this common pattern which might or might not suggest refactoring : we have several places where we `do x and return unless something`

lib/audit/catalog_to_moab.rb
```diff
@@ -142,4 +143,23 @@ def online_moab_found?(druid, storage_dir)
     false
   end
 
+  # given whether the caller found the expected version of preserved_copy on disk, this will perform
+  # other validations of what's on disk, and will update the status accordingly.
```
ndushay: since this method essentially duplicates a method in pohandler results, please add the comment implying that?  (see other private methods)

lib/audit/catalog_to_moab.rb
```diff
@@ -142,4 +143,23 @@ def online_moab_found?(druid, storage_dir)
...
+  def set_status_as_seen_on_disk(found_expected_version)
+    unless moab_validation_errors.empty?
```
ndushay: per tony's comment below, what if we did something like

```ruby
if moab_validation_errors.any?
  update_status(...)
  return
end
```

Do we prefer this?

If we like this, it seems to occur only in the analogous place in POHandler, so add a commit here changing that one too?  I would find it easier than "unless" statements ...

lib/audit/catalog_to_moab.rb
```diff
@@ -142,4 +143,23 @@ def online_moab_found?(druid, storage_dir)
...
+      update_status(PreservedCopy::INVALID_MOAB_STATUS)
+      return
+    end
+
+    unless found_expected_version
```
ndushay: this is a little less cognitive load ... but for some reason `unless` makes my face scrunch up.  Yet it is awesomely useful.

spec/lib/audit/catalog_to_moab_instance_spec.rb
```diff
@@ -75,8 +77,47 @@
...
         )
+        expect(pohandler_results).to receive(:add_result).with(
+          PreservedObjectHandlerResults::PC_STATUS_CHANGED, old_status: "ok", new_status: "online_moab_not_found"
+        )
         c2m.check_catalog_version
       end
+      context 'updates status correctly' do
+        before do
+          allow(Moab::StorageObject).to receive(:new).with(druid, instance_of(String)).and_return(nil)
+          # allow(pohandler_results).to receive(:add_result)
```
ndushay: this comment is probably no longer needed?

spec/lib/audit/catalog_to_moab_instance_spec.rb
```diff
@@ -75,8 +77,47 @@
...
+        end
+        it 'had OK_STATUS, but is now ONLINE_MOAB_NOT_FOUND_STATUS' do
```
ndushay: what do you think about

```ruby
[
  PreservedCopy::OK_STATUS,
  ...
  PreservedCopy::VALIDITY_UNKNOWN_STATUS
].each do |orig_status
  it "changes status to ONLINE_MOAB_NOT_FOUND_STATUS from orig status: #{orig_status}" do 
    pres_copy.status = orig_status
    pres_copy.save!
    c2m.check_catalog_version
    expect(pres_copy.reload.status).to eq PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS
end
```

spec/lib/audit/catalog_to_moab_instance_spec.rb
```diff
@@ -109,6 +150,55 @@
         )
         c2m.check_catalog_version
       end
+
+      context 'check whether PreservedCopy already has a status other than OK_STATUS, re-check status if so' do
```
ndushay: similar to above comment  about condensing the written code, but not what is actually checked ...

spec/lib/audit/catalog_to_moab_instance_spec.rb
ndushay: I'm okay with your suggestion;  it's really up to @julianmorley .   I'd suggest coding what you think is sensible here, following up with Julian, and make a separate ticket if he wants something different?  (Or just get an answer from him RIGHT NOW)

spec/lib/audit/catalog_to_moab_instance_spec.rb
```diff
@@ -133,6 +223,56 @@
         expect(pohandler).to receive(:update_version_after_validation)
         c2m.check_catalog_version
       end
+
+      context 'check whether PreservedCopy already has a status other than OK_STATUS, re-check status if so' do
+        it 'had OK_STATUS, version increased, should still have OK_STATUS' do
```
ndushay: this is another place to consider condensing code to read without reducing the tests executed?

spec/lib/audit/catalog_to_moab_instance_spec.rb
```diff
@@ -199,6 +339,55 @@
         allow(PreservedObjectHandlerResults).to receive(:new).and_return(pohandler_results)
         c2m.check_catalog_version
       end
+
+      context 'check whether PreservedCopy already has a status other than OK_STATUS, re-check status if so' do
```
ndushay: same again?   shared example??  something.  It's tricky looking at all these and figuring out which cases group together and which don't, and so on.

app/services/preserved_object_handler.rb
```diff
@@ -270,9 +270,6 @@ def update_pc_invalid_moab
   # given a PreservedCopy instance and whether the caller found the expected version of it on disk, this will perform
   # other validations of what's on disk, and will update the status accordingly
   def set_status_as_seen_on_disk(pres_copy, found_expected_version)
-    # TODO: when we implement C2M, we should add a check here to look for the object on disk, and set PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS
```
ndushay: yay!  removal of a TODO with no action required!

lib/audit/catalog_to_moab.rb
ndushay: See below - basically:  I'm ok with whatever;  it's kinda up to @julianmorley, no?

lib/audit/catalog_to_moab.rb
ndushay: good observation, @tallenaz .  I feel it is a bit too early to refactor;  I also figure it's a common thing more broadly than this code base (like, out in ruby land)

lib/audit/catalog_to_moab.rb
jmartin-sul: > I think .save! will only do its thing if .changed? returns true. I just experimented in irb and that's what I got.

yup.  as i recall, that was actually why we checked the `changed?` flag, so we could `touch` if the object hadn't actually changed (thus forcing `updated_at` to get updated).

spec/lib/audit/catalog_to_moab_instance_spec.rb
jmartin-sul: oh sure, yeah, i like this approach, i think it's a fine style to use when testing lists of things that all have the same expected behavior.

lib/audit/catalog_to_moab.rb
jmartin-sul: i don't have strong feelings about any of the above suggestions, would be fine with more consistency between names (and/or shorter names as per #372).

i agree with @tallenaz's point that using both "not found" and "does not exist" is a bit confusing, since we mean the same thing with both of those terms.  but i also agree with @ndushay's overarching goals (PreservedCopy model enum makes sense in its own context, devs of the future can easily see what's going on).

would also agree with @ndushay that this is outside the scope of this ticket, and that any follow on work along these lines could be done under #372 (or some other result class refactoring ticket).

lib/audit/catalog_to_moab.rb
jmartin-sul: discussed in story time, @julianmorley says fine to just set the status regardless when C2M finds that the version on disk has increased compared to the catalog.

lib/audit/catalog_to_moab.rb
jmartin-sul: i prefer it as is, but i was not initially a fan of `unless`, so i get the aversion.  my preference here is because in both the usages that get changed, there's a neighboring `unless` statement.  having both be `unless` feels more immediately readable to me.

but i don't feel that strongly about it, so i pushed a commit with the suggested change.  i'd prefer to remove it before merging, but if everyone prefers this style, feel free to merge (assuming no new complaints otherwise).

jmartin-sul:
i think all the suggestions have been implemented, should be ready for review again (@ndushay)

app/services/preserved_object_handler.rb
ndushay: uh oh - I forgot you did this here -- if #544 gets merged first there will be a conflict here.  Sorry!

app/services/preserved_object_handler.rb
```diff
@@ -270,10 +270,7 @@ def update_pc_invalid_moab
...
-    # if needed.  or maybe the caller will have done that already, since it needs to try reading the moab to obtain incoming_version?
-
-    unless moab_validation_errors.empty?
+    if moab_validation_errors.any?
```
ndushay: i do like ```if xxx.any?``` better than ```unless xxx.empy?```

lib/audit/catalog_to_moab.rb
```diff
@@ -66,28 +66,28 @@ def check_catalog_version
     # TODO: anything special if preserved_copy.status is not OK_STATUS? - see #480
```
ndushay: is this not the TODO message that needs to go away with this PR?

spec/lib/audit/catalog_to_moab_instance_spec.rb
```diff
@@ -76,8 +78,31 @@
...
+        end
+
+        [
+          PreservedCopy::VALIDITY_UNKNOWN_STATUS,
+          PreservedCopy::OK_STATUS,
+          PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS,
+          PreservedCopy::INVALID_MOAB_STATUS,
+          PreservedCopy::EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS
+        ].each do |orig_status|
+          it "had #{orig_status}, and still has ONLINE_MOAB_NOT_FOUND_STATUS" do
```
ndushay: "still has" ?   "ends with"?

lib/audit/catalog_to_moab.rb
```diff
@@ -139,4 +139,24 @@ def online_moab_found?(druid, storage_dir)
     false
   end
 
+  # given whether the caller found the expected version of preserved_copy on disk, this will perform
+  # other validations of what's on disk, and will update the status accordingly.
+  # TODO: near duplicate of method in POHandler - extract superclass or moab wrapper class??
```
ndushay: dang it!  OC me also puts a similar message in POHandler by the corresponding method ...

app/services/preserved_object_handler.rb
jmartin-sul: git dealt with this with no human intervention (did the whole rebase of this branch with no human intervention after #544 was merged to master).  not that it'd have been a terrible conflict anyway.

jmartin-sul:
new feedback incorporated, ready for re-review (@ndushay)

lib/audit/catalog_to_moab.rb
ndushay: @tallenaz if you feel strongly enough about the naming mismatch, please make a ticket.  I'm okay with it dying off without further discussion, so it's up to you if you want to champion it.

tallenaz: I agree the model object's status enum makes sense in its own context, and I'm happy for the devs of the future and their low cognitive overload, so for me this can fade away into the depths of github issue history.

PR #534, opened 2018-01-30
(CV) PreservedCopy scopes: fixity_check_expired, and related minor refactoring

### LINKED ISSUE #515: (CV) SQL query to select objects to be checksum validated
Analogous to C2M, we need a SQL query to get the objects we want, performantly:

- last_checked_b4_date
- endpoint

Note that if preservation policy TTL comes into play, this will likely affect the SQL queries.

jmartin-sul:
~~on hold for a bit since we aren't working on checksum validation quite yet (i just happened to be playing with scopes and this query at the end of the day last friday).  just making a PR so we can easily link the branch to the issue it addresses.~~

adds `fixity_check_expired`, and a couple utility scopes for filtering pres copies by endpoint.  also refactors an existing scope to use the new endpoint filter scope (`by_storage_location`) instead of taking a storage location param itself.

closes #515

### DIFF
--- a/app/models/preserved_copy.rb
+++ b/app/models/preserved_copy.rb
@@ -38,6 +38,21 @@ class PreservedCopy < ApplicationRecord
       .where(endpoints: { storage_location: storage_dir })
       .where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date)
       .order('last_version_audit IS NOT NULL, last_version_audit ASC')
+    # possibly counter-intuitive: the .order sorts so that null values come first (because IS NOT NULL evaluates
+    # to 0 for nulls, which sorts before 1 for non-nulls, which are then sorted by last_version_audit)
+  }
+
+  scope :fixity_check_expired, lambda {
+    joins(:preserved_object)
+      .joins(
+        'INNER JOIN preservation_policies'\
+        ' ON preservation_policies.id = preserved_objects.preservation_policy_id'\
+        ' AND (last_checksum_validation + (fixity_ttl * INTERVAL \'1 SECOND\')) < CURRENT_TIMESTAMP'\
+        ' OR last_checksum_validation IS NULL'
+      )
+      .order('last_checksum_validation IS NOT NULL, last_checksum_validation ASC')
+    # possibly counter-intuitive: the .order sorts so that null values come first (because IS NOT NULL evaluates
+    # to 0 for nulls, which sorts before 1 for non-nulls, which are then sorted by last_version_audit)
   }
 
   def update_audit_timestamps(moab_validated, version_audited)
--- a/app/models/preserved_copy.rb
+++ b/app/models/preserved_copy.rb
@@ -32,6 +32,10 @@ class PreservedCopy < ApplicationRecord
   validates :status, inclusion: { in: statuses.keys }
   validates :version, presence: true
 
+  scope :by_storage_location, lambda { |storage_dir|
+    joins(:endpoint).where(endpoints: { storage_location: storage_dir })
+  }
+
   scope :least_recent_version_audit, lambda { |last_checked_b4_date, storage_dir|
     last_checked_b4_date = normalize_date(last_checked_b4_date)
     joins(:endpoint)
--- a/app/models/preserved_copy.rb
+++ b/app/models/preserved_copy.rb
@@ -36,11 +36,9 @@ class PreservedCopy < ApplicationRecord
     joins(:endpoint).where(endpoints: { storage_location: storage_dir })
   }
 
-  scope :least_recent_version_audit, lambda { |last_checked_b4_date, storage_dir|
+  scope :least_recent_version_audit, lambda { |last_checked_b4_date|
     last_checked_b4_date = normalize_date(last_checked_b4_date)
-    joins(:endpoint)
-      .where(endpoints: { storage_location: storage_dir })
-      .where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date)
+    where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date)
       .order('last_version_audit IS NOT NULL, last_version_audit ASC')
     # possibly counter-intuitive: the .order sorts so that null values come first (because IS NOT NULL evaluates
     # to 0 for nulls, which sorts before 1 for non-nulls, which are then sorted by last_version_audit)
--- a/app/models/preserved_copy.rb
+++ b/app/models/preserved_copy.rb
@@ -32,6 +32,10 @@ class PreservedCopy < ApplicationRecord
   validates :status, inclusion: { in: statuses.keys }
   validates :version, presence: true
 
+  scope :by_endpoint_name, lambda { |endpoint_name|
+    joins(:endpoint).where(endpoints: { endpoint_name: endpoint_name })
+  }
+
   scope :by_storage_location, lambda { |storage_dir|
     joins(:endpoint).where(endpoints: { storage_location: storage_dir })
   }
--- a/lib/audit/catalog_to_moab.rb
+++ b/lib/audit/catalog_to_moab.rb
@@ -7,14 +7,19 @@ class CatalogToMoab
   # allows for sharding/parallelization by storage_dir
   # use model scope query (which contains ordering), limit for batching, and .each within a while loop to
   # process records in order in batches.  Note that .find_each does batches, but disregards order from
-  # the scope, so we must use .each
+  # the scope, so we must use .each on a limit to process a batch after ordering over the whole set.
   def self.check_version_on_dir(last_checked_b4_date, storage_dir, limit=Settings.c2m_sql_limit)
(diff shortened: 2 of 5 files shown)


lib/audit/catalog_to_moab.rb
```diff
@@ -7,14 +7,19 @@ class CatalogToMoab
...
   def self.check_version_on_dir(last_checked_b4_date, storage_dir, limit=Settings.c2m_sql_limit)
     start_msg = "#{Time.now.utc.iso8601} C2M check_version starting for #{storage_dir}"
     puts start_msg
     Rails.logger.info start_msg
-    num_to_process = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count
+
+    # all_processable_copies is an AR relation; fine to run it with a .count or a .limit tacked on, but
+    # don't call .each directly on it and get the whole result set at once.
+    all_processable_copies =
+      PreservedCopy.least_recent_version_audit(last_checked_b4_date).by_storage_location(storage_dir)
```
jmartin-sul: did some testing to satisfy my own curiosity, and the chained scope version does generate essentially the same sql query as the version where `least_recent_version_audit` to `storage_dir`:
```sql
-- [5] pry(main)> PreservedCopy.least_recent_version_audit(Time.now.utc).by_storage_location('spec/fixtures/storage_root01/moab_storage_trunk')
SELECT "preserved_copies".*
FROM "preserved_copies"
INNER JOIN "endpoints"
ON "endpoints"."id" = "preserved_copies"."endpoint_id"
WHERE (last_version_audit IS NULL or last_version_audit < '2018-02-24 05:42:05.055725')
AND "endpoints"."storage_location" = $1
ORDER BY last_version_audit IS NOT NULL, last_version_audit ASC  [["storage_location", "spec/fixtures/storage_root01/moab_storage_trunk"]]

-- [1] pry(main)> PreservedCopy.least_recent_version_audit(Time.now.utc, 'spec/fixtures/storage_root01/moab_storage_trunk')
SELECT "preserved_copies".*
FROM "preserved_copies"
INNER JOIN "endpoints"
ON "endpoints"."id" = "preserved_copies"."endpoint_id"
WHERE "endpoints"."storage_location" = $1
AND (last_version_audit IS NULL or last_version_audit < '2018-02-24 05:42:39.866743')
ORDER BY last_version_audit IS NOT NULL, last_version_audit ASC  [["storage_location", "spec/fixtures/storage_root01/moab_storage_trunk"]]
```

(sql formatting edited for readability)

jmartin-sul:
saved the refactoring i wasn't as sure about for the last commit.  but i figure that filter will be useful generally (i'd have ended up re-implementing the same sort of endpoint filter in the scope that drove this PR, so seemed better to share it as a small scope of its own).

jmartin-sul:
> Dont know if we need both `by_endpoint_name` and `by_storage_location` scopes

yeah, i was a bit undecided on this.  both seemed useful, but only `by_storage_location` is actually used at the moment, and i totally understand the impulse to not write too much ahead of time.  figured we got `by_endpoint_name` pretty much for free since it was so simple and so similar to `by_storage_location`, but happy to remove if people think it adds too much clutter.

app/models/preserved_copy.rb
```diff
@@ -32,14 +32,31 @@ class PreservedCopy < ApplicationRecord
...
+    where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date)
       .order('last_version_audit IS NOT NULL, last_version_audit ASC')
   }
 
+  scope :fixity_check_expired, lambda {
+    joins(:preserved_object)
+      .joins(
+        'INNER JOIN preservation_policies'\
+        ' ON preservation_policies.id = preserved_objects.preservation_policy_id'\
+        ' AND (last_checksum_validation + (fixity_ttl * interval \'1 second\')) < CURRENT_TIMESTAMP'\
```
ndushay: double checking:  "interval" is a reserved word for scope?   And this is checking that the last_checksum_validation is at least 1 second older than the current timestamp, or it is null?  And we're sure it's pp AND (a OR b) rather than  (pp AND a) OR b ?

app/models/preserved_copy.rb
```diff
@@ -32,14 +32,31 @@ class PreservedCopy < ApplicationRecord
...
+        ' OR last_checksum_validation IS NULL'
+      )
+      .order('last_checksum_validation IS NOT NULL, last_checksum_validation ASC')
```
ndushay: ~~shouldn't we be doing the null ones FIRST?~~  I see the tests confirm we are getting the right order.

spec/models/preserved_copy_spec.rb
```diff
@@ -259,4 +259,153 @@
...
+        size: 1,
+        last_checksum_validation: (Time.now.utc - 1.day)
+      )
+    end
+    let!(:pcs_ordered_by_query) { PreservedCopy.fixity_check_expired }
+
+    it 'returns PreservedCopies with nils first, then old to new timestamps' do
+      expect(pcs_ordered_by_query).to eq [never_checked_pc, checked_before_threshold_pc1, checked_before_threshold_pc2]
+    end
+    it 'returns no PreservedCopies with timestamps indicating fixity check in the last week' do
```
ndushay: I would like `expect(preservation_policy.ttl_expire).to be a week` to be added to show where the value of "week" comes from ...

app/models/preserved_copy.rb
jmartin-sul: yeah, this is an unfortunately confusing sql thing.  oracle lets you just give a sort order and say `NULLS FIRST`.  other RDBMS software doesn't, though, as `NULLS FIRST` is an oracle-ism and not part of the vanilla sql standard.

since we've run into this counter-intuitive bit of sql before before, and since everyone (me included) first reads this `order by` as the opposite of what it really does, maybe a comment is in order?  see also https://github.com/sul-dlss/preservation_catalog/pull/519#discussion_r164555083

app/models/preserved_copy.rb
jmartin-sul: `interval` is actually just plain SQL (and following my own loose formatting convention, i probably should've capitalized it like the other SQL keywords).  the ActiveRecord `joins` method lets you do a join by either:
* specifying a table to join to (and a hash of field/value criteria on which to join if the join isn't on the obvious foreign key column, of if you'd like the join to be more restrictive than just the FK col).  from which it will magically generate your `join` subclause when building a sql statement.
* specifying raw SQL to use for the join if you want to do something more clever than what ActiveRecord allows you to do by way of a table name and a parameter hash.  so here, i've just written out the entire `INNER JOIN` sub-clause for the join to the `preservation_policies` table.

a bit more info on `interval` (which i'm just using here to allow for some simple date math, making the `fixity_ttl` integer addable to the `last_checksum_validation` timestamp, so that the sum can be compared to the current time):
* https://www.postgresql.org/docs/8.2/static/functions-datetime.html
* https://stackoverflow.com/questions/5879448/interval-function-in-oracle#5880358

jmartin-sul: cc'ing @atz here, since he's got some ActiveRecord experience.  if there is a way to do this that generates sane sql and doesn't resort to writing the sql ourselves, i'd be happy to do that instead, since that might be more maintainable in the long run.  i'm not super experience with ActiveRecord, so while i felt like i had a reasonable idea of what the SQL for this should look like, i couldn't figure out how to make AR write that SQL for me.

spec/models/preserved_copy_spec.rb
jmartin-sul: good call, i thought the 1 week threshold was non-obvious in the context of these tests, and was about to leave a comment, but then that felt like overcommenting.  i like the `expect` approach because it leaves the comment in such a way that the test breaks when it's no longer true 😃

app/models/preserved_copy.rb
ndushay: yeah, a comment would be lovely.  Feel free to add one to analogous places elsewhere..

app/models/preserved_copy.rb
ndushay: you got it working, and tests prove it's working, yes?   I like the idea of capitalizing INTERVAL and apologize for my SQL ignorance.

jmartin-sul: yes and yes and will do and no worries at all (because this stuff is not particularly intuitive despite what rails would like you to believe)

jmartin-sul:
@ndushay, i think i've implemented all the suggested feedback

spec/models/preserved_copy_spec.rb
jmartin-sul: i was a bit reluctant to test a settings value or behavior from the PreservationPolicy class in this spec.  so i ended up doing something slightly more clever, but also hopefully more illustrative of where the threshold comes from.

jmartin-sul:
ok, i had removed the `by_endpoint_name` scope after mild objections from both @SaravShah and @ndushay, but it turns out sarav will likely need it in work he's currently doing, so i added it back in a follow-on commit.

PR #541, opened 2018-01-31
stub for ChecksumValidator service class and tests

### LINKED ISSUE #511: (CV) create checksum validator service class
To start, we will want something roughly like `app/services/checksum_validator.rb`

```ruby
class ChecksumValidator
  def c2m_validation(druid, endpoint, algorithm)
    # TODO: implement this;  we begin with a placeholder
  end
end

### PR DESCRIPTION
**Pull request by jmartin-sul**
something like this?

closes #511

### DIFF
--- a/spec/services/checksum_validator_spec.rb
+++ b/spec/services/checksum_validator_spec.rb
@@ -1,0 +1,7 @@
+require 'rails_helper'
+
+RSpec.describe ChecksumValidator do
+  it 'has some behaviors' do
+    skip 'we should figure out what they are and test them'
+  end
+end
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: spec/services/checksum_validator_spec.rb [hunk 1]
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe ChecksumValidator do
+  it 'has some behaviors'
```
ndushay: wonder if this should be a skip to show up as "needs work" rather than a false pass??

jmartin-sul: @ndushay: fixed!

PR #542, opened 2018-01-31
Remove duplicate checksum validation failure statuses

### LINKED ISSUE #372: PreservedCopy has duplicate statuses for checksum validation failure

jmartin-sul:
this is branched off of c2m-handle-pc-status-already-bad from PR #526, because that PR includes a comment that references one of the statuses being removed here (because it ripped off a method from POH that did the same, also fixed in this PR).  so we should merge that PR and rebase this one on master before merging it (hence the WIP).

closes #372

### DIFF
--- a/app/models/preserved_copy.rb
+++ b/app/models/preserved_copy.rb
@@ -5,7 +5,7 @@ class PreservedCopy < ApplicationRecord
   INVALID_MOAB_STATUS = 'invalid_moab'.freeze
   INVALID_CHECKSUM_STATUS = 'invalid_checksum'.freeze
   ONLINE_MOAB_NOT_FOUND_STATUS = 'online_moab_not_found'.freeze
-  EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS = 'unexpected_version_on_storage'.freeze
+  UNEXPECTED_VERSION_ON_STORAGE_STATUS = 'unexpected_version_on_storage'.freeze
   VALIDITY_UNKNOWN_STATUS = 'validity_unknown'.freeze
 
   # NOTE:  DO NOT change the underlying constants for enum values that have been merged to
@@ -15,7 +15,7 @@ class PreservedCopy < ApplicationRecord
     INVALID_MOAB_STATUS => 1,
     INVALID_CHECKSUM_STATUS => 2,
     ONLINE_MOAB_NOT_FOUND_STATUS => 3,
-    EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS => 4,
+    UNEXPECTED_VERSION_ON_STORAGE_STATUS => 4,
     VALIDITY_UNKNOWN_STATUS => 6
   }
 
--- a/app/models/preserved_copy.rb
+++ b/app/models/preserved_copy.rb
@@ -10,6 +10,9 @@ class PreservedCopy < ApplicationRecord
 
   # NOTE:  DO NOT change the underlying constants for enum values that have been merged to
   # master/used in prod db (or at least, consider the necessary migration)
+  # If an enum value is removed, it's probably easiest to just not re-use the underlying int code it used before.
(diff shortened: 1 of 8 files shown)


jmartin-sul:
this should be ready for review, now that #526 has been merged and this has been rebased on master

app/models/preserved_copy.rb
```diff
@@ -16,8 +15,7 @@ class PreservedCopy < ApplicationRecord
     INVALID_MOAB_STATUS => 1,
     INVALID_CHECKSUM_STATUS => 2,
     ONLINE_MOAB_NOT_FOUND_STATUS => 3,
-    EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS => 4,
-    FIXITY_CHECK_FAILED_STATUS => 5,
+    UNEXPECTED_VERSION_ON_STORAGE_STATUS => 4,
     VALIDITY_UNKNOWN_STATUS => 6
```
ndushay: so we'll just skip 5  for now?    Ok with me.  Should there be a comment about it?

jmartin-sul: yeah, i'd be inclined to skip 5 permanently.  because we have things that use 6.  since nothing used the removed status, which previously mapped to 5, we could conceivably re-use that code.  but i don't think we'll have tones of new statuses, so my preference would be to just keep using the next highest enum value as we add new statuses, without re-using old ones.  just seems lower effort.

there is already the comment about not changing the underlying constants, but probably best to add a clarification about why there might be gaps in the sequence?

jmartin-sul: @ndushay, how about something like:

```ruby
# NOTE:  DO NOT change the underlying constants for enum values that have been merged to
# master/used in prod db (or at least, consider the necessary migration).
# If an enum value is removed, it's probably easiest to just not re-use the underlying int code it used before.
# Just add new statuses using the next highest unused int value, and treat the numbering as an increasing
# sequence that will likely have holes.
```

app/models/preserved_copy.rb
```diff
@@ -5,8 +5,7 @@ class PreservedCopy < ApplicationRecord
   INVALID_MOAB_STATUS = 'invalid_moab'.freeze
   INVALID_CHECKSUM_STATUS = 'invalid_checksum'.freeze
   ONLINE_MOAB_NOT_FOUND_STATUS = 'online_moab_not_found'.freeze
-  EXPECTED_VERS_NOT_FOUND_ON_STORAGE_STATUS = 'expected_vers_not_found_on_storage'.freeze
-  FIXITY_CHECK_FAILED_STATUS = 'fixity_check_failed'.freeze
+  UNEXPECTED_VERSION_ON_STORAGE_STATUS = 'unexpected_version_on_storage'.freeze
```
SaravShah: i'm guessing we will want to report this to workflow errors?

jmartin-sul: yup, but i believe that'll still be handled whenever the corresponding `AuditResults::UNEXPECTED_VERSION` result code gets processed, which should remain unchanged by this PR.  this was just a rename of the constant and the string that we use for the PresCopy status that reflects this situation.

app/models/preserved_copy.rb
jmartin-sul: fixed

PR #546, opened 2018-01-31
[WIP] ensure last_checked_b4_date is in the correct format

### LINKED ISSUE #485: (M2C) ensure last_checked_b4_date is in the correct format
In check_version_on_dir, we need to ensure that last_checked_b4_date is in the format needed.

Possibly, this check should be done first in any method with the argument, or possibly it should always fall through to check_version_on_dir method to do the testing.

Note we might want to accept certain incorrect formats and convert them, rather than reject all but the format we ultimately want.

SaravShah:
#### NOTE: Do not Merge this until #519 is merged. Will rebase on master and edit code to display changes made in #519
- probably only needs 1 pair of eyes
- adds `.normalize_date(timestamp)` so `last_checked_b4_date` can either be a Time object, String, or nil.
- RSpec tests for this method
- Rubocop exception

closes #485

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -132,6 +132,10 @@ Rails/Output:
     - 'lib/audit/catalog_to_moab.rb' # use puts statements for following progress of long job
     - 'lib/audit/moab_to_catalog.rb' # use puts statements for following progress of long job
 
+Rails/TimeZone:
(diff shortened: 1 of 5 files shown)


SaravShah:
closed in favor of #550

PR #550, opened 2018-02-01
Date format

### LINKED ISSUE #485: (M2C) ensure last_checked_b4_date is in the correct format

SaravShah:
- probably only needs 1 pair of eyes
- adds .normalize_date(timestamp) so last_checked_b4_date can either be a Time object, String, or nil.
- RSpec tests for this method
- ~~Rubocop exception~~
closes #485

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -136,6 +136,10 @@ Rails/Output:
     - 'lib/audit/catalog_to_moab.rb' # use puts statements for following progress of long job
     - 'lib/audit/moab_to_catalog.rb' # use puts statements for following progress of long job
 
+Rails/TimeZone:
+  Exclude:
+    - 'app/models/preserved_copy.rb' # use Time.parse without zone
+
 # --- RSpec ---
 
 RSpec/BeforeAfterAll:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -136,10 +136,6 @@ Rails/Output:
     - 'lib/audit/catalog_to_moab.rb' # use puts statements for following progress of long job
     - 'lib/audit/moab_to_catalog.rb' # use puts statements for following progress of long job
 
-Rails/TimeZone:
-  Exclude:
-    - 'app/models/preserved_copy.rb' # use Time.parse without zone
-
 # --- RSpec ---
 
 RSpec/BeforeAfterAll:
--- a/Rakefile
+++ b/Rakefile
@@ -128,14 +128,18 @@ task :c2m_check_version_on_dir, [:last_checked_b4_date, :storage_root, :profile]
   root = args[:storage_root].to_sym
   storage_dir = "#{Settings.moab.storage_roots[root]}/#{Settings.moab.storage_trunk}"
   last_checked = args[:last_checked_b4_date].to_sym
-
-  if args[:profile] == 'profile'
-    puts "When done, check log/profile_c2m_check_version_on_dir[TIMESTAMP].txt for profiling details"
-    CatalogToMoab.check_version_on_dir_profiled(last_checked, storage_dir)
-  elsif args[:profile].nil?
-    CatalogToMoab.check_version_on_dir(last_checked, storage_dir)
+  begin
+    if args[:profile] == 'profile'
+      puts "When done, check log/profile_c2m_check_version_on_dir[TIMESTAMP].txt for profiling details"
+      CatalogToMoab.check_version_on_dir_profiled(last_checked, storage_dir)
+    elsif args[:profile].nil?
+      CatalogToMoab.check_version_on_dir(last_checked, storage_dir)
+    end
+    puts "#{Time.now.utc.iso8601} Catalog to Moab version check on #{storage_dir} is done."
+  rescue TypeError, ArgumentError
+    p "You've entered an incorrect timestamp format #{last_checked}."
+    p "Please enter correct timestamp format (UTC) (2018-02-01T18:54:48Z)"
   end
-  puts "#{Time.now.utc.iso8601} Catalog to Moab version check on #{storage_dir} is done."
   $stdout.flush
 end
 
@@ -146,13 +150,17 @@ task :c2m_check_version_all_dirs, [:last_checked_b4_date, :profile] => [:environ
     exit
   end
   last_checked = args[:last_checked_b4_date].to_sym
-
-  if args[:profile] == 'profile'
-    puts "When done, check log/profile_c2m_check_version_all_roots[TIMESTAMP].txt for profiling details"
-    CatalogToMoab.check_version_all_dirs_profiled(last_checked)
-  elsif args[:profile].nil?
-    CatalogToMoab.check_version_all_dirs(last_checked)
+  begin
+    if args[:profile] == 'profile'
+      puts "When done, check log/profile_c2m_check_version_all_roots[TIMESTAMP].txt for profiling details"
+      CatalogToMoab.check_version_all_dirs_profiled(last_checked)
+    elsif args[:profile].nil?
+      CatalogToMoab.check_version_all_dirs(last_checked)
+    end
+    puts "#{Time.now.utc.iso8601} Catalog to Moab version check on all roots is done."
+  rescue TypeError, ArgumentError
+    p "You've entered an incorrect timestamp format #{last_checked}."
+    p "Please enter correct timestamp format (UTC) (2018-02-01T18:54:48Z)"
   end
-  puts "#{Time.now.utc.iso8601} Catalog to Moab version check on all roots is done."
   $stdout.flush
 end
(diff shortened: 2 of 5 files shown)


.rubocop.yml
```diff
@@ -136,6 +136,10 @@ Rails/Output:
     - 'lib/audit/catalog_to_moab.rb' # use puts statements for following progress of long job
     - 'lib/audit/moab_to_catalog.rb' # use puts statements for following progress of long job
 
+Rails/TimeZone:
+  Exclude:
+    - 'app/models/preserved_copy.rb' # use Time.parse without zone
+
 # --- RSpec ---
```
tallenaz: I can get all specs/cops to pass without this exception by following one of Rubocop's suggestions -- see next comment.

app/models/preserved_copy.rb
```diff
@@ -63,4 +64,14 @@ def update_status(new_status)
...
     version == preserved_object.current_version
   end
+
+  # Input: Takes a Time object, String, or Nil
+  # Output: If String method will return a time object,
+  # if nil, method will return the start of the Epoch Time Object
+  # and if Time object, it will return the same Time object.
+  private_class_method def self.normalize_date(timestamp)
+    timestamp = '1970-01-01T00:00:00Z' if timestamp.nil?
+    timestamp = Time.parse(timestamp) unless timestamp.instance_of?(Time)
```
tallenaz: With `timestamp = Time.parse(timestamp).utc unless timestamp.instance_of?(Time)`, all specs/cops pass, without the Rubocop exception. Is there another reason not to add `.utc`?

SaravShah:
@tallenaz i've made the change and is ready for review again!

tallenaz:
The only other concern I have is that, in some places, for example (https://github.com/sul-dlss/preservation_catalog/blob/master/spec/lib/audit/catalog_to_moab_class_spec.rb#L5), we take pains to make sure timestamps are in `iso8601`. Do we care about that in this method?

tallenaz:
`.iso8601` transforms a `DateTime` object into a `String`, and we don't want a string.

lib/audit/catalog_to_moab.rb
```diff
@@ -6,7 +6,6 @@ class CatalogToMoab
 
   # allows for sharding/parallelization by storage_dir
   def self.check_version_on_dir_of_batch(last_checked_b4_date, storage_dir, limit)
-    # TODO: ensure last_checked_version_b4_date is in the right format for query - see #485
```
ndushay: There is also a similar comment on line #30 or so.

app/models/preserved_copy.rb
```diff
@@ -63,4 +64,14 @@ def update_status(new_status)
   def matches_po_current_version?
     version == preserved_object.current_version
   end
+
+  # Input: Takes a Time object, String, or Nil
+  # Output: If String method will return a time object,
+  # if nil, method will return the start of the Epoch Time Object
+  # and if Time object, it will return the same Time object.
```
ndushay: I think the comments are largely unnecessary -- they read like the code itself, no?   What do you think about something like

```ruby
# Wrangles time object, string or nil into what is needed for SQL query
# Note that nil defaults to the start of the Epoch so only PreservedCopy objects without a last_version_audit timestamp will be processed
```

In some ways the above comments are in the code already, or more accurately belong in the README or ... somwhere else, perhaps?  That last one could be a `@param` comment for public methods taking date arg ...

spec/models/preserved_copy_spec.rb
```diff
@@ -234,5 +234,17 @@
...
+    context '.normalize_date(timestamp)' do
+      it 'given a String timestamp, returns a Time object' do
+        expect(described_class.send(:normalize_date, '2018-01-22T18:54:48')).to be_an_instance_of(Time)
+      end
+      it 'given a Time Object, returns the same Time object' do
+        expect(described_class.send(:normalize_date, Time.now.utc)).to be_an_instance_of(Time)
+      end
+      it 'given nil, returns the start of teh Epoch Time object' do
+        expect(described_class.send(:normalize_date, nil)).to be_an_instance_of(Time)
+      end
```
ndushay: Might we have a test for
- unparseable (as a date)
- day only (no time) -- show what happens (specific Time, not just Time object)
- month only (no day or time)
- year only

For "fun", if you want to see some ugly date strings, coming from our actual real production data, check out https://github.com/sul-dlss/stanford-mods/blob/master/spec/date_parsing_spec.rb.  Note that the cases I mention above are likely sufficient for exercising what might get typed into a rake task in this context ...

app/models/preserved_copy.rb
SaravShah: i think you're right. The comment is in the code already - and is not needed. Looking through the code you can tell that all of that is happening. I'll remove it :)

SaravShah:
-Edit removed setting default date to `1970-01-01T00:00:00Z` if input was nil.
If the user enters nil, nothing would get returned and they would have to dig through the code to figure out what happened.

- I think it would be better if we returned a message saying "You entered an incorrect timestamp format, please enter a correct timestamp format"

Rakefile
```diff
@@ -128,14 +128,18 @@ task :c2m_check_version_on_dir, [:last_checked_b4_date, :storage_root, :profile]
...
-    CatalogToMoab.check_version_on_dir(last_checked, storage_dir)
+  begin
+    if args[:profile] == 'profile'
+      puts "When done, check log/profile_c2m_check_version_on_dir[TIMESTAMP].txt for profiling details"
+      CatalogToMoab.check_version_on_dir_profiled(last_checked, storage_dir)
+    elsif args[:profile].nil?
+      CatalogToMoab.check_version_on_dir(last_checked, storage_dir)
+    end
+    puts "#{Time.now.utc.iso8601} Catalog to Moab version check on #{storage_dir} is done."
+  rescue StandardError
```
jmartin-sul: here i think i'd be more specific and `rescue ArgumentError, TypeError`

Rakefile
```diff
@@ -146,13 +150,17 @@ task :c2m_check_version_all_dirs, [:last_checked_b4_date, :profile] => [:environ
...
-    CatalogToMoab.check_version_all_dirs(last_checked)
+  begin
+    if args[:profile] == 'profile'
+      puts "When done, check log/profile_c2m_check_version_all_roots[TIMESTAMP].txt for profiling details"
+      CatalogToMoab.check_version_all_dirs_profiled(last_checked)
+    elsif args[:profile].nil?
+      CatalogToMoab.check_version_all_dirs(last_checked)
+    end
+    puts "#{Time.now.utc.iso8601} Catalog to Moab version check on all roots is done."
+  rescue StandardError
```
jmartin-sul: same thing here about the more specific `rescue`

Rakefile
SaravShah: that sounds good, i'll add it. I'm wondering if there are any other Error's I'm leaving out that we don't know about.

jmartin-sul: maybe, but i think better to let them bubble up.  i think you only want to catch stuff you expect and know how to handle.

Rakefile
```diff
@@ -128,14 +128,18 @@ task :c2m_check_version_on_dir, [:last_checked_b4_date, :storage_root, :profile]
...
+    if args[:profile] == 'profile'
+      puts "When done, check log/profile_c2m_check_version_on_dir[TIMESTAMP].txt for profiling details"
+      CatalogToMoab.check_version_on_dir_profiled(last_checked, storage_dir)
+    elsif args[:profile].nil?
+      CatalogToMoab.check_version_on_dir(last_checked, storage_dir)
+    end
+    puts "#{Time.now.utc.iso8601} Catalog to Moab version check on #{storage_dir} is done."
+  rescue TypeError, ArgumentError
+    p "You've entered an incorrect timestamp format #{last_checked}."
+    p "Please enter correct timestamp format (2018-02-01T18:54:48)"
```
ndushay: is this UTC time?  We probably want to say so here ... and imply so with the Z showing (Z on the end implies UTC).

Rakefile
```diff
@@ -146,13 +150,17 @@ task :c2m_check_version_all_dirs, [:last_checked_b4_date, :profile] => [:environ
...
+    if args[:profile] == 'profile'
+      puts "When done, check log/profile_c2m_check_version_all_roots[TIMESTAMP].txt for profiling details"
+      CatalogToMoab.check_version_all_dirs_profiled(last_checked)
+    elsif args[:profile].nil?
+      CatalogToMoab.check_version_all_dirs(last_checked)
+    end
+    puts "#{Time.now.utc.iso8601} Catalog to Moab version check on all roots is done."
+  rescue TypeError, ArgumentError
+    p "You've entered an incorrect timestamp format #{last_checked}."
+    p "Please enter correct timestamp format (2018-02-01T18:54:48)"
```
ndushay: ditto

Issue #539, opened 2018-01-31
(PC) AuditResults - generalize it -- message prefix and instance vars needs rework (see also #528)


ndushay:
- [x] do we actually want the message prefix on EACH result?  Could we just use it for logging and for workflow reporting as needed?  (#545)

prefix:
- [x] class is used by more than PreservedObjectHandler:  should we pass in the calling class name?  should we derive it?  should we forget it?  (#545)
~~- [ ] should endpoint info be storage_location instead of endpoint_name?~~

attributes (coded in branch audit-results-revisions)
- [x] "incoming_version" is less meaningful when it's a C2M check ... -- change to 'actual_version' as var name and in result messages (#544)
     - [x] ensure result messages have actual_version correct for C2M (#544)

- [x]  "incoming_size" -- kill this (#544)

related to #528

PR #551, opened 2018-02-01
audit_results: constant names made less confusing and shorter

ndushay:
(again, probably everyone should see this)

AuditResults constant names were confusing me:
- "OBJECT" referring to db vs. moab.
- "actual" which I neglected to migrate to the constants
- THE_INCREDIBLY_LONG_NAMES_I_WROTE_WERE_BOTHERING_ME

Connects to #539

### DIFF
--- a/app/services/audit_results.rb
+++ b/app/services/audit_results.rb
@@ -14,12 +14,12 @@ class AuditResults
 
   INVALID_ARGUMENTS = :invalid_arguments
   VERSION_MATCHES = :version_matches
-  ARG_VERSION_GREATER_THAN_DB_OBJECT = :arg_version_greater_than_db
-  ARG_VERSION_LESS_THAN_DB_OBJECT = :arg_version_less_than_db
+  ACTUAL_VERS_GT_DB_OBJ = :actual_vers_gt_db_obj
+  ACTUAL_VERS_LT_DB_OBJ = :actual_vers_lt_db_obj
   CREATED_NEW_OBJECT = :created_new_object
   DB_UPDATE_FAILED = :db_update_failed
-  OBJECT_ALREADY_EXISTS = :object_already_exists
-  OBJECT_DOES_NOT_EXIST = :object_does_not_exist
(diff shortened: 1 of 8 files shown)


app/services/audit_results.rb
```diff
@@ -14,12 +14,12 @@ class AuditResults
 
   INVALID_ARGUMENTS = :invalid_arguments
   VERSION_MATCHES = :version_matches
-  ARG_VERSION_GREATER_THAN_DB_OBJECT = :arg_version_greater_than_db
-  ARG_VERSION_LESS_THAN_DB_OBJECT = :arg_version_less_than_db
+  ACTUAL_VERS_GT_DB_OBJ = :arg_version_greater_than_db
+  ACTUAL_VERS_LT_DB_OBJ = :arg_version_less_than_db
   CREATED_NEW_OBJECT = :created_new_object
```
tallenaz: `ARG` changes to `ACTUAL` on the left, but `arg` stays `arg` on the right?

app/services/audit_results.rb
```diff
@@ -14,12 +14,12 @@ class AuditResults
...
   DB_UPDATE_FAILED = :db_update_failed
-  OBJECT_ALREADY_EXISTS = :object_already_exists
-  OBJECT_DOES_NOT_EXIST = :object_does_not_exist
+  DB_OBJ_ALREADY_EXISTS = :object_already_exists
+  DB_OBJ_DOES_NOT_EXIST = :object_does_not_exist
   PC_STATUS_CHANGED = :pc_status_changed
```
tallenaz: Similarly, `OBJECT` changes to `DB_OBJ` on the left, but `object` stays `object` on the right?

app/services/audit_results.rb
ndushay: hahahaha - oops.  thanks.

app/services/audit_results.rb
ndushay: also ... oops.  Will fix.

ndushay:
@tallenaz fixed!  thanks for spotting it.

PR #552, opened 2018-02-01
Refactor c2m sql query

### LINKED ISSUE #549: (C2M) Tidy up shameless green sql query code
See https://github.com/sul-dlss/preservation_catalog/pull/519#pullrequestreview-93136679 for requirements.

tallenaz:
to reduce number of methods involved in the version check
  and to call least_recent_version_audit only once

  Added a spec to test that the version check respects limit

  Closes #549 
  
  Shouldn't need > 1 reviewer.


tallenaz:
If #550 gets merged first, then this should be rebased.

lib/audit/catalog_to_moab.rb
```diff
@@ -5,18 +5,14 @@
...
-    pcs.each do |pc|
-      c2m = CatalogToMoab.new(pc, storage_dir)
-      c2m.check_catalog_version
-    end
-  end
-
-  def self.check_version_on_dir(last_checked_b4_date, storage_dir)
-    unless PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count.zero?
-      check_version_on_dir_of_batch(last_checked_b4_date, storage_dir, Settings.c2m_sql_limit)
+    pcs = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir)
```
ndushay: once again, please pardon my ignorance here.

Based on what @jmartin-sul said in the earlier PR https://github.com/sul-dlss/preservation_catalog/pull/519#discussion_r165276153 and F2F at standup:

- doing a "count" query is cheap
- getting back a LOT of actual objects is expensive

So is this a step in the wrong direction?

Not sure if the following s expensive or not:
- "order by" on a large set of objects (order by date)


The following could avoid repeating the count query (which probably does NOT matter):

```ruby
num_to_process = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count
while num_to_process > 0
  check_version_on_dir_of_batch(last_checked_b4_date, storage_dir, Settings.c2m_sql_limit)
  num_to_process = num_to_process - Settings.c2m_sql_limit
end
```

If @jmartin-sul agrees, I think leaving the code as it was will likely have better performance.  I would guess the performance hits are in sorting a huge number of db entries, in retrieving a huge number of db entries, and in instantiating ActiveRecord objects.

jmartin-sul: ah, ok, some clarification:
* my understanding, not empirically tested here, but which i have some confidence in, is that count queries are indeed cheaper.
  * but, my guess without yet having tried it in a console is that 
   ```ruby
    num_to_process = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count
    while num_to_process > 0
    ```
   will loop infinitely or not at all, because num_to_process will be the integer result of the `count` (i think `count` and `pluck` both cause you to run the query immediately and get a value or values back immediately, as opposed to getting an ActiveRecord Relation that executes lazily once you do something to access the result).  so you'd probably want to do the same but skip the intermediate variable.
* we shouldn't do quite what's happening on this PR now, which i believe will inadvertently load all our pres copies into memory:  i think `unless pcs.empty?` will treat the `least_recent_version_audit` `Relation` as an array, just to see whether it has any elements.  since there's no limit clause on that one, all records would be loaded, i think.  and then the query will re-run with the limit (which would likely be cheaper than the check, ironically, due to the limit).  so... there seem to me to be two options
  * what's here, minus the initial check to see whether there are elements.  just the `pcs.limit(limit).each do |pc| ...` loop.  if there are pres copies that satisfy the query, `limit` or fewer of them will be returned and iterated over, and if there are none that satisfy the query, no loop iterations occur.  no need to get a count beforehand, i don't think.
  * as naomi suggests, use the count to have a `while` loop iterate over all qualifying  pres copies in `limit` sized batches (but inline the `count` query into the loop condition, or assign it to a lambda or whatever so that it's re-evaluated every time, since `count` returns an int immediately).

i'm partial to the second, because then we don't have to get fancy with our cron job scheduling, other than maybe a lockfile.  we iterate through everything available till nothing's left to be audited.


http://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-count
http://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-pluck

jmartin-sul:
a bit of testing on stage just now:

i did this `count` and it returned what felt like instantly:
```ruby
[24] pry(main)> PreservedCopy.joins(:endpoint).where(endpoints: { storage_location: storage_dir }).where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date).order('last_version_audit IS NOT NULL, last_version_audit ASC').count
=> 45040
```

and i did this query for the 1000 least recently audited objects on disk 14 and it returned in what felt like less than a second (with all the AR objects loaded into memory, presumably, since i was paging through an array in pry's less viewer):
```ruby
[27] pry(main)> PreservedCopy.joins(:endpoint).where(endpoints: { storage_location: storage_dir }).where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date).order('last_version_audit IS NOT NULL, last_version_audit ASC').limit(1000)
=> [#<PreservedCopy:0x0000000002f401a8
  id: 186017,
  version: 1,
  preserved_object_id: 186017,
  endpoint_id: 13,
  created_at: Fri, 26 Jan 2018 21:06:32 UTC +00:00,
  updated_at: Fri, 02 Feb 2018 00:40:57 UTC +00:00,
  last_moab_validation: Fri, 26 Jan 2018 21:06:32 UTC +00:00,
  last_checksum_validation: nil,
  size: 72069545,
  status: "ok",
  last_version_audit: Fri, 02 Feb 2018 00:40:57 UTC +00:00>,
 #<PreservedCopy:0x0000000002f3bd88
  id: 186018,
  version: 1,
...
```

so i think the count is cheap as is the retrieval with a limit and a reasonable batch size.

ok, i just did the same ordered query with a limit of 10k, and that was still maybe just over a second.  would be interesting to see some runs with some profiling.

lib/audit/catalog_to_moab.rb
jmartin-sul: d'oh, sorry @ndushay, i didn't see the decrementing in your loop.  probably a safer bet than my proposal of re-running the query, since it would result in a definite finite number of iterations (whereas getting the count anew could go forever if our throughput was too low or we started failing to update audit dates).

lib/audit/catalog_to_moab.rb
```diff
@@ -5,7 +5,7 @@
 class CatalogToMoab
 
   # allows for sharding/parallelization by storage_dir
-  def self.check_version_on_dir_of_batch(last_checked_b4_date, storage_dir, limit)
+  def self.check_version_on_dir(last_checked_b4_date, storage_dir, limit)
     # TODO: ensure last_checked_version_b4_date is in the right format for query - see #485
     pcs = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).limit(limit)
```
ndushay: doesn't there need to be some sort of clause somewhere to set limit to Settings.c2m_sql_limit if it is nil?  Like maybe in the args `limit = Settings.c2m_sql_limit` ?  

I think we want this method to be directly callable.

jmartin-sul: i like the default limit idea.  would be happy to do it as a follow on ticket, but seems worth doing.

jmartin-sul:
@ndushay said:
> I would like some "proof" (maybe in the form of a spec?) that we are getting the first X of ALL the records ordered, rather than the first X records, which are then ordered.

> If this is straight up Arel or AR or other foo, please pardon my ignorance.

this sort of limit clause is pretty standard SQL, so i'd be happy to assume our DB is doing the right thing if we're constructing the query correctly.

@ndushay, to your point about a default limit, we could do something along these lines and check to see that the limit is present on the query even if the limit is omitted from the params.  but i'm not inclined to test whethere `LIMIT` works as expected in PG.

> I would also like a test showing that we process multiple batches properly. Perhaps setting limit to 1 or 2 for 3 relevant records and ensuring they are all processed, in the right order.

we don't, but neither did the prior iteration of this PR (because it checked the count using an `unless`, which doesn't loop; presumably we'd use a `while` or an until if we wanted that part of the code to loop till count hit zero).

there is a test in the current PR making sure that the processing done by`check_version_on_dir` doesn't exceed the batch size:
https://github.com/sul-dlss/preservation_catalog/blob/151434489b62dc37974c486327430d5df1b732c0/spec/lib/audit/catalog_to_moab_class_spec.rb#L22-L26

jmartin-sul:
per the SQL standard, the result set is constructed by first considering the `where` clause, then the `order`, then the `limit`:
https://stackoverflow.com/questions/39680598/querying-with-limit-and-offset-is-there-a-chance-to-have-different-results-with#39680636
https://stackoverflow.com/questions/4708708/mysql-order-by-limit

side note, probably inconsequential:  both postgres and mysql seem to say that they don't promise that `limit` will return the same N rows as the first N rows of the non-`limit`ed version of the same query.  but only in cases where there's no deterministic way to break a tie in the sort.  which is business as usual since the sql standard never guarantees any sort of consistent ordering beyond what's needed to fulfill the `order` clause's restrictions.  which i think is not a thing that should matter to us (because we probably won't have things with exactly the same audit time stamps, and if we do, we don't really have something to deterministically break that tie with anyway).


these are good bits of re-assurance that a `limit`ed query will be more performant than the same query sans `limit` (as well as some more official substantiation of the above aside):
https://www.postgresql.org/docs/8.1/static/queries-limit.html

> The query optimizer takes LIMIT into account when generating a query plan, so you are very likely to get different plans (yielding different row orders) depending on what you give for LIMIT and OFFSET. Thus, using different LIMIT/OFFSET values to select different subsets of a query result will give inconsistent results unless you enforce a predictable result ordering with ORDER BY. This is not a bug; it is an inherent consequence of the fact that SQL does not promise to deliver the results of a query in any particular order unless ORDER BY is used to constrain the order.

https://www.postgresql.org/docs/8.0/static/performance-tips.html
> Estimated total cost (If all rows were to be retrieved, which they may not be: a query with a LIMIT clause will stop short of paying the total cost, for example.)

tallenaz:
Paired with @jmartin-sul and this is ready for review -- which would be @SaravShah today

lib/audit/catalog_to_moab.rb
```diff
@@ -5,17 +5,17 @@
...
-      c2m = CatalogToMoab.new(pc, storage_dir)
-      c2m.check_catalog_version
-    end
-  end
-
-  def self.check_version_on_dir(last_checked_b4_date, storage_dir)
-    unless PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count.zero?
-      check_version_on_dir_of_batch(last_checked_b4_date, storage_dir, Settings.c2m_sql_limit)
+  def self.check_version_on_dir(last_checked_b4_date, storage_dir, limit=Settings.c2m_sql_limit)
+    # This runs through everything added since last_checked_b4_date at the time .count is run
```
jmartin-sul: on second thought... maybe best to leave this comment out?  this wording is concise and readable, but i realized it's misleading, since we're really getting everything that's never been checked (regardless of when it was added) and everything that hasn't been checked since prior to `last_checked_b4_date`.

i think the reason i wanted to leave a comment is because the alternative would be to just use the `.count` call directly in the loop condition, in which case the loop would keep running through anything added to the repository while the loop was already running.  that's what i assumed the default behavior should be when i was looking at an earlier version of this PR, but that feels less safe (since there's a slim chance it could loop forever), and so actually i don't think there's a reason to highlight the behavior in this PR as weird or unexpected.

i can't think of a good concise explanation for what the behavior is that's easier than just reading the code and reasoning about it, and it's not all that hard to reason about anyway, so yeah, i vote for striking this line.

sorry for requesting a comment unnecessarily!

ndushay: I agree that this comment can go.

lib/audit/catalog_to_moab.rb
```diff
@@ -5,17 +5,17 @@
...
+    num_to_process = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count
+    while num_to_process > 0
+      pcs = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).limit(limit)
+      # This processes limit size batches, as find_each disregards order from the scope
```
ndushay: This comment seems misplaced somehow.  

What do you think of a *method* comment being the only comment, in which we say something like:

```ruby
# use model scope query (which contains ordering), limit for batching, and .each within a while loop to 
# process records in order in batches.  Note that .find_each does batches, but disregards order from 
# the scope, so we must use .each.
```

spec/lib/audit/catalog_to_moab_class_spec.rb
```diff
@@ -6,31 +6,41 @@
...
-      described_class.check_version_on_dir_of_batch(last_checked_version_b4_date, storage_dir, limit)
-    end
-  end
+    context 'when there are PreservedCopies to check' do
+      let(:c2m_mock) { instance_double(described_class) }
 
-  context ".check_version_on_dir" do
-    include_context 'fixture moabs in db'
-    let(:subject) { described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir) }
+      it 'creates an instance and calls #check_catalog_version within the limit' do
```
ndushay: "within the limit" doesn't make sense to me.  ?? `when query results are in single batch, calls #check_catalog_version for every result` ???   Is that the gist??

spec/lib/audit/catalog_to_moab_class_spec.rb
```diff
@@ -6,31 +6,41 @@
...
+        allow(described_class).to receive(:new).and_return(c2m_mock)
+        expect(c2m_mock).to receive(:check_catalog_version).exactly(3).times
+        described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir, limit)
+      end
 
-    it "calls check_version_on_dir_of_batch when there are objects to audit" do
-      expect(described_class).to receive(:check_version_on_dir_of_batch).exactly(1).times
-      subject
+      it 'creates an instance and calls #check_catalog_version on everything in batches' do
+        # we use 2 separate loops as we have to do all of our .new calls on CatalogToMoab before the allowance
```
ndushay: what's "the allowance" ?  I wonder if this comment can be deleted.   Basically, what I see is "here's a bunch of setup, with allows and expects, and here's the call to the method.   That's pretty standard rspec fare, to me.  I'm open to other opinions.

spec/lib/audit/catalog_to_moab_class_spec.rb
```diff
@@ -6,31 +6,41 @@
...
+        pcs_from_scope = PreservedCopy.least_recent_version_audit(last_checked_version_b4_date, storage_dir)
+        c2m_list = pcs_from_scope.map do |pc|
+          described_class.new(pc, storage_dir)
+        end
+        c2m_list.each do |c2m|
+          allow(described_class).to receive(:new).with(c2m.preserved_copy, storage_dir).and_return(c2m)
+          # we use and_call_original to get the audit timestamps updated
```
ndushay: this comment, to me, is overexplaining -- it's explaining what `.and_call_original` does, which is standard rspec.  I'm open to other opinions.

spec/lib/audit/catalog_to_moab_class_spec.rb
```diff
@@ -6,31 +6,41 @@
...
-      expect(described_class).to receive(:check_version_on_dir_of_batch).exactly(1).times
-      subject
+      it 'creates an instance and calls #check_catalog_version on everything in batches' do
+        # we use 2 separate loops as we have to do all of our .new calls on CatalogToMoab before the allowance
+        pcs_from_scope = PreservedCopy.least_recent_version_audit(last_checked_version_b4_date, storage_dir)
+        c2m_list = pcs_from_scope.map do |pc|
+          described_class.new(pc, storage_dir)
+        end
+        c2m_list.each do |c2m|
+          allow(described_class).to receive(:new).with(c2m.preserved_copy, storage_dir).and_return(c2m)
```
ndushay: to prove that it's doing the processing in batches -- would it make sense to change line 29 to "expect" ending in "at_least(2).times ??

spec/lib/audit/catalog_to_moab_class_spec.rb
jmartin-sul: the specific tricky part is that the allocation of real `CatalogToMoab` objects has to all happen up front in the first loop, because otherwise you set an `allow` on `C2M#new` for the first of the c2m test instances, but then immediately call `C2M#new` with parameters for the next c2m instance, and then rspec complains that you called `#new` with unexpected params.  that bit us and took us a few minutes to figure out.  our first inclination was to put everything in one loop.  the intent was just to explain why the intuitive thing doesn't work.

spec/lib/audit/catalog_to_moab_class_spec.rb
jmartin-sul: yeah, this is another thing that bit us (our test was failing because we weren't actually updating audit timestamps, so each batch just kept trying to grab the same first couple pres copies).

but this one's maybe more obvious than the two separate loops thing?

spec/lib/audit/catalog_to_moab_class_spec.rb
jmartin-sul: i don't think that'll work with the current test setup, since each pres copy needs its own specific c2m, and so no specific `#new` call will actually be made more than once.

i'm really not sure how to prove that it's processing `limit` pres copies at a time.  maybe... abstract the `.limit` query into a method call to return the AR Relation to get the next batch, and test it separately to ensure that it only returns `limit` results even if more are available?  i think i'd prefer to get this PR merged and ticket that as a follow on.  i could also bring it up as a question in code club today.

lib/audit/catalog_to_moab.rb
jmartin-sul: i like that idea

spec/lib/audit/catalog_to_moab_class_spec.rb
jmartin-sul: i think so

spec/lib/audit/catalog_to_moab_class_spec.rb
jmartin-sul: you could change it to an expect, and that'd get you the expectation you want, i think, because it'd show you're calling the method once for each pres copy.  which is less ambiguous than the `at_least(2).times` approach.  though that'd seem to me to test essentially the same thing as `expect(c2m).to receive(:check_catalog_version).exactly(1).times.and_call_original`, which seems a bit more to me to be the point of the test.

spec/lib/audit/catalog_to_moab_class_spec.rb
tallenaz: I'm in agreement with @jmartin-sul -- the setup was unintuitive, and the comment is hopefully informative in that light.

spec/lib/audit/catalog_to_moab_class_spec.rb
tallenaz: I hadn't run into `.and_call_original` before, so from that point of view, I feel like it's helpful, but since it's standard rspec, I'll delete it, and if we can discuss it further, if necessary.

spec/lib/audit/catalog_to_moab_class_spec.rb
tallenaz: +1 to ticketing this as follow on work.

spec/lib/audit/catalog_to_moab_class_spec.rb
tallenaz: Sounds good to me

tallenaz:
I believe I've addressed all the latest concerns. Should be ready for re-review.

jmartin-sul:
i think this just needs another round of review from @ndushay to confirm or not whether her suggestions were addressed?

lib/audit/catalog_to_moab.rb
```diff
@@ -5,17 +5,18 @@
...
-  def self.check_version_on_dir(last_checked_b4_date, storage_dir)
-    unless PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count.zero?
-      check_version_on_dir_of_batch(last_checked_b4_date, storage_dir, Settings.c2m_sql_limit)
+  def self.check_version_on_dir(last_checked_b4_date, storage_dir, limit=Settings.c2m_sql_limit)
+    num_to_process = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).count
+    while num_to_process > 0
+      pcs = PreservedCopy.least_recent_version_audit(last_checked_b4_date, storage_dir).limit(limit)
+      # use model scope query (which contains ordering), limit for batching, and .each within a while loop to
+      # process records in order in batches.  Note that .find_each does batches, but disregards order from
+      # the scope, so we must use .each
```
ndushay: ah -sorry I wasn't clear enough.  I meant for this to be a METHOD comment, not inside the method.  mainly because it refers to the whole strategy, not just the each loop.

spec/lib/audit/catalog_to_moab_class_spec.rb
ndushay: ah - ok.  I'm struggling with the word "allowance" - i don't connect it with "allow" statements.  How about

```ruby
# we must set up all the described_class instance objects ahead of any process calling CatalogToMoab.new
```

spec/lib/audit/catalog_to_moab_class_spec.rb
ndushay: It's fine for this to be ticketed as follow on work.  My main concern is. `it "creates an instance and calls #check_catalog_version on everything in batches"` is the description, but I'm not finding that the test is proving that things are in batches.    Maybe add a comment along these lines?

```
# note that there are 3 objects to be processed, and we are setting batch limit to 2, and we are ensuring that we call #check_catalog_version on all 3 objects.
```

???

Ok with this as follow on work

tallenaz:
Thanks @ndushay I've amended this commit with changes I hope satisfy your concerns about comments.

PR #559, opened 2018-02-03
Workflows error messages get more info

### LINKED ISSUE #509: (C2M) Add additional information to the ONLINE_MOAB_DOES_NOT_EXIST error code (blocked by #528)
Ask @julianmorley and @blalbrit  to see if there is any preference for what additional information we should send like..... 
~~- endpoint directory - unless it is already in the message prefix~~ done with #555
- preservedcopy version in catalog?
- updated_at or created_at timestamps??

ndushay:
You might want to review this by individual commits. 

Closes #509


app/services/audit_results.rb
```diff
@@ -39,7 +39,7 @@ class AuditResults
     UNEXPECTED_VERSION => "actual version (%{actual_version}) has unexpected relationship to %{addl} db version; ERROR!",
     INVALID_MOAB => "Invalid moab, validation errors: %{addl}",
     PC_PO_VERSION_MISMATCH => "PreservedCopy online moab version %{pc_version} does not match PreservedObject current_version %{po_version}",
-    ONLINE_MOAB_DOES_NOT_EXIST => "db has moab that is not found online"
+    ONLINE_MOAB_DOES_NOT_EXIST => "db PreservedCopy (created %{db_created_at}; last updated %{db_updated_at}) exists but Moab not found"
```
SaravShah: 👍

app/services/preserved_object_handler.rb
```diff
@@ -343,12 +343,25 @@ def with_active_record_transaction_and_rescue
 
   # expects @incoming_version to be numeric
   def version_comparison_results(db_object, version_symbol)
+    db_version =
+      case db_object.class
+      when PreservedCopy
+        db_object.version
+      when PreservedObject
+        db_object.current_version
+      end
```
jmartin-sul: i think instead you could just extract the result of `db_object.send(version_symbol)` into a variable, right?  would probably help make the conditional below more readable too, since it appears to be asking for that same piece of info in each of its branches.

alternatively, we could keep the above change and use `db_version` in place of the `db_object.send(version_symbol)`, and get rid of the `version_symbol` param, which always did feel a little icky from an object oriented design perspective.  i might actually be in favor of that.

ndushay: Good 👀 - I somehow missed that whole :version_symbol thing.   I added a commit that changes the argument passed to the actual db_version, which cleans everything up.

ndushay:
@jmartin-sul I believe I addressed your concern with an additional commit, so this is ready for re-review.

PR #575, opened 2018-02-08
Timestamp in c2m rake task argument is a string

tallenaz:
and profile log name is capitalized


SaravShah:
Will merge after tests pass

PR #580, opened 2018-02-18
db migration to index po and pc created_at and updated_at dates

### LINKED ISSUE #442: should we index some/all of the updated_at columns? (i.e. will we query on them?)
i expect that we'll query or sort by these columns at least sometimes?  maybe not?

ndushay:
2 reviewers?

It turns out we have been using updated_at when we query the db for timings and state via the rails console.

So far we've only used `PreservedCopy.updated_at` in those queries, by and large, but I thought we might play it safe.  Happy to index fewer fields with this PR if there's a reason not to index.  (Performance concerns?)

Closes #442

If we deploy to stage and prod with cap, this migration will happen appropriately.  No db data manipulation needed.


SaravShah:
yep! merging :)

PR #598, opened 2018-02-28
miscellaneous polish (shared ActiveRecord helper methods, readme, log file names)

### LINKED ISSUE #573: (C2M) refine class spec re processing in batches
``` ruby
it 'creates an instance and calls #check_catalog_version on everything in batches' do
        # we use 2 separate loops as we have to do all of our .new calls on CatalogToMoab before the allowance
        pcs_from_scope = PreservedCopy.least_recent_version_audit(last_checked_version_b4_date, storage_dir)
        c2m_list = pcs_from_scope.map do |pc|
          described_class.new(pc, storage_dir)
        end
        c2m_list.each do |c2m|
          allow(described_class).to receive(:new).with(c2m.preserved_copy, storage_dir).and_return(c2m)
          expect(c2m).to receive(:check_catalog_version).exactly(1).times.and_call_original
        end
        described_class.check_version_on_dir(last_checked_version_b4_date, storage_dir, 2)
      end
```
Refine this spec according to https://github.com/sul-dlss/preservation_catalog/pull/552#discussion_r166504761

- [x] move comment inside .check_version_on_dir to be a method level comment.
- [x]  change spec comment using the word "allowance"
- [ ]  add some clarification how the test on line 22 is truly checking batches.

### LINKED ISSUE #609: (CV, C2M) add tests to assert that we process objects in the right order
See discussion in #606 ("checksum.validate_disk" tests).  Possibly will be addressed by #598.

similarly for C2M.

jmartin-sul:
* CV: profiling output should use name of profile method in file name like similar M2C and C2M methods (also fix tests and documentation)
* centralize some re-useable activerecord code
* have C2M use transactions when it does updates that should be grouped atomically

TODO:
- [x] could use some tests for the new class
- [ ] i'm also not thrilled with the name for the new class, but it'll do if no one's got better ideas.
- [x] would also like to use the newly factored out transaction handler to add transactionality to C2M.
  - [x] which will also need tests
- [x] refactor the `limit`/`each` construct out of checksum validation, as was done in this PR already for C2M

fine to do any of those things in this ticket or in follow on tickets.

also, i know utility classes are a thing some people don't like.  happy to hear suggestions for a better approach, though i think this is better than rewriting the code in this PR multiple times across different classes that need to do very similar things.

~~this is now rebased on #606, which should get merged first (because this will refactor a thing from that PR).~~

closes #609 
closes #573

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -226,3 +226,7 @@ Style/StringLiterals:
 Style/SymbolArray:
   Exclude:
     - 'Rakefile' # because [:spec, :rubocop] isn't a big deal
+
+Style/TernaryParentheses:
+  Exclude:
+    - spec/lib/active_record_utils_spec.rb # reads better with parens
--- a/Rakefile
+++ b/Rakefile
@@ -174,7 +174,7 @@ task :cv_single_endpoint, [:storage_root, :profile] => [:environment] do |_t, ar
   end
   storage_root = args[:storage_root].to_sym
   if args[:profile] == 'profile'
-    puts "When done, check log/profile_cv_single_endpoint[TIMESTAMP] for profiling details"
+    puts "When done, check log/profile_cv_validate_disk[TIMESTAMP] for profiling details"
     Checksum.validate_disk_profiled(storage_root)
   elsif args[:profile].nil?
     Checksum.validate_disk(storage_root)
@@ -190,7 +190,7 @@ task :cv_all_endpoints, [:profile] => [:environment] do |_t, args|
     exit
   end
   if args[:profile] == 'profile'
-    puts "When done, check log/profile_cv_all_endpoints[TIMESTAMP].txt for profiling details"
+    puts "When done, check log/profile_cv_validate_disk_all_endpoints[TIMESTAMP].txt for profiling details"
     Checksum.validate_disk_all_endpoints_profiled
   elsif args[:profile].nil?
     Checksum.validate_disk_all_endpoints
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -7,6 +7,7 @@
 # inspired by http://www.thegreatcodeadventure.com/smarter-rails-services-with-active-record-modules/
 class PreservedObjectHandler
 
+  require 'active_record_utils.rb'
   require 'audit_results.rb'
 
   include ActiveModel::Validations
@@ -319,20 +320,9 @@ class PreservedObjectHandler
     handler_results.remove_db_updated_results unless transaction_ok
   end
 
-  # performs passed code wrapped in ActiveRecord transaction via yield
-  # @return true if transaction completed without error; false if ActiveRecordError was raised
+  # this wrapper reads a little nicer in this class, since POH is always doing this the same way
   def with_active_record_transaction_and_rescue
-    begin
-      ApplicationRecord.transaction { yield }
-      return true
-    rescue ActiveRecord::RecordNotFound => e
-      handler_results.add_result(AuditResults::DB_OBJ_DOES_NOT_EXIST, e.inspect)
-    rescue ActiveRecord::ActiveRecordError => e
-      handler_results.add_result(
-        AuditResults::DB_UPDATE_FAILED, "#{e.inspect} #{e.message} #{e.backtrace.inspect}"
-      )
-    end
-    false
+    ActiveRecordUtils.with_transaction_and_rescue(handler_results) { yield }
   end
 
   # expects @incoming_version to be numeric
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -285,6 +285,10 @@ class PreservedObjectHandler
     end
 
     # TODO: do the check that'd set INVALID_CHECKSUM_STATUS
+    #  and actually, maybe we should either 1) trigger it async, or 2) break out the status
+    #  update, otherwise checksumming could get enclosed in a DB transaction, and that seems
+    #  like a real bad idea, cuz that transaction might be open a looooooong time.
+    # see https://github.com/sul-dlss/preservation_catalog/issues/612
 
     update_status(pres_copy, PreservedCopy::OK_STATUS)
   end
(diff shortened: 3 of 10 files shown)


jmartin-sul:
my hope is that @ndushay's concerns from #609 might be addressed by one of the tests in the second commit:
https://github.com/sul-dlss/preservation_catalog/pull/598/commits/e0e22c23e164223f1f358e9e83a4a3e60fe9e7fb#diff-a4c754f158c0a1ba32651fb058bcf56fR53

...which i think shows that the results were processed in the expected number of batches, in the expected order.

spec/lib/audit/catalog_to_moab_instance_spec.rb
```diff
@@ -109,6 +109,16 @@
...
         c2m.check_catalog_version
       end
+      context 'DB transaction handling' do
+        it 'on transaction failure, completes without raising error, removes PC_STATUS_CHANGED result code' do
+          allow(Moab::StorageObject).to receive(:new).with(druid, instance_of(String)).and_return(nil)
+          allow(pres_copy).to receive(:save!).and_raise(ActiveRecord::ConnectionTimeoutError)
+          c2m.check_catalog_version
+          expect(c2m.results.result_array).to include(a_hash_including(AuditResults::MOAB_NOT_FOUND))
+          expect(c2m.results.result_array).not_to include(a_hash_including(AuditResults::PC_STATUS_CHANGED))
+          expect(pres_copy.reload.status).not_to eq PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS
```
jmartin-sul: i felt a little strange setting an expectation that the status would be wrong, but this is what should happen.  i mostly wanted to make it clear in the test why the result code was getting removed, which i think is the more important part of the test.

spec/lib/audit/catalog_to_moab_instance_spec.rb
```diff
@@ -332,5 +342,25 @@
...
+        allow(moab).to receive(:current_version_id).and_return(2)
+      end
+
+      context 'DB transaction handling' do
+        it 'on transaction failure, completes without raising error, removes PC_STATUS_CHANGED result code' do
+          allow(pres_copy).to receive(:save!).and_raise(ActiveRecord::ConnectionTimeoutError)
+          c2m.check_catalog_version
+          expect(c2m.results.result_array).to include(a_hash_including(AuditResults::UNEXPECTED_VERSION))
+          expect(c2m.results.result_array).not_to include(a_hash_including(AuditResults::PC_STATUS_CHANGED))
+          expect(pres_copy.reload.status).not_to eq PreservedCopy::UNEXPECTED_VERSION_ON_STORAGE_STATUS
```
jmartin-sul: same hand-wringing about the status check on this test

.rubocop.yml
```diff
@@ -226,3 +226,7 @@ Style/StringLiterals:
 Style/SymbolArray:
   Exclude:
     - 'Rakefile' # because [:spec, :rubocop] isn't a big deal
+
+Style/TernaryParentheses:
+  Exclude:
+    - spec/lib/active_record_utils_spec.rb # ternary expression contains a little math, reads better with parens
```
ndushay: don't need to bother changing, but fine to shorten comment to `# reads better with parens`

jmartin-sul:
@ndushay, would you say it addresses your testing concerns from #573 and #609?  or is there more you'd like to see along those lines before those issues get closed?

ndushay:
It definitely addresses #573 and #606.  **Least effort forward is to merge** ( ... but we could also have added tests like that to existing code and avoided the indirection.)

(Spoken as one who has been whacking at too much indirection in a variety of code bases)

lib/audit/catalog_to_moab.rb
```diff
@@ -72,36 +72,43 @@ def check_catalog_version
     end
 
     unless online_moab_found?(druid, storage_dir)
-      update_status(PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS)
+      transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
```
SaravShah: 👍

app/services/preserved_object_handler.rb
```diff
@@ -285,6 +285,10 @@ def set_status_as_seen_on_disk(pres_copy, found_expected_version)
     end
 
     # TODO: do the check that'd set INVALID_CHECKSUM_STATUS
+    #  and actually, maybe we should either 1) trigger it async, or 2) break out the status
+    #  update, otherwise checksumming could get enclosed in a DB tx, and that seems like a real
+    #  bad idea, cuz that tx might be open a looooooong time.
```
SaravShah: tx?

SaravShah: oh ticket?

jmartin-sul: oh, uh, yeah, colloquialism for transaction, i've tried to avoid that outside of my own notes since it's not always clear (case in point), but i slipped here.

whoops, will correct and re-push...

Rakefile
```diff
@@ -174,7 +174,7 @@ task :cv_single_endpoint, [:storage_root, :profile] => [:environment] do |_t, ar
   end
   storage_root = args[:storage_root].to_sym
   if args[:profile] == 'profile'
-    puts "When done, check log/profile_cv_single_endpoint[TIMESTAMP] for profiling details"
+    puts "When done, check log/profile_CV_validate_disk[TIMESTAMP] for profiling details"
```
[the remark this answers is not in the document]
jmartin-sul: since i was already in there fixing a comment pointed out by sarav, i went ahead and fixed this too (and squashed it into the first commit).

.rubocop.yml
jmartin-sul: also fixed this (and the lack of line break) at the end

app/services/preserved_object_handler.rb
jmartin-sul: fixed the comment

jmartin-sul:
@SaravShah said:
> Would we want to have this be in lib/utilities/active_record_utils.rb?

i'd be fine with that.  follow on ticket?  or wait till we have more utils? (which hopefully we won't because stuff will fit more neatly into models or services, but seems plausible that there might be more utility method stuff later)

or i can fix in this PR, up to you.

i think i addressed all your other feedback, though.

SaravShah:
We can just wait till we have more utils. Merging.

PR #594, opened 2018-02-26
Code, tests, and fixtures for sig cat CV

tallenaz:
~~Marked as WIP since it is branched off of `cv-pairing`, and shouldn't get merged until #591 does.~~

Changes are in one commit: https://github.com/sul-dlss/preservation_catalog/commit/b4bdb71bbf31a04cd75344832e1de38e8f65aeb0

- Rubocop exclusions for class length of ChecksumValidator, and AbcSize of shameless green methods
- Adds `validate_signature_catalog` that calls some other private methods
- Adds pared down fixtures, which required touching up M2C specs
- Rspec for `validate_signature_catalog`, and integration-type specs

Closes #522

PR #600, opened 2018-03-01
Add MOAB_CHECKSUM_VALID code

### LINKED ISSUE #596: (CV) Create a success result code in AuditResults for ChecksumValidator
-  have an AuditResult for checksums passing validation
-  will be nice to have an explicit "it passed" -- not only to make this code clearer, but to make it easier for any ReST hookups and log perusing.

tallenaz:
for when a moab passes CV
  ~~also adds a skipped test as a reminder~~
  set up `validate_checksum` method and specs

wait till #594  is merge (still needs to rebase off master too)

  connects #596

### DIFF
--- a/app/services/audit_results.rb
+++ b/app/services/audit_results.rb
@@ -26,6 +26,7 @@ class AuditResults
   PC_PO_VERSION_MISMATCH = :pc_po_version_mismatch
   MOAB_NOT_FOUND = :moab_not_found
   MOAB_FILE_CHECKSUM_MISMATCH = :moab_file_checksum_mismatch
+  MOAB_CHECKSUM_VALID = :moab_checksum_valid
   FILE_NOT_IN_MOAB = :file_not_in_moab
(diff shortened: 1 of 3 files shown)


tallenaz:
Not sure either, but my first thought was that the existing `validate_checksum` method would use it, since I imagine we're going to use it to call the manifest inventory and signature catalog methods, but I'm speculating...

SaravShah:
Should we wait to merge this, and have this PR be the PR that implements it into manifestInventory (once PR#591 is merged ofcourse)????

tallenaz:
If we use `ChecksumValidator#checksum_validate` (I guess as a method to wrap `validate_manifest_inventories` and `validate_signature_catalog`), then would this new `MOAB_CHECKSUM_VALID` live there?

SaravShah:
i think this can actually be merged by itself? my PR # 606 ) is rebased off this

PR #605, opened 2018-03-03
compare data files against signature catalog entries

### LINKED ISSUE #604: (CV) for data files, compare against latest sigCat entries for FILE_NOT_IN_MANIFEST
We want behavior that generates a list of files present in any version's data dir, and for each file, see if it is in the latest version's signatureCatalog, and if not, report `FILE_NOT_IN_MANIFEST`. Presumably, this would live in `verify_signature_catalog`.

tallenaz:
I separated this step out into it's own ~~awkwardly named~~ method, ~~contrasting with~~ and wrapped it up inside `validate_signature_catalog` ~~which now also seems awkwardly named to me, so I'm wide open to discussing better names, and this is pretty much why I'm making it WIP for now.~~

This seemed like a quicker approach, because housing the functionality inside of `validate_signature_catalog` broke tests for other methods in the spec, and this way didn't.

I'm assuming that comparing checksums is the way to determine that a file on disk is "in" the signatureCatalog, but that may be controversial.

I know we'd rather not disable cops inside of code, but where I did, I felt justified, or just righteous indignation :)

I took some refactoring and styling opportunities I thought were worth the time.

I'd definitely like more than one pair of eyes on this, because of the naming issue, and how I set up `Find.find`, mostly. It's probably easier to review commit by commit.

Closes #604


tallenaz:
@jmartin-sul I went with your naming suggestions, I think they're an improvement both because they're shorter and helpfully less general. 

The new `Find.find` approach you suggested makes the method a lot shorter and allows us to get rid of the inline rubocop enable/disabling. 

Maybe the last thing to note is that I made the new methods that `validate_signature_catalog` wraps up private, and I kept the tests even though it seems like there's some consensus around not testing private methods. If it's controversial to keep then around, I'd want to argue for a "the more tests the better" approach.

I think this is ready for re-review.

app/services/checksum_validator.rb
```diff
@@ -47,7 +50,10 @@ def validate_signature_catalog
     checksum_results.report_results
   end
 
-  private
+  def flag_unexpected_data_content_files
```
ndushay: actually, this needs to check in both "data/content" and "/data/metadata".   manifestsInventory is for top level "manifests" directory, and signatureCatalog is for everything under the "data" directory, no?

(it's confusing because the word "manifest" is similar to the word "metadata" and I get them confused all the time)

I would rename to `flag_unexpected_data_files`

app/services/checksum_validator.rb
```diff
@@ -28,17 +28,20 @@ def validate_checksum
...
+  private
+
+  def validate_signature_catalog_listing
     begin
-      latest_signature_catalog_entries.each do |signature_catalog_entry|
-        validate_signature_catalog_entry(signature_catalog_entry)
-      end
+      latest_signature_catalog_entries.each { |entry| validate_signature_catalog_entry(entry) }
     rescue Errno::ENOENT
       checksum_results.add_result(AuditResults::MANIFEST_NOT_IN_MOAB, manifest_file_path: latest_signature_catalog_path)
```
ndushay: I think this is the wrong error code,no?   It's not a *manifest* that isn't in the Moab, it's the signature catalog file.  We may need to add some more error codes to AuditResults.   Let's not be too stingy with them.

spec/services/checksum_validator_spec.rb
```diff
@@ -203,24 +203,60 @@
...
+        cv.send(:flag_unexpected_data_content_files)
+      end
+    end
+  end
+
+  context '#validate_signature_catalog' do
+    let(:druid) { 'bj102hs9687' }
+    let(:cv) { described_class.new(druid, endpoint_name) }
+
+    it 'calls validate_signature_catalog_against_data_content_files' do
```
jmartin-sul: looks like the test description should get updated to match the renamed method

spec/services/checksum_validator_spec.rb
```diff
@@ -203,24 +203,60 @@
...
+      expect(cv).to receive(:validate_signature_catalog_listing)
+      cv.validate_signature_catalog
+    end
+
+    it 'calls validate_data_content_files_against_signature_catalog' do
```
jmartin-sul: ditto this one

app/services/checksum_validator.rb
```diff
@@ -140,4 +145,30 @@ def latest_signature_catalog_entries
   def latest_moab_version
     moab_storage_object.version_list.last
   end
+
+  def validate_against_signature_catalog(data_content_file)
+    absent_from_manifest_data = { file_path: data_content_file, manifest_file_path: latest_signature_catalog_path }
+    paths_from_manifest = latest_signature_catalog_entries.map { |entry| signature_catalog_entry_path(entry) }
```
ndushay: you call this method for each file;  it seems like you could have a method to memoize the paths from signature catalog. (and I would call it paths_from_signature_catalog)

app/services/checksum_validator.rb
```diff
@@ -140,4 +145,30 @@ def latest_signature_catalog_entries
   def latest_moab_version
     moab_storage_object.version_list.last
   end
+
+  def validate_against_signature_catalog(data_content_file)
+    absent_from_manifest_data = { file_path: data_content_file, manifest_file_path: latest_signature_catalog_path }
```
ndushay: avoid "manifest" in var names.  You're working with the signature_catalog, and with files somewhere under "data/", not with manifests.

app/services/checksum_validator.rb
```diff
@@ -140,4 +145,30 @@ def latest_signature_catalog_entries
...
+    paths_from_manifest = latest_signature_catalog_entries.map { |entry| signature_catalog_entry_path(entry) }
+    file_in_manifest = paths_from_manifest.any? { |entry| entry == data_content_file }
```
ndushay: "manifest" again.  How about "sig_cat_has_file?" as a separate method or something like that?   Would it make the code easier to grok?

app/services/checksum_validator.rb
```diff
@@ -140,4 +145,30 @@ def latest_signature_catalog_entries
...
+    checksum_results.add_result(AuditResults::FILE_NOT_IN_MANIFEST, absent_from_manifest_data) unless file_in_manifest
```
ndushay: again, I don't like the error code here;  file_not_in_signature_catalog would be better and rename the interpolation keys, too.

app/services/checksum_validator.rb
```diff
@@ -140,4 +145,30 @@ def latest_signature_catalog_entries
...
+  end
+
+  def data_content_files
```
ndushay: I think this method is the right idea.  Two concerns:

1.  data/content   vs.   data/metedata
2.  nested subdirs under content   (need a test for this?  we ended up allowing this in MV, yes???)

spec/services/checksum_validator_spec.rb
```diff
@@ -203,24 +203,60 @@
...
+        file_path = 'spec/fixtures/checksum_root01/moab_storage_trunk/zz/555/zz/5555/zz555zz5555/v0001/data/content/not_in_sigcat.txt'
+        manifest_file_path = 'spec/fixtures/checksum_root01/moab_storage_trunk/zz/555/zz/5555/zz555zz5555/v0002/manifests/signatureCatalog.xml'
+        results = instance_double(AuditResults, report_results: nil, check_name: nil)
+        allow(AuditResults).to receive(:new).and_return(results)
+        cv = described_class.new(druid, endpoint_name)
+        expect(results).to receive(:add_result).with(
+          AuditResults::FILE_NOT_IN_MANIFEST, file_path: file_path, manifest_file_path: manifest_file_path
+        )
+        cv.send(:flag_unexpected_data_content_files)
+      end
```
ndushay: I would like to see more tests here:
- for a metadata file
- for a file in a content subdir
- ...?

spec/services/checksum_validator_spec.rb
```diff
@@ -203,24 +203,60 @@
...
+        expect(cv).to receive(:validate_against_signature_catalog).with(file)
+      end
+      cv.send(:flag_unexpected_data_content_files)
     end
 
-    it 'returns error codes for a druid that fails validation' do
-      cv = described_class.new(druid, endpoint_name)
-      cv.validate_checksum
-      expect(cv.checksum_results.result_array.first).to have_key(:file_not_in_manifest)
+    it 'calls AuditResults.report_results' do
```
ndushay: I believe this shouldn't happen here but in #validate_checksum, no?

Also, not your "fault" but `validate_checksums` is more accurate as we don't validate only ONE.

spec/services/checksum_validator_spec.rb
```diff
@@ -203,24 +203,60 @@
...
-  context '#validate_checksum' do
+  context '#flag_unexpected_data_content_files' do
     let(:druid) { 'bj102hs9687' }
-    let(:cv) { described_class.new(druid, 'fixture_sr1') }
+    let(:cv) { described_class.new(druid, endpoint_name) }
 
-    it 'returns a positive result for a druid that passes validation' do
-      cv.validate_checksum
-      expect(cv.checksum_results.result_array.first).to have_key(:moab_checksum_valid)
+    it 'calls validate_against_signature_catalog on each of the data_content_files' do
```
ndushay: so this might need to show it's called:
- for a metadata file
- for a top level content file
- for a file in a content subdir 
- ...

spec/services/checksum_validator_spec.rb
```diff
@@ -203,24 +203,60 @@
...
+      files = ['spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/data/content/eric-smith-dissertation-augmented.pdf',
```
ndushay: pro tip:  put `'spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/data/` into a variable (maybe in a let statement) and then make it easy to read files:

```ruby
files = [
    "#{object_version_data_dir}metadata/asdfasf.xml",
    "#{object_version_data_dir}content/ugh.xml",
    "#{object_version_data_dir}content/subdir/foo.xml"
]
```
Then it's really easy to see what you are actually testing.

tallenaz:
Closed in favor of #610

Issue #595, opened 2018-02-27
(CV) lib/audit/checksum.rb implement .validate_disk (hook up to checksum_validator)


SaravShah:
- Checksum.validate_disk need to be coded to  connect to ChecksumValidator class

- "We could put in the timestamp code and so on analogous to M2C / C2M checks now??" - from Naomi

jmartin-sul:
> "We could put in the timestamp code and so on analogous to M2C / C2M checks now??"
  -- @ndushay 

maybe the stdout and log statements that say "started running"/"finished running"?

PR #606, opened 2018-03-06
Checksum validate disk

SaravShah:
- ~~Rebased off Tony's PR so that should get merged first?? (PR #600)~~

- Codes Checksum.validate_disk w/ RSpec tests
- ChecksumValidator takes in pres_copy instead of druid - refactor tests w/ that change
- renamed fixtures so we can use them as test fixtures 

connects #595

Multiple reviewers?


jmartin-sul:
re: the rubocop failure, i think you could just switch the extension on that file to something like `.txt` or `.md`, and i'd expect you'd be ok?

lib/audit/checksum.rb
```diff
@@ -3,12 +3,23 @@
...
-  # rubocop:disable Lint/UnusedMethodArgument
-  # rubocop:disable Style/EmptyMethod:
-  # TODO: implement this;  we begin with a placeholder
-
-  def self.validate_disk(endpoint_name)
+  def self.validate_disk(endpoint_name, limit=Settings.c2m_sql_limit)
+    start_msg = "#{Time.now.utc.iso8601} CV validate_disk starting for #{endpoint_name}"
+    puts start_msg
+    Rails.logger.info start_msg
+    all_processable_copies = PreservedCopy.by_endpoint_name(endpoint_name).fixity_check_expired
```
ndushay: total quibble:  "pres_copies_to_validate" instead of "all_processable_copies" ?
- pres_copies_to_validate
- pres_copies_to_process
- pcs_to_validate
- pcs_w_expired_fixity_check
...

spec/lib/audit/checksum_spec.rb
```diff
@@ -1,14 +1,48 @@
 require 'rails_helper'
 require_relative "../../../lib/audit/checksum.rb"
+require_relative '../../load_fixtures_helper.rb'
+
 # FIXME: remove this rubocop once we start writing tests
 # TODO: implement this;  we begin with a placeholder
```
ndushay: I imagine we can remove the above comments?  Look for any others marked TODO or FIXME in the spec and in the class?

spec/lib/audit/checksum_spec.rb
```diff
@@ -1,14 +1,48 @@
...
+      it 'creates an instance and calls #validate_checksum for every result when results are in a single batch' do
+        allow(ChecksumValidator).to receive(:new).and_return(cv_mock)
+        expect(cv_mock).to receive(:validate_checksum).exactly(3).times
+        described_class.validate_disk(endpoint_name, limit)
+      end
 
-  describe ".validate_disk" do
-    described_class.validate_disk('fixture_sr3')
-    skip 'we should figure out what they are and test them'
+      it 'creates an instance and calls #validate_checksum on everything in batches' do
```
ndushay: should we check the ordering here, since changing it to .for_each would mess up the ordering?

lib/audit/checksum.rb
```diff
@@ -3,12 +3,23 @@
 # Checksum validator code
 class Checksum
 
-  # FIXME: remove rubocop exception once we start filling these methods in
-  # rubocop:disable Lint/UnusedMethodArgument
-  # rubocop:disable Style/EmptyMethod:
-  # TODO: implement this;  we begin with a placeholder
-
-  def self.validate_disk(endpoint_name)
+  def self.validate_disk(endpoint_name, limit=Settings.c2m_sql_limit)
```
ndushay: add comment akin to C2M indicating that we must use .each and not .for_each to preserve ordering??

spec/lib/audit/checksum_spec.rb
SaravShah: @jmartin-sul maybe you can help me answer this. I used the same tests as C2m_class_spec.rb... and it doesn't look like c2m is checking order either?

lib/audit/checksum.rb
jmartin-sul: well, the relation as defined would actually return all processable copies, but the `.count` and `.limit` methods chained onto `all_processable_copies` keep it from actually doing that when the final query is constructed.

ndushay: my quibble is more with "processable" because technically, we could process any PreservedCopy object;  these are the ones we've SELECTED to be processed based on fixity expiration.

spec/lib/audit/checksum_spec.rb
ndushay: maybe there should be such a test in both places to ensure that future refactoring doesn't mess it up ...

lib/audit/checksum.rb
jmartin-sul: ah, yeah, i see, i figured it was the `all` you didn't like.

jmartin-sul: in that case, i like you're last one, `pcs_w_expired_fixity_check`

spec/lib/audit/checksum_spec.rb
jmartin-sul: so in the refactoring i'm doing where the `limit`/`each` construct gets factored out into a common method, i add some tests for that method, one of which checks ordering.  so maybe that will suffice, since i'll rebase on this PR and refactor a bit to pull out the `limit`/`each` that's being added inline for this work?

i'd agree that we should test ordering, though.  and indeed, it seems we missed testing that in the C2M specs (though we do make sure that all expected records are processed).

jmartin-sul: see https://github.com/sul-dlss/preservation_catalog/pull/598/files#diff-a4c754f158c0a1ba32651fb058bcf56fR44

SaravShah: Should I add a test for ordering, or do we think John's test that checks the ordering will suffice?

jmartin-sul: i'd vote for letting the PR i linked take care of it, and trusting that the tests for the commonly used method will be sufficient, but i'm open to testing ordering in both of the clients too, if e.g. @ndushay has a strong preference for that.

spec/lib/audit/moab_to_catalog_spec.rb
```diff
@@ -246,15 +246,16 @@
     end
 
     it 'drops PreservedCopies that correspond to the given endpoint' do
-      expect(PreservedCopy.count).to eq 9
+      expect(PreservedCopy.count).to eq 15
+      p PreservedObject.all.pluck(:druid).count
```
jmartin-sul: stray debug statement

spec/fixtures/checksum_root01/moab_storage_trunk/INVALID_MOABS_HERE.md
```diff
@@ -0,0 +1 @@
+# Invalid Moabs Here.
```
jmartin-sul: maybe add a bit of explanation (e.g. druid paths don't match druids in manifests).

or not, i don't feel strongly about this at all.

spec/lib/audit/checksum_spec.rb
```diff
@@ -1,14 +1,45 @@
...
+
+      it 'creates an instance and calls #validate_checksum on everything in batches' do
+        pcs_from_scope = PreservedCopy.by_endpoint_name(endpoint_name).fixity_check_expired
+        cv_list = pcs_from_scope.map do |pc|
+          ChecksumValidator.new(pc, endpoint_name)
+        end
+        cv_list.each do |cv|
+          allow(ChecksumValidator).to receive(:new).with(cv.preserved_copy, endpoint_name).and_return(cv)
+          expect(cv).to receive(:validate_checksum).exactly(1).times.and_call_original
+        end
```
ndushay: I have a thought about our batch testing:

```ruby
      it 'creates an instance and calls #validate_checksum on everything in batches' do
        pcs_from_scope = PreservedCopy.by_endpoint_name(endpoint_name).fixity_check_expired
        cv_list = pcs_from_scope.map do |pc|
          ChecksumValidator.new(pc, endpoint_name)
        end
        call_count = 0
        cv_list.each do |cv|
          allow(ChecksumValidator).to receive(:new).with(cv.preserved_copy, endpoint_name).and_return(cv)
          expect(cv).to receive(:validate_checksum).exactly(1).times.and_call_original
          call_count += 1
        end
        expect(described_class).to receive(:pcs_w_expired_fixity_check).exactly(2).times.and_call_original # 2 batches
        described_class.validate_disk(endpoint_name, 2)
         expect(call_count).to eq cv_list.size # each druid was called
      end
    end
```

If it works, we should put it on C2M as well -- that confirms that ALL expected ids were checked, in 2 batches.

See issue #573.  If you don't want to try this now, just copy this to that ticket.  If it works, maybe do a separate PR for C2M fix and close #573 with that PR?

Issue #587, opened 2018-02-22
(CV) rake task to run against a single druid at a time?  (or ReST call?  asynch?)


tallenaz:
Requested by @julianmorley -- looking ahead to archiving/replication, this will help him move moabs to the cloud.

ndushay:
rake task or ReST call?

Issue #263, opened 2017-11-03
(PC) update_version - what if there is more than one pres_copy for the endpoint?


ndushay:

atz:
With #798, the uniqueness of druid/version/endpoint in `PreservedCopy` is guaranteed.

PR #621, opened 2018-03-09
Rake task druid

SaravShah:
Update ReadME w/ usage info
Add Rake task
Add method in `checksum.rb` w/ Spec

connects #587

### DIFF
--- a/lib/audit/checksum.rb
+++ b/lib/audit/checksum.rb
@@ -54,16 +54,16 @@ class Checksum
     start_msg = "#{Time.now.utc.iso8601} CV validate_druid starting for #{druid}"
     puts start_msg
     Rails.logger.info start_msg
-    po = PreservedObject.find_by(druid: druid)
-    pres_copy = PreservedCopy.find_by(preserved_object: po)
-    endpoint_name = pres_copy.endpoint.endpoint_name
-    cv = ChecksumValidator.new(pres_copy, endpoint_name)
-    cv.validate_checksums
+    pres_copies = PreservedCopy.joins(:preserved_object).where(preserved_objects: { druid: druid })
+    Rails.logger.error("Found #{pres_copies.size} preserved copies.") if pres_copies.empty?
+    pres_copies.each do |pc|
+      endpoint_name = pc.endpoint.endpoint_name
+      cv = ChecksumValidator.new(pc, endpoint_name)
+      cv.validate_checksums
+    end
     end_msg = "#{Time.now.utc.iso8601} CV validate_druid ended for #{druid}"
     puts end_msg
     Rails.logger.info end_msg
-  rescue NoMethodError => e
-    Rails.logger.error("Undefined method #{e.inspect}")
   end
 
 end
--- a/lib/audit/checksum.rb
+++ b/lib/audit/checksum.rb
@@ -62,6 +62,8 @@ class Checksum
     end_msg = "#{Time.now.utc.iso8601} CV validate_druid ended for #{druid}"
     puts end_msg
     Rails.logger.info end_msg
+  rescue NoMethodError => e
+    Rails.logger.error("Undefined method #{e.inspect}")
   end
 
 end
--- a/lib/audit/checksum.rb
+++ b/lib/audit/checksum.rb
@@ -50,4 +50,18 @@ class Checksum
     profiler.print_results_flat('CV_checksum_validation_all_endpoints')
   end
 
+  def self.validate_druid(druid)
+    start_msg = "#{Time.now.utc.iso8601} CV validate_druid starting for #{druid}"
+    puts start_msg
+    Rails.logger.info start_msg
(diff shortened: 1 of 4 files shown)


SaravShah:
@ndushay  i will create 3 tickets for M2C, C2M, and MV. :)

lib/audit/checksum.rb
```diff
@@ -62,6 +62,8 @@ def self.validate_druid(druid)
     end_msg = "#{Time.now.utc.iso8601} CV validate_druid ended for #{druid}"
     puts end_msg
     Rails.logger.info end_msg
+  rescue NoMethodError => e
```
jmartin-sul: if we're taking this approach (assuming there's one entry in `preserved_copies` per druid, which is safe as long as we haven't gotten to archive copies yet), we should do a `find_by!` instead of a `find_by`, and then we get a nice active record error about the object not being found, instead of getting a `nil` back and inferring from the `NoMethodError`.

lib/audit/checksum.rb
```diff
@@ -50,4 +50,18 @@ def self.validate_disk_all_endpoints_profiled
     profiler.print_results_flat('CV_checksum_validation_all_endpoints')
   end
 
+  def self.validate_druid(druid)
+    start_msg = "#{Time.now.utc.iso8601} CV validate_druid starting for #{druid}"
+    puts start_msg
+    Rails.logger.info start_msg
+    po = PreservedObject.find_by(druid: druid)
+    pres_copy = PreservedCopy.find_by(preserved_object: po)
```
jmartin-sul: one thing worth noting is that this assumes that there's only one pres copy per druid.  which is true for now, but won't be true once we have archive copies.

one option:  query like this:
```ruby
PreservedCopy.joins(:preserved_object).where(preserved_objects: {druid: druid})
```

and then loop over the results.

for now, the behavior would be equivalent.  and when we do have archive copies, we'd have to filter them out, or have `ChecsumValidator#validate_checksums` skip them or otherwise do something sensible.

re: the query we were just talking about F2F, where you needed to get PCs for a given druid, i think you could turn the `joins(:preserved_object).where(preserved_objects: {druid: druid})` part of the above AR query into a scope on `PreservedCopy`, and chain it with the other scopes you're using.

ndushay: See #263 re: one pres_copy re druid-endpoint

lib/audit/checksum.rb
ndushay: See #263 re: one pres_copy re druid-endpoint

jmartin-sul: i'm totally fine with assuming one PC per druid for now since that's the way things currently are, and then revisiting under #263.  and i'd approve and merge this as-is if people thought the above comment was too nitpicky.  but using `find_by!` instead of `find_by` is more consistent with our other code (and is a thing we deliberately standardized on in POH), and i think `find_by!` would provide a more informative exception, so i figured it'd be worth proposing that if we did stick with the `find_by` approach in general.

jmartin-sul: (and then if we wanted to trap and log something, we could specifically trap `ActiveRecord::RecordNotFound`)

lib/audit/checksum.rb
```diff
@@ -54,16 +54,18 @@ def self.validate_druid(druid)
     start_msg = "#{Time.now.utc.iso8601} CV validate_druid starting for #{druid}"
     puts start_msg
     Rails.logger.info start_msg
-    po = PreservedObject.find_by(druid: druid)
-    pres_copy = PreservedCopy.find_by(preserved_object: po)
-    endpoint_name = pres_copy.endpoint.endpoint_name
-    cv = ChecksumValidator.new(pres_copy, endpoint_name)
-    cv.validate_checksums
+    # po = PreservedObject.find_by(druid: druid)
+    # pres_copy = PreservedCopy.find_by(preserved_object: po)
```
jmartin-sul: 👍 overall, but looks like a bit of debris from the old version is still there 🙂

SaravShah: you always catch me

lib/audit/checksum.rb
```diff
@@ -50,4 +50,20 @@ def self.validate_disk_all_endpoints_profiled
     profiler.print_results_flat('CV_checksum_validation_all_endpoints')
   end
 
+  def self.validate_druid(druid)
+    start_msg = "#{Time.now.utc.iso8601} CV validate_druid starting for #{druid}"
+    puts start_msg
+    Rails.logger.info start_msg
+    pres_copies = PreservedCopy.joins(:preserved_object).where(preserved_objects: { druid: druid })
+    Rails.logger.error("Found #{pres_copies.size} preserved copies.") if pres_copies.empty?
```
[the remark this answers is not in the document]
SaravShah: Will remove interpolation.

Issue #624, opened 2018-03-13
(C2M) Rake task to run single druid


SaravShah:
- Update README w/ usage info
- Add rake task that allows for druid input

ndushay:
a ReST call would be even better!  (separate ticket?)

SaravShah:
#629

atz:
This is just:
```ruby
Audit::CatalogToMoab.new(complete_moab, storage_dir).check_catalog_version
```
Anybody can do that on the console. `storage_dir` is a `MoabStorageRoot#storage_location` value.

Definitely do NOT need a rake task.  Since the REST call is separately ticketed (twice?), closing this one.

PR #632, opened 2018-03-15
[WIP] C2M check version via single druid

SaravShah:
- Did not add profiling for single druid check
- Updated README usage information
- Added scope ```by_druid```
- Added ```  def self.check_version_druid method ``` w/ RSpec tests

(Rebased off master)

connects #624

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -166,6 +166,19 @@ task :c2m_check_version_all_dirs, [:last_checked_b4_date, :profile] => [:environ
   $stdout.flush
 end
 
+desc "Fire off C2M for a single druid"
+task :c2m_check_druid, [:last_checked_b4_date, :druid] => [:environment] do |_t, args|
+  begin
+    last_checked = args[:last_checked_b4_date].to_s
+    druid = args[:druid].to_sym
+    CatalogToMoab.check_version_druid(last_checked, druid)
+    puts "#{Time.now.utc.iso8601} Catalog to Moab version check on all roots is done."
+  rescue TypeError, ArgumentError
+    p "You've entered an incorrect timestamp format #{last_checked}."
+    p "Please enter correct timestamp format (UTC) (2018-02-01T18:54:48Z)"
+  end
+end
+
 desc "Fire off checksum validation on a single storage root"
 task :cv_single_endpoint, [:storage_root, :profile] => [:environment] do |_t, args|
   unless args[:profile] == 'profile' || args[:profile].nil?
--- a/app/models/preserved_copy.rb
+++ b/app/models/preserved_copy.rb
@@ -40,6 +40,10 @@ class PreservedCopy < ApplicationRecord
     joins(:endpoint).where(endpoints: { storage_location: storage_dir })
   }
 
+  scope :by_druid, lambda { |druid|
+    joins(:preserved_object).where(preserved_objects: { druid: druid })
+  }
+
   scope :least_recent_version_audit, lambda { |last_checked_b4_date|
     last_checked_b4_date = normalize_date(last_checked_b4_date)
     where('last_version_audit IS NULL or last_version_audit < ?', last_checked_b4_date)
--- a/lib/audit/catalog_to_moab.rb
+++ b/lib/audit/catalog_to_moab.rb
@@ -49,6 +49,28 @@ class CatalogToMoab
     profiler.print_results_flat('C2M_check_version_all_dirs')
   end
 
+  def self.check_version_druid(last_checked_b4_date, druid, limit=Settings.c2m_sql_limit)
+    start_msg = "#{Time.now.utc.iso8601} C2M check_version starting for #{druid}"
+    puts start_msg
+    Rails.logger.info start_msg
+
+    # all_processable_copies is an AR relation; fine to run it with a .count or a .limit tacked on, but
+    # don't call .each directly on it and get the whole result set at once.
+    all_processable_copies =
+      PreservedCopy.least_recent_version_audit(last_checked_b4_date).by_druid(druid)
+    num_to_process = all_processable_copies.count
+    while num_to_process > 0
+      pcs = all_processable_copies.limit(limit)
+      pcs.each do |pc|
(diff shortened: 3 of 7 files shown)


app/models/preserved_copy.rb
```diff
@@ -40,6 +40,10 @@ class PreservedCopy < ApplicationRecord
     joins(:endpoint).where(endpoints: { storage_location: storage_dir })
   }
 
+  scope :by_druid, lambda { |druid|
+    joins(:preserved_object).where(preserved_objects: { druid: druid })
+  }
+
```
ndushay: Should there be tests for this scope?  Things like:

- no copies
- one copy
- multiple copies

Should this scope be used by CV check?

lib/audit/catalog_to_moab.rb
```diff
@@ -49,6 +49,28 @@ def self.check_version_all_dirs_profiled(last_checked_b4_date)
     profiler.print_results_flat('C2M_check_version_all_dirs')
   end
 
+  def self.check_version_druid(last_checked_b4_date, druid, limit=Settings.c2m_sql_limit)
```
ndushay: Copy-pasta brain freeze.

This seems like a lot of code for this.  Maybe you just need:

```
preserved_copies = PreservedCopy.least_recent_version_audit(last_checked_b4_date).by_druid(druid)
preserved_copies.each do |pc|
    c2m = CatalogToMoab.new(pc, pc.endpoint.storage_location)
    c2m.check_catalog_version
end
```

??    Since a single druid will only have ... 50 versions x 5 endpoints  preserved copies, max?

spec/lib/audit/catalog_to_moab_class_spec.rb
```diff
@@ -89,4 +89,46 @@
...
+
+  context '.check_version_druid' do
+    include_context 'fixture moabs in db'
+    let(:druid) { 'bj102hs9687' }
+    let(:subject) { described_class.check_version_druid(last_checked_version_b4_date, druid, limit) }
+
+    context 'when there are PreservedCopies to check' do
+      let(:c2m_mock) { instance_double(described_class) }
+
+      it 'creates an instance and calls #check_catalog_version for every result when results are in a single batch' do
```
ndushay: same copy pasta brain freeze.   Seems like you may want to affirm that you process each preserved copy associated with the druid.  batches ... not so much.

lib/audit/catalog_to_moab.rb
SaravShah: >.<

SaravShah: Right......whoops

lib/audit/checksum.rb
```diff
@@ -51,7 +51,7 @@ def self.validate_druid(druid)
     start_msg = "#{Time.now.utc.iso8601} CV validate_druid starting for #{druid}"
     puts start_msg
     Rails.logger.info start_msg
-    pres_copies = PreservedCopy.joins(:preserved_object).where(preserved_objects: { druid: druid })
+    pres_copies = PreservedCopy.by_druid(druid)
```
ndushay: OMG this is awesome.  Would you please make a ticket to look for other places in our code and tests where we could use this and simplify the code?

Rakefile
```diff
@@ -166,6 +166,19 @@ task :c2m_check_version_all_dirs, [:last_checked_b4_date, :profile] => [:environ
   $stdout.flush
 end
 
+desc "Fire off C2M for a single druid"
+task :c2m_check_druid, [:last_checked_b4_date, :druid] => [:environment] do |_t, args|
+  begin
+    last_checked = args[:last_checked_b4_date].to_s
+    druid = args[:druid].to_sym
+    CatalogToMoab.check_version_druid(last_checked, druid)
+    puts "#{Time.now.utc.iso8601} Catalog to Moab version check on all roots is done."
```
ndushay: if you're going to have these puts statements, there should be one before, and one after, and they should have language appropriate to running a single druid.  "... on all roots is done" is a bit off base.

spec/lib/audit/catalog_to_moab_class_spec.rb
```diff
@@ -89,4 +89,38 @@
...
+    context 'when there are PreservedCopies to check' do
+      let(:c2m_mock) { instance_double(described_class) }
+
+      it 'creates an instance and calls #check_catalog_version for every result' do
+        allow(described_class).to receive(:new).and_return(c2m_mock)
+        expect(c2m_mock).to receive(:check_catalog_version).exactly(1).times
+        described_class.check_version_druid(last_checked_version_b4_date, druid)
+      end
+
+      it 'creates an instance and calls #check_catalog_version if there are multiple druids' do
```
ndushay: uh ... I'm a little confused.  There should only be one druid, though there may be multiple preserved_copy objects ... so is this description off, or am I confused?

Assuming the former, this is the same situation as batching -- it would be useful to have some sort of counter and use expectations to show:  1.  there is more than one preserved_copy  2.  the number of times the method is called is equal to the number of preserved copies.

spec/models/preserved_copy_spec.rb
```diff
@@ -329,10 +330,43 @@
...
+                              status: status,
+                              size: 1
+                            }])
+    end
+
+    it 'returns the expected preserved copy given a druid' do
+      expect(PreservedCopy.by_druid('bj102hs9687').length).to eq 0
+      expect(PreservedCopy.by_druid('xx000xx0000').length).to eq 1
+    end
+    it 'returns the expected preserved copy given multiple druids' do
```
ndushay: again, is it not that there are multiple preserved_copy objects associated with a single druid?  Or am I confused?

lib/audit/checksum.rb
SaravShah: #644

atz:
Need a rebase here

atz:
I think this might be obsoleted by parallel work?

SaravShah:
seems like we don't need this check. closing for now.

Issue #109, opened 2017-09-20
(PC) ReST call for preserved_object/_copy CREATE methods


ndushay:
no yet clear if online Moab vs. archived object is an argument or a rest endpoint, as args may be slightly different.

See

https://docs.google.com/document/d/1_ejJE0Ve1gEezQWFnqJAsl1FI4rpREW9lSL0ZSbD62U

jmartin-sul:
would be used by both on VM inventory and by REST.

see #113

PR #643, opened 2018-03-29
[WIP] Create route ReST Calls (first pass)

SaravShah:
This is just a first pass PR to see what other devs think and if they have any thoughts on what I should change. 

`catalog_controller.rb` - `def create` method w/ the http_status return codes. Thought it would be useful if we returned useful http codes...I can change them to other http codes if someone doesn't like them.  I added these codes because I was getting a 204 (no content) return code whether success or failure of object creation. 

`audit_results.rb` - `contains_result_code?(code)` is used by `CatalogController#create` ( PR John put up I merged into my branch)

connects #109

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -3,36 +3,34 @@
 # to add an existing moab object to the catalog, or to update an entry for a moab object
 # that's already in the catalog.
 class CatalogController < ApplicationController
-  before_action :set_id
 
-  def add_preserved_object
-    PreservedObject.create!(
-      druid: @id, size: moab_size(@id), current_version: Stanford::StorageServices.current_version(@id)
-    )
-  end
+  attr_accessor :poh
 
-  def update_preserved_object
-    preserved_obj = PreservedObject.find_by(druid: @id)
-    preserved_obj.update_attributes(
-      druid: @id, size: moab_size(@id), current_version: Stanford::StorageServices.current_version(@id)
-    )
+  # POST /catalog
+  def create
+    druid = poh_params[:druid]
+    incoming_version = poh_params[:incoming_version].to_i
+    incoming_size = poh_params[:incoming_size].to_i
+    endpoint = Endpoint.find_by(endpoint_name: poh_params[:endpoint_name])
+    @poh = PreservedObjectHandler.new(druid, incoming_version, incoming_size, endpoint)
+    poh.create
+    status_code =
+      if poh.handler_results.contains_result_code?(:created_new_object)
+        :created # 201
+      elsif poh.handler_results.contains_result_code?(:db_obj_already_exists)
+        :conflict # 409
+      elsif poh.handler_results.contains_result_code?(:invalid_arguments)
+        :not_acceptable # 406
+      else
+        :internal_server_error # 500
+      end
+    render status: status_code, json: poh.handler_results.to_json
   end
 
   private
 
-  def set_id
-    @id = catalog_params[:id]
-    head(:unprocessable_entity) if @id.blank?
-  end
-
-  def catalog_params
-    params.permit(:id)
-  end
-
-  def moab_size(id)
-    # TODO: make this actually get the size once there's a method for that.
-    # for now, just using the id to quiet rubocop complaint.
-    # https://github.com/sul-dlss/moab-versioning/issues/21
-    42 || id
+  # strong params / whitelist params
+  def poh_params
+    params.permit(:druid, :incoming_version, :incoming_size, :endpoint_name)
   end
 end
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -1,46 +1,95 @@
 require 'rails_helper'
-
 RSpec.describe CatalogController, type: :controller do
-  describe 'POST #add_preserved_object' do
-    context 'with valid params' do
-      it 'saves the object with the expected field values' do
-        allow(Stanford::StorageServices).to receive(:current_version).and_return(6)
-        allow(subject).to receive(:moab_size).and_return(555)
-        id = 'ab123cd45678'
+  describe 'POST #create' do
+    let(:size) { 2342 }
+    let(:ver) { 3 }
+    let(:endpoint_name) { 'fixture_sr1' }
+    let(:druid) { 'bj102hs9687' }
 
-        allow(PreservedObject).to receive(:create!)
-        post :add_preserved_object, params: { id: id }
-        expect(PreservedObject).to have_received(:create!).with(druid: id, size: 555, current_version: 6)
+    context 'with valid params' do
+      let(:pres_copy) { PreservedCopy.first }
+      let(:pres_obj) { PreservedObject.first }
+
+      before do
+        post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, endpoint_name: endpoint_name }
+      end
+
+      it 'saves PreservedObject and PreservedCopy in db' do
+        po = PreservedObject.find_by(druid: druid)
+        pc = PreservedCopy.find_by(preserved_object: po)
+        expect(po).to be_an_instance_of PreservedObject
+        expect(pc).to be_an_instance_of PreservedCopy
+      end
+      it 'PreservedCopy and PreservedObject have correct attributes' do
+        expect(pres_copy.endpoint.endpoint_name).to eq endpoint_name
+        expect(pres_copy.version).to eq ver
+        expect(pres_copy.size).to eq size
+        expect(pres_obj.druid).to eq druid
+      end
+
+      it 'response contains create_new_object code ' do
+        exp_msg = [{ AuditResults::CREATED_NEW_OBJECT => "added object to db as it did not exist" }]
+        expect(response.body).to eq exp_msg.to_json
+      end
+
+      it 'returns a created response code' do
+        expect(response).to have_http_status(:created)
       end
     end
 
     context 'with invalid params' do
-      it 'returns the unprocessable entity http status code' do
-        post :add_preserved_object
-        expect(response).to have_http_status(422)
+      before do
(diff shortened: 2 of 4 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -3,36 +3,32 @@
...
-    preserved_obj = PreservedObject.find_by(druid: @id)
-    preserved_obj.update_attributes(
-      druid: @id, size: moab_size(@id), current_version: Stanford::StorageServices.current_version(@id)
-    )
+  # POST /catalog
+  def create
+    druid = poh_params[:druid]
+    incoming_version = poh_params[:incoming_version].to_i
+    incoming_size = poh_params[:incoming_size].to_i
+    endpoint = Endpoint.find(poh_params[:endpoint])
```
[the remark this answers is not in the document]
SaravShah: Ah good point. I wasn't sure what the robots would know, so I left it as endpoint. Changing it to endpoint_name will be straight forward though :)

spec/controllers/catalog_controller_spec.rb
```diff
@@ -1,46 +1,81 @@
...
+        expect(pres_obj.druid).to eq druid
+      end
 
-        allow(PreservedObject).to receive(:create!)
-        post :add_preserved_object, params: { id: id }
-        expect(PreservedObject).to have_received(:create!).with(druid: id, size: 555, current_version: 6)
+      it 'response contains create_new_object code ' do
+        code = AuditResults::CREATED_NEW_OBJECT
+        exp_msg = "added object to db as it did not exist"
+        post_response = response.request.env['action_controller.instance'].poh.handler_results.result_array
```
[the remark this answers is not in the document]
SaravShah: Are you saying that in my CatalogController#create method instead of 
```
if poh.handler_results.contains_result_code?(:created_new_object)
      render status: 201, plain: ":created"
    elsif poh.handler_results.contains_result_code?(:db_obj_already_exists)
      render status: 409, plain: ":conflict"
    elsif poh.handler_results.contains_result_code?(:invalid_arguments)
      render status: 406, plain: ":not_acceptable"
    else
      render status: 404, plain: ":not_found"
    end
```

you're saying i should do something like this:
```
if poh.handler_results.contains_result_code?(:created_new_object)
      render status: 201, plain: result_array.json
    elsif poh.handler_results.contains_result_code?(:db_obj_already_exists)
      render status: 409, plain: result_array.json
    elsif poh.handler_results.contains_result_code?(:invalid_arguments)
      render status: 406, plain: result_array.json
    else
      render status: 404, plain: result_array.json
    end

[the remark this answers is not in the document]
Comment (SaravShah) [reply spec/controllers/catalog_controller_spec.rb]: Also oops, I should have read your comment below...

File: app/controllers/catalog_controller.rb [hunk 3]
```diff
@@ -3,36 +3,32 @@
...
+    @poh.create
+    if @poh.handler_results.contains_result_code?(:created_new_object)
+      render status: 201, plain: ":created"
+    elsif @poh.handler_results.contains_result_code?(:db_obj_already_exists)
+      render status: 409, plain: ":conflict"
+    elsif @poh.handler_results.contains_result_code?(:invalid_arguments)
+      render status: 406, plain: ":not_acceptable"
+    else
+      render status: 404, plain: ":not_found"
+    end
```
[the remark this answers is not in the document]
SaravShah: I like it. and i can do `poh.handler_results` right since its an `attr_accessor` ?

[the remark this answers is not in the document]
jmartin-sul: yup, good point, `poh` should work

jmartin-sul:
oh, one other thing:  i have a simple `#to_json` implementation for the results here:
https://github.com/sul-dlss/preservation_catalog/pull/653
https://github.com/sul-dlss/preservation_catalog/pull/653/commits/99c76c716225f09a996f1829108d040ac93db96b

SaravShah:
@jmartin-sul just pushed up changed, should be ready for review :)

spec/controllers/catalog_controller_spec.rb
```diff
@@ -1,46 +1,97 @@
...
+        expect(pres_obj.druid).to eq druid
+      end
 
-        allow(PreservedObject).to receive(:create!)
-        post :add_preserved_object, params: { id: id }
-        expect(PreservedObject).to have_received(:create!).with(druid: id, size: 555, current_version: 6)
+      it 'response contains create_new_object code ' do
+        exp_msg = [{ AuditResults::CREATED_NEW_OBJECT => "added object to db as it did not exist" }]
+        post_response = response.body
+        expect(post_response).to eq exp_msg.to_json
```
jmartin-sul: i'd be inclined to do `expect(response.body)` directly without the temp var, but i don't have strong feelings and if you fine this style more readable, i'm totally fine with that.

SaravShah: i changed it in the other places too

SaravShah:
@jmartin-sul just pushed up w/ changes. Rdy for review :)

spec/controllers/catalog_controller_spec.rb
```diff
@@ -1,46 +1,92 @@
...
-        id = 'ab123cd45678'
+      let(:pres_copy) { PreservedCopy.first }
+      let(:pres_obj) { PreservedObject.first }
+
+      before do
+        post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, endpoint_name: endpoint_name }
+      end
+
+      it 'saves PreservedObject and PreservedCopy in db' do
+        expect(PreservedObject).to exist
```
ndushay: is this the class or the object?   should it be `PreservedObject.find(druid)` instead of `PreservedObject`?

spec/controllers/catalog_controller_spec.rb
```diff
@@ -1,46 +1,92 @@
...
+        expect(pres_copy.size).to eq size
+        expect(pres_obj.druid).to eq druid
+      end
 
-        allow(PreservedObject).to receive(:create!)
-        post :add_preserved_object, params: { id: id }
-        expect(PreservedObject).to have_received(:create!).with(druid: id, size: 555, current_version: 6)
+      it 'response contains create_new_object code ' do
+        exp_msg = [{ AuditResults::CREATED_NEW_OBJECT => "added object to db as it did not exist" }]
+        expect(response.body).to eq exp_msg.to_json
```
ndushay: does the response body contain the druid?  It might be nice.  Could be a separate ticket.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -1,46 +1,92 @@
...
     context 'with invalid params' do
-      it 'returns the unprocessable entity http status code' do
-        post :add_preserved_object
-        expect(response).to have_http_status(422)
+      before do
+        post :create, params: { druid: nil, incoming_version: ver, incoming_size: size, endpoint_name: endpoint_name }
+      end
+
+      it 'does not save PreservedObject or PreservedCopy in db' do
+        expect(PreservedObject).not_to exist
```
ndushay: similar question to above ... is this testing for the object or the class?

spec/controllers/catalog_controller_spec.rb
SaravShah: From RSpec docs:
```
  expect(obj).to exist # passes if obj.exist? or obj.exists?
```

When I try:

```
it 'saves PreservedObject and PreservedCopy in db' do
  pres_obj = PreservedObject.find_by(druid: druid)
  expect(pres_obj).to exist
end
```
it fails with:
```
Failure/Error: expect(pres_obj).to exist
       expected #<PreservedObject id: 1, druid: "bj102hs9687", current_version: 3, created_at: "2018-04-05 16:30:46", updated_at: "2018-04-05 16:30:46", preservation_policy_id: 1> to exist but it does not respond to either `exist?` or `exists?`
```

However; I did try in the console that when there are no objects in PreservedObject or PreservedCopy table, ` PreservedCopy.exists?` and `PreservedObject.exists?` both return false.

```
input> PreservedObject.all
=>   PreservedObject Load (3.1ms)  SELECT "preserved_objects".* FROM "preserved_objects"
[]
input> PreservedObject.exists?
  PreservedObject Exists (0.4ms)  SELECT  1 AS one FROM "preserved_objects" LIMIT $1  [["LIMIT", 1]]
=> false
```

and then when I create 1 object in the PreservedCopy and PreservedObject table I get:

```
input> PreservedObject.exists?
  PreservedObject Exists (0.4ms)  SELECT  1 AS one FROM "preserved_objects" LIMIT $1  [["LIMIT", 1]]
=> true
```

ndushay: er ... wouldn't you do something like

```
it 'saves PreservedObject and PreservedCopy in db' do
  pres_obj = PreservedObject.find_by(druid: druid)
  expect(pres_obj).to be_an_instance of PreservedObject
end
```

Basically, you are proving the specific object is there.  Does find_by return nil if it is NOT there?

Note that `PreservedObject.exist?` says nothing about WHICH object -- it just says there is an entry in the table.

SaravShah: Yes find_by returns `nil` if it is NOT there.

True - I'll change it to yours. I like it better.

spec/controllers/catalog_controller_spec.rb
SaravShah: ticket #656

spec/controllers/catalog_controller_spec.rb
```diff
@@ -1,46 +1,96 @@
...
-        expect(response).to have_http_status(422)
+      before do
+        post :create, params: { druid: nil, incoming_version: ver, incoming_size: size, endpoint_name: endpoint_name }
+      end
+
+      it 'does not save PreservedObject or PreservedCopy in db' do
+        po = PreservedObject.find_by(druid: druid)
+        pc = PreservedCopy.find_by(preserved_object: po)
+        expect(po).not_to be_an_instance_of PreservedObject
+        expect(pc).not_to be_an_instance_of PreservedCopy
```
ndushay: ... expect(xx).to be_nil?

PR #650, opened 2018-03-30
HostSettings specifies storage_roots by hostname for sharding

### LINKED ISSUE #617: (CV) multiple prod VMs need unique yet similar configs
Can we do this in single shared_config branch -- perhaps detect machine name and use that to indicate which storage roots, etc?

vs. having diff configs for each instance (16+ configs for prod ...)

Tony thinks there may be an example of this in existing shared_configs (not sure which branch)

See also #619

tallenaz:
- HostSettings class (with specs) gets storage_roots for the host configured in `storage_root_map`
- Replaces calls to `Settings.moab.storage_roots` with `HostSettings.storage_roots`
- Adds `storage_root_map` to settings

Paired with @atz and @jmartin-sul 

Early on, `HostSettings.storage_roots` was memoized, but then the test environment would use the laptop's hostname instead of the `default` host configured in `storage_root_map`.  No memoization worked with the existing tests, options for maintaining memoization seemed ugly, so I think of this as a shameless green PR.

I tested this out on a live deployment to the new prod server, and verified that `HostSettings.storage_roots` returns the same roots configured in `shared_configs`.

Maybe needs only one pair of eyes.

Closes #617

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -1,8 +1,15 @@
-moab:
-  # this is a list of named storage roots.  the names are used to seed the endpoints table.
-  storage_roots:
-    # name: '/location/on/disk'  # e.g.
+# the keys of storage_root_map represent deployed hosts and point to their storage_roots
+# these host values are just examples or fixtures, for real values, see sul-dlss/shared_configs
+storage_root_map:
+  default:
     fixture_sr1: 'spec/fixtures/storage_root01'
     fixture_sr2: 'spec/fixtures/storage_root02'
     fixture_sr3: 'spec/fixtures/checksum_root01'
+  preservation_catalog_prod:
+    fixture_sr1: 'spec/fixtures/storage_root01'
+    fixture_sr2: 'spec/fixtures/storage_root02'
+    fixture_sr3: 'spec/fixtures/checksum_root01'
+  preservation_catalog_prod_a:
+    fixture_sr1: 'spec/fixtures/storage_root01'
+    fixture_sr2: 'spec/fixtures/storage_root02'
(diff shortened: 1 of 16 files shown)


config/settings/development.yml
```diff
@@ -1,8 +1,13 @@
...
     fixture_sr1: 'spec/fixtures/storage_root01'
     fixture_sr2: 'spec/fixtures/storage_root02'
     fixture_sr3: 'spec/fixtures/checksum_root01'
+  preservation_catalog_prod:
+    fixture_sr1: 'spec/fixtures/storage_root01'
+    fixture_sr2: 'spec/fixtures/storage_root02'
+    fixture_sr3: 'spec/fixtures/checksum_root01'
+  preservation_catalog_prod_a:
+    fixture_sr1: 'spec/fixtures/storage_root01'
+    fixture_sr2: 'spec/fixtures/storage_root02'
```
SaravShah: Are we not adding the checksum_root01 b/c the objects in there arent valid?

tallenaz: The blocks for `preservation_catalog_prod` and `prod_a` here are just to show an example.  They could just as easily be like

```yaml
preservation_catalog_prod_a:
    fixture_sr1: 'first/storage/root/path'
    fixture_sr2: 'second/storage/root/path'
    fixture_sr3: 'third/storage/root/path'
```
Would it help to write them up like that, and maybe include a comment?

SaravShah: Ohhhh gotcha that makes sense. Yeah sure either or works for me :)

ndushay:
review in progress ...

Issue #655, opened 2018-04-04
(CV) need to set PreservedCopy::INVALID_CHECKSUM_STATUS for failed checksum validation


jmartin-sul:
we add the right AuditResults entries along the way, but we're not updating status when checksum validation fails

PR #664, opened 2018-04-09
update pres copy status

jmartin-sul:
* ~~easiest version, sets INVALID_CHECKSUM_STATUS on failure, unsets it on success (but doesn't set OK_STATUS, because we don't really know it's ok)~~
* see comments below for implementation summary
* tests

connects to #655

### DIFF
--- a/app/services/checksum_validator.rb
+++ b/app/services/checksum_validator.rb
@@ -1,7 +1,7 @@
 # code for validating Moab checksums
 class ChecksumValidator
 
-  attr_reader :checksum_results, :endpoint, :full_druid, :preserved_copy, :bare_druid
+  attr_reader :results, :endpoint, :full_druid, :preserved_copy, :bare_druid
 
   DATA = 'data'.freeze
   MANIFESTS = 'manifests'.freeze
@@ -18,7 +18,7 @@ class ChecksumValidator
     @preserved_copy = preserved_copy
     @bare_druid = preserved_copy.preserved_object.druid
     @endpoint = Endpoint.find_by(endpoint_name: endpoint_name)
-    @checksum_results = AuditResults.new(bare_druid, nil, endpoint)
+    @results = AuditResults.new(bare_druid, nil, endpoint)
     @full_druid = "druid:#{bare_druid}"
   end
 
@@ -26,8 +26,8 @@ class ChecksumValidator
     validate_manifest_inventories
     validate_signature_catalog
     preserved_copy.update!(last_checksum_validation: Time.current)
-    checksum_results.add_result(AuditResults::MOAB_CHECKSUM_VALID) if checksum_results.result_array.empty?
-    checksum_results.report_results
+    results.add_result(AuditResults::MOAB_CHECKSUM_VALID) if results.result_array.empty?
+    results.report_results
   end
 
   def validate_manifest_inventories
@@ -44,9 +44,9 @@ class ChecksumValidator
   def validate_signature_catalog_listing
     latest_signature_catalog_entries.each { |entry| validate_signature_catalog_entry(entry) }
   rescue Errno::ENOENT
-    checksum_results.add_result(AuditResults::SIGNATURE_CATALOG_NOT_IN_MOAB, signature_catalog_path: latest_signature_catalog_path)
+    results.add_result(AuditResults::SIGNATURE_CATALOG_NOT_IN_MOAB, signature_catalog_path: latest_signature_catalog_path)
   rescue Nokogiri::XML::SyntaxError
-    checksum_results.add_result(AuditResults::INVALID_MANIFEST, manifest_file_path: latest_signature_catalog_path)
+    results.add_result(AuditResults::INVALID_MANIFEST, manifest_file_path: latest_signature_catalog_path)
   end
 
   def flag_unexpected_data_files
@@ -60,9 +60,9 @@ class ChecksumValidator
       verification_result = moab_version.verify_manifest_inventory
       parse_verification_subentities(verification_result)
     rescue Nokogiri::XML::SyntaxError
-      checksum_results.add_result(AuditResults::INVALID_MANIFEST, manifest_file_path: manifest_file_path)
+      results.add_result(AuditResults::INVALID_MANIFEST, manifest_file_path: manifest_file_path)
     rescue Errno::ENOENT
-      checksum_results.add_result(AuditResults::MANIFEST_NOT_IN_MOAB, manifest_file_path: manifest_file_path)
+      results.add_result(AuditResults::MANIFEST_NOT_IN_MOAB, manifest_file_path: manifest_file_path)
     end
   end
 
@@ -85,7 +85,7 @@ class ChecksumValidator
         file_path: "#{subentity.details['other']}/#{details['basis_path']}",
         version: subentity.details['basis']
       }
-      checksum_results.add_result(AuditResults::MOAB_FILE_CHECKSUM_MISMATCH, mismatch_error_data)
+      results.add_result(AuditResults::MOAB_FILE_CHECKSUM_MISMATCH, mismatch_error_data)
     end
   end
 
@@ -95,7 +95,7 @@ class ChecksumValidator
         file_path: "#{subentity.details['other']}/#{details['other_path']}",
         manifest_file_path: "#{subentity.details['other']}/#{MANIFESTS_XML}"
       }
-      checksum_results.add_result(AuditResults::FILE_NOT_IN_MANIFEST, absent_from_manifest_data)
+      results.add_result(AuditResults::FILE_NOT_IN_MANIFEST, absent_from_manifest_data)
     end
   end
 
@@ -105,19 +105,19 @@ class ChecksumValidator
         manifest_file_path: "#{subentity.details['other']}/#{MANIFESTS_XML}",
         file_path: "#{subentity.details['other']}/#{details['basis_path']}"
       }
-      checksum_results.add_result(AuditResults::FILE_NOT_IN_MOAB, absent_from_moab_data)
+      results.add_result(AuditResults::FILE_NOT_IN_MOAB, absent_from_moab_data)
     end
   end
 
   def validate_signature_catalog_entry(entry)
     unless entry.signature.eql?(calculated_signature(signature_catalog_entry_path(entry)))
       mismatch_error_data = { file_path: signature_catalog_entry_path(entry), version: entry.version_id }
-      checksum_results.add_result(AuditResults::MOAB_FILE_CHECKSUM_MISMATCH, mismatch_error_data)
+      results.add_result(AuditResults::MOAB_FILE_CHECKSUM_MISMATCH, mismatch_error_data)
     end
   rescue Errno::ENOENT
     absent_from_moab_data = { manifest_file_path: latest_signature_catalog_path,
                               file_path: signature_catalog_entry_path(entry) }
-    checksum_results.add_result(AuditResults::FILE_NOT_IN_MOAB, absent_from_moab_data)
+    results.add_result(AuditResults::FILE_NOT_IN_MOAB, absent_from_moab_data)
   end
 
   def moab_storage_object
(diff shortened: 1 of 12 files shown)


app/services/checksum_validator.rb
```diff
@@ -26,7 +26,13 @@ def validate_checksums
     validate_manifest_inventories
     validate_signature_catalog
     preserved_copy.update!(last_checksum_validation: Time.current)
-    checksum_results.add_result(AuditResults::MOAB_CHECKSUM_VALID) if checksum_results.result_array.empty?
+    if checksum_results.result_array.empty?
+      checksum_results.add_result(AuditResults::MOAB_CHECKSUM_VALID)
+      # we don't know enough to set it to OK_STATUS, but we can unset INVALID_CHECKSUM_STATUS
```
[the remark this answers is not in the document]
atz: Yeah, this is a modeling problem.  We are talking about more than one aspect or characteristic and then pretending the status is a single unified thing.  Fine to punt for now, but something to think about.

app/services/checksum_validator.rb
```diff
@@ -26,7 +26,15 @@ def validate_checksums
     validate_manifest_inventories
     validate_signature_catalog
     preserved_copy.update!(last_checksum_validation: Time.current)
-    checksum_results.add_result(AuditResults::MOAB_CHECKSUM_VALID) if checksum_results.result_array.empty?
+    if checksum_results.result_array.empty?
+      checksum_results.add_result(AuditResults::MOAB_CHECKSUM_VALID)
+      # we don't know enough to set it to OK_STATUS, but we can unset INVALID_CHECKSUM_STATUS and ONLINE_MOAB_NOT_FOUND_STATUS
+      if [PreservedCopy::INVALID_CHECKSUM_STATUS, PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS].include?(preserved_copy.status)
+        update_status(PreservedCopy::VALIDITY_UNKNOWN_STATUS)
```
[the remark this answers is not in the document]
ndushay: this seems like a story-time discussion ...   your question is a good one and we should answer it properly.

jmartin-sul:
decision from story time:
* each validation check (M2C, C2M, CV) will look at status first for each pres copy it looks at.
  * if the status is OK, VALIDITY_UNKNOWN, or any status in the specific check's purview (e.g. INVALID_CHECKSUM for CV, INVALID_MOAB for MV, ONLINE_MOAB_NOT_FOUND for any of them), the check will proceed.
    * if the status was OK, and the validation that's being run finds no errors, the status will stay OK.
    * if the status was something else in the purview of the check, and the validation that's being run finds no errors, the status will be set back to OK.
    * if the status was something outside of the specific check's purview (e.g. INVALID_CHECKSUM for the M2C check), the check will stop, notify (log? honeybadger alert?), and leave status unaltered.  ("stop, drop, and log").  the assumption is that an alert will trigger manual remediation.

when a person manually remediates the object with a bad status, they should run all validations on the pres copy, starting with the validation for the known bad status.  the validations will then set the pres copy status back to OK.

so basically, put guard clauses in the current checks, such that the guard clause halts further checking when the check encounters a status isn't in that check's purview.  alert.  rely on manual remediation, and less self-healing for catalog entries.

SaravShah:
[Uploading IMG_20180411_143953.jpg...]

SaravShah:
img_20180411_143953

atz:
I believe the design questions here are unblocked by the consensus we had in meeting.

app/services/checksum_validator.rb
```diff
@@ -1,7 +1,13 @@
+require 'moab_validation_handler.rb'
+
 # code for validating Moab checksums
 class ChecksumValidator
+  include ::MoabValidationHandler
+
+  attr_reader :results, :endpoint, :full_druid, :preserved_copy, :bare_druid
 
-  attr_reader :checksum_results, :endpoint, :full_druid, :preserved_copy, :bare_druid
+  alias druid bare_druid
```
tallenaz: I didn't know about alias, so I did some reading on it and see that some folks advocate for `alias_method` over `alias`. For ex. (https://blog.bigbinary.com/2012/01/08/alias-vs-alias-method.html). Do you think there's any value in switching over?

spec/services/checksum_validator_spec.rb
```diff
@@ -208,27 +208,213 @@
...
+      ].each do |initial_status|
+        it "sets status to OK_STATUS if it was previously #{initial_status}" do
+          pres_copy.status = initial_status
+          pres_copy.save!
+          expect { cv.validate_checksums }.to change { pres_copy.status }.to PreservedCopy::OK_STATUS
+          expect(pres_copy.reload.status).to eq PreservedCopy::OK_STATUS
+        end
+      end
+
+      it "leaves status as OK_STATUS as-is" do
```
tallenaz: Nitpicky, but this reads to me "leaves status of OK_STATUS as-is", but not sure if that was your intent.

spec/services/checksum_validator_spec.rb
```diff
@@ -208,27 +208,213 @@
...
-        expect(cv.checksum_results.result_array.first).to have_key(:file_not_in_manifest)
+        expect(cv.results.result_array.first).to have_key(:file_not_in_manifest)
+      end
+
+      [
+        PreservedCopy::OK_STATUS,
+        PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS,
+        PreservedCopy::INVALID_MOAB_STATUS,
+        PreservedCopy::UNEXPECTED_VERSION_ON_STORAGE_STATUS,
+        PreservedCopy::VALIDITY_UNKNOWN_STATUS
```
tallenaz: I see we use this list at least twice. Do you think there's any value in factoring out a common `let` or something similar? I'm not convinced myself, but just saw the opportunity.

spec/services/checksum_validator_spec.rb
```diff
@@ -208,27 +208,213 @@
...
+        PreservedCopy::ONLINE_MOAB_NOT_FOUND_STATUS,
+        PreservedCopy::INVALID_MOAB_STATUS,
+        PreservedCopy::UNEXPECTED_VERSION_ON_STORAGE_STATUS,
+        PreservedCopy::INVALID_CHECKSUM_STATUS,
+        PreservedCopy::VALIDITY_UNKNOWN_STATUS
+      ].each do |initial_status|
+        it "sets status to OK_STATUS if it was previously #{initial_status}" do
+          pres_copy.status = initial_status
+          pres_copy.save!
+          expect { cv.validate_checksums }.to change { pres_copy.status }.to PreservedCopy::OK_STATUS
```
tallenaz: Cool, I didn't know about the `change` matcher.

app/services/checksum_validator.rb
ndushay: I don't have strong feelings either way.  Lowest effort - leave it as it is?

jmartin-sul: i went and implemented this suggestion, switching over to `alias_method`, but then rubocop complained that i should prefer alias in a class body (i guess as opposed to in a method call, as in the example link):
```
Style/Alias: Use alias instead of alias_method in a class body.
```

i'm not inclined to make a class method to do the aliasing as in the example in the above link, just because it seems a little convoluted.  though, since we're using inheritance of a sort here, the argument in that link is sorta convincing.

@atz has good decomposition and OO instincts and is more of a rubyist than me, so maybe he has an opinion?

in the interest of getting in what's hopefully an improvement, i'm inclined to ticket this discussion as follow-on work, unless joe weighs in with clear explanation as to why we should definitely go one way or the other.

spec/services/checksum_validator_spec.rb
jmartin-sul: good catch, definitely a typo, artifact of repeated editing.

spec/services/checksum_validator_spec.rb
jmartin-sul: yeah, totally, i had similar thoughts about that and discomfort with the repetition.

some of the lists differ in that many are 5 of the 6 possible statuses, but i'm sure some are repeated.  what might be nice is if pres copy provided a set of all possible statuses, and then we could just subtract from that as needed (e.g. `PreservedCopy.all_statuses - PreservedCopy::OK_STATUS` gets you the list you care about).

@atz was also just a bit uncomfortable with this test construction (declaring examples inside of a loop), because it can make the failed test output a bit harder to read.  not terrible, but a shared example would likely be cleaner and more idiomatic.

what if we generate another tech debt follow-on ticket or two from this thread?

spec/services/checksum_validator_spec.rb
jmartin-sul: yeah, i'd seen that in the past and forgotten about it, and was reminded of its existence when pairing w/ @atz

app/services/checksum_validator.rb
jmartin-sul: filed #686

spec/services/checksum_validator_spec.rb
jmartin-sul: ok, filed https://github.com/sul-dlss/preservation_catalog/issues/687

jmartin-sul:
i believe i've addressed all of @tallenaz's feedback, in one case by making a small fix and re-pushing, in the other cases by filing issues for follow on work.

spec/services/checksum_validator_spec.rb
jmartin-sul: fixed and repushed

app/services/moab_validation_handler.rb
```diff
@@ -0,0 +1,79 @@
...
+      update_status(PreservedCopy::INVALID_MOAB_STATUS)
+      return
+    end
+
+    unless found_expected_version
+      update_status(PreservedCopy::UNEXPECTED_VERSION_ON_STORAGE_STATUS)
+      return
+    end
+
+    # TODO: this still isn't quite honest in the cases where checksum validation hasn't been performed
```
ndushay: have we ticketed this missing work yet?

jmartin-sul:
the branch with the start of the follow on work for making `set_status_as_seen_on_disk` more truthful (as per https://github.com/sul-dlss/preservation_catalog/pull/664#issuecomment-382200890) is here:  https://github.com/sul-dlss/preservation_catalog/compare/cv-fail-upd-pc-status...cv-fail-upd-pc-status--honest-set_status_as_seen_on_disk?expand=1

the part of the linked comment that's relevant to that branch is:
> one downside to this approach:  as it is, if M2C or C2M run, and they hit something with `VALIDITY_UNKNOWN_STATUS`, they will set it to `OK_STATUS` if their validations pass.  this is a logical consequence of the approach we discussed last week.  but it is less than honest, because, for example, a moab which would fail checksum validation, but not structural moab validation, will be set to `OK_STATUS` incorrectly for the time between the first M2C validation execution and the first checksum validation execution.  not the end of the world, but less than desirable behavior.  i have a branch based on this branch which i hope will be able to address that issue with a minimum of further code churn (and which leverages the refactoring in this PR nicely, IMHO).  but i'm saving that as a follow-on PR, in order to hopefully get this reviewed and merged, since it's dragged on a couple days longer than i'd initially anticipated.

app/services/preserved_object_handler.rb
```diff
@@ -163,36 +164,24 @@ def update_version
...
-      end
+  def pres_object
+    @pres_object ||= PreservedObject.find_by!(druid: druid)
   end
 
-  # TODO: duplicate of method in catalog_to_moab - extract superclass or moab wrapper class??
-  def ran_moab_validation?
-    @ran_moab_validation ||= false
+  def pres_copy
+    # FIXME: what if there is more than one associated pres_copy?
```
ndushay: same: have we ticketed this missing work yet?

app/services/moab_validation_handler.rb
jmartin-sul: as of a few moments ago:  https://github.com/sul-dlss/preservation_catalog/issues/690

app/services/preserved_object_handler.rb
jmartin-sul: that's an old fixme that i just carried over, but just created a ticket:  https://github.com/sul-dlss/preservation_catalog/issues/691

PR #666, opened 2018-04-11
Selectable checksum test

### LINKED ISSUE #663: (CV) Use Moab::Versioning config to select checksums (md5 only)
When https://github.com/sul-dlss/moab-versioning/issues/142 is in a released version of `Moab::Versioning`, bump to include it and set checksum selection accordingly.

SaravShah:
1st Commit is some RSpec refactoring that was done during code club w/ @atz. I can separate this out into another commit if wanted.

2nd Commit contains moab-versioning update and adding the configurable checksum w/ RSpec tests.
 ( I can move the whole`context 'checksums are configurable' do..` bock into the `context #validate_checkums' do` block if wanted as well. )

ad hoc paired w/ @jmartin-sul  for 2nd commit.

closes #663

### DIFF
--- a/spec/services/checksum_validator_spec.rb
+++ b/spec/services/checksum_validator_spec.rb
@@ -304,4 +304,29 @@ RSpec.describe ChecksumValidator do
       end
     end
   end
+
+  context 'checksums are configurable' do
+    it 'all three checksums at once' do
+      allow(Moab::Config).to receive(:checksum_algos).and_return(%i[md5 sha1 sha256])
+      expect(Digest::MD5).to receive(:new).and_call_original.at_least(:once)
+      expect(Digest::SHA1).to receive(:new).and_call_original.at_least(:once)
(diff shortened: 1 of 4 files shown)


spec/services/checksum_validator_spec.rb
```diff
@@ -3,19 +3,19 @@
 
 RSpec.describe ChecksumValidator do
   include_context 'fixture moabs in db'
+  let(:druid) { 'zz102hs9687' } # most common druid
```
ndushay: "most common druid" is apparently a "note of interest" but not actually meaningful.  Tee hee.

Oh wait - I see -- this is trying to say WHY you put zz102hs9687 here.  I would omit the comment, but it's not  important.

SaravShah: i'll remove it! - repushed now

PR #668, opened 2018-04-12
Introduce okcomputer, 2 types of checks and tests to match

### LINKED ISSUE #574: add ok_computer to preservation catalog
(high priority in that this should happen before we go to prod)

we should probably have some status checks that nagios can look at.  what i can think of to start:
* are the service disks available?
* is postgres available?
* is workflow services available?

bonus stuff:
* figure out something about current audit throughput?  alert if below some threshold?  is that overkill?

atz:
- RSpec matchers from okcomputer
- use `spec/support` dir for add'l test scaffolding
- Amend activerecord scope to accept `ActiveSupport::TimeWithZone` objects, which are of course compatible with ActiveRecord querying. These are returned by statements like `2.days.ago`.

Fixes #574

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,7 @@ gem 'config' # Settings to manage configs on different instances
 gem 'faraday' # ReST calls
 gem 'honeybadger' # for error reporting / tracking / notifications
 gem 'jbuilder', '~> 2.5' # Build JSON APIs with ease.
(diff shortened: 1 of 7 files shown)


jmartin-sul:
pretty straight-forward, what's there seems like an improvement, so i'll merge and then if there are suggestions, we can file follow-on tickets.

PR #699, opened 2018-04-23
Set up resque and capistrano for stage worker machine

tallenaz:


config/routes.rb
```diff
@@ -1,6 +1,10 @@
+require 'resque/server'
+
 Rails.application.routes.draw do
   resources :catalog, param: :druid, only: %i[create update]
   resources :moab_storage, only: %i[index show]
 
+  mount Resque::Server.new, at: "/resque"
```
ndushay: do you want to make this conditional?  like behind a feature flag?  Or does it not matter if we add this all the time.

config/deploy/stage.rb
```diff
@@ -1,7 +1,10 @@
 server 'preservation-catalog-stage.stanford.edu', user: 'pres', roles: %w[app db web]
+server 'preservation-worker-stage.stanford.edu', user: 'pres', roles: %w[app db web]
 
 Capistrano::OneTimeKey.generate_one_time_key!
 set :rails_env, 'production'
 set :bundle_without, 'deploy test'
 set :deploy_to, '/opt/app/pres/preservation_catalog'
-append :linked_files, "config/newrelic.yml"
+
+# uncomment if/when we turn newrelic on
```
ndushay: 👍

config/routes.rb
tallenaz: I found one way to do this is with Rails route constraints, like:

`mount Resque::Server.new, at: '/resque', constraints: { host: /worker-machine-name-here/ }`

tallenaz:
Closing this in favor of an upcoming fresh PR. From there, I'll link back to this.

PR #700, opened 2018-04-24
Diagram stubs

### LINKED ISSUE #674: (R) explicit API for s3 actions?
Current notion is that there will be an archive service class in pres_cat that will ultimately implement:
- Exists?  (sends druid, version) (returns metadata)
- Upload? (sends file, MD5, etc)  (returns MD5)

not currently critical path (i.e. postpone impl):
- Delete
- Download

### LINKED ISSUE #694: (R) Scaffold Replication queues/workers
Based on diagram:
https://www.lucidchart.com/documents/view/45ded815-9fec-46fa-b935-1e26f3819f0e

Stub in basic workers and queue namespaces, tests.  Don't have to fully integrate right now or even implement each worker's logic, just get placeholder pieces in with the relationships to each other.

atz:
Implements the skeleton of our flow diagram:
   https://www.lucidchart.com/documents/edit/45ded815-9fec-46fa-b935-1e26f3819f0e/0

Fixes #694


app/jobs/s3_endpoint_delivery_job.rb
```diff
@@ -0,0 +1,17 @@
...
+# Upload zip if needed.
+# Notify ResultsRecorderJob.
+class S3EndpointDeliveryJob < EndpointDeliveryBase
+  queue_as :s3_enpoint_delivery
+  # note: EndpointDeliveryBase gives us `zip`
+
+  # @param [String] druid
+  # @param [Integer] version
+  def perform(druid, version)
+    ResultsRecorderJob.perform_later(druid, version, 's3', '12345ABC') # value will be from zip.checksum
```
jmartin-sul: i think i'd prefer to leave something so example-like commented out.

_edit_ ok, after getting to the tests, i can see how this makes the tests easier for now, fine with this as a start.

spec/jobs/integration_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  let(:checksum) { '12345ABC' }
+  let(:hash) do
+    { druid: druid, version: version, endpoint: endpoint, checksum: checksum }
+  end
+
+  it 'gets from zipmaker queue to replication result message' do
+    ActiveJob::Base.queue_adapter = :inline
+    expect(PlexerJob).to receive(:perform_later).with(druid, version).and_call_original
+    expect(S3EndpointDeliveryJob).to receive(:perform_later).with(druid, version).and_call_original
+    # other enpoints as added...
```
jmartin-sul: maybe a `TODO` here, if only to keep us from leaving the stray comment in once it's not needed?

spec/jobs/plexer_job_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+describe PlexerJob, type: :job do
+  let(:druid) { 'bj102hs9687' }
+  let(:version) { 1 }
+
+  it 'descends from ApplicationJob' do
+    expect(described_class.new).to be_an(ApplicationJob)
+  end
+
+  # future: does some replication logic
```
jmartin-sul: `TODO` just to make it more conventional/searchable?

atz: sure

atz: I didn't conceive of there being a large gap of time between this PR and the uptake of future work such that it would be easy to "miss" that the pieces are still skeletal.

atz:
The scope of this PR was to implement the diagram that already had established broad consensus (as a first pass), so until there is some other diagram to implement that has more consensus, I personally don't have clarity to implement.

The *purpose* was to unblock parallel work such that it would not be in conflict.  Without the class names, queue names and signatures established ahead of time, parts of parallel work will block each other or fail to integrate.  So the idea of _only_ doing zipmaker does not address the fundamental concern I had going in, since parallel work would not be unblocked.

PR #701, opened 2018-04-25
Logger chain

atz:
Previously, code was calling both `Rails.logger` (to file) and `puts` (to `STDOUT`) with the same messages.  I consolidated those lines by having a single logger that outputs to both.

This cleans up:
- the code
- our test output, by keeping the `puts` statements from firing during `rspec` or CI

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -17,5 +17,12 @@ module PreservationCatalog
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+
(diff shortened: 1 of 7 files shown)


config/application.rb
```diff
@@ -17,5 +17,12 @@ class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+
+    # Double-output logging, both to Rails.logger and STDOUT.  Helps avoid puts statements.
```
SaravShah: i think you could remove "helps avoid puts statements" and just leave that in the PR description

PR #716, opened 2018-05-02
Host settings robustness

atz:
Add baseline empty storage_root_map to Settings.yml and provide failover.

Because the lookup is based on system name, we should anticipate that it will be invoked on systems not yet present in the configuration (even if that is a dev system). Returning an empty hash is preferable to raising an uninformative `NoMethodError`.

### DIFF
--- a/app/models/host_settings.rb
+++ b/app/models/host_settings.rb
@@ -1,8 +1,8 @@
 # Lets us consolidate shared_configs for almost identical machines
 class HostSettings
-  # @return [Config::Options] key/value pairs of name: file_path
+  # @return [Config::Options, Hash] key/value pairs of name: file_path
   def self.storage_roots
-    Settings.storage_root_map[storage_root_lookup_name]
+    Settings.storage_root_map[storage_root_lookup_name] || {}
   end
 
   # @return [String] hostname depending on environment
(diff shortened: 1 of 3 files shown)


jmartin-sul:
@atz... on further thought, is it worth maybe a honeybadger error when the lookup fails to find something in prod?  or do you think that'd be overkill?  i guess i'm worried about mis-configuration going unnoticed.

jmartin-sul:
discussed F2F, joe pointed out that it's nice to be able to easily run as any environment on a dev laptop, including prod, even if it's just making rails think it's the prod environment, without connecting to other real prod services or databases.

will merge this PR, and file a follow on ticket for alerting on possible mis-configuration, since neither of us saw an obviously right or low-effort way to deal with that situation.

jmartin-sul:
also worth noting, this PR was the result of joe trying to debug an issue locally after it only surfaced on prod (the behavior here prevented running as `RAILS_ENV=prod` on a laptop).

atz:
I agree with the possibility of adding checking with a `raise` for an environment where we want it (i.e. prod), but it should be in an initializer or `production.rb`, not in the `HostSettings` class.  And we shouldn't rely on an unexpected `NoMethodError` as our mode of protection.

jmartin-sul:
filed #717

PR #718, opened 2018-05-02
Remove spring, which was locking up rails console in dev

atz:
We don't benefit from using spring since we don't have ongoing browser-based interactions w/ this app during development.

I implemented the removal instructions from here:
  https://github.com/rails/spring#removal


jmartin-sul:
👍

spring also kept rails console from properly re-loading my REPL history between sessions.  i'd been disabling spring locally, which took care of it, but happy to see this fix for the project in general.

PR #719, opened 2018-05-02
normalize inputs to controller

SaravShah:
Robots sends druid as `druid:xx000xx0000`. preserved_object_handler needs the druid in `xx000xx0000` format. So remove `druid:` by using split.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -8,7 +8,7 @@ class CatalogController < ApplicationController
 
   # POST /catalog
   def create
-    druid = poh_params[:druid]
+    druid = poh_params[:druid].split(':', 2).last
     incoming_version = poh_params[:incoming_version].to_i
     incoming_size = poh_params[:incoming_size].to_i
     endpoint = Endpoint.find_by(storage_location: "#{poh_params[:storage_location]}/#{Moab::Config.storage_trunk}")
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -8,7 +8,7 @@ class CatalogController < ApplicationController
 
   # POST /catalog
   def create
-    druid = poh_params[:druid]
+    druid = poh_params[:druid].split(':').last
```
atz: Should go ahead and use `split(':', 2)` since we only ever want to split off at most 1 chunk.

tallenaz:
Apart from the change @atz is requesting, this LGTMerge.

app/controllers/catalog_controller.rb
SaravShah: when i do this, it breaks all my tests

SaravShah: oh nvm. i got it

spec/controllers/catalog_controller_spec.rb
```diff
@@ -96,14 +97,14 @@
...
-      expect(response.body).to include(druid)
+      expect(response.body).to include(druid_no_pre)
     end
   end
 
   describe 'PATCH #update' do
     before do
       po = PreservedObject.create!(
-        druid: "bj102hs9687", current_version: ver, preservation_policy: PreservationPolicy.default_policy
+        druid: "druid:bj102hs9687", current_version: ver, preservation_policy: PreservationPolicy.default_policy
```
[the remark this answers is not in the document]
SaravShah: fixed in subsequent PR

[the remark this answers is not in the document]
jmartin-sul: fixed by #721

PR #725, opened 2018-05-02
Reduce unnecessary requires

atz:
Inside `app` everything is auto/eager-loaded by Rails, so you never need to explicitly require anything else that lives inside `app`.

The way `lib` is invoked most typically is via `autoload_path` or in pure Ruby, `ruby -I ./lib` or the `RUBYLIB` ENV var. So components should be available to each other if naming conventions are followed.

We don't actually follow naming conventions fully (e.g. `Checksum` is not in an `Audit` module), *but* for referencing peers in the same directory that doesn't matter.

### DIFF
--- a/app/services/checksum_validator.rb
+++ b/app/services/checksum_validator.rb
@@ -1,5 +1,3 @@
-require 'moab_validation_handler.rb'
-
 # code for validating Moab checksums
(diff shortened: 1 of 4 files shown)


atz:
Hold off on this for a second, still debugging...

Issue #724, opened 2018-05-02
(PC)  rspec tests are ordered - maybe make them work in random order?


SaravShah:
when running `be rspec --order random` we get a lot of failed tests.

SaravShah:
i dont think so. i think backlog is fine ?

ndushay:
See also #728

atz:
Random order reveals where tests side-effect each other, which affects the reproducibility and validity of the tests.  For example, renaming a test file (thereby changing it's sort order position) should not spontaneously cause the tests to fail, but order dependence creates just that scenario.  

An author can intentionally construct tests w/ cascading side effects, but they cannot plausibly maintain it is more readable and maintainable than isolated atomic tests.  In particular, it is not helpful to any subsequent author of tests who is now tasked with understanding complex, invisible and inherited state from **unrelated** examples.

atz:
Sometimes `--order=random` still passes, but sometimes it fails with, e.g. `257` errors.   A lot of them are ` WebMock::NetConnectNotAllowedError` when trying to connect to `sul-lyberservices-test.stanford.edu`.   There is intelligence built-in to `rspec` for identifying which minimum set of tests are order-entangled.

atz:
Got it down to `20` with `--order=random --seed=1485`, then let `--bisect` crunch away on it.  It repeatedly runs  smaller and smaller groups of tests (like binary search) until it finds the smallest order dependent entanglement.  

Final output:
```
Bisect complete! Reduced necessary non-failing examples from 765 to 1 in 2 minutes 31.5 seconds.

The minimal reproduction command is:
  rspec ./spec/controllers/catalog_controller_spec.rb[1:1:3:1,1:1:3:2,1:1:4:1,1:1:4:2,1:2:4:1,1:2:4:2,1:2:5:1,1:2:5:2,1:2:6:1,1:2:6:2,1:2:7] ./spec/services/preserved_object_handler_create_spec.rb[1:1:3,1:1:5:1:1,1:1:5:1:2,1:1:5:2,1:2:6:1,1:2:6:2,1:2:6:3:1:1,1:2:6:3:1:2,1:2:6:3:2] ./spec/services/workflow_errors_reporter_spec.rb[1:1:1] --order=random --seed=1485
```

So there it is: 3 things that are side-effecting each other.

atz:
In `workflow_errors_reporter_spec`, the interaction is obvious, because it does:
```ruby
[...]

PR #727, opened 2018-05-03
Audit module

atz:
Branched from #725.  This puts the classes that already lived in `lib/audit/` into an `Audit` module namespace, in order to correspond with dependency loading expectations.  A small `audit.rb` file is added to group the three classes and load common dependencies, simplifying usage for consumers:

```ruby
require 'audit'
Audit::Checksum.validate_disk(...)
```

Less `require_relative` backtracking all around and hopefully, eventually, less circular dependency loading.

Note that the dependency structure is as aspect of order-dependence in tests, since some `lib` tests fail if they are run before rails is loaded, but are fine otherwise.  Breaking test order dependence out to separate work, re: #724


atz:
Rebased

atz:
Rebased again and resolved minor conflict.

atz:
Rebased again and resolve somewhat larger conflict.  Also corrected calls to logger to use the class logger, not `Rails.logger` directly.

app/lib/active_record_utils.rb
```diff
@@ -1,6 +1,5 @@
-##
-# some usefule re-usable ActiveRecord patterns
-class ActiveRecordUtils
+# some useful re-usable ActiveRecord patterns
+module ActiveRecordUtils
```
SaravShah: 👍

app/lib/audit.rb
```diff
@@ -0,0 +1,16 @@
...
+
+require_relative 'audit/catalog_to_moab'
+require_relative 'audit/checksum'
+require_relative 'audit/moab_to_catalog'
+
+# @example Usage
+#  require 'audit'
+#  ...
+#  Audit::Checksum.validate_disk(storage_root)
+module Audit
```
SaravShah: why do we need an empty `module Audit` here?

spec/lib/audit/catalog_to_moab_class_spec.rb
```diff
@@ -1,8 +1,8 @@
 require 'rails_helper'
-require_relative '../../../lib/audit/catalog_to_moab.rb'
+require 'audit'
```
SaravShah: can we just `require 'audit' ` in `rails_helper.rb`?

app/lib/audit.rb
atz: Just convention to define the namespace.  Not strictly necessary, and I'd remove it if ppl prefer.

spec/lib/audit/catalog_to_moab_class_spec.rb
atz: In fact, I think we might not even have to require anything since we can rely on Rails auto/eager-loading.

atz:
@SaravShah removed the requires from spec files (inside `app`, Rails autoloading takes care of it) and removed the empty `module` declaration.

SaravShah:
will wait for one more set of 👀 before merging!

app/lib/audit/checksum.rb
```diff
@@ -10,9 +10,9 @@ def self.validate_disk(endpoint_name, limit=Settings.c2m_sql_limit)
       # pcs_w_expired_fixity_check is an AR Relation; it could return a lot of results, so we want to process it in
```
ndushay: whoops - looks like the comment wasn't updated with the var name chagne.  Also not sure why var name change was necessary, but whatever.

spec/lib/audit/moab_to_catalog_spec.rb
```diff
@@ -19,8 +19,8 @@
     end
 
     it 'calls check_existence_for_dir with the right arguments' do
-      HostSettings.storage_roots.each do |storage_root|
-        expect(described_class).to receive(:check_existence_for_dir).with("#{storage_root[1]}/#{Settings.moab.storage_trunk}")
+      HostSettings.storage_roots.to_h.each_value do |path|
```
ndushay: i'd rather we went with `storage_root_location` rather than `path`, mimicing the other tests changed.

spec/load_fixtures_helper.rb
```diff
@@ -17,7 +17,7 @@
   end
   after do
     # TODO: danger - if there are objects in the test db we want to keep
-    HostSettings.storage_roots.each_value do |storage_root|
+    HostSettings.storage_roots.to_h.each_value do |storage_root|
```
ndushay: similarly, while in here, change to storage_root_location??

atz: Seems unrelated to my purpose.

spec/lib/audit/moab_to_catalog_spec.rb
atz: Because it is a path, and that is sufficient for understanding the test on the *very next line*.  We do pathy things with it.  The value in question was not even named previously, it was just `storage_root[1]`.  This feels like an incredibly random nit after a month of time and multiple rebases.

spec/load_fixtures_helper.rb
atz: @ndushay You realize this line is completely broken as written, right?  Literally the block never executes now.  Example:
```ruby
> HostSettings.storage_roots.each_value { puts "XXX" }
=> nil
> HostSettings.storage_roots.to_h.each_value { puts "XXX" }
XXX
XXX
XXX
...
```

atz:
Rebased again.

atz:
Rebased again to incorporate small upstream changes to logging

atz:
Blocking the concrete console instructions in #833

PR #733, opened 2018-05-04
Account for rubocop exceptions in the right places

atz:
By definition "shameless green" style exceptions are not the same as others we may intend permanently.  The right place for those to live is the TODO file. That way when they get fixed (intentionally or coincidentally), the exclusion can just be mechanically deleted/overwritten by `--auto-gen-config`.

The `puts` statements exceptions are a good example of why storing many of these cases in `.rubocop.yml` is misguided.  Those cases where already cleaned up.  If they lived in `.rubocop_todo.yml`, they would never have needed to be revisited manually, since auto-generating the TODO file would do the trick.


ndushay:
I'm not sure I agree with you here.   Because rubocop_todo is auto-genned, we don't control the granularity of how it deals with exceptions.  Maybe it says:

```
Exclude:
   - 'lib/**/*
```

but our  intent is not to exclude a rule for ALL of our specs -- it's just rubocop's auto-gen algorithm that does it.

If we are *intentionally* excluding something, we want to be clear about that, and why.  

Your concern seems to be around ensuring we clean up our rubocop.yml after we address a problem.  We could work with that by having comments in the code indicating there is a rubocop exception.

Or, honestly, we could just turn rubocop off.  I feel like we spend a lot of effort pleasing our rubocop master and it's not necessarily creating significantly better code -- it's creating churn.  You have strong feelings about rubocop and I think they are getting in our way.  These cycles would be better spent on critical path tasks.

SaravShah:
oh thats awesome. didn't know that but def makes more sense now!

atz:
@ndushay I'm not sure I can relate your comment to any of the lines in this PR.  None of the exceptions that we intend to keep were altered.  

I feel like Rubocop exceptions were thrown in in the absence of fully understanding what they were about, masking some legitimate problems.  Keeping them in `.rubocop.yml` conveys that they are permanent, but I gave some concrete examples of why to prefer the TODO file at least for the stuff we **know** we intend to revise (i.e. "shameless green").  

The auto-generation has heuristics to minimize the scope of exceptions while not bloating the file. 
It does not add exceptions where none are needed.  My concern is that we are using the tool the wrong way.  I don't want comments doing multiple accounting for the same rubocop exceptions (that would be worse).

You can help by not blocking minor and appropriate work with handwringing about cycles and "critical path".

ndushay:
Joe,

I disagree with you.  I have never been on a project that used an autogen file this way.  I think we need to resolve this in person -- the tone here is already devolving, and that's not a way forward.

- Naomi

atz:
OK, lets do that.  The project currently treats **every** exception as permanent, which is at odds w/ the tooling and actual intent.  We have plenty of project that do use rubocop this way (literally every previous project I've been on, including the community ones).  

I don't think we *actually* disagree about intention, I think some unfamiliarity is getting in the way.

atz:
rebased

atz:
rebased and resolved a small conflict.  this time the addition of a new exception wouldn't have been necessary if the `Style/SymbolArray.MinSize: 4` setting were already in place.

SaravShah:
close in favor of #1069 1069

PR #734, opened 2018-05-04
remove rubocop from default rake task

ndushay:
Let's try a few weeks with mandatory rubocop NOT a part of the build.  Let's see if it helps keep our focus on critical path tasks.


atz:
Consensus is to defer on this one.

Issue #739, opened 2018-05-07
(R) Need a setting configuration for sdr_transfer volume


SaravShah:
Setting config for sdr_transfer vol in settings.yml (and shared_configs)

Story Time Notes: 
https://docs.google.com/document/d/10HMhTMWlZaM1DJChFr2r3coOVDZKe0EDAH8PPumb-2E/edit#

See also #678

PR #740, opened 2018-05-08
add zip storage setting to preservation_catalog

tallenaz:
Connects to #739


config/settings.yml
```diff
@@ -40,3 +40,5 @@ workflow_services_url: ''
 c2m_sql_limit: 1000
 
 checksum_algos: ['md5'] # 'sha1' 'sha256'
+
+zip_storage: '' # empty here, override in #{RAILS_ENV}.yml
```
ndushay: actually, might I suggest "tmp"  or something for a local setting, but keep the second half of your comment?  Then we can use it for tests.

PR #743, opened 2018-05-09
Add db:seed to cap

### LINKED ISSUE #736: (PC) add rake db:seed to cap deploy
in order to ensure we seed new endpoints, we should add rake db:seed to cap deploy.

So when a new endpoint is added to shared configs, then cap deploy should be run.

This should also be documented in
- consul add a storage root page
- PC README (?)

tallenaz:
Adds a capistrano task to call seed_catalog
  and perform it after migrations on a deploy

  Fixes #736

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,3 +216,15 @@ rake spec
 ## Deploying
 
 Capistrano is used to deploy.
+
+To run `rake db:seed` in a deploy environment:
+
+```sh
+bundle exec cap stage db_seed # for the stage servers
+```
+
+or
+
+```sh
+bundle exec cap prod db_seed # for the prod servers
+```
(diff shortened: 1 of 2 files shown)


config/deploy.rb
```diff
@@ -36,3 +36,14 @@
...
 # Default value for keep_releases is 5
 # set :keep_releases, 5
+
+after 'deploy:migrate', 'seed_catalog'
+
+desc 'seed the catalog'
+task :seed_catalog do
+  on roles(:all) do
+    within current_path do
+      execute :rake, 'seed_catalog'
```
SaravShah: i think this should `rake db:seed`, cause `seed_catalog` will seed the storage roots. `rake db:seed` will seed the Endpoints, Preservation Policy, and EndpointType

tallenaz: D'oh

README.md
```diff
@@ -216,3 +216,15 @@ rake spec
 ## Deploying
 
 Capistrano is used to deploy.
+
+To `rake db:seed` in a deploy environment:
```
ndushay: To run ?   it reads a bit oddly, but isn't a blocker.

PR #748, opened 2018-05-11
Update README for redis prerequisite

atz:
Corrected `ruby` code blocks that are actually shell commands and other markdown formatting issues.


README.md
```diff
@@ -145,61 +150,55 @@ this will generate a log at, for example, ```log/profile_check_existence_for_all
...
 RAILS_ENV=production bundle exec rake cv_single_endpoint[fixture_sr3,profile]
 ```
-this will generate a log at, for example,
-```log/profile_cv_validate_disk2018-01-01T14:25:31-flat.txt```
+this will generate a log at, for example, `log/profile_cv_validate_disk2018-01-01T14:25:31-flat.txt`
 
 #### All Roots
 - Without profiling:
-```ruby
+```sh
```
SaravShah: good call

Issue #722, opened 2018-05-02
(IR) remove objects with druid: prefix in prod


SaravShah:
created some objects in prod that have a druid name of `druid:xx000xx0000` 
need to remove those prefix 

John: i'm not sure what the exact syntax is for AR and PG (and it's been a while since i've done it in oracle), but it should be pretty easy to search for pres objects that start with the `druid:` prefix, and update them programmatically.... we could even do it as a proper AR migration in a PR, and make it happen by deploying (instead of just going cowboy on the rails console on prod)

jmartin-sul:
* [ ] query to find druids with 'druid:' prefix
* [ ] loop to normalize (prob don't need transaction, just get through what you can?)
* [ ] simple test (create some bad druids, run clean-up, show query returns zero results after cleanup)
* [ ] migration calls normalize method

PR #749, opened 2018-05-11
Normalize prod data

SaravShah:
connects #722

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'honeybadger' # for error reporting / tracking / notifications
 gem 'jbuilder', '~> 2.5' # Build JSON APIs with ease.
 gem 'okcomputer'
 gem 'pg' # postgres database
+gem "factory_bot_rails", "~> 4.0"
 # pry is useful for debugging, even in prod
 gem 'pry-byebug' # call 'binding.pry' anywhere in the code to stop execution and get a pry-byebug console
 gem 'pry-rails' # use pry as the rails console shell instead of IRB
--- a/app/models/preserved_object.rb
+++ b/app/models/preserved_object.rb
@@ -9,4 +9,21 @@ class PreservedObject < ApplicationRecord
   validates :druid, presence: true, uniqueness: true, format: { with: DruidTools::Druid.pattern }
   validates :current_version, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :preservation_policy, null: false
+
+  def self.normalize_druid_name
+    targets = PreservedObject.where("druid LIKE 'druid:%'")
+    druids_no_prefix = targets.pluck(:druid).map { |d| d.split(':', 2).last }
+    targets_by_druid = targets.map { |rec| [rec.druid, rec] }.to_h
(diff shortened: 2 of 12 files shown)


app/models/preserved_object.rb
```diff
@@ -9,4 +9,21 @@ class PreservedObject < ApplicationRecord
   validates :druid, presence: true, uniqueness: true, format: { with: DruidTools::Druid.pattern }
   validates :current_version, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :preservation_policy, null: false
+
+  def self.normalize_druid_name
+    targets = PreservedObject.where("druid LIKE 'druid:%'")
```
tallenaz: Is this sanitized enough for our purposes? Maybe it doesn't matter, since this isn't exposed to the world, but:

```ruby
PreservedObject.where("druid LIKE ?", "druid:%")
```

feels safer. Not a blocker.

Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'honeybadger' # for error reporting / tracking / notifications
 gem 'jbuilder', '~> 2.5' # Build JSON APIs with ease.
 gem 'okcomputer'
 gem 'pg' # postgres database
+gem "factory_bot_rails", "~> 4.0"
```
tallenaz: Hadn't known `factory_girl` got renamed. 👍

db/migrate/20180511210539_noramlize_druid_name.rb
```diff
@@ -0,0 +1,5 @@
+class NoramlizeDruidName < ActiveRecord::Migration[5.1]
+  def change
+    PreservedObject.normalize_druid_name
+  end
+end
```
tallenaz: I like the approach. I hadn't thought much about using migrations to remediate data, as opposed to changing a schema.

app/models/preserved_object.rb
SaravShah: i _think_ we use that syntax (`PreservedObject.where("druid LIKE ?", "druid:%")
)` if we are going to interpolate a variable and are worried about a sql injection, but since we are just hard coding `'druid:%'` it doesn't matter...?

@jmartin-sul  is that right?

jmartin-sul: yeah, since it's a hard coded string that we know is safe, there's no risk of injection, and so the bind variable seems unnecessary to me.

spec/support/job_config.rb
```diff
@@ -3,6 +3,10 @@
     next unless RSpec.current_example.metadata[:type] == :job
     ActiveJob::Base.queue_adapter = :test
     allow(ActiveJob::Base.logger).to receive(:info) # keep the default logging quiet
-    Resque.redis.redis.flushall # clear queues and locks
+    begin
+      Resque.redis.redis.flushall # clear queues and locks
+    rescue Redis::CannotConnectError
+      p "we are rescuing!"
```
ndushay: this is not really a useful message ... but since it is just a spec, probably not worth addressing???

ndushay: It would please me if you changed it;  a quick PR and I'll merge it fast!

PR #750, opened 2018-05-11
catalog controller can take druid with or without prefix

### LINKED ISSUE #729: (PC) Add test where catalog_controller can take `druid:xx000xx0000` or `xx000xxx0000`
Add a test where druid can take druid with prefix in the catalog controller for #create and #update

SaravShah:
Via the `#create` and `#update` ReST calls in the `catalog_controller`, the `druid` parameter should be allowed to take in either `druid:xx000xx0000` or `xx000xx0000`.

Change `druid` to `prefixed_druid` and `druid_no_pre` to `bare_druid` - better variable naming

Closes #729

### DIFF
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -2,22 +2,22 @@ require 'rails_helper'
 RSpec.describe CatalogController, type: :controller do
   let(:size) { 2342 }
   let(:ver) { 3 }
-  let(:druid) { 'druid:bj102hs9687' }
-  let(:druid_no_pre) { 'bj102hs9687' }
+  let(:prefixed_druid) { 'druid:bj102hs9687' }
+  let(:bare_druid) { 'bj102hs9687' }
   let(:storage_location) { "#{storage_location_param}/moab_storage_trunk" }
   let(:storage_location_param) { "spec/fixtures/storage_root01" }
 
   describe 'POST #create' do
     context 'with valid params' do
-      let(:pres_obj) { PreservedObject.find_by(druid: druid_no_pre) }
+      let(:pres_obj) { PreservedObject.find_by(druid: bare_druid) }
       let(:pres_copy) { PreservedCopy.find_by(preserved_object: pres_obj) }
 
       before do
-        post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
+        post :create, params: { druid: prefixed_druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
       end
 
       it 'saves PreservedObject and PreservedCopy in db' do
-        po = PreservedObject.find_by(druid: druid_no_pre)
+        po = PreservedObject.find_by(druid: bare_druid)
         pc = PreservedCopy.find_by(preserved_object: po)
         expect(po).to be_an_instance_of PreservedObject
         expect(pc).to be_an_instance_of PreservedCopy
@@ -26,7 +26,7 @@ RSpec.describe CatalogController, type: :controller do
         expect(pres_copy.endpoint.storage_location).to eq storage_location
         expect(pres_copy.version).to eq ver
         expect(pres_copy.size).to eq size
-        expect(pres_obj.druid).to eq druid_no_pre
+        expect(pres_obj.druid).to eq bare_druid
       end
 
       it 'response contains create_new_object code ' do
@@ -45,7 +45,7 @@ RSpec.describe CatalogController, type: :controller do
       end
 
       it 'does not save PreservedObject or PreservedCopy in db' do
-        po = PreservedObject.find_by(druid: druid)
+        po = PreservedObject.find_by(druid: prefixed_druid)
         pc = PreservedCopy.find_by(preserved_object: po)
         expect(po).to be_nil
         expect(pc).to be_nil
@@ -64,8 +64,8 @@ RSpec.describe CatalogController, type: :controller do
 
     context 'object already exists' do
       before do
-        post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
-        post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
+        post :create, params: { druid: prefixed_druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
+        post :create, params: { druid: prefixed_druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
       end
 
       it 'response contains error message' do
@@ -80,9 +80,9 @@ RSpec.describe CatalogController, type: :controller do
 
     context 'db update failed' do
       before do
-        allow(PreservedObject).to receive(:create!).with(hash_including(druid: druid_no_pre))
+        allow(PreservedObject).to receive(:create!).with(hash_including(druid: bare_druid))
                                                    .and_raise(ActiveRecord::ActiveRecordError, 'foo')
-        post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
+        post :create, params: { druid: prefixed_druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
       end
 
       it 'response contains error message' do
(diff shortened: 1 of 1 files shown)


spec/controllers/catalog_controller_spec.rb
```diff
@@ -99,6 +99,21 @@
       post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
       expect(response.body).to include(druid_no_pre)
     end
+    context 'can take two forms of druid input' do
+      let(:druid) { 'druid:jj925bx9565' }
+      let(:druid_no_pre) { 'jj925bx9565' }
+
+      it 'druid:jj925bx9565' do
+        post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
+        expect(response.body).to include(druid_no_pre)
```
tallenaz: I'm a little unclear about this expectation: the response body equals `druid` and includes `druid_no_pre`. Would a more rigorous expectation be that the body equals `druid`? I understand that the body includes the String `jj925bx9565`, but `druid_no_pre` makes me think it's a druid, not a String, and then I get hung up on the weird thought that a druid includes a druid. Or do we need this expectation at all? Is it enough to expect a `:created` status?

SaravShah: So, i'm just trying to show that you can give the `POST` or `PATCH` request either `druid:jj925bx9565` or `jj925bx9565` and the request will go through, and not throw an error. The catalog controller code will take either input do a `.split(':', 2)` and return `jj925bx9565` in both cases. So its just like oh yeah send the create params w/ `druid:jj925bx9565`, but we'll end up w/ `jj925bx9565` bc of internal code.

I could also add a test thats like 
`expect(response.body).no_to include(druid)`


I can remove this test and just keep the status code. If you think that would be more clear.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -221,5 +236,23 @@
...
+    context 'can take two forms of druid input' do
+      before do
+        post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
+      end
+      let(:druid) { 'druid:jj925bx9565' }
+      let(:druid_no_pre) { 'jj925bx9565' }
+
+      it 'druid:jj925bx9565' do
+        patch :update, params: { druid: druid, incoming_version: 5, incoming_size: size, storage_location: storage_location_param }
+        expect(response.body).to include(druid_no_pre)
```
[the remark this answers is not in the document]
SaravShah: the `response.body` is equal to `"{\"druid\":\"jj925bx9565\",\"result_array\":[{\"actual_vers_gt_db_obj\":\"actual version (5) greater than PreservedCopy db version (3)\"}]}"`

Just thought i'm not testing for the result array so I could just test for the druid - happy to change it though

spec/controllers/catalog_controller_spec.rb
tallenaz: The additional `expect(response.body).not_to include(druid)`makes it clearer to my brain.

spec/controllers/catalog_controller_spec.rb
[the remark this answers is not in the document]
tallenaz: Looking this over again, I actually don't have the same worry, since the expectation in on `druid_no_pre`, I was probably just overly concerned from the previous context. This one is ok by me.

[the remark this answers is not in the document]
atz: Testing for the exact body response makes the test pretty brittle when something trivial like a new key/value is added to the response.  Tests should try to avoid that when possible.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -99,6 +99,21 @@
       post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
       expect(response.body).to include(druid_no_pre)
     end
+    context 'can take two forms of druid input' do
+      let(:druid) { 'druid:jj925bx9565' }
+      let(:druid_no_pre) { 'jj925bx9565' }
```
ndushay: a more commonly used name for this would be `bare_druid`  :-)   You might even consider `bare_druid` and `prefixed_druid` to be super-duper clear.   And even better:

```ruby
let(:bare_druid) {'asdfasdfsadf'}
let(:prefixed_druid {"druid:#{bare_druid}"}
```

spec/controllers/catalog_controller_spec.rb
```diff
@@ -99,6 +99,21 @@
       post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
       expect(response.body).to include(druid_no_pre)
     end
+    context 'can take two forms of druid input' do
```
ndushay: `can take druid with or without "druid:" prefix`  ?

As came up in devops today, think about reading this code without any context.  Like, me two months after I'm done with this project.  Or you, in a year.  Or ...

spec/controllers/catalog_controller_spec.rb
ndushay: and here, in the expectations, using the names `bare_druid` and `prefixed_druid` would help make this clearer, too.  Also, I agree with @tallenaz

spec/controllers/catalog_controller_spec.rb
```diff
@@ -99,6 +99,21 @@
...
+      let(:druid) { 'druid:jj925bx9565' }
+      let(:druid_no_pre) { 'jj925bx9565' }
+
+      it 'druid:jj925bx9565' do
```
ndushay: you can use a variable:  

```ruby
it prefixed_druid do
  blah
end
```

if that doesn't work, try

```ruby
it "#{prefixed_druid}" do
```

spec/controllers/catalog_controller_spec.rb
```diff
@@ -221,5 +236,23 @@
       post :update, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
       expect(response.body).to include(druid_no_pre)
     end
+    context 'can take two forms of druid input' do
```
ndushay: ditto on this description -- it isn't telling us enough.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -221,5 +236,23 @@
...
+      before do
+        post :create, params: { druid: druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
+      end
+      let(:druid) { 'druid:jj925bx9565' }
+      let(:druid_no_pre) { 'jj925bx9565' }
```
ndushay: just declare these once and use them for both Patch and Post?  with better names?

spec/controllers/catalog_controller_spec.rb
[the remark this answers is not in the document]
ndushay: are we expecting it NOT to include prefixed_druid  as well as  including  bare_druid?    You could write expectations to make this abundantly clear.

spec/controllers/catalog_controller_spec.rb
ndushay: are we expecting it NOT to include prefixed_druid  as well as  including  bare_druid?    You could write expectations to make this abundantly clear.

spec/controllers/catalog_controller_spec.rb
SaravShah: tried both - didn't work

  `prefixed_druid` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).

spec/controllers/catalog_controller_spec.rb
SaravShah: i could declare it at the top but we already have a bare_druid and prefixed_druid (bj102hs9687) being declared. 
i'm trying to think of a name that I could use that is not `bare_druid` or `prefixed_druid` at the top...can't seem to think of a good one

spec/controllers/catalog_controller_spec.rb
```diff
@@ -201,25 +218,45 @@
...
+      let(:prefixed_druid) { "druid:#{bare_druid}" }
+      let(:bare_druid) { 'jj925bx9565' }
+
+      it 'prefixed_druid' do
+        patch :update, params: { druid: prefixed_druid, incoming_version: 5, incoming_size: size, storage_location: storage_location_param }
+        expect(response.body).to include(bare_druid)
+        expect(response.body).not_to include(prefixed_druid)
+        expect(response).to have_http_status(:ok)
+      end
+      it 'druid' do
```
ndushay: bare druid ?

spec/controllers/catalog_controller_spec.rb
```diff
@@ -96,8 +96,25 @@
...
+      let(:bare_druid) { 'jj925bx9565' }
+      let(:prefixed_druid) { "druid:#{bare_druid}" }
+
+      it 'prefixed_druid' do
+        post :create, params: { druid: prefixed_druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param }
+        expect(response.body).to include(bare_druid)
+        expect(response.body).not_to include(prefixed_druid)
+        expect(response).to have_http_status(:created)
+      end
+      it 'druid' do
```
ndushay: bare druid?

spec/controllers/catalog_controller_spec.rb
SaravShah: done!

PR #755, opened 2018-05-14
add logging to checksum.validate_disk

SaravShah:
While trying to reproduce status ok to unexpected version on storage, thought this would help as to why if you run cv twice w/in a short period of time - you will probably not find any PC's

### DIFF
--- a/lib/audit/checksum.rb
+++ b/lib/audit/checksum.rb
@@ -12,6 +12,7 @@ class Checksum
     # batches.  we can't use ActiveRecord's .find_each, because that'll disregard the order .fixity_check_expired
     # specified.  so we use our own batch processing method, which does respect Relation order.
     pcs_w_expired_fixity_check = PreservedCopy.by_endpoint_name(endpoint_name).fixity_check_expired
+    logger.info "Number of Preserved Copies to be checksum validated: #{pcs_w_expired_fixity_check.count}"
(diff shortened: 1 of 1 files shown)


lib/audit/checksum.rb
```diff
@@ -12,6 +12,7 @@ def self.validate_disk(endpoint_name, limit=Settings.c2m_sql_limit)
     # batches.  we can't use ActiveRecord's .find_each, because that'll disregard the order .fixity_check_expired
     # specified.  so we use our own batch processing method, which does respect Relation order.
     pcs_w_expired_fixity_check = PreservedCopy.by_endpoint_name(endpoint_name).fixity_check_expired
+    logger.info "Preserved Copies to be checksum validated: #{pcs_w_expired_fixity_check.count}"
```
tallenaz: How about "Number of Preserved Copies"? or "Count of"? It would make it clearer to me that I'm not looking at, for example, a single database id...

SaravShah:
@tallenaz @ndushay rdy for review

jmartin-sul:
@ndushay counts are generally a cheap thing to do (much cheaper than returning the results being counted, both because the info isn't being returned over the network and because the query planner can be smarter and do less work than if it had to actually fetch all the rows; plus, no need to instantiate any ActiveRecord objects).

so i think this is a safe thing to do.

PR #759, opened 2018-05-15
Target archive endpoints

### LINKED ISSUE #713: (R) write shared replication logic ("what endpoints *should* a PC go to?")
likely a scope (on PreservedCopy?)  and maybe some helper methods or additional scopes for chaining for each use case?

the places where this will be used:
* create the `PreservedCopy` entries for as-yet unreplicated druid-versions (in `UNREPLICATED_STATUS` or similar), before a `ZipmakerJob` is queued.
* `PlexerJob` will need to determine what endpoints a zipped archive version should be sent to.
* `ResultsRecorderJob` will need to determine whether the result it just got indicates that the last remaining archive copy has been uploaded.

jmartin-sul:
still needs:
* [x] tests for archive endpoint seeding
* [x] tests for scopes
* [x] tests for pres copy creation
  * [x] make sure PCs get made for default seeded endpoint
  * [x] make sure only PCs for given version are made
  * [x] add endpoint and call creation and confirm a new PC gets made
  * [x] version check
  * [x] transactionality

i'd also like to:
* [ ] index endpoint_types.type_name and .endpoint_class ; preserved_copies.version ; preserved_objects.current_version
* [ ] take care of the small TODOs:
  * [ ] move the default endpoint type methods into the `EndpointType` class and specs
  * [ ] enum for `endpoint_class` ~maybe constants or something for endpoint type~.

closes #713

### DIFF
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -18,6 +18,31 @@ class Endpoint < ApplicationRecord
   validates :storage_location, presence: true
   validates :recovery_cost, presence: true
 
+  scope :archive, lambda {
+    # TODO: maybe endpoint_class should be an enum or a constant?
+    joins(:endpoint_type).where(endpoint_types: { endpoint_class: 'archive' })
+  }
+
+  # for the given druid, which archive endpoints should have preserved copies?
+  scope :archive_targets, lambda { |druid|
+    archive.joins(preservation_policies: [:preserved_objects]).where(preserved_objects: { druid: druid })
+  }
+
+  scope :which_need_archive_copy, lambda { |druid, version|
+    # cast version to int for nicer errors in the case of bad input.  though ActiveRecord/ARel would still protect
+    # against injection attacks even without using a bind var (e.g. a string passed in for an int col query would
+    # silently be inserted as 0, an INTEGER).
+    pc_table = PreservedCopy.arel_table
+    ep_table = Endpoint.arel_table
+    endpoint_has_pres_copy_subquery =
+      PreservedCopy.where(
+        pc_table[:endpoint_id].eq(ep_table[:id])
+          .and(pc_table[:version].eq(version.to_i))
+      ).exists
+
+    archive_targets(druid).where.not(endpoint_has_pres_copy_subquery)
+  }
+
   # iterates over the storage roots enumerated in settings, creating an endpoint for each if one doesn't
   # already exist.
   # returns an array with the result of the ActiveRecord find_or_create_by! call for each settings entry (i.e.,
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -36,6 +36,20 @@ class Endpoint < ApplicationRecord
     end
   end
 
+  def self.seed_archive_endpoints_from_config(preservation_policies)
+    Settings.archive_endpoints.map do |endpoint_name, endpoint_config|
+      find_or_create_by!(endpoint_name: endpoint_name.to_s) do |endpoint|
+        endpoint.endpoint_type = EndpointType.find_by!(type_name: endpoint_config.endpoint_type_name)
+        endpoint.endpoint_node = endpoint_config.endpoint_node
+        endpoint.storage_location = endpoint_config.storage_location
+        endpoint.access_key = endpoint_config.access_key
+        endpoint.recovery_cost = endpoint_config.recovery_cost
+        endpoint.preservation_policies = preservation_policies
+      end
+    end
+  end
+
+  # TODO: move to EndpointType class?  e.g. .default_for_storage_root
   def self.default_storage_root_endpoint_type
     EndpointType.find_by!(type_name: Settings.endpoints.storage_root_defaults.endpoint_type_name)
   end
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -16,7 +16,6 @@ class Endpoint < ApplicationRecord
   validates :endpoint_type, presence: true
   validates :endpoint_node, presence: true
   validates :storage_location, presence: true
-  validates :recovery_cost, presence: true
 
   scope :archive, lambda {
     # TODO: maybe endpoint_class should be an enum or a constant?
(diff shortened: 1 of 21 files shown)


app/models/endpoint.rb
```diff
@@ -16,7 +16,31 @@ class Endpoint < ApplicationRecord
   validates :endpoint_type, presence: true
   validates :endpoint_node, presence: true
   validates :storage_location, presence: true
-  validates :recovery_cost, presence: true
+
+  scope :archive, lambda {
+    # TODO: maybe endpoint_class should be an enum or a constant?
```
atz: Agree.  That would let us do `Endpoint_type.archive` as an auto-magic activerecord scope.

jmartin-sul:
this is ready for actual review.

i ended up finishing all the tests i'd set out to do.  will ticket the follow-on stuff and start implementing that in a branch off of this branch.

app/models/endpoint.rb
```diff
@@ -16,7 +16,31 @@ class Endpoint < ApplicationRecord
...
+    # cast version to int for nicer errors in the case of bad input.  though ActiveRecord/ARel would still protect
+    # against injection attacks even without using a bind var (e.g. a string passed in for an int col query would
+    # silently be inserted as 0, an INTEGER).
+    pc_table = PreservedCopy.arel_table
+    ep_table = Endpoint.arel_table
+    endpoint_has_pres_copy_subquery =
+      PreservedCopy.where(
+        pc_table[:endpoint_id].eq(ep_table[:id])
+          .and(pc_table[:version].eq(version.to_i))
+      ).exists
```
jmartin-sul: another way to write this subquery would be:
```ruby
  pc_table = PreservedCopy.arel_table
  ep_table = Endpoint.arel_table
  endpoint_has_pres_copy_subquery =
    PreservedCopy.where(
      'preserved_copies.endpoint_id = endpoints.id and preserved_copies.version = #{version.to_i}'
    ).exists
```

from the reading i found online (stack exchange posts and blog posts), the `arel_table` approach in this PR seemed slightly more idiomatic, and *maybe* more maintainable in theory, but it might also be slightly less readable?

do others have strong preferences?

jmartin-sul:
filed #793 for the remaining follow-on work

app/models/endpoint.rb
atz: In either case, the subquery only makes sense in the context of the macro-query `.joins`.  I can take a stab at some other forms if you want.

app/models/endpoint.rb
ndushay: Created #797 re: this.  Put it in backlog.

app/models/endpoint.rb
ndushay: I don't have an opinion on which way to go with this.  Are we likely to care about performance for this?  I think the two methods John proposes are equally readable/maintainable.

app/models/preserved_object.rb
```diff
@@ -26,4 +26,19 @@ def self.normalize_druid_name
       po.save!
     end
   end
+
+  def create_archive_copies(archive_vers)
```
ndushay: any chance this would be misinterpreted as "it's going to actually do the replication" ?     Add a comment?  Rename?  Or just let folks read the code?

Other naming suggestions:  `create_archive_preserved_copies` ?

app/models/preserved_object.rb
```diff
@@ -26,4 +26,19 @@ def self.normalize_druid_name
...
+    unless archive_vers > 0 && archive_vers <= current_version
+      raise ArgumentError, "archive_vers (#{archive_vers}) must be between 0 and current_version (#{current_version})"
+    end
+
+    ApplicationRecord.transaction do
+      Endpoint.which_need_archive_copy(druid, archive_vers).map do |ep|
+        # TODO: remember to update size at some later point, after zip is created
```
ndushay: should this be ticketed?

spec/models/preserved_object_spec.rb
```diff
@@ -106,4 +106,72 @@
...
+      }.from([archive_ep]).to([])
+      expect(archive_pcs_for_druid.where(version: current_version).count).to eq 1
+
+      new_archive_ep.preservation_policies = [PreservationPolicy.default_policy]
+      expect { po.create_archive_copies(current_version) }.to change {
+        Endpoint.which_need_archive_copy(druid, current_version).to_a
+      }.from([new_archive_ep]).to([])
+      expect(archive_pcs_for_druid.where(version: current_version).count).to eq 2
+    end
+
```
ndushay: should we have these tests?

- does not create pres copy if it already exists
- archive copies created have status unreplicated
...

jmartin-sul: > * does not create pres copy if it already exists

hmm... i was thinking that i'd covered that in the tests that re-run ~~`create_archive_copies`~~ `create_archive_preserved_copies` and then confirm the pres copy counts are right, but it's true that they don't test invocation on the same druid/version combo multiple times.  i will add that in #796, if you don't mind.

> * archive copies created have status unreplicated

good catch, will also do that in 796

atz: You can also punt on this and just merge the DB constraint.

app/models/endpoint.rb
atz: Query structure ends up looking like:
```
SELECT "endpoints".* FROM "endpoints" 
INNER JOIN "endpoint_types" ON "endpoint_types"."id" = "endpoints"."endpoint_type_id" 
INNER JOIN "endpoints_preservation_policies" ON "endpoints_preservation_policies"."endpoint_id" = "endpoints"."id" 
INNER JOIN "preservation_policies" ON "preservation_policies"."id" = "endpoints_preservation_policies"."preservation_policy_id" 
INNER JOIN "preserved_objects" ON "preserved_objects"."preservation_policy_id" = "preservation_policies"."id" 
WHERE "endpoint_types"."endpoint_class" = $1 
AND "preserved_objects"."druid" = $2 
AND (NOT (EXISTS (
  SELECT "preserved_copies".* FROM "preserved_copies" 
  WHERE ("preserved_copies"."endpoint_id" = "endpoints"."id" 
  AND "preserved_copies"."version" = 1)
)))
```

Example [`EXPLAIN ANALYZE`](https://explain.depesz.com/s/sAH8) for my tiny dev db.

jmartin-sul: yup, that looks like the SQL i remember seeing get generated.  and that plan looks almost exactly like the one in my notes, with very slightly different overall costs, but the same join etc plan (which makes sense, since we have the same fixture data on similar laptops).

PR #761, opened 2018-05-15
Constrain resque dashboard set up to workers

tallenaz:
This is a naive approach that works, at the cost of hard-coding parts of machine names into a regex. 

If the team is uncomfortable with it, one other way is to set up an "advanced constraint", such as the example at http://guides.rubyonrails.org/routing.html#advanced-constraints. The TLDR is to "provide an object that responds to `matches?` that Rails should use" in the constraint. This might be a small class that pulls from `Settings`, or...

Or there might be another approach that I'm unaware of.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,6 +1,10 @@
+require 'resque/server'
+
 Rails.application.routes.draw do
   resources :catalog, param: :druid, only: %i[create update]
   resources :moab_storage, only: %i[index show]
+
+  mount Resque::Server.new, at: '/resque', constraints: { host: /stage-02|prod-03/ }
 
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
 end


tallenaz:
Some discussion already on this issue is here: https://github.com/sul-dlss/preservation_catalog/pull/699

config/routes.rb
```diff
@@ -1,6 +1,10 @@
+require 'resque/server'
+
 Rails.application.routes.draw do
   resources :catalog, param: :druid, only: %i[create update]
   resources :moab_storage, only: %i[index show]
 
+  mount Resque::Server.new, at: '/resque', constraints: { host: /stage-02|prod-03/ }
```
SaravShah: So only prod-03 VM has access to the `resque` route?

tallenaz: Yep, the dashboard will be set up on stage-02 and prod-03 -- the worker machines.

SaravShah:
seems like it would be a good idea to use advanced constraints?? We wouldn’t have to hard code in the VM’s (for any reason they change) and then we can create a class (like you mentioned above) and set up the matcher method like “hey, does this host match the worker_vm?”
```
class WorkerConstraint
  def matches?(host)
    /prod_03|stage-02/ =~ Setting.something
  end
end
```

what do others think?

tallenaz:
I think there's another approach to document here. If we decide to alias the worker machines with a name that distinguishes them as such, we can use that as a match for the constraint. Aliasing the prod machines is on deck for a story time discussion, so the results of that may have implications here...

config/routes.rb
ndushay: I thought it was:   
- prod-01:   ReST (including Resque dashboard) and M2C, C2M
- prod-02:  CV
- prod-03:  workers

???

tallenaz:
Closing this for a PR that uses an advanced constraint with Settings to restrict the dashboard setup, and that adds configuration to wire up the application to resque/redis. Will link back to this ticket

PR #763, opened 2018-05-15
Additional logging for when we change status 

### LINKED ISSUE #752: (CV?) status change and logging for UNEXPECTED_VERSION_ON_STORAGE_STATUS
we noticed a bunch of objects in `PreservedCopy::UNEXPECTED_VERSION_ON_STORAGE_STATUS`, but no corresponding `AuditResults::UNEXPECTED_VERSION` message.

we currently suspect the following sequence of events to be the problem:
- update catalog robot wasn't hooked up for a while, and was also briefly talking to the wrong hostname for pres cat.
  - as such, there are moabs that have had a version increase on disk, but the version increase was not properly reflected in the catalog.
- CV (which we've been running) detects and sets `UNEXPECTED_VERSION_ON_STORAGE_STATUS`, but does not properly add the corresponding audit result.
  - as such, we suspect that CV found some moabs where the catalog version lagged the version on disk.  rightly from its perspective, it set the status to unexpected version, but it didn't add the corresponding audit result.

if our above understanding is correct, it seems like the right thing might be to do the `AuditResult` addition in the shared `MoabValdiationHandler` mixin (likely in the `set_status_as_seen_on_disk` method, since that's where the status setting happens).  and then we could remove the inline audit result addition from the `CatalogToMoab` and `PreservedObjectHandler` classes (well, we might not be able to remove all instances of the audit result addition in POH, just the ones that go along with `set_status_as_seen_on_disk`).

SaravShah:
When status is set to `PreservedCopy::UNEXPECTED_VERSION_ON_STORAGE`, we want to log more information than just the status information. We want to also long the corresponding `AuditResults::UNEXPECTED_VERSION` message.

closes #752

### DIFF
--- a/app/services/checksum_validator.rb
+++ b/app/services/checksum_validator.rb
@@ -36,6 +36,9 @@ class ChecksumValidator
         results.add_result(AuditResults::MOAB_CHECKSUM_VALID)
         found_expected_version = moab.current_version_id == preserved_copy.version
         set_status_as_seen_on_disk(found_expected_version)
+        unless found_expected_version
+          results.add_result(AuditResults::UNEXPECTED_VERSION, actual_version: moab.current_version_id, db_obj_name: 'PreservedCopy', db_obj_version: preserved_copy.version)
+        end
       else
         update_status(PreservedCopy::INVALID_CHECKSUM_STATUS)
       end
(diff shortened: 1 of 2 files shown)


jmartin-sul:
@ndushay good catch on the audit check name, we should be setting that in the `CV.initialize` method when the `AuditResults` instance is created.  filed an issue for that:  https://github.com/sul-dlss/preservation_catalog/issues/767

jmartin-sul:
https://github.com/sul-dlss/preservation_catalog/issues/768 for the refactoring suggestion

Issue #737, opened 2018-05-07
(R) Which ZIP lib to use?


SaravShah:
Should we use Zlib, rubyzip, or shell out?

Story Time Notes:
https://docs.google.com/document/d/10HMhTMWlZaM1DJChFr2r3coOVDZKe0EDAH8PPumb-2E/edit#

atz:
According to notes, previous rough implementation used:
```
zip -v0r -s 10g
```
Some reading:
- https://wiki.debian.org/ReproducibleBuilds/Howto
- https://content.pivotal.io/blog/barriers-to-deterministic-reproducible-zip-files 
- Notes about our intended structure [here](https://docs.google.com/document/d/1ofCfmeZ_lx8cpMUFbCb1Bt2R3TSna48oPuQT8YphSvg/edit#heading=h.y9oczez8cqg3)

Takeaway:
- We should also use `-X` or equivalent to avoid putting access-timestamps in the zip

atz:
My local zip:
```
 zip -v
Copyright (c) 1990-2008 Info-ZIP - Type 'zip "-L"' for software license.
This is Zip 3.0 (July 5th 2008), by Info-ZIP.
Currently maintained by E. Gordon.  Please send bug reports to
the authors using the web page at www.info-zip.org; see README for details.

Latest sources and executables are at ftp://ftp.info-zip.org/pub/infozip,
as of above date; see http://www.info-zip.org/ for other sites.

Compiled with gcc 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34) for Unix (Mac OS X) on Oct 14 2016.

Zip special compilation options:
	USE_EF_UT_TIME       (store Universal Time)
	SYMLINK_SUPPORT      (symbolic links supported)
	LARGE_FILE_SUPPORT   (can read and write large files on file system)
	ZIP64_SUPPORT        (use Zip64 to store large files in archives)
	STORE_UNIX_UIDs_GIDs (store UID/GID sizes/values using new extra field)
	UIDGID_16BIT         (old Unix 16-bit UID/GID extra field also used)
	[encryption, version 2.91 of 05 Jan 2007] (modified for Zip 3)
```

Point being `zip` is a decade old and that *is* the most recent release.  While that might seem like a drawback, in this case it suggests incredible stability.  Compare to [`rubyzip`](https://github.com/rubyzip/rubyzip), which has 19 releases in the same period.

atz:
Using:
```sh
[...]

PR #771, opened 2018-05-17
Add code, fixture, spec for zipmaker worker

### LINKED ISSUE #707: (R) Implement ZipMaker Worker (blocked by #737)
- How to create zip file? ( Are using lib zip or shell out ) 
  - Ask Julian or ops what he thinks we should use
  - Zip with no compression
  - Main concern is reproducibility for a zip file.

- Does Zip already exist? 
  - if yes do a unix touch (updates timestamp) - can be implemented later since we aren't cleaning up old zips yet (we think clean up will remove least-recently used first)
  - if no, create a zip
    - and queue the message for zip maker success (already done)

tallenaz:
Fixes #707 

  paired with @SaravShah on rspec, with some of @ndushay 's help 

  This PR assumes a zip storage file system layout that doesn't include the bucket name, and may need some team discussion to decide whether it's correct.

  The form of the zip command draws from https://github.com/sul-dlss/preservation_catalog/issues/737#issuecomment-388504059

### DIFF
--- a/app/jobs/zipmaker_job.rb
+++ b/app/jobs/zipmaker_job.rb
@@ -10,26 +10,26 @@ class ZipmakerJob < ApplicationJob
   # @param [String] druid
   # @param [Integer] version
   def perform(druid, version)
-    binary_path = Moab::StorageServices.object_version_path(druid, version)
+    moab_version_path = Moab::StorageServices.object_version_path(druid, version)
     zip_path = DruidVersionZip.new(druid, version).file_path
     unless File.exist?(zip_path)
-      ZipmakerJob.zip_binary(zip_path, binary_path) if binary_path
+      ZipmakerJob.create_zip!(zip_path, moab_version_path) if moab_version_path
     end
     PlexerJob.perform_later(druid, version)
   end
 
-  # @param [String] path to druid version zip
-  # @param [String] path to druid version
+  # @param [String] path to zip file to be made
+  # @param [String] path to druid moab version directory
   # @todo calculate md5 of zip for plexer
-  def self.zip_binary(zip_path, binary_path)
-    _output, error, status = Open3.capture3(zip_command(zip_path, binary_path))
+  def self.create_zip!(zip_path, moab_version_path)
+    _output, error, status = Open3.capture3(zip_command(zip_path, moab_version_path))
     raise "zipmaker failure #{error}" unless status.success?
   end
 
-  # @param [String] path to druid version zip
-  # @param [String] path to druid version
+  # @param [String] path to zip file to be made
+  # @param [String] path to druid moab version directory
   # @return [String]
-  def self.zip_command(zip_path, binary_path)
-    "zip -vr0X -s 10g #{zip_path} #{binary_path}"
+  def self.zip_command(zip_path, moab_version_path)
+    "zip -vr0X -s 10g #{zip_path} #{moab_version_path}"
   end
 end
--- a/app/jobs/zipmaker_job.rb
+++ b/app/jobs/zipmaker_job.rb
@@ -1,3 +1,4 @@
+require 'open3'
 # Responsibilities:
 # locate files
 # zip files
@@ -9,6 +10,26 @@ class ZipmakerJob < ApplicationJob
   # @param [String] druid
   # @param [Integer] version
   def perform(druid, version)
+    binary_path = Moab::StorageServices.object_version_path(druid, version)
+    zip_path = DruidVersionZip.new(druid, version).file_path
+    unless File.exist?(zip_path)
+      ZipmakerJob.zip_binary(zip_path, binary_path) if binary_path
+    end
     PlexerJob.perform_later(druid, version)
+  end
+
+  # @param [String] path to druid version zip
+  # @param [String] path to druid version
+  # @todo calculate md5 of zip for plexer
+  def self.zip_binary(zip_path, binary_path)
+    _output, error, status = Open3.capture3(zip_command(zip_path, binary_path))
+    raise "zipmaker failure #{error}" unless status.success?
+  end
+
+  # @param [String] path to druid version zip
+  # @param [String] path to druid version
+  # @return [String]
+  def self.zip_command(zip_path, binary_path)
+    "zip -vr0X -s 10g #{zip_path} #{binary_path}"
   end
 end
--- a/spec/jobs/zipmaker_job_spec.rb
+++ b/spec/jobs/zipmaker_job_spec.rb
@@ -3,7 +3,7 @@ require 'rails_helper'
 describe ZipmakerJob, type: :job do
   let(:druid) { 'bj102hs9687' }
   let(:version) { 1 }
-  let(:binary_path) { Moab::StorageServices.object_version_path(druid, version) }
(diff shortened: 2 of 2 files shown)


app/jobs/zipmaker_job.rb
```diff
@@ -9,6 +9,11 @@ class ZipmakerJob < ApplicationJob
   # @param [String] druid
   # @param [Integer] version
   def perform(druid, version)
+    binary_path = Moab::StorageServices.object_version_path(druid, version)
+    zip_path = "#{Settings.zip_storage}/transfers/#{druid}#{format('.v%04d.zip', version)}"
```
atz: I thought we would still use druid-tree subdirs for zip storage for performance reasons.  I don't think it should be necessary to do `#{Settings.zip_storage}/transfers`.  If a `transfers` (or any other) subdirectory is needed, it should be part of `Settings.zip_storage`.  This logic has to match (or ideally reuse) the zip-locating logic in `DruidVersionZip` or our pieces wont fit together.

When the zip exists we should `FileUtils.touch` the file to prevent it being cleaned out by whatever cache-maintenance code we write later.

ndushay: I would say dealing with zip files needing to be cleaned out should be a separate ticket;  o.w. I agree with @atz.

tallenaz:
I believe this is ready for another round of reviewing. 

Some things to note:

- We're punting on having the zipmaker generate an md5 until a later PR
- The worker determines `zip_path` via `DruidVersionZip`
- We're using `open3` instead of a `system` call
- Some additional specs to check for whether the system level `zip` utility morphs.

app/jobs/zipmaker_job.rb
```diff
@@ -9,6 +10,26 @@ class ZipmakerJob < ApplicationJob
...
   def perform(druid, version)
+    binary_path = Moab::StorageServices.object_version_path(druid, version)
+    zip_path = DruidVersionZip.new(druid, version).file_path
+    unless File.exist?(zip_path)
+      ZipmakerJob.zip_binary(zip_path, binary_path) if binary_path
+    end
     PlexerJob.perform_later(druid, version)
   end
+
+  # @param [String] path to druid version zip
```
ndushay: super nit-picky;  looking at these comments with fresh eyes, they are a bit confusing.  how about something more like
- "path to zip file we are making" 
and
- "path to druid's Moab version directory" 
???

Here and below ...

app/jobs/zipmaker_job.rb
```diff
@@ -9,6 +10,26 @@ class ZipmakerJob < ApplicationJob
...
+  # @param [String] path to druid version
+  # @todo calculate md5 of zip for plexer
+  def self.zip_binary(zip_path, binary_path)
```
ndushay: would a better name for this method be ...   create_zip ?     Why "binary"?

app/jobs/zipmaker_job.rb
```diff
@@ -9,6 +10,26 @@ class ZipmakerJob < ApplicationJob
   # @param [String] druid
   # @param [Integer] version
   def perform(druid, version)
+    binary_path = Moab::StorageServices.object_version_path(druid, version)
```
ndushay: why "binary"?  what do you think of "moab_version_path" ?

(binary appears in a lot of places here ... )

ndushay: Also:  instead of passing the moab version path and zip path around .. should they be attributes?   Memoized methods, with druid and version being the attributes?

Because I noticed you are passing the same arguments to 2 methods.

spec/jobs/zipmaker_job_spec.rb
```diff
@@ -3,13 +3,77 @@
...
+
+    describe 'succeeds in zipping the binary' do
+      it 'does not raise an error' do
+        expect { described_class.zip_binary(zip_path, binary_path) }.not_to raise_error
+        File.delete(zip_path)
+      end
+    end
+
+    describe 'fails to zip the binary' do
+      describe 'when inpath is incorrect' do
```
ndushay: would this be the same as "when the moab version path doesn't exist"?

spec/jobs/zipmaker_job_spec.rb
```diff
@@ -3,13 +3,77 @@
...
+    let(:version) { 2 }
+
+    describe 'succeeds in zipping the binary' do
+      it 'does not raise an error' do
+        expect { described_class.zip_binary(zip_path, binary_path) }.not_to raise_error
+        File.delete(zip_path)
+      end
+    end
+
+    describe 'fails to zip the binary' do
```
ndushay: fails to create the zip file ?

spec/jobs/zipmaker_job_spec.rb
```diff
@@ -3,13 +3,77 @@
...
+    it 'invokes PlexerJob' do
+      expect(PlexerJob).to receive(:perform_later).with(druid, version)
+      described_class.perform_now(druid, version)
+    end
+  end
+
+  describe '.zip_binary' do
+    let(:version) { 2 }
+
+    describe 'succeeds in zipping the binary' do
```
ndushay: "succeeds in zipping the moab version directory" ?

spec/jobs/zipmaker_job_spec.rb
```diff
@@ -3,13 +3,77 @@
...
+  end
 
   it 'descends from ApplicationJob' do
     expect(described_class.new).to be_an(ApplicationJob)
   end
 
-  it 'invokes PlexerJob' do
-    expect(PlexerJob).to receive(:perform_later).with(druid, version)
-    described_class.perform_now(druid, version)
+  describe 'zip already exists in zip storage' do
```
ndushay: shouldn't it still call the plexer, even if the zip already exists?

app/jobs/zipmaker_job.rb
atz: It is markedly harder to do attributes on jobs, since you don't get to control their initialization.  I think what you are suggesting is total overkill.

app/jobs/zipmaker_job.rb
atz: Or rather, `create_zip!`, since we `raise` unless successful.

spec/jobs/zipmaker_job_spec.rb
```diff
@@ -3,13 +3,77 @@
...
+
+    context 'succeeds in zipping the binary' do
+      it 'does not raise an error' do
+        expect { described_class.zip_binary(zip_path, binary_path) }.not_to raise_error
+        File.delete(zip_path)
+      end
+    end
+
+    context 'fails to zip the binary' do
+      context 'when inpath is incorrect' do
```
atz: `inpath`?

tallenaz: I wasn't sure of what to call the argument, and found `inpath` in the zip manual:

```
Command format.  The basic command format is

              zip options archive inpath inpath ...

       where  archive  is a new or existing zip archive and inpath is a direc-
       tory or file path optionally including wildcards.
```

atz:
@ndushay The current code looks like it does call plexer reliably whether the zip is newly created or not.  The only time that doesn't happen is when zip creation fails (via exception), which seems logically correct.

PR #772, opened 2018-05-18
Zip checksum table

### LINKED ISSUE #754: (R) store archive zip checksums in pres cat database
we'll create a new table (`preserved_copy_zip_checksums`, or something less unwieldy).

the table will store an MD5 checksum for each archive zip we create.  so, for example, if druid `ab123cd4567` is on version 3, has an online moab, and has archive copies on one archive endpoint, it will have:
* no entries in `preserved_copy_zip_checksums` for the online pres copy.
* 3 entries in `preserved_copies`, 1 for each archived version.
  * at least 3 entries in `preserved_copy_zip_checksums`, one for each zip file that was generated along the way (e.g., if the zip file for a given version was re-created and re-pushed to an endpoint after being uploaded for the first time, it will have two entries in `preserved_copy_zip_checksums`, one for each zip that was created for the version).

so, at the least, `preserved_copy_zip_checksums` should have 3 fields:
* `preserved_copy_id`:  non-null FK to `preserved_copies.id`.
*  `MD5`:  the MD5 hash for the particular archive zip being logged.
* `creation_info`:  IMPORTANT: this is, e.g., the shell command and OS info under which the zip was created.  should we have to recreate a specific zip we created in the past (for purposes of checksum comparison), we should be able to use this info, as underlying system dependencies may shift over time, and the same input to the zip command may not produce the same output on different systems or under different configurations.

SaravShah:
Create new table that will store the `md5` checksum for each archive zip created. Table contains `md5`, `create_info` - which will have the shell command and OS info under which the zip was created), and the `preserved_copy_id` (preserved_copies has many zip_checksums)

Also introduce zip_checksum factory as the last commit, but the `zip_checksum_spec.rb` uses it before hand.

closes #754

### DIFF
--- a/app/models/zip_checksum.rb
+++ b/app/models/zip_checksum.rb
@@ -5,6 +5,6 @@ class ZipChecksum < ApplicationRecord
   belongs_to :preserved_copy
 
   validates :preserved_copy, presence: true
-  validates :md5, presence: true
+  validates :md5, presence: true, format: { with: /\A[0-9a-f]{32}\z/ }
   validates :create_info, presence: true
 end
--- a/app/models/zip_checksum.rb
+++ b/app/models/zip_checksum.rb
@@ -1,0 +1,10 @@
+##
+# ZipChecksum contains the MD5 checksum and creation info for every preserved copy (druid, version)
+# that has been archived.
+class ZipChecksum < ApplicationRecord
+  belongs_to :preserved_copy
+
+  validates :preserved_copy, presence: true
+  validates :md5, presence: true
+  validates :create_info, presence: true
+end
--- a/spec/models/zip_checksum_spec.rb
+++ b/spec/models/zip_checksum_spec.rb
@@ -1,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe ZipChecksum, type: :model do
+  let(:po) { create(:preserved_object) }
+  let(:pres_copy) { create(:preserved_copy, preserved_object: po, endpoint: Endpoint.first) }
+  let(:zip_checksum) do
+    ZipChecksum.create(md5: '387a8ds87ea87', create_info: 'shell command and OS info', preserved_copy: pres_copy)
+  end
+
+  it 'is not valid without valid attributes' do
+    zc = ZipChecksum.new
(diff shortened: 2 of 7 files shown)


app/models/zip_checksum.rb
```diff
@@ -0,0 +1,10 @@
+##
+# ZipChecksum contains the MD5 checksum and creation info for every preserved copy (druid, version)
+# that has been archived.
+class ZipChecksum < ApplicationRecord
+  belongs_to :preserved_copy
+
+  validates :preserved_copy, presence: true
+  validates :md5, presence: true
```
ndushay: we know more about md5 that we can validate for.   We could use:
- length validator
- format validator to ensure it is always a well-formed MD5 (more expensive to check than length, but not too bad?)
- uniqueness ?
- strict?

SaravShah: Strict Validation: You can also specify validations to be strict and raise ActiveModel::StrictValidationFailed when the object is invalid.

Do we want this?

ndushay: uh - how do we report now?  how would our reporting change w strict?

SaravShah: For now, I just added the format validator.

spec/models/zip_checksum_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  it 'is not valid unless it has all required attributes' do
+    zc = ZipChecksum.new(preserved_copy_id: pres_copy.id)
+    expect(zc).not_to be_valid
+    expect { zc.save! }.to raise_error(ActiveRecord::RecordInvalid)
+  end
+
+  it 'is valid with valid attributes' do
+    expect(zip_checksum).to be_valid
+  end
+  it { is_expected.to validate_presence_of(:md5) }
```
ndushay: dang it - you forgot to update the test ... I think?

SaravShah: do you mean, add a test that tests the format of the md5?

ndushay: yes, precisely.  :-)

PR #774, opened 2018-05-19
Plexer logic and tests

### LINKED ISSUE #709: (R) Implement Plexer Worker (blocked by #713)
- given a druid what endpoints should i live on?

- will need a scope and or a helper method on models? which we will use that query to help determine where the zip file should live (endpoints)

- also use that query to create the initial pres_copy before queueing to the zipmaker worker and at the end to see what we should have vs what we actually have (Not related to the implementation for the plexer worker, but we will be using this elsewhere.)

atz:
Initially based off John's branch, but I didn't end up using anything from it, so rebased on master.

Adds more factories.  

Closes #709

### DIFF
--- a/app/jobs/plexer_job.rb
+++ b/app/jobs/plexer_job.rb
@@ -11,6 +11,23 @@ class PlexerJob < ApplicationJob
   # @param [String] druid
   # @param [Integer] version
   def perform(druid, version)
-    S3EndpointDeliveryJob.perform_later(druid, version)
+    targets(druid, version).each do |worker|
+      worker.perform_later(druid, version)
+    end
+  end
+
+  # @return [Array<Class>] EndpointDeliveryBase-descending classes to be targeted
+  def targets(druid, version)
+    Endpoint
+      .joins(:endpoint_type, preserved_copies: [:preserved_object])
+      .where(
+        'endpoint_types.endpoint_class' => 'archive',
+        'preserved_objects.druid' => druid,
+        'preserved_copies.version' => version
+      )
+      .map do |ep|
+        Rails.logger.error("Archive Endpoint (id: #{ep.id}) has no delivery_class") unless ep.delivery_class
+        ep.delivery_class
+      end.compact
   end
 end
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -6,6 +6,10 @@ class Endpoint < ApplicationRecord
   belongs_to :endpoint_type
   has_and_belongs_to_many :preservation_policies
 
+  enum delivery_class: {
+    S3EndpointDeliveryJob => 1
+  }
+
   validates :endpoint_name, presence: true, uniqueness: true
   validates :endpoint_type, presence: true
   validates :endpoint_node, presence: true
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -6,6 +6,8 @@ class Endpoint < ApplicationRecord
   belongs_to :endpoint_type
   has_and_belongs_to_many :preservation_policies
 
+  # @note Hash values cannot be modified without migrating any associated persisted data.
+  # @see [enum docs] http://api.rubyonrails.org/classes/ActiveRecord/Enum.html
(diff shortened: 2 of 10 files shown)


app/models/endpoint.rb
```diff
@@ -6,6 +6,10 @@ class Endpoint < ApplicationRecord
   belongs_to :endpoint_type
   has_and_belongs_to_many :preservation_policies
 
+  enum delivery_class: {
```
SaravShah: add a comment saying:
` # NOTE: DO not change the underlying contents for enum values that have been merged to master?`

just as a reminder to anyone viewing this code?
what do you think?

spec/factories/preserved_copy.rb
```diff
@@ -3,7 +3,7 @@
     version 1
     status 'ok'
     size 231
-    endpoint { "fake endpoint" }
+    endpoint
```
SaravShah: yay!

app/jobs/plexer_job.rb
```diff
@@ -11,6 +11,23 @@ class PlexerJob < ApplicationJob
...
   # @param [Integer] version
   def perform(druid, version)
-    S3EndpointDeliveryJob.perform_later(druid, version)
+    targets(druid, version).each do |worker|
+      worker.perform_later(druid, version)
+    end
+  end
+
+  # @return [Array<Class>] EndpointDeliveryBase-descending classes to be targeted
+  def targets(druid, version)
```
SaravShah: shouldn't the preservation_policy be used somewhere here? (cause thats where we will know which endpoint to send the druid/version?

SaravShah: talked F2F - the plexer worker should already know which endpoint to target based on the pres_copy rows being existing. The preservation_policy logic can be implemented else where (like if it is not `default` and it is only `s3-west`)

jmartin-sul: i think the approach here is "any archive endpoint that has a preserved copy for the given druid/version combo", under the assumption that the preserved copies have already been created in `UNREPLICATED_STATUS` (according to the endpoints specified by the pres policy).  so... i think this is ok?

SaravShah: yup

app/models/endpoint.rb
SaravShah: yeah, and we did it w/ the status enum's so just trying to stay consistent without commenting

atz: I would delete the comments in the other location if consistency is required.  I dislike the other comment because it seems like notes scribbled in the margin of a textbook example.  IMO, the merged-to-master-ness of the `enum` isn't something that needs to be warned about... particularly in code that is *already* merged to master.  Code should never have to know what branch it is on.  Also, the branch 100% doesn't matter: the blocker is persisted **data** attached to the enum that would require migration.

The scenario of a maintainer familiar w/ specifying only the keys is one where there is *even less* freedom to modify the code, since the order of elements must be maintained.  Not sure why that person would now feel inclined to change integer values.  (We don't put similar warnings in our migration files or the schema definition for the table, though the constraints are even stronger there.)

I can add a link to the docs page for active-record enums, if that suffices.  It is appropriate to document dependencies in the app and point to clarifying docs.  But to the extent that we are using common dependencies in a highly conventional manner, IMO, it is mildly inappropriate for our comments to assume the role of educating others (or ourselves!) how those common dependencies themselves work in conventional application.  That isn't directly part of the app and we aren't really the authority on it anyway!

jmartin-sul: fair points.  i like your wording better than what i'd left on the other enum, so thanks for changing it there too 🙂.  that comment was indeed conflating a merge to master with the existence of data that needs migration.

anyway, i pretty much entirely agree, other than that i don't feel bad occasionally stating what might be obvious or conventional -- in this case, i'm just imagining a rails neophyte or someone who's been away for a while not thinking through the consequences of changing the mapping.  maybe that's exceptionally unlikely, making this comment noise, but my gut feeling was that this was an appropriate place for some over-explanation.

PR #776, opened 2018-05-21
Add resque config and add routes for dashboard

tallenaz:
Previous discussion at https://github.com/sul-dlss/preservation_catalog/pull/761

This adds resque config files to wire the app up to redis/resque, and restricts the resque dashboard to what we specify via a setting. 

If the settings approach is agreeable, I'll have PRs in to shared configs that set up the resque dashboards in stage/prod

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -45,3 +45,7 @@ c2m_sql_limit: 1000
 checksum_algos: ['md5'] # 'sha1' 'sha256'
 
 zip_storage: '/tmp' # override in #{RAILS_ENV}.yml
+
+resque_dashboard_hostnames: # tells the router where to mount the resque dashboard
+  - 'worker-hostname-01.example.com'
+  - 'worker-hostname-02.example.com'
--- a/config/deploy/stage.rb
+++ b/config/deploy/stage.rb
@@ -5,4 +5,4 @@ Capistrano::OneTimeKey.generate_one_time_key!
 set :rails_env, 'production'
 set :bundle_without, 'deploy test'
 set :deploy_to, '/opt/app/pres/preservation_catalog'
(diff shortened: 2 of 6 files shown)


config/settings.yml
```diff
@@ -45,3 +45,7 @@ c2m_sql_limit: 1000
 checksum_algos: ['md5'] # 'sha1' 'sha256'
 
 zip_storage: '/tmp' # override in #{RAILS_ENV}.yml
+
+worker_dashboards:
```
ndushay: would like a comment clarifying what these are for, and/or change this key to be something more descriptive, like `resque_dashboard_hostnames`

ndushay:
Also, see:  https://github.com/resque/resque-web -- looks like we may want to go in that direction?

config/settings.yml
tallenaz: - I included the commented out config files that are used in shared_configs to wire resque up, to give the reader some added context, but if they're confusing, I'm happy to leave them out. 

- I don't know if we need to limit resque route actions. From a glance at the dashboard, I don't see any "dangerous" looking tabs.

tallenaz:
I'm unable to get the `sassc` gem dependency for `resque-web` installed. There's an open issue on that gem where it can't be installed on CentOs 6.x, which is what we've got on the current vms.

tallenaz:
@ndushay this is ready for another round of review. I've avoided commented out config files by opting to set up the initializer in this repo, and a resque config file in development.

AFAICT, we can't get `resque-web` going b/c it is not installable in CentOS 6.

I'm in favor of ticketing the investigation of restricting the dashboard routes further -- after a brief look, I don't see anything dangerously open.

PR #796, opened 2018-05-23
target archive endpoints followons pt 1

### LINKED ISSUE #215: why does **Endpoint** have default_storage_root_endpoint_type method?
why isn't this in EndpointType  as  default_endpoint_type ??

At miminum, it might be good to add a comment in the code as I may not be the only one confused.

### LINKED ISSUE #793: (R, etc) misc follow on work from target archive endpoints PR (#759)
kind of a grab bag, we can break out separate tickets if we want, though i think these things are all small enough and related enough that doing them all at once in a single PR would still be sane.

* [x] move `Endpoint.default_storage_root_endpoint_type` to `EndpointType.default_for_storage_root`
* [x] make `EndpointType.endpoint_class` an enum
* [x] test for `Endpoint.which_need_archive_copy` that passes SQL injection attempt for `version` param
* [x] index `endpoint_types.type_name` and `.endpoint_class`; `preserved_copies.version`; `preserved_objects.current_version`
* [x] ~~test that `create_archive_preserved_copies` does not create pres copy if it already exists~~ (_obviated by #798_: can't be done if there's a unique constraint on druid/version/endpoint combo; method queries for endpoints that need a druid/version combo, and creates the preserved copies that don't exist, all in one shot in a transaction.  even if the code tried to insert dupes, e.g. bad query logic, the DB would reject them.)
* [x] test that `create_archive_preserved_copies` creates copies with `UNREPLICATED_STATUS`

spawned by #759

jmartin-sul:
follow-on work from #759

connects to #793
closes #215

### DIFF
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -70,11 +70,6 @@ class Endpoint < ApplicationRecord
     end
   end
 
-  # TODO: move to EndpointType class?  e.g. .default_for_storage_root
-  def self.default_storage_root_endpoint_type
-    EndpointType.find_by!(type_name: Settings.endpoints.storage_root_defaults.endpoint_type_name)
-  end
-
   def to_h
     {
       endpoint_name: endpoint_name,
--- a/app/models/endpoint_type.rb
+++ b/app/models/endpoint_type.rb
@@ -1,11 +1,19 @@
 ##
 # metadata about a specific replication endpoint
 class EndpointType < ApplicationRecord
+  # TODO: this should eventually become a regular int Enum.  String Enums work and are tested in Rails 5,
+  # but are neither documented nor officially supported, and may be de-supported in the future.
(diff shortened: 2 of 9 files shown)


app/models/endpoint.rb
```diff
@@ -70,11 +70,6 @@ def self.seed_archive_endpoints_from_config(preservation_policies)
     end
   end
 
-  # TODO: move to EndpointType class?  e.g. .default_for_storage_root
-  def self.default_storage_root_endpoint_type
-    EndpointType.find_by!(type_name: Settings.endpoints.storage_root_defaults.endpoint_type_name)
-  end
-
```
ndushay: See ticket #215 -- I raised this question a while back.

ndushay: See ticket #215 -- I raised this question a while back.

jmartin-sul: sorry it took me so long to see the error of my ways and circle back 🙂

PR #804, opened 2018-05-26
touchups to Endpoint & PreservedObject tests, some indexes, some comment touchups and other misc refinement

### LINKED ISSUE #793: (R, etc) misc follow on work from target archive endpoints PR (#759)

jmartin-sul:
closes #793

### DIFF
--- a/app/jobs/plexer_job.rb
+++ b/app/jobs/plexer_job.rb
@@ -21,9 +21,9 @@ class PlexerJob < ApplicationJob
     Endpoint
       .joins(:endpoint_type, preserved_copies: [:preserved_object])
       .where(
-        'endpoint_types.endpoint_class' => 'archive',
-        'preserved_objects.druid' => druid,
-        'preserved_copies.version' => version
+        endpoint_types: { endpoint_class: 'archive' },
+        preserved_objects: { druid: druid },
+        preserved_copies: { version: version }
       )
       .map do |ep|
         Rails.logger.error("Archive Endpoint (id: #{ep.id}) has no delivery_class") unless ep.delivery_class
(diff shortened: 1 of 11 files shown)


jmartin-sul:
@atz makes a convincing case that indexing pres copy version and pres object current version isn't worth the cost (because, e.g., many many things will all map to version 1).

will fix that and then take off the WIP tag...

jmartin-sul:
ready for review again...

app/jobs/plexer_job.rb
```diff
@@ -21,9 +21,9 @@ def targets(druid, version)
     Endpoint
       .joins(:endpoint_type, preserved_copies: [:preserved_object])
       .where(
-        'endpoint_types.endpoint_class' => 'archive',
-        'preserved_objects.druid' => druid,
-        'preserved_copies.version' => version
+        endpoint_types: { endpoint_class: 'archive' },
+        preserved_objects: { druid: druid },
+        preserved_copies: { version: version }
```
atz: 👍

PR #808, opened 2018-05-30
Update moab-versioning to v4.2.2

tallenaz:


jmartin-sul:
i saw that we were on 806 yesterday, and i was hoping to get [808](https://en.wikipedia.org/wiki/Roland_TR-808) for myself, but i suppose i'll have to settle for the merge commit 🥁

Issue #810, opened 2018-05-31
(C2M, M2C) make sure cron jobs ran successfully


jmartin-sul:
we scheduled the running of checks on a regular basis.  we should make sure they ran.

ndushay:
Two concerns: 

1.  both M2C and C2M ran at the same time.  I must've messed up the cron job/whenever:

#### M2C ####

```
[pres@preservation-catalog-prod-01 log]$ cat m2c.log.1
I, [2018-05-26T05:00:04.055431 #18961]  INFO -- : 2018-05-26T12:00:04Z M2C check_existence for all storage roots starting'
I, [2018-05-26T05:00:04.055596 #18961]  INFO -- : 2018-05-26T12:00:04Z M2C check_existence starting for '/services-disk02/sdr2objects'
I, [2018-05-27T02:42:48.576926 #18961]  INFO -- : 2018-05-27T09:42:48Z M2C check_existence ended for '/services-disk02/sdr2objects'
I, [2018-05-27T02:42:48.576992 #18961]  INFO -- : 2018-05-27T09:42:48Z M2C check_existence starting for '/services-disk03/sdr2objects'
I, [2018-05-27T13:50:10.978088 #18961]  INFO -- : 2018-05-27T20:50:10Z M2C check_existence ended for '/services-disk03/sdr2objects'
I, [2018-05-27T13:50:10.978142 #18961]  INFO -- : 2018-05-27T20:50:10Z M2C check_existence starting for '/services-disk04/sdr2objects'
I, [2018-05-27T22:55:27.568950 #18961]  INFO -- : 2018-05-28T05:55:27Z M2C check_existence ended for '/services-disk04/sdr2objects'
I, [2018-05-27T22:55:27.569046 #18961]  INFO -- : 2018-05-28T05:55:27Z M2C check_existence starting for '/services-disk05/sdr2objects'
I, [2018-05-28T09:56:47.122212 #18961]  INFO -- : 2018-05-28T16:56:47Z M2C check_existence ended for '/services-disk05/sdr2objects'
I, [2018-05-28T09:56:47.122317 #18961]  INFO -- : 2018-05-28T16:56:47Z M2C check_existence starting for '/services-disk06/sdr2objects'
I, [2018-05-28T16:46:19.739198 #18961]  INFO -- : 2018-05-28T23:46:19Z M2C check_existence ended for '/services-disk06/sdr2objects'
[...]

PR #811, opened 2018-05-31
schedule: cron updates

ndushay:
Turns out you can't use ranges and commas in same expression in a crontab.   I'm also not 100% certain about Sat vs. 6 for day of week.

Connects to #810


tallenaz:
Github hasn't caught up yet with coveralls, but coveralls says all is well.

PR #818, opened 2018-06-02
Update more timestamps in checksum validator

### LINKED ISSUE #688: (CV) set all validation timestamps
`ChecksumValidator` should also update `last_moab_validation` and `last_version_audit` (in addition to `last_checksum_validation`), since it now validates those things.

tallenaz:
Fixes #688 

I'm open to adding to tests, if they seem insufficient, say, for a change in timestamp value, or testing that timestamps don't get updated in the else condition.

### DIFF
--- a/spec/services/checksum_validator_spec.rb
+++ b/spec/services/checksum_validator_spec.rb
@@ -239,6 +239,15 @@ RSpec.describe ChecksumValidator do
         expect(pres_copy.reload.status).to eq PreservedCopy::OK_STATUS
       end
 
+      it 'updates audit timestamps' do
+        expect(pres_copy.last_moab_validation).to be nil
+        expect(pres_copy.last_version_audit).to be nil
+        approximate_validation_time = Time.current
+        cv.validate_checksums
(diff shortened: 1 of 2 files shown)


spec/services/checksum_validator_spec.rb
```diff
@@ -239,6 +239,11 @@
         expect(pres_copy.reload.status).to eq PreservedCopy::OK_STATUS
       end
 
+      it 'updates audit timestamps' do
+        expect(pres_copy).to receive(:update_audit_timestamps).with(true, true)
```
jmartin-sul: i'd be inclined to query for the pres copy and check that the timestamps had the expected values (i.e. newer than when `cv.validate_checksums` got called), as opposed to checking that the update method got called.

ndushay: good suggestion, @jmartin-sul - check the desired outcome, not the methodology

tallenaz:
@jmartin-sul @ndushay this is ready for another round -- the tests no longer rely on the method called. I wonder if there's a better way to capture the time cv takes place, but I think this works OK.

PR #822, opened 2018-06-04
M2C and C2M crons only run on correct Saturdays

ndushay:
It turns out this cron:
```
0 5 1-7 * 6
```
runs each day numbered between 1-7 as well as on every Saturday.   So I took a kludge from the interwebs to fix it,  so we *only* run M2C and C2M on the appropriate Saturdays.

I also tweaked the README:
- added a linked TOC
- made the Headings more skimmable
- moved the seeding section down

Connects to #810

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,22 @@
 
 Rails application to track, audit and replicate archival artifacts associated with SDR objects.
 
-## Getting Started
+## Table of Contents
 
-### PostgreSQL
+* [Getting Started](#getting-started)
+    * [PostgreSQL](#postgreSQL)
+    * [Redis](#redis)
+* [Usage Instructions](#usage)
+    * [Moab to Catalog](#m2c) (M2C) existence/version check
+    * [Catalog to Moab](#c2m) (C2M) existence/version check
+    * [Checksum Validation](#cv) (CV)
+    * [Seed the catalog](#seeding)
+* [Development](#development)
+* [Deploying](#deploying)
+
+## <a name="getting-started"/>Getting Started
+
+### <a name="postgresSQL"/>PostgreSQL
 
 #### Installing Postgres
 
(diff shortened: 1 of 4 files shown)


config/schedule.rb
```diff
@@ -1,22 +1,24 @@
 # Use this file to easily define all of your cron jobs.
 # Learn more: http://github.com/javan/whenever
 
-# run these task only on servers with the correct roles in Capistrano
-# these tasks append to existing logs
-every '0 5 1-7 * 6', roles: [:m2c] do
+# also overriding b/c whenever adds --silent (https://github.com/javan/whenever/issues/524, #541)
+job_type :sat_only_rake, "cd :path && bin/is_it_saturday.sh && :environment_variable=:environment :bundle_command rake :task :output"
```
atz: I don't know if it is an issue here (can't see the file mode in github view), but `bash bin/is_it_saturday.sh` will succeed when the script is just readable, whereas direct invocation requires it to be executable.

atz: Ah, OK, if I go view the file itself, GitHub shows it as an `Executable File`, so no problem.

README.md
```diff
@@ -206,7 +172,66 @@ this will generate a log at, for example, `log/profile_cv_validate_disk_all_endp
...
+RAILS_ENV=production bundle exec rake drop[fixture_sr1]
+```
+
+#### Populate the catalog:
+
+```sh
+RAILS_ENV=production bundle exec rake populate[fixture_sr1]
+```
+
+## <a name="development"/>Development
```
atz: Headings automatically have anchors added, so when the anchor isn't changing, it is redundant to add explicitly.  Github already renders the markdown with an anchor, e.g. for `### Redis`, like:
```html
<h3><a id="user-content-redis" class="anchor" aria-hidden="true" href="#redis"></a>Redis</h3>
```

ndushay: ah - I wasn't sure, and couldn't think of a good way to test without pushing to github.

jmartin-sul: merged, filed #832 for this loose end.

PR #823, opened 2018-06-05
[WIP] M2C, C2M, CV ensure they are only for online copies

### LINKED ISSUE #778: ensure M2C, C2M, CV queries only for online objects
That is, we don't want to do these things on archival PreservedCopy objects

ndushay:
WIP - replaced by PRs #836 and #839

~~[WIP - what is needed for ChecksumValidation may affect this]~~

Assure that our existing M2C, C2M and CV checks only affect online PreservedCopy objects, using @jmartin-sul 's the elegant solution.

Closes #778

### DIFF
--- a/app/services/checksum_validator.rb
+++ b/app/services/checksum_validator.rb
@@ -1,6 +1,15 @@
 # code for validating Moab checksums
 class ChecksumValidator
   include ::MoabValidationHandler
+  include ActiveModel::Validations
+
+  validates_each :endpoint do |record, attr, value|
+    if value.is_a?(Endpoint)
+      record.errors.add(attr, "must be an online Endpoint for #{name}") unless value.endpoint_type.online?
+    else
+      record.errors.add(attr, 'must be an actual Endpoint')
+    end
+  end
 
   attr_reader :results, :endpoint, :full_druid, :preserved_copy, :bare_druid
 
@@ -22,6 +31,7 @@ class ChecksumValidator
     @preserved_copy = preserved_copy
     @bare_druid = preserved_copy.preserved_object.druid
     @endpoint = Endpoint.find_by(endpoint_name: endpoint_name)
+    valid?
     @results = AuditResults.new(bare_druid, nil, endpoint, 'validate_checksums')
     @full_druid = "druid:#{bare_druid}"
   end
--- a/app/services/preserved_object_handler.rb
+++ b/app/services/preserved_object_handler.rb
@@ -9,12 +9,15 @@ class PreservedObjectHandler
   include ::MoabValidationHandler
   include ActiveModel::Validations
 
-  # Note: supplying validations here to allow validation before use, e.g. incoming_version in numeric logic
   validates :druid, presence: true, format: { with: DruidTools::Druid.pattern }
   validates :incoming_version, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :incoming_size, numericality: { only_integer: true, greater_than: 0 }
   validates_each :endpoint do |record, attr, value|
-    record.errors.add(attr, 'must be an actual Endpoint') unless value.is_a?(Endpoint)
+    if value.is_a?(Endpoint)
+      record.errors.add(attr, "must be an online Endpoint for #{name}") unless value.endpoint_type.online?
+    else
+      record.errors.add(attr, 'must be an actual Endpoint')
+    end
   end
 
   attr_reader :druid, :incoming_version, :incoming_size, :endpoint, :results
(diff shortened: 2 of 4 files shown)


app/services/checksum_validator.rb
```diff
@@ -1,6 +1,15 @@
 # code for validating Moab checksums
 class ChecksumValidator
   include ::MoabValidationHandler
+  include ActiveModel::Validations
+
```
ndushay: I'm not in love with the syntax for the below, but it gets the job done.  Improvements welcome.

app/services/checksum_validator.rb
```diff
@@ -1,6 +1,15 @@
 # code for validating Moab checksums
 class ChecksumValidator
   include ::MoabValidationHandler
+  include ActiveModel::Validations
```
ndushay: It seemed preferable to add in validations and perform it in the initialize method, since there are multiple public methods that could be called ...

app/services/preserved_object_handler.rb
```diff
@@ -9,12 +9,15 @@ class PreservedObjectHandler
   include ::MoabValidationHandler
   include ActiveModel::Validations
 
-  # Note: supplying validations here to allow validation before use, e.g. incoming_version in numeric logic
   validates :druid, presence: true, format: { with: DruidTools::Druid.pattern }
   validates :incoming_version, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :incoming_size, numericality: { only_integer: true, greater_than: 0 }
   validates_each :endpoint do |record, attr, value|
-    record.errors.add(attr, 'must be an actual Endpoint') unless value.is_a?(Endpoint)
+    if value.is_a?(Endpoint)
```
ndushay: ditto on syntax:  I'm not in with this, but it gets the job done. Improvements welcome.

atz:
In general, model validations should not cross AR associations (that cause additional DB queries), since model validations have to validate the **in memory** state of an object as a precondition of getting to the database.  

This is one of the arguments for subsuming `EndpointType` back into `Endpoint`.

spec/services/checksum_validator_spec.rb
```diff
@@ -7,10 +7,8 @@
   let(:endpoint_name) { "fixture_sr3" }
   let(:endpoint) { Endpoint.find_by(endpoint_name: endpoint_name) }
   let(:object_dir) { "#{endpoint.storage_location}/#{DruidTools::Druid.new(druid).tree.join('/')}" }
-  let(:pres_copy) do
-    po = PreservedObject.find_by(druid: druid)
-    PreservedCopy.find_by(preserved_object: po, endpoint: endpoint)
-  end
+  let(:po) { PreservedObject.find_by(druid: druid) }
```
atz: All these seeded `find_by` calls would be better served as factory `create` statements (removing the need for many of them).

spec/services/checksum_validator_spec.rb
```diff
@@ -526,4 +524,27 @@
...
   end
+
+  describe 'endpoint validation' do
+    it 'errors when endpoint is not an Endpoint object' do
+      cv = described_class.new(pres_copy, 'foo')
+      cv.valid?
+      expect(cv.errors.messages).to include(endpoint: ["must be an actual Endpoint"])
+    end
+    it 'errors when endpoint_type is not online' do
+      endpoint_type = EndpointType.create(type_name: 'foo', endpoint_class: 'archive')
```
atz: Why not use a factory?  We already have a special one for this case, like:
```ruby
let(:endpoint) { create(:archive_endpoint) }
...
  cv = described_class.new(pc, endpoint.endpoint_name)
```
Then you don't need to create `endpoint_type` directly at all.

app/services/checksum_validator.rb
```diff
@@ -1,6 +1,15 @@
...
+
+  validates_each :endpoint do |record, attr, value|
+    if value.is_a?(Endpoint)
```
atz: Is this object valid if the Endpoint isn't also valid?  Type-checking and cascading validity checks are the kind of thing that an ActiveRecord association would deliver inherently.

app/services/checksum_validator.rb
```diff
@@ -22,6 +31,7 @@ def initialize(preserved_copy, endpoint_name)
     @preserved_copy = preserved_copy
     @bare_druid = preserved_copy.preserved_object.druid
     @endpoint = Endpoint.find_by(endpoint_name: endpoint_name)
+    valid?
```
atz: If this is important mid-initialization, shouldn't it `raise` if invalid?  It seems counterproductive to front-load the validity check and then do nothing with the result.  `endpoint` is not assignable by the caller, so they have no possibility of putting the object into a valid state.  Seems likely they shouldn't have the object at all.

Alternatively, check validity in an `after_initialize` callback, and raise there if necessary.  

The way we have to lookup `Endpoint` by name **per PC** is a performance waste.  Callers should just pass the `Endpoint` themselves, which gives `initialize` the chance to interrogate it for correctness.  Even if we don't subsume `EndpointType` (and we should!), we should add a `Endpoint#archive_type?` that caches the answer, so that the combination of passing the `Endpoint` and calling that gives us:
- the validation we want
- saves millions of `find_by` queries

atz: Upon reflection, the method isn't even necessary if we pass the Endpoint object, since the associated EndpointType object will be cached with it.

ndushay:
Some excellent points and suggestions, Joe, thanks.  Completely dopey of me to not actually check the result of `valid?` - d'oh.

ndushay:
this PR replaced by PRs #836 and #839.  Nothing to see here;  move along.

PR #825, opened 2018-06-05
only set ok status if all validations (incl CV) have passed

### LINKED ISSUE #690: MoabValidationHandler#set_status_as_seen_on_disk should set status to unknown when it falls through and checksums haven't been validated
at the moment, when this method is called from M2C or C2M, it can be less than honest if the validations that those classes perform pass, but checksum validation would fail.  since C2M and M2C can't call out to checksum validation (which we schedule to be less frequent due to it being more resource intensive), the best they can do is set status to validity unknown.

see this WIP branch:  https://github.com/sul-dlss/preservation_catalog/compare/master...cv-fail-upd-pc-status--honest-set_status_as_seen_on_disk?expand=1

and this comment (and PR in general) for more context/explanation:  https://github.com/sul-dlss/preservation_catalog/pull/664#issuecomment-382567704

jmartin-sul:
otherwise, status can be set to any errors that may have been found (e.g. `'invalid_moab'`, `'unexpected_version_on_storage'`), or `'validity_unknown'` if e.g. no errors were found but checksum validation wasn't run.

closes #690

### DIFF
--- a/app/services/moab_validation_handler.rb
+++ b/app/services/moab_validation_handler.rb
@@ -73,7 +73,15 @@ module MoabValidationHandler
       return
     end
 
-    # TODO: this still isn't quite honest in the cases where checksum validation hasn't been performed
-    update_status(PreservedCopy::OK_STATUS)
+    # NOTE: subclasses which override this method should NOT perform checksum validation inside of this method!
+    # CV is expensive, and can run a while, and this method should likely be called from within a DB transaction,
+    # but CV definitely shouldn't happen inside a DB transaction.
+    if results.contains_result_code?(AuditResults::MOAB_CHECKSUM_VALID)
+      update_status(PreservedCopy::OK_STATUS)
+    elsif can_validate_checksums?
+      update_status(PreservedCopy::INVALID_CHECKSUM_STATUS)
+    else
+      update_status(PreservedCopy::VALIDITY_UNKNOWN_STATUS)
+    end
   end
 end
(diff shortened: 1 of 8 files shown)


app/services/moab_validation_handler.rb
```diff
@@ -73,7 +73,15 @@ def set_status_as_seen_on_disk(found_expected_version)
...
 
-    # TODO: this still isn't quite honest in the cases where checksum validation hasn't been performed
-    update_status(PreservedCopy::OK_STATUS)
+    # NOTE: subclasses which override this method should NOT perform checksum validation inside of this method!
+    # CV is expensive, and can run a while, and this method should likely be called from within a DB transaction,
+    # but CV definitely shouldn't happen inside a DB transaction.
+    if results.contains_result_code?(AuditResults::MOAB_CHECKSUM_VALID)
+      update_status(PreservedCopy::OK_STATUS)
+    elsif can_validate_checksums?
+      update_status(PreservedCopy::INVALID_CHECKSUM_STATUS)
```
tallenaz: I'm not clear on lines 81-2. I see that `can_validate_checksums?` is false, and so the update on 82 won't happen. I'm wondering why that update is called at all, unless there's a case where `can_validate_checksums?` returns true, which I don't see. Can you give me some context here? Otherwise, it LGTM.

atz: I guess the idea is that a subclass could override that method and return true.

atz: Ah, yeah, I see `ChecksumValidator` does [exactly this](https://github.com/sul-dlss/preservation_catalog/blob/master/app/services/checksum_validator.rb#L63)

tallenaz: Ah ok, thanks!

PR #835, opened 2018-06-06
Second S3 endpoint: us-east-1

### LINKED ISSUE #788: (R) add addl AWS regional endpoint
- new worker class with a separate queue ("queue_as")
    - extend the other AWS class and use a diff queue name

- add new delivery worker as a value in the endpoint deliver class enum (NOTE:  worker is getting creds supplied via ENV;  we would need to specify ENV var at worker startup (e.g. rake take invocation - `rake resque:work`)

- add an endpoint that has the enum value and is named appropriately

- get creds for AWS new region into puppet (can be split out into separate ticket)

atz:
closes #788

### DIFF
--- a/spec/jobs/s3_east_delivery_job_spec.rb
+++ b/spec/jobs/s3_east_delivery_job_spec.rb
@@ -1,6 +1,8 @@
+require 'rails_helper'
+
 describe S3EastDeliveryJob, type: :job do
-  it 'descends from S3EndpointDeliveryJob' do
-    expect(described_class.new).to be_an(S3EndpointDeliveryJob)
+  it 'descends from S3WestDeliveryJob' do
+    expect(described_class.new).to be_an(S3WestDeliveryJob)
   end
 
   it 'uses a different queue' do
--- a/spec/jobs/s3_east_delivery_job_spec.rb
+++ b/spec/jobs/s3_east_delivery_job_spec.rb
@@ -1,0 +1,9 @@
+describe S3EastDeliveryJob, type: :job do
+  it 'descends from S3EndpointDeliveryJob' do
+    expect(described_class.new).to be_an(S3EndpointDeliveryJob)
+  end
+
+  it 'uses a different queue' do
+    expect(described_class.new.queue_name).to eq 's3_us_east_1_delivery'
+  end
+end
--- a/spec/models/endpoint_spec.rb
+++ b/spec/models/endpoint_spec.rb
@@ -1,48 +1,32 @@
 require 'rails_helper'
 
 RSpec.describe Endpoint, type: :model do
-
-  let!(:endpoint_type) { EndpointType.create(type_name: 'aws', endpoint_class: 'archive') }
+  let(:default_pres_policies) { [PreservationPolicy.default_policy] }
+  let(:endpoint_type) { build(:endpoint_type, type_name: 'aws', endpoint_class: 'archive') }
   let!(:endpoint) do
-    Endpoint.create(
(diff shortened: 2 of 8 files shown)


spec/jobs/s3_east_delivery_job_spec.rb
```diff
@@ -0,0 +1,9 @@
+describe S3EastDeliveryJob, type: :job do
```
SaravShah: if you run `be rspec spec/job/s3_east_delivery_job_spec.rb` by it self, it fails. is it bc this is inheriting from s3_west, so it needs to run s3_west_spec before the s3_east_spec is ran?

spec/models/endpoint_spec.rb
```diff
@@ -1,48 +1,36 @@
...
+      :endpoint,
       endpoint_name: 'aws-us-east-2',
-      endpoint_type_id: endpoint_type.id,
+      endpoint_type: endpoint_type,
       endpoint_node: 's3.us-east-2.amazonaws.com',
       storage_location: 'sdr-bucket-01'
     )
   end
 
-  it 'is not valid without valid attributes' do
```
SaravShah: 👍

spec/models/endpoint_spec.rb
```diff
@@ -175,15 +161,10 @@
...
-    it 'does not allow SQL injection' do
-      # a bit too whitebox, but oh well...  specifically testing version field
-      # because we know we're not using the usual AR params for bind vars approach
-      # in the subquery.  we cast version.to_i, which will silently convert anything
-      # it can't figure out how to properly translate into zero.
-      bogus_version = "0)))); GRANT ALL PRIVILEGES ON DATABASE pres TO icanhaznjctionattk; DROP TABLE students; --"
-      generated_sql = Endpoint.which_need_archive_copy(druid, bogus_version).to_sql
-      expect(generated_sql).not_to match(/#{Regexp.escape(bogus_version)}/)
-      expect(generated_sql).to match(/"preserved_copies"."version" = 0/)
+    it 'Casts version to integer' do
```
ndushay: I would like to at least retain a comment about why we care (avoid SQL injection b/c if the version is being cast to an integer, it isn't possible to deliver any valuable SQL payload via it) in the code itself.

spec/jobs/s3_east_delivery_job_spec.rb
atz: ah, just forgot to require `rails_helper`, thx

atz:
Fixed both of those points, thanks.

spec/models/endpoint_spec.rb
```diff
@@ -175,15 +161,11 @@
...
-    it 'does not allow SQL injection' do
-      # a bit too whitebox, but oh well...  specifically testing version field
-      # because we know we're not using the usual AR params for bind vars approach
-      # in the subquery.  we cast version.to_i, which will silently convert anything
-      # it can't figure out how to properly translate into zero.
-      bogus_version = "0)))); GRANT ALL PRIVILEGES ON DATABASE pres TO icanhaznjctionattk; DROP TABLE students; --"
-      generated_sql = Endpoint.which_need_archive_copy(druid, bogus_version).to_sql
-      expect(generated_sql).not_to match(/#{Regexp.escape(bogus_version)}/)
-      expect(generated_sql).to match(/"preserved_copies"."version" = 0/)
+    # since we build AREL subquery, the cast is a guarantee against SQL injection
```
ndushay: 👍

PR #836, opened 2018-06-07
M2C, C2M: ensure they are only for online pres_copies

ndushay:
Ensure that our existing M2C and C2M checks only affect *online* PreservedCopy objects, using @jmartin-sul 's the elegant solution.

supersedes the M2C and C2M part of PR #823

Connected to #778

### DIFF
--- a/spec/services/preserved_object_handler_spec.rb
+++ b/spec/services/preserved_object_handler_spec.rb
@@ -114,7 +114,7 @@ RSpec.describe PreservedObjectHandler do
   describe 'endpoint validation' do
     it 'errors when endpoint is not an Endpoint object' do
       poh = described_class.new(druid, incoming_version, incoming_size, 1)
-      poh.valid?
+      expect(poh).to be_invalid
       expect(poh.errors.messages).to include(endpoint: ["must be an actual Endpoint"])
     end
     it 'errors when endpoint_type is not online' do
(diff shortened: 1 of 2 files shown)


spec/services/preserved_object_handler_spec.rb
```diff
@@ -106,4 +106,26 @@
...
+      expect(poh.errors.messages).to include(endpoint: ["must be an actual Endpoint"])
+    end
+    it 'errors when endpoint_type is not online' do
+      endpoint_type = EndpointType.create(type_name: 'foo', endpoint_class: 'archive')
+      archival_endpoint = Endpoint.create(endpoint_name: 'archive1',
+                                          endpoint_type_id: endpoint_type.id,
+                                          endpoint_node: 'node',
+                                          storage_location: 'somewhere')
+      poh = described_class.new(druid, incoming_version, incoming_size, archival_endpoint)
+      poh.valid?
```
atz: For these, how about:
`expect(poh).not_to be_valid` ?

That calls `.valid?` **and** makes sure it is returning the correct value.  Since we will be checking `.valid?` (or `.invalid?`) before we ever go looking at `.errors`, it seems reasonable to have expectations on it.

ndushay:
@SaravShah rails magic - when you validate, if it fails an ActiveXXX::XXXError is raised, which will show the errors listed.

ndushay:
@atz thanks for the 👀 ;  I've updated those specs per your suggestion.

spec/services/preserved_object_handler_spec.rb
```diff
@@ -110,4 +110,25 @@
...
+      poh = described_class.new(druid, incoming_version, incoming_size, 1)
+      expect(poh).to be_invalid
+      expect(poh.errors.messages).to include(endpoint: ["must be an actual Endpoint"])
+    end
+    it 'errors when endpoint_type is not online' do
+      endpoint_type = EndpointType.create(type_name: 'foo', endpoint_class: 'archive')
+      archival_endpoint = Endpoint.create(endpoint_name: 'archive1',
+                                          endpoint_type_id: endpoint_type.id,
+                                          endpoint_node: 'node',
+                                          storage_location: 'somewhere')
```
atz: You could replace both these 5 lines with one factory invocation.  We have one just for archival endpoints.
```ruby
archival_endpoint = create(:archive_endpoint)
```

jmartin-sul: https://github.com/sul-dlss/preservation_catalog/pull/873

PR #838, opened 2018-06-07
Fix #728 profiler_spec: no rails, do cleanup, order independence

### LINKED ISSUE #728: rspec spec/lib/profiler_spec.rb fails out of order due to dependency on Rails `Time.current`
Cannot run `rspec spec/lib/profiler_spec.rb` in isolation.

Error is:
```
  1) Profiler#print_results_flat returns the printer and prints to the path we pass in
     Failure/Error: expected_filepath = "log/profile_test#{Time.current.localtime.strftime('%FT%T')}-flat.txt"
     
     NoMethodError:
       undefined method `current' for Time:Class
     # ./spec/lib/profiler_spec.rb:23:in `block (3 levels) in <top (required)>'
```

`Time.current` is a [Rails-ism](https://apidock.com/rails/Time/current/class), so this test only succeeds in a given order relative to other tests that have loaded Rails.  

Relates to #724.

atz:
`Time.current` relies on Rails, but this `lib` class doesn't have any guarantee of Rails being loaded. Here I convert the call to use only the functionality available from the un-enhanced Ruby class.

This resolves a test order-dependency and test-independence failure (the spec could not previously run in isolation).

Would you prefer not to have 200 log files leftover by running specs on your local machine?  Me, too. I also made the test clean up after itself.

Fix #728

### DIFF
--- a/lib/profiler.rb
+++ b/lib/profiler.rb
@@ -1,13 +1,16 @@
 # adapted from argo
 # takes a block of code, starts the profiler, runs the code, then
 # stops the profiler and returns the results of the profiling.
-# example usage:
+# @example usage
 #  profiler = Profiler.new
 #  profiler.prof { MoabToCatalog.seed_catalog_for_dir(storage_dir) }
 #  profiler.print_results_flat(out_file_id)
 class Profiler
-
   attr_accessor :results
+
+  def output_path(out_file_id)
+    "log/profile_#{out_file_id}#{Time.now.utc.localtime.strftime('%FT%T')}-flat.txt" # start w/ UTC per rubocop
+  end
 
   def prof
     RubyProf.start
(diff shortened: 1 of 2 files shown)


atz:
Not a refactor, an actual already documented bug.

ndushay:
I called it a refactor because I don't view "must be able to run tests in random order" nor "must be able to run tests independently" as a group-agreed requirement.

atz:
Tests must be able to run independently, or they are not repeatable (or reusable).  Tests *should* cleanup after themselves.  

I don't think these principles require any kind of novel group interpretation.  For example, Sarav recently _correctly_ requested changes on a PR of mine because the new spec file would not pass when run individually (I had forgotten to require rails_helper).  That is the standard already.

More to the point, this code will FAIL if invoked without Rails loaded. As a regular `lib` file, Profiler should have been able to be used independently (not everything we want to profile is a Rails operation), but it currently cannot be.  That is a real bug being fixed here that has nothing to do with test *anything*.

atz:
And not to belabor the point, but that bug was **concealed** by order-dependent test execution when random order would have exposed it.  That doesn't mean random order is **required**, it means that it is helpful for revealing bogus assumptions made in code.

PR #840, opened 2018-06-07
Use dor-workflow-service for rest calls

### LINKED ISSUE #809: use dor-workflow-service gem for workflow ReST calls
Audit checks currently use faraday to talk directly to the workflow service.   It would be better, infrastructure-wise, to use the dor-workflow-service gem (as we do in robots):
- more robust if workflow service changes
- lets us know w/ gem dependency graph which code is using workflow services

This is especially useful if we add additional calls to workflow service (e.g.  when we change from some sort of invalid preserved_copy status to "ok")

tallenaz:
Fixes #809 

  - installs and configures `dor-workflow-service`
  - replaces a lot of homegrown faraday code with a call to `dor-workflow-service`, with a spec (for a method that unexceptionally returns `true`)
  - adds to `Settings` (would be nice to get the actual workflow service URL out of there)
  - makes other specs aware of the `dor-workflow-service` call

### DIFF
--- a/app/services/workflow_errors_reporter.rb
+++ b/app/services/workflow_errors_reporter.rb
@@ -3,44 +3,11 @@ require 'retries'
 # send errors to preservationAuditWF workflow for an object via ReST calls.
 class WorkflowErrorsReporter
 
+  # this method will always return true because of the dor-workflow-service gem
+  # see issue sul-dlss/dor-workflow-service#50 for more context
   def self.update_workflow(druid, process_name, error_message)
-    response = http_workflow_request(druid, process_name, error_message)
-    if response
-      if response.status == 204
-        Rails.logger.debug("#{druid} - sent error to workflow service for preservationAuditWF #{process_name}")
-      else
-        # Note: status == 400 will be handled by the rescue clause
-        Rails.logger.error("#{druid} - unable to update workflow for preservationAuditWF #{process_name}: #{response.body}. Error message: #{error_message}")
-      end
-    end
-  rescue StandardError => e
-    Rails.logger.error("#{druid} - unable to update workflow for preservationAuditWF #{process_name} #{e.inspect}. Error message: #{error_message}")
-  end
-
-  private_class_method def self.http_workflow_request(druid, process_name, error_message)
-    return unless conn
-    handler = proc do |exception, attempt_number, total_delay|
-      Rails.logger.debug("Handler saw a #{exception.class}; retry attempt #{attempt_number}; #{total_delay} seconds have passed.")
-    end
-    with_retries(max_tries: 3, handler: handler, rescue: Faraday::Error) do
-      conn.put do |request|
-        request_params(request, druid, process_name, error_message)
-      end
-    end
-  end
-
-  private_class_method def self.request_params(request, druid, process_name, error_message)
-    request.headers['content-type'] = "application/xml"
-    request.url  "/workflow/dor/objects/druid:#{druid}/workflows/preservationAuditWF/#{process_name}"
-    request.body = "<process name='#{process_name}' status='error' errorMessage='#{CGI.escapeHTML(error_message)}'/>"
-  end
-
-  private_class_method def self.conn
     if Settings.workflow_services_url.present?
-      @connection ||= Faraday.new(url: Settings.workflow_services_url) do |c|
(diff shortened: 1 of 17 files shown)


app/services/workflow_errors_reporter.rb
```diff
@@ -4,43 +4,8 @@
...
-    request.body = "<process name='#{process_name}' status='error' errorMessage='#{CGI.escapeHTML(error_message)}'/>"
-  end
-
-  private_class_method def self.conn
     if Settings.workflow_services_url.present?
-      @connection ||= Faraday.new(url: Settings.workflow_services_url) do |c|
-        c.use Faraday::Response::RaiseError
-        c.use Faraday::Adapter::NetHttp
-      end
+      Dor::WorkflowService.update_workflow_error_status('dor', "druid:#{druid}", 'preservationAuditWF', process_name, error_message)
```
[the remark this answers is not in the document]
SaravShah: oh the method returns `true` always, even if given wrong info? or returns a 400?

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -4,58 +4,20 @@
...
   let(:result) { "Invalid moab, validation error...ential version directories." }
-  let(:body) { "<process name='moab-valid' status='error' errorMessage='#{result}'/>" }
+  let(:body) { "<?xml version=\"1.0\"?>\n<process name=\"moab-valid\" status=\"error\" errorMessage=\"#{result}\"/>\n" }
   let(:druid) { 'jj925bx9565' }
   let(:process_name) { 'moab-valid' }
 
   context '.update_workflow' do
-    it '204 response' do
-      allow(Settings).to receive(:workflow_services_url).and_return('https://sul-lyberservices-test.stanford.edu/workflow/')
+    it 'returns true' do
```
ndushay: please add a comment that it returns true because that is what we ALWAYS get from dor-workflow-service gem and reference https://github.com/sul-dlss/dor-workflow-service/issues/50 in the comment.

app/services/workflow_errors_reporter.rb
```diff
@@ -4,43 +4,8 @@
 class WorkflowErrorsReporter
```
ndushay: Please add a method comment that this will always return true because that is what we ALWAYS get from dor-workflow-service gem and reference https://github.com/sul-dlss/dor-workflow-service/issues/50 in the comment.

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -4,58 +4,20 @@
   let(:full_url) do
```
ndushay: I would move these "let" statements to be variables inside the test, since they are only used in that one place now ...  YMMV

app/services/workflow_errors_reporter.rb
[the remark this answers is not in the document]
tallenaz: Yep, it always returns true...

spec/services/workflow_errors_reporter_spec.rb
```diff
@@ -4,58 +4,20 @@
...
-      allow(Settings).to receive(:workflow_services_url).and_return('https://sul-lyberservices-test.stanford.edu/workflow/')
+    it 'returns true' do
       stub_request(:put, full_url)
-        .with(body: body, headers: headers)
-        .to_return(status: 204, body: "", headers: {})
-      expect(Rails.logger).to receive(:debug).with("#{druid} - sent error to workflow service for preservationAuditWF moab-valid")
-      described_class.update_workflow(druid, process_name, result)
-    end
-
-    it '400 response' do
```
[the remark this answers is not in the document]
tallenaz: ....it always return true, but it does throw a `Dor::WorkflowException` against a bad url. I'll look into testing that.

app/services/workflow_errors_reporter.rb
```diff
@@ -3,44 +3,11 @@
 # send errors to preservationAuditWF workflow for an object via ReST calls.
 class WorkflowErrorsReporter
 
+  # this method will always return true because of the dor-workflow-service gem
+  # see issue sul-dlss/dor-workflow-service#50 for more context
```
ndushay: 👍

PR #855, opened 2018-06-09
Fix zipmaker bugs and gaps

### LINKED ISSUE #845: Zipmaker has to `mkdir -p` the destination directory for the zip
Currently the druid-tree directory structure doesn't exist in zip_storage, so we need to put it there. 

Use `zip.file_path` as the basis of the directory info.

Consider using a `before_perform` callback.

### LINKED ISSUE #849: Zip command must start from "inside" the druid tree
Incorrect Example:
```
zip -vr0X -s 10g here.zip spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001
```
Produces structure like:
```
unzip -l here.zip 
Archive:  here.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  04-20-2018 12:06   spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/
        0  04-20-2018 12:06   spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/data/
        0  04-20-2018 12:06   spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/data/content/
   905566  04-20-2018 12:06   spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/data/content/eric-smith-dissertation-augmented.pdf
  1000217  04-20-2018 12:06   spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/data/content/eric-smith-dissertation.pdf
        0  04-20-2018 12:06   spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/data/metadata/
     1325  04-20-2018 12:06   spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/data/metadata/contentMetadata.xml
     3882  04-20-2018 12:06   spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0001/data/metadata/descMetadata.xml

```
whereas we want just:
```
        0  04-20-2018 12:06   bj102hs9687/v0001/
        0  04-20-2018 12:06   bj102hs9687/v0001/data/
        0  04-20-2018 12:06   bj102hs9687/v0001/data/content/
   905566  04-20-2018 12:06   bj102hs9687/v0001/data/content/eric-smith-dissertation-augmented.pdf
...
```

After all, we don't want the replicated object to hardcode info about what storage root a MOAB lives on.  That isn't object data (just an artifact of our storage at a given moment in time).

atz:
- Move last auxiliary method from zipmaker to DruidVersionZip
- Handles chdir for zip creation
- Handles `mkdir -p` for zip destination directory
- Performs `touch` on existing zip file to prevent premature cleanup

Fix #849
Fix #845

### DIFF
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -1,3 +1,5 @@
+require 'open3'
+
 # Just a regular model, not an ActiveRecord-backed model
 class DruidVersionZip
   attr_reader :druid, :version
@@ -10,11 +12,27 @@ class DruidVersionZip
     @version = version
   end
 
+  # Changes directory so that the storage root (and druid tree) are not part of
+  # the archival directory structure, just the object.
+  def create_zip!
+    ensure_zip_directory!
+    Dir.chdir(work_dir.to_s) do
+      combined, status = Open3.capture2e(zip_command)
+      raise "zipmaker failure #{combined}" unless status.success?
+    end
+  end
+
+  # Ensure the directory the zip will live in exists
+  # @return [Pathname] the existing or created directory
+  def ensure_zip_directory!
+    Pathname.new(file_path).tap { |pn| pn.dirname.mkpath }
+  end
+
   # @return [String]
   # @see [S3 key name performance implications] https://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html
   # @example return 'ab/123/cd/4567/ab123cd4567/ab123cd4567.v0001.zip'
   def s3_key
-    druid.tree.join('/') + format(".v%04d.zip", version)
+    druid.tree.join('/') + ".#{v_version}.zip"
   end
 
   # @return [File] opened zip file
@@ -54,14 +72,29 @@ class DruidVersionZip
     @size ||= FileTest.size(file_path)
   end
 
+  # @return [String] "v" with zero-padded 4-digit version, e.g., v0001
+  def v_version
+    format("v%04d", version)
+  end
+
+  # @return [Pathname] The proper directory in which to execute zip_command
+  def work_dir
+    Pathname.new(moab_version_path).parent.parent
+  end
+
   # @return [String] shell command to unzip
   # def unzip_command
   #   "unzip #{file_path} -d #{place_to_unzip}"
   # end
 
+  # Presumes execution just "above" the druid dir in the druid tree, i.e. if the Moab is:
+  #   /storage_trunk_01/bj/102/hs/9687/bj102hs9687/v0003/...
+  # This command should execute from:
+  #   /storage_trunk_01/bj/102/hs/9687/
+  # @see #work_dir
   # @return [String] shell command to create this zip
   def zip_command
-    "zip -vr0X -s 10g #{file_path} #{moab_version_path}"
+    "zip -vr0X -s 10g #{file_path} #{druid.id}/#{v_version}"
   end
 
   # We presume the system guts do not change underneath a given class instance.
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -12,8 +12,10 @@ class DruidVersionZip
     @version = version
   end
 
+  # Creates a zip of Druid-Version content.
   # Changes directory so that the storage root (and druid tree) are not part of
-  # the archival directory structure, just the object.
+  # the archival directory structure, just the object, e.g. starting at 'ab123cd4567/...' directory,
+  # not 'ab/123/cd/4567/ab123cd4567/...'
   def create_zip!
     ensure_zip_directory!
     Dir.chdir(work_dir.to_s) do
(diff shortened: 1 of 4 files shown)


atz:
Need this to push stage replication further

spec/models/druid_version_zip_spec.rb
```diff
@@ -63,7 +106,7 @@
     let(:zip_path) { '/tmp/bj/102/hs/9687/bj102hs9687.v0001.zip' }
 
     it 'returns zip string to execute for this druid/version' do
-      expect(dvz.zip_command).to eq "zip -vr0X -s 10g #{zip_path} #{moab_version_path}"
+      expect(dvz.zip_command).to eq "zip -vr0X -s 10g #{zip_path} bj102hs9687/v0001"
```
SaravShah: don't need `let(:moab_version_path)` anymore

spec/models/druid_version_zip_spec.rb
```diff
@@ -1,8 +1,7 @@
 require 'rails_helper'
 
 describe DruidVersionZip do
-  subject(:dvz) { described_class.new(druid, version) }
```
SaravShah: why change it to `let`? just wondering

spec/models/druid_version_zip_spec.rb
atz: good point!

spec/models/druid_version_zip_spec.rb
atz: Because at one point I mock a piece of it, which rubocop-rspec dislikes.

spec/models/druid_version_zip_spec.rb
atz: Removed it.

app/models/druid_version_zip.rb
```diff
@@ -10,11 +12,27 @@ def initialize(druid, version)
     @version = version
   end
 
+  # Changes directory so that the storage root (and druid tree) are not part of
+  # the archival directory structure, just the object.
```
ndushay: What do you think about adding an example of the directory structure we're going for here?  The comment assumes the reader has the Moab druid-tree layout readily loaded into their human memory.

Also, what do you think about starting the comment "Creates zip for Moab version, but changes ..."  The comment as it stands reads a little orthogonal to the name of the method.

spec/models/druid_version_zip_spec.rb
```diff
@@ -25,6 +24,50 @@
     end
   end
 
+  describe '.create_zip!' do
```
ndushay: it seems we should have tests somewhere that show the directory structure inside the zip is what we want.

And possibly to show that it writes it to the right directory?  (-p option added)

atz: The specs are mainly concerned with the making the zip command what we expect.  They do not intend to test `zip` itself deeply.  While possible, testing one version of `zip` is meager reassurance, given that we have a system anticipating using multiple different versions of `zip` over time (and simultaneously!).  I'll put some work in this direction, but given the flexibility of the interface w/ the system utils, I don't expect to ever make this bulletproof.

Line 38 already does test that the zip is created in the right place, which implies that the necessary directory was created (else it would raise, like it did when we tried to use it in stage).

atz: @ndushay To do this correctly seems like it would require including a gem like `zip`... in test only, used only for one test.  Does that seem worthwhile?

ndushay: I guess what struck me is that the internal directory structure is really importan, so having a test to show it might be a good thing, long term maint-wise.  What if we just shell out to unpack it and then read the unpacked dir with ruby to check the structure is what we expect?

If we start having multiple zip commands, then the zipmaker code will have to accommodate ... so new code then implies additional tests then as well, no?

atz: Yeah, this is what I mentioned to you at close last night, I think I have a workaround (using `unzip -l ...` process return value like `ls`).  Pursuing that. 

We don't intend to have multiple zip **commands**, we expect that over time, the `zip` utility itself will be updated (or differ on how it was compiled on a given OS).

atz: added test of internal structure

app/models/druid_version_zip.rb
atz: IMO, I'm not assuming the reader knows anything about druid trees.  I'm saying only the object is preserved and no other structure.  

I agree this comment is more about the `chdir` than a description of the method itself, so I'll add a broader description.

PR #872, opened 2018-06-13
Add test to show sdk picking up credentials from files

### LINKED ISSUE #847: Investigate setting up > 1 set of aws credentials for worker machines

tallenaz:
through the AWS_PROFILE env var

  Fixes #847

### DIFF
--- a/spec/lib/preservation_catalog/s3_spec.rb
+++ b/spec/lib/preservation_catalog/s3_spec.rb
@@ -21,26 +21,73 @@ describe PreservationCatalog::S3 do
   end
 
   describe 'config' do
-    let(:config) { described_class.client.config }
-    let(:trio) do
-      {
-        'AWS_SECRET_ACCESS_KEY' => 'secret',
-        'AWS_ACCESS_KEY_ID' => 'some_key',
-        'AWS_REGION' => 'us-east-1'
-      }
+    context 'with access key and region env vars' do
+      let(:config) { described_class.client.config }
+      let(:envs) do
+        {
+          'AWS_SECRET_ACCESS_KEY' => 'secret',
+          'AWS_ACCESS_KEY_ID' => 'some_key',
+          'AWS_REGION' => 'us-east-1'
+        }
+      end
+
+      around do |example|
+        old_vals = envs.keys.zip(ENV.values_at(*envs.keys)).to_h
+        envs.each { |k, v| ENV[k] = v }
+        example.run
+        old_vals.each { |k, v| ENV[k] = v }
+      end
+      it 'pulls from ENV vars' do
+        expect(config.region).to eq 'us-east-1'
+        expect(config.credentials).to be_an(Aws::Credentials)
+        expect(config.credentials).to be_set
+        expect(config.credentials.access_key_id).to eq 'some_key'
+      end
     end
 
-    around do |example|
-      old_vals = trio.keys.zip(ENV.values_at(*trio.keys)).to_h
-      trio.each { |k, v| ENV[k] = v }
-      example.run
-      old_vals.each { |k, v| ENV[k] = v }
-    end
-    it 'pulls from ENV vars' do
-      expect(config.region).to eq 'us-east-1'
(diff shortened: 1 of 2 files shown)


spec/lib/preservation_catalog/s3_spec.rb
```diff
@@ -21,28 +21,80 @@
...
+      it 'pulls from ENV vars' do
+        expect(config.region).to eq 'us-east-1'
+        expect(config.credentials).to be_an(Aws::Credentials)
+        expect(config.credentials).to be_set
+        expect(config.credentials.access_key_id).to eq 'some_key'
+      end
     end
   end
+  #
+  #   context 'pointing the client to shared credentials' do
```
ndushay: why is this in here but commented out?

atz:
I don't think you can unilaterally update the global `Aws.config` without undoing it, since it bleeds over and causes the live S3 tests to fail.  

Also `DEPRECATION WARNING: called deprecated method `access_key_id' of an Aws::CredentialProvider, use #credentials instead`

tallenaz:
@ndushay @atz this is ready for review. The commented out stuff was work in progress that I've rebased, and I've reset the aws config object properly in an after block, and worked around the `credentials.access_key_id` deprecation with the `credentials.credentials.access_key_id` call.

spec/lib/preservation_catalog/s3_spec.rb
```diff
@@ -21,26 +21,81 @@
...
+          expect(config.region).to eq 'us-west-2'
+          expect(config.credentials.credentials.access_key_id).to eq 'foo'
+          expect(config.credentials.credentials.secret_access_key).to eq 'bar'
+        end
+      end
+      context 'profile us_east_1' do
+        let(:uno) do
+          {
+            'AWS_PROFILE' => 'us_east_1'
+          }
```
ndushay: this is a place where you could do

```ruby
let(:uno) { 'AWS_PROFILE' => 'us_east_1' }
```
I believe

spec/lib/preservation_catalog/s3_spec.rb
```diff
@@ -21,26 +21,81 @@
...
+
+      after do
+        Aws.config = {}
+      end
+
+      context 'profile us_west_2' do
+        let(:uno) do
+          {
+            'AWS_PROFILE' => 'us_west_2'
+          }
```
ndushay: see comment below ...

spec/lib/preservation_catalog/s3_spec.rb
tallenaz: rspec doesn't like that, but it will accept

```
let(:uno) { { 'AWS_PROFILE' => 'us_west_2' } }
```

I'm happy to change it to a one-liner that way, or leave it as is...

spec/lib/preservation_catalog/s3_spec.rb
tallenaz: see above....

spec/lib/preservation_catalog/s3_spec.rb
ndushay: meh.

atz: I prefer concision, but it isn't critical here.  `let(:uno) { Hash['AWS_PROFILE' => 'us_east_1'] }` works too, if you don't like the double curlies.  (Double curlies are fine by me.)  I might be tempted to rename all these (including the old `trio` as `envs`) since we treat them the same way.

spec/lib/preservation_catalog/s3_spec.rb
```diff
@@ -21,26 +21,81 @@
...
+        end
+
+        around do |example|
```
atz: Can move both these `around` blocks up one level and consolidate them.  This helps make it clear what each test is testing.

Or make all the `envs` names consistent and move ALL the `around` blocks to the top level.

tallenaz:
I avoided the double curlies with by calling `Hash`, and renamed environment variable `let`s to `envs`.

atz:
thx, Tony

PR #875, opened 2018-06-13
Update workflow to completed if PC.status is OK

SaravShah:
- This is a pretty long PR, so if you would like me to break it up please let me know
- Best to review this PR commit by commit
- last 2 commits are refactor commits, changing name of class and class method names in `WorkflowReporter` and switch the context block to describe block, for our RSpec tests. 

needs to deal with M2C and C2M still but created a story time question for that here -> (https://docs.google.com/document/u/1/d/10HMhTMWlZaM1DJChFr2r3coOVDZKe0EDAH8PPumb-2E/edit#) 

fixes #827

PR #881, opened 2018-06-14
line in between constants and first method, make string use in update…

jmartin-sul:
…_workflow_status call consistent w/ other strings in the class

using `COMPLETED` for `'completed'` seemed most consistent with the rest of the class, to my eyes.  happy to get rid of that if others think it's overkill.  would've made it a separate commit to make it easier to lop off, but two commits that small felt silly to me.

spawned by #875


tallenaz:
Just checked on coveralls manually, and it's OK.

PR #883, opened 2018-06-14
Follow-up removal of unnecessary method and attribute

atz:
MoabValidationHandler already does what we want here.

`full_druid` isn't used anywhere anymore.


jmartin-sul:
coveralls hadn't reported, but spec results showed coverage remained the same at 97.99%

Issue #877, opened 2018-06-14
(R) set delivery class on archive endpoints, create archive endpoints on stage


SaravShah:
currently, we have an `Endpoint#delivery_class`, but no way to set it when archive endpoints are seeded.

we will also have to add archive endpoints for stage (and eventually prod) to shared_configs.

jmartin-sul:
filed #892 for adding the prod endpoints.  closing this.

PR #887, opened 2018-06-14
Endpoint.seed_archive_endpoints_from_config should set delivery class

jmartin-sul:
update specs, update dev and test configs.

connects to #877

### DIFF
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -80,6 +80,7 @@ class Endpoint < ApplicationRecord
         endpoint.endpoint_node = endpoint_config.endpoint_node
         endpoint.storage_location = endpoint_config.storage_location
         endpoint.preservation_policies = preservation_policies
+        endpoint.delivery_class = Endpoint.delivery_classes[endpoint_config.delivery_class.constantize]
       end
     end
   end
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -17,6 +17,7 @@ archive_endpoints:
     endpoint_type_name: 'aws_s3'
     endpoint_node: 'localhost'
     storage_location: 'bucket_name'
+    delivery_class: 'S3WestDeliveryJob'
 workflow_services_url: 'https://sul-lyberservices-test.stanford.edu/workflow/'
 resque_dashboard_hostnames:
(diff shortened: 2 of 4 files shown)


app/models/endpoint.rb
```diff
@@ -80,6 +80,7 @@ def self.seed_archive_endpoints_from_config(preservation_policies)
         endpoint.endpoint_node = endpoint_config.endpoint_node
         endpoint.storage_location = endpoint_config.storage_location
         endpoint.preservation_policies = preservation_policies
+        endpoint.delivery_class = Endpoint.delivery_classes[endpoint_config.delivery_class.constantize]
```
jmartin-sul: i didn't expect the extra indirection of referring to the enum mapping, but when i just set equal to `endpoint_config.delivery_class.constantize`, i got this error in the specs:

```
ActiveRecord::StatementInvalid:
  PG::InvalidTextRepresentation: ERROR:  invalid input syntax for integer: "S3WestDeliveryJob"
  : INSERT INTO "endpoints" ("endpoint_name", "created_at", "updated_at", "endpoint_node", "storage_location", "endpoint_type_id", "delivery_class") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"
```

which makes me think AR is trying to cast the class to a string before doing the enum lookup and assignment?  i found that slightly unexpected.  anyway, i'd greatly prefer this indirection so that we can put the delivery class name in the settings, because putting raw int values in the settings seems error prone, and counter to the purpose of human readable enums.

would welcome suggestions that simplify this.

atz: I get the same behavior.  Your way is fine.  The idea of having actual classes as the enum keys is probably subverting some of the rails internals.  Like we definitely can't use the enum to do `endpoint.(::S3WestDeliveryJob)!` the same way we can do `pres_copy.ok!`.  

Maybe we should have strings in the enum and a wrapping getter that does the `constantize`?  I think that would make the guts work more conventionally.  But not a huge deal to me, since we will only ever be setting this value in one place.

jmartin-sul: filed #890 as backlog tech debt

PR #896, opened 2018-06-16
Add `PreservedCopy#replicate!` with tests

atz:
This is both a convenience method (all a caller has to know about is the PC) and the decision point for whether to put the object into the existing pipeline. Subsequently this would be where logic to reroute to a different pipeline ties in.

This is just a *part* of the overall strategy for large objects, but we can establish this piece independently.

### DIFF
--- a/app/models/preserved_copy.rb
+++ b/app/models/preserved_copy.rb
@@ -71,6 +71,16 @@ class PreservedCopy < ApplicationRecord
     # to 0 for nulls, which sorts before 1 for non-nulls, which are then sorted by last_checksum_validation)
   }
 
+  # Send to asynchronous replication pipeline
+  # @raise [RuntimeError] if object is unpersisted or too large (>=~10GB)
+  # @todo reroute to large object pipeline instead of raise
+  def replicate!
+    raise 'PreservedCopy must be persisted' unless persisted?
+    raise 'Only online PCs can be replicated' unless endpoint.endpoint_type.online?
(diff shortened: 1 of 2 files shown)


app/models/preserved_copy.rb
```diff
@@ -71,6 +71,15 @@ class PreservedCopy < ApplicationRecord
     # to 0 for nulls, which sorts before 1 for non-nulls, which are then sorted by last_checksum_validation)
   }
 
+  # Send to asynchronous replication pipeline
+  # @raise [RuntimeError] if object is unpersisted or too large (>=~10GB)
+  # @todo reroute to large object pipeline instead of raise
+  def preserve!
```
ndushay: it might be less confusing to rename this method `replicate!`

atz: yeah, I was torn between the two names.  `replicate!` is fine by me, though.

atz:
@ndushay: renamed

atz:
@jmartin-sul that doesn't quite make sense to me, since we have a has_many relationship to zip_checksums that includes size metadata.  So one PC can end up having been zipped at different times and one `size` doesn't necessarily apply to all its zips.  

In the context of this method though, it is expecting to be called on an `online` object (since those are the things we replicate: we don't currently replicate *from* one archive object to others).  

Really sick of the double duty modeling we do between archive/online types overall.  Should be different models, IMO.  

Anyway, I added a guard clause to enforce online-ness.

Issue #817, opened 2018-06-02
robots are creating preserved_copies with validity_unknown status


ndushay:
I believe that the (4479) objects with preserved_copy status of validity_unknown are because the catalog_controller is calling pohandler.create before a PreservedObject exists for the object.

Two thoughts:
- have catalog controller create PreservedObject before calling pohandler.create
OR
- have catalog_controller call pohandler.create_after_validation

ndushay:
the reason for NOT validating with robots is there is beaucoup validation happening as part of robot processing;  it was considered redundant and unnecessary processing.

ndushay:
@jmartin-sul argues that the object is validity unknown because we haven't run checksum validation yet.  

Thus, things are validity_unknown until we run CV.   As of 2018-06-01, it is unclear if we will be running CV regularly or not - so TBD

ndushay:
Further note:  using rake task to run CV manually on a couple of the validity_unknown druids made the count of said druids go down.

ndushay:
Per storytime today (https://docs.google.com/document/d/10HMhTMWlZaM1DJChFr2r3coOVDZKe0EDAH8PPumb-2E ), we discovered that the robots do checksum validation, so robots *should* set status to OK

ndushay:
From Storytime doc:  "Since prez robots *do* perform checksum validation for new Moabs / new versions of Moabs, we can let pohandler create and update_version actions 1) set the checksum validation timestamp to now and 2)  set the status to OK b/c we HAVE performed a flavor of checksum validation."

julianmorley:
Ideally, I'd like to see this set 'OK' on initial ingest (trust the Robots to not give us bad data) but then immediately add the new druid to the CV queue for validation (trust... but verify).

PR #897, opened 2018-06-16
[WIP] POH: optional flag to public update and create methods to indicate caller verified checksums

jmartin-sul:
called out some things i wasn't sure about in TODO comments.  did a little refactoring that felt like it made existing tests a little easier to read, and that i thought made the new tests easier to read (by showing more clearly what differed from parent contexts).  also got rid of an instance var in the specs.

another open question besides the TODOs about what timestamps to update:  do we queue druids for CV in the methods that got modified here?  should we just always add that functionality?  or should we let regularly scheduled CV deal with that as it gets to the least recently checked objects continuously?  @julianmorley [said](https://github.com/sul-dlss/preservation_catalog/issues/817#issuecomment-396678967):
> Ideally, I'd like to see this set 'OK' on initial ingest (trust the Robots to not give us bad data) but then immediately add the new druid to the CV queue for validation (trust... but verify).

connects to #817
(doesn't close it, because we need to wire these changes up to controllers, and have robots pass the right values in their calls to the controllers)


ndushay:
closing this in favor of #982

Issue #899, opened 2018-06-18
(R) can't get what needs to be replicated


SaravShah:
```
=> #<PreservedCopy:0x00000000078913e8
 id: 33,
 version: 3,
 preserved_object_id: 33,
 endpoint_id: 7,
 created_at: Thu, 07 Jun 2018 23:37:06 UTC +00:00,
 updated_at: Thu, 07 Jun 2018 23:37:06 UTC +00:00,
 last_moab_validation: Thu, 07 Jun 2018 23:37:06 UTC +00:00,
 last_checksum_validation: nil,
 size: 150321967,
 status: "validity_unknown",
 last_version_audit: Thu, 07 Jun 2018 23:37:06 UTC +00:00>
[44] pry(main)> pc.status = 'ok'
=> "ok"
[45] pry(main)> pc.save!
=> true
[46] pry(main)> pc.preserved_object.create_archive_preserved_copies(3)
=> []
[47] pry(main)> pc.preserved_object.create_archive_preserved_copies(2)
=> []
```

Maybe last_checksum_validation needs to be not nil? After running CV status was changed to `ok` as well.

```
[3] pry(main)> pc
=> #<PreservedCopy:0x0000000004447600
 id: 33,
 version: 3,
 preserved_object_id: 33,
 endpoint_id: 7,
 created_at: Thu, 07 Jun 2018 23:37:06 UTC +00:00,
 updated_at: Mon, 18 Jun 2018 21:19:32 UTC +00:00,
 last_moab_validation: Mon, 18 Jun 2018 21:19:32 UTC +00:00,
 last_checksum_validation: Mon, 18 Jun 2018 21:19:32 UTC +00:00,
 size: 150321967,
 status: "ok",
 last_version_audit: Mon, 18 Jun 2018 21:19:32 UTC +00:00>
[4] pry(main)> pc.preserved_object
=> #<PreservedObject:0x0000000006169d00
 id: 33,
 druid: "bc656gn1417",
 current_version: 3,
 created_at: Thu, 07 Jun 2018 23:37:06 UTC +00:00,
 updated_at: Thu, 07 Jun 2018 23:37:06 UTC +00:00,
 preservation_policy_id: 1>
[5] pry(main)> pc.preserved_object.create_archive_preserved_copies(2)
=> []
```

SaravShah:
I do have the preservation_policy set up correctly:
```
[15] pry(main)> Endpoint.archive_targets('bc656gn1417')
[...]

PR #915, opened 2018-06-20
C2A Logic. Checks the existence of s3 object and the md5.

### LINKED ISSUE #813: (C2A) audit check(s) for replicated objects; does replicated object exist? and match the MD5?
probably will spawn mult tickets:

- for each archival pres_copy in the catalog, does the replicated object exist?  (also match the MD5?)
    - scaffolding described below
    - aws call would live with the worker

SaravShah:
- Created a stub method in `PreservedObject` that will go off of what @atz  is working on (#899). `PreservedObject.check_endpoints!' will answer the question: `We have a PC, which endpoints were you supposed to replicate to. given the preservation policy? ` 
something like this: `whatever_logic_gets_me_PCs.each { |pc| EndpointCheckJob.perform_later(pc) }`
Ticket #920

- Introduced a helper method to `PreservedCopy` that will easily instantiate the `DruidVersionZip` and then delegate it to `s3_key` to get the file path of the zip. 
- Also add PC status, this will have to change once the online and archival PC's are broken out into separate tables. Put a TODO in there. 

fixes #813

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -59,6 +59,7 @@ Lint/RescueWithoutErrorClass:
 
 Metrics/AbcSize:
   Exclude:
+    - 'app/jobs/replication_check_job.rb' # [18.49/15] method still readable
     - 'app/models/preserved_object.rb' # methods still readable
     - 'app/services/checksum_validator.rb' # methods still readable; shameless green
     - 'app/services/preserved_object_handler.rb' # methods still readable; shameless green
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -59,7 +59,7 @@ Lint/RescueWithoutErrorClass:
 
 Metrics/AbcSize:
   Exclude:
-    - 'app/jobs/replication_check_job.rb' # [18.49/15] method still readable
+    - 'app/jobs/replicated_file_check_job.rb' # [18.49/15] method still readable
     - 'app/models/preserved_object.rb' # methods still readable
     - 'app/services/checksum_validator.rb' # methods still readable; shameless green
     - 'app/services/preserved_object_handler.rb' # methods still readable; shameless green
--- a/app/jobs/endpoint_check_job.rb
+++ b/app/jobs/endpoint_check_job.rb
@@ -1,26 +1,25 @@
-# Confirms existence of PreservedCopy on an endpoint
-# Notify ResultsRecorder, with success or failure of existence.
+# Confirms existence of PreservedCopy on an endpoint.
+# Confirms the MD5 checksum matches in database and s3.
 class EndpointCheckJob < ApplicationJob
+  # This queue is never expected to be used. For example:
+  # EndpointCheckJob.set(queue: :endpoint_check_us_west_2).perform_later(pc)
+  # EndpointCheckJob.set(queue: :endpoint_check_us_east_1).perform_later(pc)
   queue_as :endpoint_check
   delegate :bucket, :bucket_name, to: PreservationCatalog::S3
 
   # @param [PreservedCopy] verify that the archived preserved_copy exists on an endpoint
   def perform(preserved_copy)
-    return if preserved_copy.status == PreservedCopy::UNREPLICATED_STATUS # FIXME: What to do covered in storytime
+    return if preserved_copy.status == PreservedCopy::UNREPLICATED_STATUS # FIXME: storytime (Monday 07/02)
     aws_s3_object = bucket.object(preserved_copy.s3_key)
-    if aws_s3_object.exists?
-      stored_checksums = stored_checksums(preserved_copy)
-      replicated_checksum = replicated_checksum(aws_s3_object)
-      if stored_checksums.include?(replicated_checksum)
-        preserved_copy.update_status(PreservedCopy::OK_STATUS)
-        preserved_copy.update(last_checksum_validation: Time.zone.now)
-      else
-        preserved_copy.update_status(PreservedCopy::CHECKSUM_MISMATCH)
-      end
-    else
-      preserved_copy.update_status(PreservedCopy::FILE_NOT_FOUND)
+    stored_checksums = stored_checksums(preserved_copy)
+    replicated_checksum = replicated_checksum(aws_s3_object)
+    return preserved_copy.replicated_copy_not_found! unless aws_s3_object.exists?
+    unless stored_checksums.include?(replicated_checksum)
+      preserved_copy.update(last_checksum_validation: Time.zone.now)
+      return preserved_copy.invalid_checksum!
     end
-    preserved_copy.save!
+    preserved_copy.update(last_checksum_validation: Time.zone.now)
+    preserved_copy.ok!
   end
 
   # returns array
@@ -32,5 +31,4 @@ class EndpointCheckJob < ApplicationJob
   def replicated_checksum(aws_s3_object)
     aws_s3_object.metadata["checksum_md5"]
   end
-
 end
--- a/app/jobs/endpoint_check_job.rb
+++ b/app/jobs/endpoint_check_job.rb
@@ -1,0 +1,36 @@
+# Confirms existence of PreservedCopy on an endpoint
+# Notify ResultsRecorder, with success or failure of existence.
+class EndpointCheckJob < ApplicationJob
+  queue_as :endpoint_check
+  delegate :bucket, :bucket_name, to: PreservationCatalog::S3
+
+  # @param [PreservedCopy] verify that the archived preserved_copy exists on an endpoint
+  def perform(preserved_copy)
+    return if preserved_copy.status == PreservedCopy::UNREPLICATED_STATUS # FIXME: What to do covered in storytime
+    aws_s3_object = bucket.object(preserved_copy.s3_key)
+    if aws_s3_object.exists?
+      stored_checksums = stored_checksums(preserved_copy)
+      replicated_checksum = replicated_checksum(aws_s3_object)
+      if stored_checksums.include?(replicated_checksum)
+        preserved_copy.update_status(PreservedCopy::OK_STATUS)
+        preserved_copy.update(last_checksum_validation: Time.zone.now)
+      else
+        preserved_copy.update_status(PreservedCopy::CHECKSUM_MISMATCH)
+      end
+    else
+      preserved_copy.update_status(PreservedCopy::FILE_NOT_FOUND)
+    end
+    preserved_copy.save!
+  end
+
+  # returns array
+  def stored_checksums(preserved_copy)
+    preserved_copy.zip_checksums.map(&:md5)
+  end
+
+  # returns string
+  def replicated_checksum(aws_s3_object)
+    aws_s3_object.metadata["checksum_md5"]
+  end
+
+end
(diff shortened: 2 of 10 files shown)


app/models/preserved_copy.rb
```diff
@@ -71,6 +71,8 @@ class PreservedCopy < ApplicationRecord
     # to 0 for nulls, which sorts before 1 for non-nulls, which are then sorted by last_checksum_validation)
   }
 
+  delegate :s3_key, to: :druid_version_zip
```
atz: This should probably move up towards the top of the file, like line 27.

atz:
We'll need to do more than just check if the file exists.  We need to get the MD5 back from the metadata (header) and compare against `pc.zip_checksums.map(&:md5)` to make sure it matches something (valid) that we sent it.

SaravShah:
@atz right that makes sense, i'll add a commit to this PR. Thanks.

app/models/preserved_copy.rb
```diff
@@ -8,23 +8,30 @@ class PreservedCopy < ApplicationRecord
   UNEXPECTED_VERSION_ON_STORAGE_STATUS = 'unexpected_version_on_storage'.freeze
   VALIDITY_UNKNOWN_STATUS = 'validity_unknown'.freeze
   UNREPLICATED_STATUS = 'unreplicated'.freeze
+  FILE_NOT_FOUND = 'file_not_found'.freeze
+  CHECKSUM_MISMATCH = 'checksum_mismatch'.freeze
```
SaravShah: maybe rename this? db_s3_checksum_mismatch? i dunno..

app/jobs/endpoint_check_job.rb
```diff
@@ -0,0 +1,36 @@
+# Confirms existence of PreservedCopy on an endpoint
+# Notify ResultsRecorder, with success or failure of existence.
+class EndpointCheckJob < ApplicationJob
+  queue_as :endpoint_check
+  delegate :bucket, :bucket_name, to: PreservationCatalog::S3
+
+  # @param [PreservedCopy] verify that the archived preserved_copy exists on an endpoint
+  def perform(preserved_copy)
+    return if preserved_copy.status == PreservedCopy::UNREPLICATED_STATUS # FIXME: What to do covered in storytime
```
SaravShah: whoops will remove!

app/models/preserved_copy.rb
atz: This is one of those things that will only apply to archive PCs when that model is split out.  In that context, I think the meaning of the status is a bit clearer, but it will still be sorta wrong to think about it at this level.  The zip **part** is what will be not found or mismatched.  The APC might have parts not found and mismatched simultaneously.

Not a blocker, but worth other people on the team weighing in on, maybe not worth adding now if we know we have to nuke it soon.

app/jobs/endpoint_check_job.rb
```diff
@@ -0,0 +1,36 @@
+# Confirms existence of PreservedCopy on an endpoint
+# Notify ResultsRecorder, with success or failure of existence.
```
atz: This doesn't notify ResultsRecorder

SaravShah: Oops! thanks.

SaravShah:
Question: Do I need to add, the error statuses (checksum mismatch and file not found) to the `AuditResults` ?

app/models/preserved_copy.rb
SaravShah: Yeah, i have note in this PR, that when the model is split out to fix the statuses. But yeah any thoughts on renaming?

app/jobs/endpoint_check_job.rb
SaravShah: oh wait  - this is a story time question - i'll update the comment w/ the date of next story time

app/jobs/endpoint_check_job.rb
```diff
@@ -0,0 +1,36 @@
...
+class EndpointCheckJob < ApplicationJob
+  queue_as :endpoint_check
```
atz: In practice we will need to separate queues per AWS region (like the delivery workers).  We can represent that at the model level w/ different classes (like delivery workers), or we can just invoke this job like:
```ruby
EndpointCheckJob.set(queue: :endpoint_check_us_west_2).perform_later(pc)
EndpointCheckJob.set(queue: :endpoint_check_us_east_1).perform_later(pc)
...
```
The code doesn't need to change here, but it probably is good to mention that this queue name itself is never expected to be used since it affects how we spin up resque workers.

SaravShah: Cool. i can add a follow on commit or ticket the addition of r the AWS region queues.

app/models/preserved_copy.rb
atz: The names themselves are OK.

app/models/preserved_copy.rb
```diff
@@ -8,23 +8,30 @@ class PreservedCopy < ApplicationRecord
...
     OK_STATUS => 0,
     INVALID_MOAB_STATUS => 1,
     INVALID_CHECKSUM_STATUS => 2,
     ONLINE_MOAB_NOT_FOUND_STATUS => 3,
     UNEXPECTED_VERSION_ON_STORAGE_STATUS => 4,
     VALIDITY_UNKNOWN_STATUS => 6,
-    UNREPLICATED_STATUS => 7
+    UNREPLICATED_STATUS => 7,
+    FILE_NOT_FOUND => 8,
+    CHECKSUM_MISMATCH => 9
```
ndushay: how is CHECKSUM_MISMATCH diff from INVALID_CHECKSUM? Should we pick one to use in both types of PreservedCopies?

app/models/preserved_object.rb
```diff
@@ -15,6 +15,16 @@ class PreservedObject < ApplicationRecord
   validates :current_version, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :preservation_policy, null: false
 
+  # Spawn asynchronous checks of all existing archive preserved_copies.
+  # This logic is similar to PlexerJob, for a different purpose.
+  # This logic should also include the archive_ttl, for pres_policies that have the oldest ttl.
+  # Calls EndpointCheckJob
+  # This builds off of #917
+  def check_endpoints!
```
ndushay: so this is PO2PC, conceptually?

app/jobs/endpoint_check_job.rb
```diff
@@ -0,0 +1,36 @@
...
+  delegate :bucket, :bucket_name, to: PreservationCatalog::S3
+
+  # @param [PreservedCopy] verify that the archived preserved_copy exists on an endpoint
+  def perform(preserved_copy)
+    return if preserved_copy.status == PreservedCopy::UNREPLICATED_STATUS # FIXME: What to do covered in storytime
+    aws_s3_object = bucket.object(preserved_copy.s3_key)
+    if aws_s3_object.exists?
+      stored_checksums = stored_checksums(preserved_copy)
```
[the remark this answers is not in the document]
SaravShah: `stored_checksums(preserved_copy).include?(replicated_checksum(aws_s3_object))` in my opinion is not that clear. So to make it clear-er i decided to separate out 2 variables.

app/jobs/endpoint_check_job.rb
```diff
@@ -0,0 +1,36 @@
...
+      replicated_checksum = replicated_checksum(aws_s3_object)
+      if stored_checksums.include?(replicated_checksum)
+        preserved_copy.update_status(PreservedCopy::OK_STATUS)
+        preserved_copy.update(last_checksum_validation: Time.zone.now)
+      else
+        preserved_copy.update_status(PreservedCopy::CHECKSUM_MISMATCH)
```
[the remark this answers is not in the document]
SaravShah: yeah, i was thinking that INVALID_CHECKSUM_STATUS was like the md5 didn't have 32 characters or had characters from `f` onwards. but yeah i could see them combining. I'll do that for now and will update the timestamp. thanks

app/models/preserved_object.rb
SaravShah: yes! :)

atz:
@ndushay you say:
`The way "endpoint" is being used in the name of the class and the queue, etc, I keep thinking it's specific to a particular endpoint. `

You were correct: each queue used IS specific to an endpoint (for the same reasons as the delivery workers).  I think that moots any proposed renaming.

SaravShah:
@atz @ndushay ready for review - just added 2 follow on commits. I can squash if you would like but just thought this would be easier.

.rubocop.yml
```diff
@@ -59,6 +59,7 @@ Lint/RescueWithoutErrorClass:
 
 Metrics/AbcSize:
   Exclude:
+    - 'app/jobs/replication_check_job.rb' # [18.49/15] method still readable
```
SaravShah: ah crap did the squash wrong (i'm just going to leave this here) 
......this should actually be `endpoint_check_job.rb` since the name changes in the last commit...but im not sure what i did here

SaravShah:
I'm happy to keep the name as `EndpointCheckJob` or to change it to `ReplicationCheckJob` - whatever the team decides.

app/jobs/replication_check_job.rb
```diff
@@ -0,0 +1,34 @@
+# Confirms existence of PreservedCopy on an endpoint.
+# Confirms the MD5 checksum matches in database and s3.
+class ReplicationCheckJob < ApplicationJob
+  # This queue is never expected to be used. For example:
+  # ReplicationCheck.set(queue: :endpoint_check_us_west_2).perform_later(pc)
+  # ReplicationCheck.set(queue: :endpoint_check_us_east_1).perform_later(pc)
+  queue_as :replication_check
```
ndushay: what if the queue name here is 'override-this-queue'  and the usage of the class is part of hte class comment.

SaravShah:
ReplicatedFileCheck name has been chosen!

app/jobs/replication_check_job.rb
```diff
@@ -0,0 +1,34 @@
...
+  delegate :bucket, :bucket_name, to: PreservationCatalog::S3
+
+  # @param [PreservedCopy] verify that the archived preserved_copy exists on an endpoint
+  def perform(preserved_copy)
+    return if preserved_copy.status == PreservedCopy::UNREPLICATED_STATUS # FIXME: storytime (Monday 07/02)
```
atz: Not important, but you can ask `preserved_copy.unreplicated?` via the enum and avoid the constant.

app/jobs/replication_check_job.rb
```diff
@@ -0,0 +1,34 @@
...
+    aws_s3_object = bucket.object(preserved_copy.s3_key)
+    stored_checksums = stored_checksums(preserved_copy)
+    replicated_checksum = replicated_checksum(aws_s3_object)
+    return preserved_copy.replicated_copy_not_found! unless aws_s3_object.exists?
+    unless stored_checksums.include?(replicated_checksum)
+      preserved_copy.update(last_checksum_validation: Time.zone.now)
```
atz: These should probably be straight assignment:
```
preserved_copy.last_checksum_validation = Time.zone.now
```
The `!` dynamic methods from the enum themselves call `update!` (which does a `save!`).  So we should just set things once and avoid saving twice.

PR #917, opened 2018-06-21
[WIP] Replace #which_need_archive_copy and #create_archive_preserved_copies

### LINKED ISSUE #899: (R) can't get what needs to be replicated

atz:
WIP - likely will be closed in favor of a diff PR after db refactor

The new logic operates in a more comprehensible and discernible manner.
- No more arel_table subquery across potentially millions of records
- `create_archive_preserved_copies!` now handles backfilling for ALL version on ALL endpoints
- `Endpoint.ids_to_versions_found` issues a single query and returns a simple data structure (Hash), independently tested, representing all targeted endpoints and the versions already found. That allows PO to iterate through per endpoint, make a single array/set comparison and know what the "missing" PCs are. All missing PCs are populated in a single `create!` statement.

Fixes #899 
Affects #915

### DIFF
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -27,28 +27,28 @@ class Endpoint < ApplicationRecord
     archive.joins(preservation_policies: [:preserved_objects]).where(preserved_objects: { druid: druid })
   }
 
-  scope :which_need_archive_copy, lambda { |druid, version|
-    # testing indicates that the Arel::Table#eq will cast the input to the appropriate type for us.  i didn't
-    # didn't see that documented, so i'm casting version.to_i to be safe (since we're not using the usual bind
-    # variable machinery).  just trying to be extra cautious about injection attacks.  we shouldn't have to
-    # worry about druid, since it gets passed via the usual ActiveRecord bind var machinery.
-    pc_table = PreservedCopy.arel_table
-    ep_table = Endpoint.arel_table
-    endpoint_has_pres_copy_subquery =
-      PreservedCopy.where(
-        pc_table[:endpoint_id].eq(ep_table[:id])
-          .and(pc_table[:version].eq(version.to_i))
-      ).exists
-
-    archive_targets(druid).where.not(endpoint_has_pres_copy_subquery)
-  }
-
   # Use a queue to validate PreservedCopy objects
   def validate_expired_checksums!
     raise 'Endpoint is not "online" type' unless endpoint_type.online?
     pcs = preserved_copies.fixity_check_expired
     Rails.logger.info "Endpoint #{id} (#{endpoint_name}), # of preserved_copies to be checksum validated: #{pcs.count}"
     pcs.find_each { |pc| ChecksumValidationJob.perform_later(pc) }
+  end
+
+  # @param [String] druid
+  # @return [Hash<Integer => Array<Integer>>] Archive Endpoint IDs mapped to found version numbers for one druid
+  # @example Endpoint.ids_to_versions_found('zz964cr9336')
+  #  { 11 => [1, 2, 3], 12 => [1, 2, 3], 13 => [] }
+  def self.ids_to_versions_found(druid)
+    archive_targets(druid)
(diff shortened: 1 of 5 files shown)


ndushay:
Sarav and I were just working with this code as deployed to stage, and we are empirically getting Endpoint.ids_to_versions_found(druid) returning the DESIRED replication copies, not the ones that exist (which would be []).  Probably easiest to talk in person.

ndushay:
Note also that we will be hooking up replication for incoming new versions before we do the backlog, so we still need a way to create unreplicated preserved_copies for archival endpoints for ONLY the given version.

atz:
I don't see any conflict (or purpose) to only replicating the "new" versions, since the Moab cannot be reconstructed from only, say, version 4 without preceding versions.  While we *could* supply logic to do that, it isn't a real use case AFAICT.

atz:
Reproduced reported errors w/ query logic.

atz:
Added extra PO w/ PCs on tests to verify query specificity.  (tests failed)  Then added an appropriate `where` clause (tests pass).

SaravShah:
the correct archival_preserved_copies are still not being created. `Endpoint.ids_to_versions_found(druid)` still not returning `{ 11 => [], 12=> [] }`.

We screenheroe'd and talked through some more test cases that could show why it isn't working on Stage but it is working in the RSpect tests. 

Will try to make the test data look like stage data, 1 PreservedObject has 1 PreservedCopy which has 1 Online Endpoint.

atz:
I misunderstood that online PCs will only exist for the most current version, so my tests diverged from stage data.  However, that is not enough to explain the behavior on stage.  The failure I get is essentially the same as the old query (despite tests passing for both).  

These queries will be directly affected by our splitting PC into 2 models, so I'm effectively punting on this until we have that.

Blocked by #927

jmartin-sul:
superseded by #958

PR #919, opened 2018-06-21
Cv by root by status

### LINKED ISSUE #912: confirm workflows errors cleared by C2M, M2C
basically, we deployed #827 and we want to ensure it works in production:

find some objects with a bad status, run a check for the druid (or disk) and see if the error count in workflows goes down.

### LINKED ISSUE #914: easy way to run CV for pcs with certain status on certain root (blocks #912)
For example, at the moment, the production robots are creating new objects with "validity_unknown" status.  We want to run *only* those objects on root 15 with "validity_unknown" status through CV.

Or, if we want to zero out all the workflow errors, to get back the appropriate messages, we want to run only the "invalid_checksum" status objects on root 2 through CV.

This will make it easier to do #912, also.

ndushay:
Fixes #914 (easy way to run CV for objects with certain status on certain storage root)

Connects to #912 (confirm workflow errors cleared by M2C, C2M) as it will make it really easy to test that for a significant number of druids.

### DIFF
--- a/app/lib/audit/checksum.rb
+++ b/app/lib/audit/checksum.rb
@@ -58,5 +58,20 @@ module Audit
         Checksum.validate_druid(row.first)
       end
     end
+
+    # validate objects with a particular status on a particular endpoint
+    def self.validate_status_root(status, endpoint_name, limit=Settings.c2m_sql_limit)
+      # pres_copies is an AR Relation; it could return a lot of results, so we want to process it in
+      # batches.  we can't use ActiveRecord's .find_each, because that'll disregard the order .fixity_check_expired
+      # specified.  so we use our own batch processing method, which does respect Relation order.
+      pres_copies = PreservedCopy.send(status).by_endpoint_name(endpoint_name).for_online_endpoints
+      desc = "Number of Preserved Copies of status #{status} from #{endpoint_name} to be checksum validated"
+      logger.info "#{desc}: #{pres_copies.count}"
+      ActiveRecordUtils.process_in_batches(pres_copies, limit) do |pc|
+        logger.info "CV beginning for #{pc.preserved_object.druid}; starting status #{pc.status}"
+        ChecksumValidator.new(pc).validate_checksums
+        logger.info "CV ended for #{pc.preserved_object.druid}; ending status #{pc.status}"
+      end
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


app/lib/audit/checksum.rb
```diff
@@ -58,5 +58,20 @@ def self.validate_list_of_druids(druid_list_file_path)
...
+    def self.validate_status_root(status, endpoint_name, limit=Settings.c2m_sql_limit)
+      # pres_copies is an AR Relation; it could return a lot of results, so we want to process it in
+      # batches.  we can't use ActiveRecord's .find_each, because that'll disregard the order .fixity_check_expired
+      # specified.  so we use our own batch processing method, which does respect Relation order.
+      pres_copies = PreservedCopy.send(status).by_endpoint_name(endpoint_name).for_online_endpoints
+      desc = "Number of Preserved Copies of status #{status} from #{endpoint_name} to be checksum validated"
+      logger.info "#{desc}: #{pres_copies.count}"
+      ActiveRecordUtils.process_in_batches(pres_copies, limit) do |pc|
+        logger.info "CV beginning for #{pc.preserved_object.druid}; starting status #{pc.status}"
+        ChecksumValidator.new(pc).validate_checksums
```
atz: While this is OK, does it make sense to have the PC model provide a method like `validate_checksums!` that wraps a `ChecksumValidator.new(self).validate_checksums`?

ndushay: Not a bad idea, but validating checksums for online pres_copies is diff from archival pres_copies ... seems perhaps better to ticket it than include it here, due to that wrinkle?  (Spoke with @SaravShah who is working on the archival copy C2A and he agrees a follow on ticket is better).  I'll make a ticket.

ndushay: Created #921

atz: thx

spec/lib/audit/checksum_spec.rb
```diff
@@ -137,4 +136,52 @@
...
+    include_context 'fixture moabs in db'
+    let(:subject) { described_class.validate_status_root('validity_unknown', endpoint_name, limit) }
+
+    context 'when there are PreservedCopies to check' do
+      let(:cv_mock) { instance_double(ChecksumValidator) }
+
+      it 'creates an instance and calls #validate_checksums for every result when results are in a single batch' do
+        allow(ChecksumValidator).to receive(:new).and_return(cv_mock)
+        expect(cv_mock).to receive(:validate_checksums).exactly(3).times
+        subject
```
atz: FYI, using `subject` for a time-dependent execution like this is something of an anti-pattern.  Since it is only invoked once, it is probably clearer to make the `.validate_status_root` call directly here.

ndushay: will do!

ndushay: and done ...

PR #922, opened 2018-06-21
deploy: update shared_configs before migrating db

### LINKED ISSUE #891: deployment: db_seed should be run after shared configs are updated (blocks #892)
trickiness:  the `db_seed` task rightfully runs after the `deploy:migrate` task.  the `shared_configs:update` runs before `deploy:restart`.

but regardless, `db_seed` needs to run on the latest configs.

i would personally not be that bothered by needing to run `db:seed` manually as needed, since shared_configs updates which add new endpoints and such should be rare.  but maybe someone with more capistrano-fu can come up with a way to do this properly and automatically.

ndushay:
Fixes #891 

For example, we want the latest endpoint data from shared_configs before the db is seeded.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -7,9 +7,6 @@ ask :branch, `git rev-parse --abbrev-ref HEAD`.chomp
 # Default deploy_to directory is /var/www/my_app_name
 set :deploy_to, "/opt/app/pres/#{fetch(:application)}"
 
-# update shared_configs before restarting app
-before 'deploy:restart', 'shared_configs:update'
-
 # Default value for :format is :airbrussh.
 # set :format, :airbrussh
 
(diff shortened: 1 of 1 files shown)


atz:
To be clear, after this change if we do `cap prod deploy:restart`, shared_configs would not be updated, whereas currently they are updated.  I'm `meh` on it, but that may affect expectations of Ops.

Long term, our config to DB-seeding strategy introduces a number of timing and orchestration complexities (like this one) that it seems best to avoid entirely.  We use it on 3 tables:
- `PreservationPolicy` - could be replaces w/ a class level method that does `find_or_create_by` w/ the sole set of seeded policy values
- `EndpointType` - unnecessary if we fold it into `Endpoint`
- `Endpoint` - the only table that presents difficulty

ndushay:
Excellent risk assessment.

@tallenaz says the risk is low as ops is not likely to expect shared_configs to update via cap deploy.  Also, the cap task to update shared configs can be used on its own via cap.

The other alternative is to update shared_configs twice -- they could be updated before restart as well.  But since risk is low, 🤷‍♀️

PR #923, opened 2018-06-22
Cache at the right level in PreservationPolicy

atz:
The valuable piece is the default policy, not (just) its ID. This requires nothing special (i.e. no special cache hash and clearing methods) and makes the model easier to use.

I moved the salient cache-behavior tests into the `. default_policy` block to ensure we still enjoy those benefits.

Everywhere that used to call `default_policy_id` can and should call `default_policy.id`.


spec/models/preservation_policy_spec.rb
```diff
@@ -54,63 +52,40 @@
...
+
     it 'returns the default preservation policy object' do
       # db already seeded
-      expect(PreservationPolicy.default_policy).to be_a_kind_of PreservationPolicy
+      expect(described_class.default_policy).to be_a_kind_of described_class
     end
 
     it "raises RecordNotFound if the default policy doesn't exist in the db" do
-      # a bit contrived, but just want to test that lack of default PreservationPolicy causes lookup to
+      # a bit contrived, but just want to test that lack of default described_class causes lookup to
```
ndushay: this comment is the one place where keeping PreservationPolicy as the string might read better

atz: yup

atz:
@ndushay corrected the comment.  Please clarify for Tony whether you are blocking or not.

PR #925, opened 2018-06-22
CV async

### LINKED ISSUE #867: (CV) set it up to run via queue
We want to run CV regularly via message queue

atz:
- Add `#validate_checksums!` to wrap hand-off to job
- Convert the logic for "large" CV runs to use asynchronous basis
- Update README

Arbitrary PCs can be CV'd asynchronously with just: 
```
PreservedCopy.where(...).find_each(&:validate_checksums!)
```
Without additional intervening logic.

Fixes #867 (Might convert other CV logic to async also, but I think this is sufficient for the major cases.)


SaravShah:
waiting for one more set of 👀

atz:
IMO, the way to get this code to have that tweak is to merge it, so that it is there when that tweak is happening.  More broadly, making this transition as early as possible is the best way to minimize overall risk to the project.  Other infrastructure configuration that is needed (basically add one queue name), is *already* required.  This PR does not introduce that requirement, it only shifts load onto it.  

Deferring the transition until some unspecified later time sounds to me like a high possibility of inducing additional work and, more disconcertingly, very high likelihood of incompletion.  That is to say, we will end up stuck doing things the "stupid" way because it seemed good enough, and that will bite us.

Making the "right(er)" way the actually-used production mechanism is the best way to guarantee that overall monitoring is representative and subsequent code doesn't repeat bad patterns.  There is no benefit to replication efforts from deferring this change and there is **immediate** benefit to monitoring from accepting it (queue depth and error queue).  

This was outstanding ticketed work.  IMO, you have introduced review criteria that are not applicable to PR, because they represent anxiety about the ticket itself.

julianmorley:
Assuming the code passes muster, I'm fine with this being merged so long as @atz can now focus on replication tickets.

app/lib/audit/checksum.rb
```diff
@@ -5,37 +5,27 @@ class << self
...
       # pres_copies is an AR Relation; it could return a lot of results, so we want to process it in
       # batches.  we can't use ActiveRecord's .find_each, because that'll disregard the order .fixity_check_expired
       # specified.  so we use our own batch processing method, which does respect Relation order.
       pres_copies = PreservedCopy.by_endpoint_name(endpoint_name).for_online_endpoints.fixity_check_expired
-      logger.info "Number of Preserved Copies to be checksum validated: #{pres_copies.count}"
-      ActiveRecordUtils.process_in_batches(pres_copies, limit) do |pc|
-        ChecksumValidator.new(pc).validate_checksums
-      end
+      logger.info "Number of Preserved Copies to be enqueued for CV: #{pres_copies.count}"
+      ActiveRecordUtils.process_in_batches(pres_copies, limit, &:validate_checksums!)
```
jmartin-sul: actually, i think you don't want to keep using this method, because it expects synchronous processing of the records being batched:  https://github.com/sul-dlss/preservation_catalog/blob/92e1a0107aa7fc4a704690c013cc789812db2892/app/lib/active_record_utils.rb#L20-L40

if the checksum validation time isn't updated in the loop, it'll just process the first batch repeatedly.  i would just use `find_each`... i can't think of a great way to get batching that respects our `order by` without coming up with a more complex windowing scheme for the query than what we currently have implemented in that helper.

if we switch to `find_each` here, we should of course also ditch the no-longer-valid comment about using our own batching method.

spec/lib/audit/checksum_spec.rb
```diff
@@ -13,87 +13,32 @@
...
-      it 'creates an instance and calls #validate_checksums on everything in batches' do
-        pcs_from_scope = PreservedCopy.by_endpoint_name(endpoint_name).fixity_check_expired
-        cv_list = pcs_from_scope.map { |pc| ChecksumValidator.new(pc) }
-        expect(cv_list.size).to eq 3
-        cv_list.each do |cv|
-          allow(ChecksumValidator).to receive(:new).with(cv.preserved_copy).and_return(cv)
-          expect(cv).to receive(:validate_checksums).exactly(1).times.and_call_original
-        end
-        described_class.validate_disk(endpoint_name, 2)
-      end
```
jmartin-sul: i would prefer to keep the old style of test that makes sure that each of the pres copies we care about was examined, not just that the CV call was made the right number of times.  i would worry less about this if we switched to `find_each`, since it feels less necessary to test ActiveRecord's batching mechanism than it does to test our own home-grown batching mechanism, with it's less-than-obvious usage guidelines.

app/lib/audit/checksum.rb
atz: Ah, interesting point.  Happy to do that!

spec/lib/audit/checksum_spec.rb
atz: Switched to `find_each`, so I think all is good.

ndushay:
I marked approved, but of course proviso on CI passing.

atz:
Weird, tests passing locally, and I'd think that a failure like `undefined local variable or method `preserved_copy'` would fail deterministically.

atz:
Yeah, `preserved_copy` cannot be undefined if rspec is working, because it is in a `let` statement at the top of the outermost `describe` block.  Color me confused.

atz:
Ok, I think I resolved all that with a rebase and some fiddling.

PR #929, opened 2018-06-27
DruidVersionZip logic for chunks

atz:
Add DVZ methods #parts and #expected_parts.  Because we will only be able to store the **count** of parts in the S3 metadata, we need a way of reconstructing what filename is expected to correspond to part X of N, or the whole set of N filenames, as here.

This logic will be utilized after workers are converted subsequent to the PC model split.

### DIFF
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -96,7 +96,7 @@ class DruidVersionZip
   # @see #work_dir
   # @return [String] shell command to create this zip
   def zip_command
-    "zip -vr0X -s 10g #{file_path} #{druid.id}/#{v_version}"
+    "zip -vr0X -sv -s #{zip_split_size} #{file_path} #{druid.id}/#{v_version}"
   end
 
   # We presume the system guts do not change underneath a given class instance.
@@ -116,6 +116,11 @@ class DruidVersionZip
     end
     return match[1] if match && match[1].present?
     raise 'No version info matched from `zip -v` ouptut'
+  end
+
+  # @return [String] the option included with "zip -s"
+  def zip_split_size
+    '10g'
   end
 
   def zip_version_regexp
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -30,6 +30,16 @@ class DruidVersionZip
     Pathname.new(file_path).tap { |pn| pn.dirname.mkpath }
   end
 
+  # @param [Integer] count Number of parts
+  # @return [Array<String>] Ordered pathnames expected to correspond to a zip broken into a given number of parts
+  def expected_parts(count = 1)
+    raise ArgumentError, "count (#{count}) must be >= 1" if count < 1
+    base = file_path.sub(/.zip\z/, '')
+    [file_path].concat(
+      (1..(count - 1)).map { |n| base + format('.z%02d', n) }
+    )
+  end
+
   # @return [String]
   # @see [S3 key name performance implications] https://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html
   # @example return 'ab/123/cd/4567/ab123cd4567/ab123cd4567.v0001.zip'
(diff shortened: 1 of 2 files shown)


app/models/druid_version_zip.rb
```diff
@@ -118,6 +133,11 @@ def fetch_zip_version
     raise 'No version info matched from `zip -v` ouptut'
   end
 
+  # @return [String] the option included with "zip -s"
+  def zip_split_size
+    '10g'
```
ndushay: should this be a config setting?  

if it changes, does it impact stuff already replicated being retrieved and reassembled?

Also, should the constant in preserved_copy.replicate! be derived from the value here?

app/models/druid_version_zip.rb
```diff
@@ -96,7 +111,7 @@ def work_dir
   # @see #work_dir
   # @return [String] shell command to create this zip
   def zip_command
-    "zip -vr0X -s 10g #{file_path} #{druid.id}/#{v_version}"
+    "zip -vr0X -sv -s #{zip_split_size} #{file_path} #{druid.id}/#{v_version}"
```
ndushay: stupidly naive question:  I assume if the file is < 10G, it will be a single file as before?

app/models/druid_version_zip.rb
```diff
@@ -30,6 +30,16 @@ def ensure_zip_directory!
     Pathname.new(file_path).tap { |pn| pn.dirname.mkpath }
   end
 
+  # @param [Integer] count Number of parts
+  # @return [Array<String>] Ordered pathnames expected to correspond to a zip broken into a given number of parts
+  def expected_parts(count = 1)
```
ndushay: i'm thinking "expected_part_names" might be more clear ...

app/models/druid_version_zip.rb
atz: According to Julian, it sounds like no.  If it becomes necessary, it is simple enough to do later.  Changing the value affects the overall logic not at all, because reassembly (unimplemented, repeatedly declared out of scope) relies only on the *count* of parts, not their size.  Once you have all the enumerated parts, you can unzip them, regardless of their size.

This is specifically an option string passed to `zip`.  While related to the enqueue constraint, that is an Integer that must anticipate overhead from zip internal structure.  The constraint and the option operate in different phases of asynchronous operation, so I'm not interested in making them tightly coupled.

app/models/druid_version_zip.rb
atz: Yes, that is default zip behavior.

spec/models/druid_version_zip_spec.rb
```diff
@@ -112,11 +128,31 @@
     end
   end
 
+  describe '#parts' do # zip splits
```
atz: @ndushay These tests seem to already be what you are asking for.  Maybe you missed them?

app/models/druid_version_zip.rb
ndushay: Let me put it this way ... I was confused when I first looked at this method;  was it going to give me the expected pieces?  The number of expected parts?  oooOOOOoooooh - the NAMES of the parts.  I'm coming at this from a different frame of reference, and I thought the name was too vague and contributed to my confusion.

app/models/druid_version_zip.rb
ndushay: Having gotten frustrated spelunking for "where are the hidden hardcoded assumptions" in moab-versioning ... is there a way to make this particular constant more obvious?

spec/models/druid_version_zip_spec.rb
ndushay: I think, again, my missing the point has to do with naming:  "parts" is perfectly clear to you, but less obvious to me.  I read 

describe "#parts"
it "lists the multiple files produced"

The test, as phrased, is essentially "show that the parts method produces the correct file names" but the concept of what I was seeking is completely subsumed ... I wouldn't think to look for it in a "#parts" describe block.

I was perhaps looking for some words closer to "ensure create_zip! creates multiple files per threshold" in a test description somewhere.

There's nothing incorrect about what you did ... but I'm not grokking the code or the tests as readily as you probably hoped.  How do we meet in the middle?

app/models/druid_version_zip.rb
atz: I think having its own overridable method makes it much more highly visible.

spec/models/druid_version_zip_spec.rb
atz: Yeah, I get what you are saying.  I can basically copy the test and put it under `create_zip!`, but that doesn't seem like a win, since I have to test it here for this method anyway.

app/models/druid_version_zip.rb
atz: - The expected pieces? Yes?
- The number of parts? Yes, indirectly, but also tautologically, because the number of parts is the documented `@param`.  

The constraint is that this logic has to operate without being able to rely on the file glob on zip transfers directory, which may no longer be comprehensive (or even have **any** matching zip files).  It is precisely symmetric to `parts` which actually does that glob, used immediately after successful zip creation.  The name reflects the relationship between `expected_parts` and `parts` and any drift obscures that.  

The return value is:
```
Ordered pathnames expected to correspond to a zip broken into a given number of parts
```
I think that is pretty descriptive and enough to relieve any confusion.  If it would be helpful, I would rename it `parts_expected`, which would move it down to live next to its counterpart.

tallenaz:
Merging for now but concerns around the outstanding changes are in https://github.com/sul-dlss/preservation_catalog/issues/938

PR #930, opened 2018-06-27
Pohandler spec refactor

ndushay:
These refactorings, which should be easy to review, were done to clear the way to adding an argument to some of the pohandler methods.  @jmartin-sul did some of them.

We need to add the argument for #817 - robots creating/updating preserved copies with status ok (instead of invalid_checksum).  I have a branch in progress for that but it still needs more love

### DIFF
--- a/spec/services/preserved_object_handler_check_exist_spec.rb
+++ b/spec/services/preserved_object_handler_check_exist_spec.rb
@@ -400,19 +400,15 @@ RSpec.describe PreservedObjectHandler do
       end
 
       it 'calls PreservedCopy.save! (but not PreservedObject.save!) if the existing record is NOT altered' do
-        po_handler = described_class.new(druid, 1, 1, ep)
-        po = instance_double(PreservedObject)
-        pc = instance_double(PreservedCopy)
+        druid = 'zy987xw6543'
+        po = create :preserved_object, druid: druid
         allow(PreservedObject).to receive(:find_by!).with(druid: druid).and_return(po)
-        allow(po).to receive(:current_version).and_return(1)
-        allow(po).to receive(:save!)
+        pc = create :preserved_copy, preserved_object: po
         allow(PreservedCopy).to receive(:find_by!).with(preserved_object: po, endpoint: ep).and_return(pc)
-        allow(pc).to receive(:matches_po_current_version?).and_return(true)
-        allow(pc).to receive(:status).and_return(PreservedCopy::OK_STATUS)
-        allow(pc).to receive(:version).and_return(1)
-        allow(pc).to receive(:update_audit_timestamps)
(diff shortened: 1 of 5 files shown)


spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -400,19 +400,15 @@
...
       it 'calls PreservedCopy.save! (but not PreservedObject.save!) if the existing record is NOT altered' do
-        po_handler = described_class.new(druid, 1, 1, ep)
-        po = instance_double(PreservedObject)
-        pc = instance_double(PreservedCopy)
+        druid = 'zy987xw6543'
+        po = create :preserved_object, druid: druid
         allow(PreservedObject).to receive(:find_by!).with(druid: druid).and_return(po)
-        allow(po).to receive(:current_version).and_return(1)
-        allow(po).to receive(:save!)
+        pc = create :preserved_copy, preserved_object: po
```
SaravShah: 🔥

spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -123,19 +123,12 @@
           context 'ActiveRecordError' do
             let(:results) do
               allow(Rails.logger).to receive(:log)
-              po = instance_double('PreservedObject')
-              allow(po).to receive(:current_version).and_return(1)
+              po = instance_double('PreservedObject', current_version: 1)
```
jmartin-sul: any particular reason for leaving `po` as a double while switching `pc` to being an actual object from the factory?  or just didn't switch it because it doesn't make a practical difference in the test?

spec/services/preserved_object_handler_update_version_spec.rb
```diff
@@ -474,19 +433,12 @@
             context 'ActiveRecordError' do
               let(:results) do
                 allow(Rails.logger).to receive(:log)
-                po = instance_double('PreservedObject')
-                allow(po).to receive(:current_version).and_return(1)
+                po = instance_double(PreservedObject, current_version: 1)
```
jmartin-sul: same question here

spec/services/preserved_object_handler_update_version_spec.rb
ndushay: mostly because it was a one liner and made no difference to the test.  I was lazy.

spec/services/preserved_object_handler_update_version_spec.rb
ndushay: same answer.  ;-)

spec/services/preserved_object_handler_update_version_spec.rb
jmartin-sul: totally reasonable, just had a preference for it being intentional and not an oversight =)

PR #932, opened 2018-06-27
Monitor resque

tallenaz:
check whether queues are backed up
  and for failed jobs

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -1,4 +1,5 @@
 require 'okcomputer'
+require Rails.root.join('config', 'initializers', 'resque.rb').to_s
 
 OkComputer.mount_at = 'status' # use /status or /status/all or /status/<name-of-check>
 OkComputer.check_in_parallel = true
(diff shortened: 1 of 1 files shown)


config/initializers/okcomputer.rb
```diff
@@ -59,6 +60,16 @@ def check
 
 OkComputer::Registry.register 'ruby_version', OkComputer::RubyVersionCheck.new
 
+# check for backed up resque queues
+Resque.queues.each do |queue|
+  OkComputer::Registry.register "feature-backed-up-#{queue}",
+                                OkComputer::ResqueBackedUpCheck.new(queue, 5_000_000)
+end
+
+# check for failed resque jobs
```
ndushay: Can we / should we do this per error queue instead?

If we keep this, what do you think of as "feature-resque-failures-threshold" as the name?

config/initializers/okcomputer.rb
```diff
@@ -59,6 +60,16 @@ def check
 
 OkComputer::Registry.register 'ruby_version', OkComputer::RubyVersionCheck.new
 
+# check for backed up resque queues
+Resque.queues.each do |queue|
+  OkComputer::Registry.register "feature-backed-up-#{queue}",
```
ndushay: what do you think of "feature-#{queue}-queue-depth" as a name?

config/initializers/okcomputer.rb
tallenaz: The native OKComputer Resque tools don't have the ability to monitor per error queue, just the total number of errors. 

I'm willing to look into writing custom checks, if we think there's enough of a need.

+1 to the name.

config/initializers/okcomputer.rb
tallenaz: Sounds good.

tallenaz:
I timeboxed 30 minutes looking into how to monitor an individual error queue, and it seems like a not-easy problem.

It's easy to get queue names with `Resque.queues`, and easy to get an individual error queue for a queue name with `Resque::Failure.failure_queue_name(Resque.queues[0])`, for example. But I don't see a straightforward way to get the count of a particular failure queue.

ndushay:
Hrmm - can you add that to storytime?  Having one check for failures for all queues seems sub-optimal.  I'm merging this but I think we're going to need to discuss ...

PR #936, opened 2018-06-29
Set up capistrano integration for resque-pool

tallenaz:
so that a deployment will set up the workers

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -13,6 +13,7 @@ require "capistrano/rails/migrations"
 require "capistrano/passenger"
 require "capistrano/honeybadger"
 require "dlss/capistrano"
+require 'capistrano-resque-pool'
 require 'whenever/capistrano'
 
 # Load custom tasks from `lib/capistrano/tasks` if you have any defined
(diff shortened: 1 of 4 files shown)


tallenaz:
I'll integrate shared_configs as follow-on work.

config/deploy/prod.rb
```diff
@@ -1,6 +1,6 @@
 server 'preservation-catalog-prod-01.stanford.edu', user: 'pres', roles: %w[m2c c2m app db web]
 server 'preservation-catalog-prod-02.stanford.edu', user: 'pres', roles: %w[cv app]
-server 'preservation-catalog-prod-03.stanford.edu', user: 'pres', roles: %w[worker app resque]
+server 'preservation-catalog-prod-03.stanford.edu', user: 'pres', roles: %w[worker app resque worker]
```
atz: These looks redundant?  `%w[worker app resque worker]`

Am I missing something?

tallenaz: No, I was missing that we weren't using the resque role for anything, even though it was there, so yeah, redundant.

PR #937, opened 2018-06-29
Archive pres copies - introduction

jmartin-sul:
at least still need to write tests.  TBD how much refactoring of existing code will fall under this PR.

### DIFF
--- a/app/models/archive_preserved_copy.rb
+++ b/app/models/archive_preserved_copy.rb
@@ -1,0 +1,19 @@
+# metadata about a zipped version directory from a Moab
+class ArchivePreservedCopy < ApplicationRecord
+  belongs_to :preserved_copy
+  belongs_to :archive_endpoint
+
+  # @note Hash values cannot be modified without migrating any associated persisted data.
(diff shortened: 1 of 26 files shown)


app/models/archive_preserved_copy.rb
```diff
@@ -0,0 +1,19 @@
...
+  enum status: {
+    'ok' => 0,
+    'unreplicated' => 1,
+    'archive_not_found' => 2,
+    'invalid_checksum' => 3
+  }
+
+  validates :archive_endpoint, presence: true
+  validates :preserved_copy, presence: true
+  validates :status, inclusion: { in: statuses.keys }
```
atz: I've been confused by this one previously, isn't this the same as `validates :status, presence: true`?  It seems like this is restating how `enum` is supposed to work already?

SaravShah:
( just adding a note here, need to add `has_many :archive_preserved_copies` to the PreservedCopy model

atz:
OK, this isn't 100% done by any measure, but it has good stuff in it and doesn't break anything, so I'm merging it as is so we can keep pushing this further.

PR #942, opened 2018-07-02
C2a logging

SaravShah:
Add some logging if archival pres copy has a status of unreplicated, if the checksums mismatch, and if the archival pres copy does not exist on s3.

### DIFF
--- a/app/jobs/replicated_file_check_job.rb
+++ b/app/jobs/replicated_file_check_job.rb
@@ -10,13 +10,24 @@ class ReplicatedFileCheckJob < ApplicationJob
 
   # @param [PreservedCopy] verify that the archived preserved_copy exists on an endpoint
   def perform(preserved_copy)
-    return if preserved_copy.unreplicated? # FIXME: storytime (Monday 07/02)
+    if preserved_copy.unreplicated?
+      Rails.logger.error("#{preserved_copy} should be replicated, but has a status of #{preserved_copy.status}.")
+      return
+    end
     aws_s3_object = bucket.object(preserved_copy.s3_key)
     stored_checksums = stored_checksums(preserved_copy)
     replicated_checksum = replicated_checksum(aws_s3_object)
-    return preserved_copy.replicated_copy_not_found! unless aws_s3_object.exists?
+    unless aws_s3_object.exists?
+      Rails.logger.error("Archival Preserved Copy: #{preserved_copy} was not found on #{bucket_name}.")
+      preserved_copy.replicated_copy_not_found!
+      return
+    end
     preserved_copy.last_checksum_validation = Time.zone.now
-    return preserved_copy.invalid_checksum! unless stored_checksums.include?(replicated_checksum)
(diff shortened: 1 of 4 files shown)


app/jobs/replicated_file_check_job.rb
```diff
@@ -10,13 +10,24 @@ class ReplicatedFileCheckJob < ApplicationJob
 
   # @param [PreservedCopy] verify that the archived preserved_copy exists on an endpoint
   def perform(preserved_copy)
-    return if preserved_copy.unreplicated? # FIXME: storytime (Monday 07/02)
+    if preserved_copy.unreplicated?
+      Rails.logger.error("Archival Preserved Copy: #{preserved_copy} has a status of #{preserved_copy.status}.")
```
tallenaz: Can we log a hint about why this is an error? Maybe, "`#{preserved_copy} should be replicated, but has status of #{preserved_copy.status}`", or something similar?

atz: Don't need to interpolate since by the conditional we already know the status is `unreplicated`.

app/jobs/replicated_file_check_job.rb
```diff
@@ -10,13 +10,24 @@ class ReplicatedFileCheckJob < ApplicationJob
...
+      return
+    end
     aws_s3_object = bucket.object(preserved_copy.s3_key)
     stored_checksums = stored_checksums(preserved_copy)
     replicated_checksum = replicated_checksum(aws_s3_object)
-    return preserved_copy.replicated_copy_not_found! unless aws_s3_object.exists?
+    unless aws_s3_object.exists?
+      preserved_copy.replicated_copy_not_found!
+      Rails.logger.error("Archival Preserved Copy: #{preserved_copy} was not found on #{bucket_name}.")
```
atz: Should do your logging off the object *before* you mutate the object.  I.E., change the order of the log and setter lines.

app/jobs/replicated_file_check_job.rb
```diff
@@ -10,13 +10,24 @@ class ReplicatedFileCheckJob < ApplicationJob
 
   # @param [PreservedCopy] verify that the archived preserved_copy exists on an endpoint
   def perform(preserved_copy)
-    return if preserved_copy.unreplicated? # FIXME: storytime (Monday 07/02)
+    if preserved_copy.unreplicated?
+      Rails.logger.error("#{preserved_copy} should be replicated, but has a status of 'unreplicated'.")
```
tallenaz: Should this pull from the actual status, instead of using a string?

SaravShah: yeah i'm not sure. since we know its `unreplicated` i just added it, i can interpolate if you'd like. What do you think?

tallenaz: I'd prefer interpolation, in the unlikely case we decide to change the name of the status at the constant level.

SaravShah: sounds good. just pushed.

Issue #944, opened 2018-07-02
Schema change, add tests to archive_preserved_copy.rb


SaravShah:
add model tests

PR #948, opened 2018-07-02
ArchivePreservedCopy Model Rspec tests

SaravShah:
connects #944

### DIFF
--- a/spec/models/archive_preserved_copy_spec.rb
+++ b/spec/models/archive_preserved_copy_spec.rb
@@ -1,5 +1,44 @@
 require 'rails_helper'
 
 RSpec.describe ArchivePreservedCopy, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+  let(:po) { build(:preserved_object) }
+  let(:pc) { build(:preserved_copy) }
+  let(:archive_endpoint) { build(:archive_endpoint) }
+  let(:apc) { build(:archive_preserved_copy, preserved_copy: pc, archive_endpoint: archive_endpoint) }
+
+  it 'is not valid without all required valid attributes' do
+    expect(described_class.new).not_to be_valid
+    expect(described_class.new(preserved_copy: pc)).not_to be_valid
+    expect(apc).to be_valid
+  end
+
+  it 'defines a status enum with the expected values' do
+    is_expected.to define_enum_for(:status).with(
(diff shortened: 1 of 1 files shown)


spec/models/archive_preserved_copy_spec.rb
```diff
@@ -1,5 +1,44 @@
...
-  pending "add some examples to (or delete) #{__FILE__}"
+  let(:po) { build(:preserved_object) }
+  let(:pc) { build(:preserved_copy) }
+  let(:archive_endpoint) { build(:archive_endpoint) }
+  let(:apc) { build(:archive_preserved_copy, preserved_copy: pc, archive_endpoint: archive_endpoint) }
+
+  it 'is not valid without all required valid attributes' do
+    expect(described_class.new).not_to be_valid
+    expect(described_class.new(preserved_copy: pc)).not_to be_valid
+    expect(archive_endpoint).to be_valid
```
jmartin-sul: i think you might want this to be `expect(described_class.new(...)).to be_valid`, where `.new()` gets passed all the required params.

atz: at a technical level, it is the same.  not sure if it is an improvement in readability or not.

SaravShah: what do you think @jmartin-sul ? should i leave it or change it

jmartin-sul: ah, sorry, i had a clearer version of this comment originally, but then i edited it for what i thought was a better suggestion, while losing my original point a bit...

what i'd originally suggested was using `expect(apc).to be_valid`.  the main point being that the expectation here is on `archive_endpoint`, but the tests are about `ArchivePreservedCopy`.

then i thought that was slightly less clear about what was being tested and exactly which fields were required, so i changed the suggestion to use `described_class.new`, but that made what i originally cared about less clear.

anyway, whichever is more readable.  as @atz points out, `apc` and `described_class.new(...)` are functionally equivalent in this case.  i leave it up to @SaravShah as the implementer to choose.  i was just pretty sure that the test shouldn't be about `archive_endpoint`.

SaravShah: I'll leave it as it is then. SO ready for re-review :D

PR #955, opened 2018-07-06
Send audit check output to separate logs

### LINKED ISSUE #858: should C2M and M2C (and CV?) write to their own logs, not Rails log?
Basically - do we want to have to grep out the results of C2M, M2C and CV checks from the Rails production.log or do we want these messages in logs specific to those jobs?

The logs, as they stand, are merely the output from the rake tasks.   Should they at least be renamed if we're going to keep them like that?  See #857

tallenaz:
Fixes #858

### DIFF
--- a/spec/lib/audit/checksum_spec.rb
+++ b/spec/lib/audit/checksum_spec.rb
@@ -11,6 +11,17 @@ RSpec.describe Audit::Checksum do
   let(:endpoint_name) { 'fixture_sr1' }
   let(:limit) { Settings.c2m_sql_limit }
 
+  describe '.logger' do
+    let(:logfile) { Rails.root.join('log', 'cv.log') }
+
+    after { FileUtils.rm_f(logfile) }
+
+    it 'writes to STDOUT and its own log' do
+      expect { described_class.logger.debug("foobar") }.to output(/foobar/).to_stdout_from_any_process
+      expect(File).to exist(logfile)
+    end
+  end
+
   describe '.validate_disk' do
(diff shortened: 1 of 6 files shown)


spec/lib/audit/checksum_spec.rb
```diff
@@ -11,6 +11,17 @@
   let(:endpoint_name) { 'fixture_sr1' }
   let(:limit) { Settings.c2m_sql_limit }
 
+  describe '.logger' do
+    let(:logfile) { Rails.root.join('log', 'cv.log') }
+
+    after { FileUtils.rm_f(logfile) }
+
+    it 'writes to STDOUT and its own log' do
+      expect { described_class.logger.debug("foobar") }.to output(/foobar/).to_stdout_from_any_process
```
SaravShah: I guess `to_stdout_from_any_process` is significantly (~30x) slower than `to_stdout`.
I forget, did we try `to_stdout` here?

tallenaz: We tried the `to_stdout` matcher, but then this expectation wouldn't pass, even though we can demonstrate the logger writing to STDOUT:

```
[1] pry(main)> Audit::CatalogToMoab.logger.debug("foobar")
D, [2018-07-06T14:50:14.472858 #36733] DEBUG -- : foobar
```

This matcher is slower, but I don't detect a significant lag when the test suite runs, and it seems to pick up on writing to STDOUT.

SaravShah: Sounds good.

Issue #973, opened 2018-07-12
(C2A) refactor and incorporate archived_preserved_copies and parts


SaravShah:
Currently using old db/schema.rb dealing with only preserved_copies. Need to refactor to use archive preserved copies and copy_parts

tallenaz:
From #974: "The hitch is that zip_checksums were used by ReplicatedFileCheckJob#stored_checksums. We can now get at the stored checksums for a "complete moab" (formerly PreservedCopy) via the archived_preserved_copies and their subsequent parts. However, the spec for stored_checksums is throwing a false positive."

PR #974, opened 2018-07-12
Remove ZipChecksum

### LINKED ISSUE #970: (R) get rid of ZipChecksum model/zip_checksums table
i believe this has been superseded by `ArchivePreservedCopyPart`, which has checksum info?  if so, zip_checksums is now obsolete/confusing.

tallenaz:
This is fairly straightforward with one hitch. 

The straightforward part is the migration, schema change, class, spec, and factory removal.

The hitch is that `zip_checksums` were used by `ReplicatedFileCheckJob#stored_checksums`. We can now get at the stored checksums for a "complete moab" (formerly `PreservedCopy`) via the `archived_preserved_copies` and their subsequent `parts`. However, the spec for `stored_checksums` is throwing a false positive. 

The proposal is to punt on fixing that spec to the upcoming `ReplicatedFileCheckJob` refactor. I've tested the re-implementation of `stored_checksums` here outside of rspec, and am pretty confident it is the correct approach. 

closes #970
connects to #973


spec/models/preserved_copy_spec.rb
```diff
@@ -65,7 +65,7 @@
   it { is_expected.to validate_presence_of(:endpoint) }
   it { is_expected.to validate_presence_of(:preserved_object) }
   it { is_expected.to validate_presence_of(:version) }
-  it { is_expected.to have_many(:zip_checksums) }
+  it { is_expected.to have_many(:archive_preserved_copies) }
```
SaravShah: nice, good catch

ndushay:
If @SaravShah and @tallenaz are confident we won't miss the false positive (and I see it is added to #973) then I am satisfied with this.  :-)

Issue #976, opened 2018-07-12
rename ArchivePreservedCopy to ZippedMoabVersion (blocked by #978)


jmartin-sul:
rename model, underlying db table, and any relevant columns and indexes

tallenaz:
I'll go ahead and branch off what's up for review in the PR for #978

ndushay:
In PreservedObject, see also:

```ruby
# given a version, create any PreservedCopy records for that version which don't yet exist for zip
#  endpoints which implement this PreservedObject's PreservationPolicy.
# @param archive_vers [Integer] the version for which preserved copies should be created.  must be between
#   1 and this PreservedObject's current version (inclusive).
# @return [Array<PreservedCopy>] the PreservedCopy records that were created
def create_archive_preserved_copies(archive_vers)
```

from PR #981.

PR #981, opened 2018-07-12
Rename Archive Endpoint to Zip endpoint

### LINKED ISSUE #975: rename ArchiveEndpoint to ZipEndpoint
rename model, underlying db table, pres policies join table, and any relevant columns and indexes.

SaravShah:
I can squash all of the commits if needed but 
First Commit: Migration and schema change
Second Commit: renames in app/ dir
Third Commit: renames in spec/ dir
Fourth Commit: renames in config/ and seeds

closes #975 

Edit need shared_config reviewed too:
https://github.com/sul-dlss/shared_configs/pull/620
https://github.com/sul-dlss/shared_configs/pull/621

### DIFF
--- a/app/models/preserved_object.rb
+++ b/app/models/preserved_object.rb
@@ -26,20 +26,4 @@ class PreservedObject < ApplicationRecord
     # FIXME: STUB
     # Ticket: 920
   end
-
-  # given a version, create any PreservedCopy records for that version which don't yet exist for zip
-  #  endpoints which implement this PreservedObject's PreservationPolicy.
-  # @param archive_vers [Integer] the version for which preserved copies should be created.  must be between
-  #   1 and this PreservedObject's current version (inclusive).
-  # @return [Array<PreservedCopy>] the PreservedCopy records that were created
-  def create_archive_preserved_copies(archive_vers)
-    unless archive_vers > 0 && archive_vers <= current_version
-      raise ArgumentError, "archive_vers (#{archive_vers}) must be between 0 and current_version (#{current_version})"
-    end
-
-    params = Endpoint.which_need_archive_copy(druid, archive_vers).map do |zep|
-      { version: archive_vers, endpoint: zep, status: PreservedCopy::UNREPLICATED_STATUS }
-    end
-    preserved_copies.create!(params)
-  end
 end
--- a/app/models/preserved_object.rb
+++ b/app/models/preserved_object.rb
@@ -22,8 +22,24 @@ class PreservedObject < ApplicationRecord
   # Log an error message.
   # Calls ReplicatedFileCheckJob
   # This builds off of #917
-  def check_endpoints!
+  def check_zip_endpoints!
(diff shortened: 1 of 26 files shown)


db/migrate/20180712203748_rename_archive_endpoint_to_zip_endpoint.rb
```diff
@@ -0,0 +1,18 @@
...
+    
+    rename_table :archive_endpoints, :zip_endpoints
+    rename_column :archive_preserved_copies, :archive_endpoint_id, :zip_endpoint_id
+
+    rename_table :archive_endpoints_preservation_policies, :preservation_policies_zip_endpoints
+    rename_column :preservation_policies_zip_endpoints, :archive_endpoint_id, :zip_endpoint_id
+
+    add_index :preservation_policies_zip_endpoints, :zip_endpoint_id, name: "index_pres_policies_zip_endpoints_on_zip_endpoint_id"
+    add_index :preservation_policies_zip_endpoints, :preservation_policy_id, name: "index_pres_policies_zip_endpoints_on_pres_policy_id"
+
```
ndushay: why all this whitespace?  (it doesn't matter and it is not worth adjusting unless you have to adjust other stuff)

SaravShah:
- removed line 80 or so in spec/models/endpoint_spec the let statement. but i think the test should stay. 

- results_recorder: changed endpoint to zip_endpoint where needed!

- job queue names -- changed to zip_endpoint_events ....will have a shared config PR up soon

- Readme has to do with the `moab_storage_root` ... so leaving it for now.

Rebased and refactored other comments @ndushay  left

app/models/preserved_object.rb
```diff
@@ -22,8 +22,24 @@ class PreservedObject < ApplicationRecord
   # Log an error message.
   # Calls ReplicatedFileCheckJob
   # This builds off of #917
-  def check_endpoints!
+  def check_zip_endpoints!
     # FIXME: STUB
     # Ticket: 920
   end
+
+  # given a version, create any PreservedCopy records for that version which don't yet exist for zip
```
ndushay: ZippedMoabVersion instead of PreservedCopy?  Could just mention this in the PreservedCopy rename ticket instead.

ndushay: put a note in #976

jmartin-sul: actually, my guess is this is an errant chunk of code that got put back in (erroneously) upon rebase.  i think this method should disappear from PO entirely, as that functionality now lives here:  https://github.com/sul-dlss/preservation_catalog/blob/1d32d8f88ab4d89d453d1153003322f48ba4350c/app/models/preserved_copy.rb#L73-L88

SaravShah: whoops! the rebase was messy...

SaravShah:
@ndushay @jmartin-sul  thanks for finding that catch. re-pushed.

PR #993, opened 2018-07-17
Zip endpoint touchups

jmartin-sul:
a couple very tiny follow-on things from #981, and another thing that probably should've happened back when we split out `ArchiveEndpoint` as a separate model.


jmartin-sul:
created #994 to keep this PR small (instead of refactoring the test that uses the deprecated factory i had to modify to get the tests for this PR to pass).

PR #996, opened 2018-07-17
catalog_controller takes checksums_validated arg

ndushay:
This hooks up the catalog_controller to accept a boolean checksums_validated arg and pass it to PreservedObjectHandler.   (See #982 for previous work).

I also took the opportunity to:
- remove a controller that never amounted to anything, from our very initial work on this app.
- add tests for incoming_size, which I thought was optional but which isn't.  :-)

Connects to #817


spec/controllers/catalog_controller_spec.rb
```diff
@@ -264,6 +264,49 @@
...
+      ['true', 'True', 'TRUE'].each do |t_val|
+        it "#{t_val} evaluates to true" do
+          expect(poh).to receive(:create).with(true)
+          post :create, params: { druid: bare_druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param, checksums_validated: t_val }
+        end
+      end
+      ['nil', '1', 'on'].each do |t_val|
+        it "#{t_val} evaluates to false" do
+          expect(poh).to receive(:update_version).with(false)
+          patch :update, params: { druid: bare_druid, incoming_version: ver, incoming_size: size, storage_location: storage_location_param, checksums_validated: 'false' }
```
jmartin-sul: i think you meant to pass `t_val` instead of `'false'` for `checksums_validated`?  although `'false'`and/or `'FALSE'` might be nice things to add to the list of values being tested.

ndushay: whoops!  good 👀 .   pushed up a fix commit.

PR #1001, opened 2018-07-19
[WIP] Set AWS_PROFILE for east/west workers

tallenaz:
and AWS_BUCKET_NAME, else the workers can't find a region to replicate to. I'm happy to discuss better ways to get the workers knowledge of their region or improvements along these lines.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -74,6 +74,8 @@ Metrics/AbcSize:
     - 'app/lib/audit/checksum.rb' # .validate_disk_all_dirs is decidedly readable
     - 'app/lib/audit/moab_to_catalog.rb' # .xxx_for_all_storage_roots is decidedly readable
     - 'app/controllers/catalog_controller.rb' # #create method is readable
+    - 'app/jobs/s3_east_delivery_job.rb' # WIP
+    - 'app/jobs/s3_west_delivery_job.rb' # WIP
 
 Metrics/BlockLength:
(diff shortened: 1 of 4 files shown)


atz:
Yeah, this seems counter to the intent of keeping all that stuff in the ENV.  I understand if the AWS_PROFILE stuff doesn't seem to work, but all the ENVs should be **supplied** by the ENV, rather than hardcoded.

tallenaz:
I agree it's not an ideal approach. We can set up env vars on the system via puppet, which populates the app user's `.bashrc` with export statements, but I don't see how we can use `AWS_PROFILE` at that point, since that will be set with a specific named profile for one worker that will leave the other worker out in the cold, which is why I started mucking with setting the different profiles up in the rails ENV.

atz:
For our purposes, we need distinct ENVs **per process**.  By my understanding, that seems below the granularity of puppet (which deals with systems, not processes).  I would consider a resque-pool config/instance per AWS_REGION.  The non-AWS jobs can be included in either one of the pools, or split off for a 3rd config.

tallenaz:
We decided on a better approach -- use the `resque-pool` custom config loader (https://github.com/nevans/resque-pool#custom-configuration-loader) and then point `resque-pool` to our custom needs with it's `-c` flag, from the init script maintained by puppet.

tallenaz:
closing b/c we're going another route

PR #1003, opened 2018-07-20
rename Endpoint model -> MoabStorageRoot

### LINKED ISSUE #979: rename Endpoint to MoabStorageRoot (blocked by #976)
rename model, as well as underlying db table, pres policies join table, and any relevant columns and indexes.

ndushay:
I did my best to break up the commits into sensible sized chunks for review, because this change affected 39 files.  The individual commits are just subsets of all the files touched, so if you'd rather review it as a single atomic thing, that works too.

Coding-wise, after the database migration and the model changes, all the rest was string replacements and addressing failing specs.

This is obviously big enough to warrant multiple reviewers.

closes #979


atz:
Seems generally OK, if bulky.  Will need to spend more time reviewing.

PR #1004, opened 2018-07-20
Custom capistrano for resque-pool

tallenaz:
The commands we were using from `capistrano-resque-pool` spun up a single pool of workers with no additional knowledge of their environment. The `resque-pool` service on the worker machines, however, spins up 3 pools of workers, where two of the pools have environment variables assigned telling them which credentials and buckets to use to replicate to. This PR intends to mirror the sevice set up through custom capistrano rake tasks, based off of what is in `capistrano-resque-pool`.

Happily, with this PR, objects make it all the way through the replication pipeline consistently across deployments.

### DIFF
--- a/lib/capistrano/tasks/capistrano-resque-pool.rake
+++ b/lib/capistrano/tasks/capistrano-resque-pool.rake
@@ -1,0 +1,95 @@
+namespace :resque do
+  namespace :pool do
+    def rails_env
+      fetch(:resque_rails_env) || fetch(:rails_env) || fetch(:stage)
+    end
+
+    desc 'Start all the workers and queues'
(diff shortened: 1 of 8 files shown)


lib/capistrano/tasks/capistrano-resque-pool.rake
```diff
@@ -0,0 +1,95 @@
+namespace :resque do
+  namespace :pool do
+    def rails_env
+      fetch(:resque_rails_env) || fetch(:rails_env) || fetch(:stage)
+    end
+
+    desc 'Start all the workers and queus'
```
atz: missing an `e` in `queues`

PR #1005, opened 2018-07-21
Replication trigger (part 1): auto-filling DMZ's on PC creation

atz:
This mostly required revision of some bulky tests' expectations.

Note: 10GB cap on `replicate!` now removed!

Next phase would include handoff to replication after ZMV rows are created.

### DIFF
--- a/app/lib/audit/moab_to_catalog.rb
+++ b/app/lib/audit/moab_to_catalog.rb
@@ -94,11 +94,11 @@ module Audit
 
     # @todo This method may not be useful anymore.  Every PC has 1..n DMZs, so either this method must
     # figure out how to specially delete them too, or we can loosen the restrictions from PC to ZMV
-    # @todo Move this method (and pouplate_endpoint/seed_catalog_for_dir) onto the Endpoint model
-    def self.drop_moab_storage_root(endpoint_name)
-      endpoint = Endpoint.find_by!(endpoint_name: endpoint_name.to_s)
+    # @todo Move this method (and pouplate_m_s_r/seed_catalog_for_dir) onto the MoabStorageRoot model
+    def self.drop_moab_storage_root(name)
+      ms_root = MoabStorageRoot.find_by!(name: name.to_s)
       ApplicationRecord.transaction do
-        endpoint.preserved_copies.destroy_all
+        ms_root.preserved_copies.destroy_all
         PreservedObject.without_preserved_copies.destroy_all
       end
     end
--- a/app/lib/audit/moab_to_catalog.rb
+++ b/app/lib/audit/moab_to_catalog.rb
@@ -92,13 +92,14 @@ module Audit
       Profiler.print_profile('M2C_check_existence_for_all_storage_roots') { check_existence_for_all_storage_roots }
     end
 
-    def self.drop_moab_storage_root(name)
+    # @todo This method may not be useful anymore.  Every PC has 1..n DMZs, so either this method must
+    # figure out how to specially delete them too, or we can loosen the restrictions from PC to ZMV
+    # @todo Move this method (and pouplate_endpoint/seed_catalog_for_dir) onto the Endpoint model
+    def self.drop_moab_storage_root(endpoint_name)
+      endpoint = Endpoint.find_by!(endpoint_name: endpoint_name.to_s)
       ApplicationRecord.transaction do
-        PreservedCopy.joins(:moab_storage_root).where(
-          "moab_storage_roots.name = :name",
-          name: name.to_s
-        ).destroy_all
-        PreservedObject.left_outer_joins(:preserved_copies).where(preserved_copies: { id: nil }).destroy_all
+        endpoint.preserved_copies.destroy_all
+        PreservedObject.without_preserved_copies.destroy_all
       end
     end
 
--- a/app/lib/audit/moab_to_catalog.rb
+++ b/app/lib/audit/moab_to_catalog.rb
@@ -92,7 +92,7 @@ module Audit
       Profiler.print_profile('M2C_check_existence_for_all_storage_roots') { check_existence_for_all_storage_roots }
     end
 
-    # @todo This method may not be useful anymore.  Every PC has 1..n DMZs, so either this method must
+    # @todo This method may not be useful anymore.  Every PC has 1..n ZMVs, so either this method must
     # figure out how to specially delete them too, or we can loosen the restrictions from PC to ZMV
     # @todo Move this method (and pouplate_m_s_r/seed_catalog_for_dir) onto the MoabStorageRoot model
     def self.drop_moab_storage_root(name)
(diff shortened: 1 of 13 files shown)


spec/lib/audit/moab_to_catalog_spec.rb
```diff
@@ -23,94 +33,60 @@
   end
 
   describe ".check_existence_for_all_storage_roots" do
-    let(:subject) { described_class.check_existence_for_all_storage_roots }
-
     it 'calls check_existence_for_dir once per storage root' do
       expect(described_class).to receive(:check_existence_for_dir).exactly(HostSettings.storage_roots.entries.count).times
-      subject
+      described_class.check_existence_for_all_storage_roots
```
tallenaz: No objection, just curious about the motivation for moving away from using `subject` here and below?

tallenaz: NM, I see your comment about it's being an antipattern, and finding discussion out there about that, for example thoughtbot advises against it.

atz: Right, the main thing is that this is really a timing-dependent invocation and not an object or value that is setup on demand like a normal `let` statement.  

If one wants `let(:subject) {...}`, they probably should `subject {...}`, but that would make it clear in these cases (to rubocop-rspec) that some non-kosher things are being done.

app/models/preserved_copy.rb
```diff
@@ -24,6 +24,8 @@ class PreservedCopy < ApplicationRecord
     REPLICATED_COPY_NOT_FOUND_STATUS => 8
   }
 
+  after_create(&:create_zipped_moab_versions!)
```
SaravShah: what if we update an existing version from 4 to 5?

spec/lib/audit/moab_to_catalog_spec.rb
```diff
@@ -4,6 +4,16 @@
 RSpec.describe Audit::MoabToCatalog do
   let(:storage_dir) { 'spec/fixtures/storage_root01/sdr2objects' }
   let(:endpoint) { Endpoint.find_by!(storage_location: storage_dir) }
+  let(:mock_profiler) do
+    prof = instance_double(Profiler, prof: nil)
+    allow(Profiler).to receive(:new).and_return(prof)
+    prof
+  end
+  let(:moab) do
+    m = instance_double(Moab::StorageObject, object_pathname: storage_dir, :storage_root= => nil)
```
SaravShah: Should we use a better variable name here? I really don't mind tho

app/lib/audit/moab_to_catalog.rb
```diff
@@ -92,13 +92,14 @@ def self.check_existence_for_all_storage_roots_profiled
       Profiler.print_profile('M2C_check_existence_for_all_storage_roots') { check_existence_for_all_storage_roots }
     end
 
+    # @todo This method may not be useful anymore.  Every PC has 1..n DMZs, so either this method must
```
SaravShah: should this be `ZMV`?

app/models/preserved_copy.rb
```diff
@@ -24,7 +24,9 @@ class PreservedCopy < ApplicationRecord
     REPLICATED_COPY_NOT_FOUND_STATUS => 8
   }
 
-  after_create(&:create_zipped_moab_versions!)
```
SaravShah: aw, i liked this better!

app/models/preserved_object.rb
```diff
@@ -15,7 +15,7 @@ class PreservedObject < ApplicationRecord
   validates :current_version, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :preservation_policy, null: false
 
-  scope :without_preserved_copies, lambda { left_outer_joins(:preserved_copies).where(preserved_copies: { id: nil }) }
+  scope :without_preserved_copies, -> { left_outer_joins(:preserved_copies).where(preserved_copies: { id: nil }) }
```
SaravShah: whats the difference here? the `lambda` vs `->`? (just want to know for personal knowledge)

app/models/preserved_copy.rb
SaravShah: also, what if we update an existing version from 4 to 5?

atz: It won't matter in "part 2" where we call 2 things from this hook, so we need the block.  When we update from 4 to 5 we create a new PC row.

app/models/preserved_object.rb
atz: afaik, no difference, just rubocop prefers one.

app/models/preserved_copy.rb
```diff
@@ -1,5 +1,5 @@
 ##
-# PreservedCopy represents a concrete instance of a PreservedObject, in physical storage on some node.
+# PreservedCopy represents a concrete instance of a PreservedObject version, in physical storage on some node.
```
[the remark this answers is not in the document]
atz: Yep, this represents my misunderstanding of the model role.

atz:
Cleaned up that misunderstanding and rebased after the big rename PR got merged.

spec/models/preserved_copy_spec.rb
```diff
@@ -320,49 +321,55 @@
...
+    before { pc.zipped_moab_versions.destroy_all } # undo auto-spawned rows from callback
 
-      expect { pc.create_zipped_moab_versions!(pc_version - 1) }.to change {
-        ZipEndpoint.which_need_archive_copy(druid, pc_version - 1).to_a
-      }.from([archive_ep]).to([])
-      expect(ZippedMoabVersion.by_druid(druid).count).to eq 2
+    it "creates ZMVs that don't yet exist for expected versions, but should" do
+      expect { pc.create_zipped_moab_versions! }.to change {
+        ZipEndpoint.which_need_archive_copy(druid, pc_version).to_a
+      }.from([archive_ep]).to([]).and change {
```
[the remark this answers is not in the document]
atz: The renaming was not part of this PR (L322 is unchanged), so yeah, I would regard that as a follow-on to the renaming.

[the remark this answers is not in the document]
SaravShah: #1015 created

spec/models/preserved_copy_spec.rb
```diff
@@ -320,49 +321,55 @@
...
     end
 
-    it "creates pres copies that don't yet exist for the given moab_storage_root, but should" do
-      expect { pc.create_zipped_moab_versions!(pc_version) }.to change {
+    it "creates ZMVs that don't yet exist for new endpoint, but should" do
+      expect { pc.create_zipped_moab_versions! }.to change {
         ZipEndpoint.which_need_archive_copy(druid, pc_version).to_a
-      }.from([archive_ep]).to([])
-      expect(ZippedMoabVersion.by_druid(druid).where(version: pc_version).count).to eq 1
+      }.from([archive_ep]).to([]).and change {
```
jmartin-sul: neat, i wasn't aware you could do these compound expectations for the `change` matcher.  very handy.

Issue #927, opened 2018-06-25
Split PC models for online and archive


atz:
Our uses and expectations for "online" and "archive" `PreservedCopy` records have diverged and introduced query complexity where we typically only want one kind or the other.  We should migrate to having different models for each.

General consensus:
- Create new model (like `ArchivePreservedCopy`) for archive PCs
- Migrate any archive data out of `preserved_copies` to the new table
- Rename `PreservedCopy` model (and table) to something like `MoabPreservedCopy` for online
- Tailor models to each case (e.g. validations, replication triggers)

Design considerations:
- In the new `ArchivePreservedCopy`, should it `belongs_to :moab_preserved_copy` rather than the PO directly?  That seems to match our process for creating the rows and our query logic based on online PC timestamps.
- Do we want to split the `Endpoint` model similarly for online/archive types also?  Separation would make strict validation very easy, and misaligning data (online PC getting an archive endpoint) impossible.  I don't think that there is much commonality between a locally mounted network disk and a remote host we pay thousands of dollars to implement a given protocol, authentication, etc.

jmartin-sul:
can also maybe get rid of `Endpoint` `#to_h`and `#to_s` and use the defaults?

other decisions after talking F2F w/ @atz:
* `MoabPreservedCopy` will replace `PreservedCopy` and will still belong to `PreservedObject`. 
  * each `ArchivePreservedCopy` instance will belong to the `MoabPreservedCopy` instance which triggered its creation.
    * each `ArchivePreservedCopy` will have 1 or more `ArchivePreservedCopyPart` objects.  each `Part` represents a chunk of a zip archive for a particular druid version living on a specific cloud endpoint.  this is because we split archive zips into 10 GB chunks.
      * maybe the checksums can live on the parts instead of in their own table?  or at least reference the parts?

atz:
Right, I think `zip_checksums` **must** `belong_to` the parts

atz:
Different constraints:
- MPC unique by druid, i.e. PO (?)
- APC unique by druid/version/endpoint.  Whichever model APC belongs_to (APC or PO) is the stand-in for druid in terms of DB uniqueness.

PR #1008, opened 2018-07-23
re-work ReplicatedFileCheckJob to deal with multi-part zips

### LINKED ISSUE #973: (C2A) refactor and incorporate archived_preserved_copies and parts

### LINKED ISSUE #994: (R) db rename/refactor: remove _deprecated factories
need to get rid of `archive_endpoint_deprecated`, `archive_copy_deprecated`, and `unreplicated_copy_deprecated`.

### LINKED ISSUE #998: (R) db refactor: remove UNREPLICATED_STATUS and REPLICATED_COPY_NOT_FOUND_STATUS from CompleteMoab model (blocked by #980)
leftovers from splitting the archive pres copy model out of pres copy.

jmartin-sul:
this branch started out as an attempt to address #973 (hence the name).  but it turns out the test code using the deprecated fixtures was testing app code that needed to be re-worked, because the app code predated the model split (#927) and the subsequent implementation of multi-part druid-version zips.

this PR will address the deprecated fixtures in the course of re-working the app and test code for the updated data model.

closes #973 
closes #994
closes #998


app/lib/audit/catalog_to_archive.rb
```diff
@@ -0,0 +1,71 @@
+module Audit
+  # Catalog to cloud archive provider (what ZipEndpoint, ZippedMoabVersion etc represent) audit code
+  class CatalogToArchive
+    # delegate :bucket, :bucket_name, to: PreservationCatalog::S3
+
+    def self.logger
+      @logger ||= Logger.new(STDOUT)
```
atz: Do we need to reproduce this weird double-logging pattern?  If you note the places we actually schedule these jobs with `whenever`, we are basically routing the STDOUT portion to `/dev/null` anyway.

jmartin-sul: discussed F2F, 👍 will just log to `c2a.log`

app/lib/audit/catalog_to_archive.rb
```diff
@@ -0,0 +1,71 @@
...
+          logger.error("Archival Preserved Copy: #{zmv.inspect} #{part.inspect} was not found on #{bucket_name}.")
+          part.update(status: 'not_found', last_existence_check: Time.zone.now)
+          next
+        end
+
+        replicated_checksum = replicated_checksum(aws_s3_object)
+        if part.md5 == replicated_checksum
+          part.update(last_checksum_validation: Time.zone.now)
+        else
+          logger.error("Stored checksum(#{part.md5}) doesn't match the replicated checksum(#{replicated_checksum}).")
```
jmartin-sul: maybe a comment making it explicit that no checksum computation is happening, this is just checking metadata that we gave AWS and that we also still store, for drift between the two systems.  we never expect that to differ, but we're here, and it's a cheap check, so why not?  because it'd be pretty weird if they did differ.

spec/lib/audit/catalog_to_archive_spec.rb
```diff
@@ -0,0 +1,313 @@
+require 'rails_helper'
+require_relative '../../load_fixtures_helper.rb'
```
jmartin-sul: this `require_relative` may be unnecessary copypasta from example spec

jmartin-sul:
ok, all suggestions from yesterday have been addressed, ready for review again.

app/models/zipped_moab_version.rb
```diff
@@ -15,21 +15,29 @@ class ZippedMoabVersion < ApplicationRecord
...
-  }
-
-  validates :complete_moab, :status, :version, :zip_endpoint, presence: true
+  validates :complete_moab, :version, :zip_endpoint, presence: true
 
   scope :by_druid, lambda { |druid|
     joins(complete_moab: [:preserved_object]).where(preserved_objects: { druid: druid })
   }
 
+  # ideally, there should be only one distinct parts_count value among a set of sibbling
```
atz: `sibling`

jmartin-sul: d'oh 🤦‍♂️

PR #1013, opened 2018-07-24
Order moved to class method for scope least_recent_version_audit and …

### LINKED ISSUE #933: Move order out of scopes
Orders in scopes introduce incompatibilities (with each other) and prevent scopes from being cleanly chainable.  For the same result, keep the query-tightening logic in the scope and wrap it with a class method that applies the order.

jermnelson:
…fixity_check_expired
closes #933

### DIFF
--- a/spec/models/preserved_copy_spec.rb
+++ b/spec/models/preserved_copy_spec.rb
@@ -170,7 +170,7 @@ RSpec.describe PreservedCopy, type: :model do
     end
   end
 
-  describe '.least_recent_version_audit' do
+  context 'ordered (by last version_audited) and unordered least_recent_version_audit' do
     let!(:newer_timestamp_pc) do
       create(:preserved_copy, args.merge(version: 6, last_version_audit: (now - 1.day)))
     end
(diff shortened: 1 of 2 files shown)


spec/models/preserved_copy_spec.rb
```diff
@@ -284,34 +306,34 @@
...
-          expect(pcs_ordered_by_query1).to eq [pc, checked_before_threshold_pc1]
-          expect(pcs_ordered_by_query2).to eq [checked_before_threshold_pc2]
+        it 'returns PreservedCopies < given date only for the chosen storage root (not ordered by last_version_audit)' do
+          expect(random_pcs_by_query1.sort).to eq [pc, checked_before_threshold_pc1]
+          expect(random_pcs_by_query2.sort).to eq [checked_before_threshold_pc2]
         end
         it 'returns no PreservedCopies with timestamps indicating fixity check in the last week' do
-          expect(pcs_ordered_by_query1).not_to include recently_checked_pc1
-          expect(pcs_ordered_by_query2).not_to include recently_checked_pc2
+          expect(random_pcs_by_query1).not_to include recently_checked_pc1
```
jermnelson: Any help with variable names would be great.

spec/models/preserved_copy_spec.rb
```diff
@@ -229,11 +244,26 @@
 
     before { pc.save! }
 
-    it 'returns PreservedCopies that need fixity check, never checked first, then least-recently to most-recently' do
-      expect(described_class.fixity_check_expired.to_a).to eq [pc, old_check_pc1, old_check_pc2]
+    describe '.fixity_check_expired' do
+      it 'returns PreservedCopies that need fixity check, never checked first, then least-recently to most-recently' do
```
jmartin-sul: i think the second part of this test title (about the order, "never checked first, then least-recently to most-recently") can be removed.

spec/models/preserved_copy_spec.rb
jmartin-sul: i think you could just drop the `random_` prefix and that'd be fine.  the var name as it is reads to me more like the set of things returned is random, as opposed to the ordering being non-deterministic (which i think will be even less obvious outside the context of this diff).

PR #1018, opened 2018-07-25
Replication trigger (part 2, take 2): auto-hook replication

### LINKED ISSUE #989: (R) only replicate CompleteMoab if status is ok
so `invalid_moab`, `invalid_checksum`, `validity_unknown`, and `unexpected_version_on_storage` moabs should *NOT* be replicated to cloud endpoints.  they should be remediated and set to `ok` before being allowed to replicate.

atz:
Added `replicatable_status?` to PC to encapsulate logic about whether it is OK to push dependent objects into pipeline.

`replicate!` doesn't raise, because that would make us unable to create PCs with non-replicatable statuses (since that auto-creates ZMVs) or ZMVs from such PCs.  Method is public on the anticipation of audit processes that might also invoke it directly.

Closes #989

### DIFF
--- a/spec/models/zipped_moab_version_spec.rb
+++ b/spec/models/zipped_moab_version_spec.rb
@@ -1,7 +1,6 @@
 require 'rails_helper'
 
 RSpec.describe ZippedMoabVersion, type: :model do
-  let(:po) { build(:preserved_object) }
   let(:pc) { build(:preserved_copy) }
   let(:zip_endpoint) { build(:zip_endpoint) }
   let(:zmv) { build(:zipped_moab_version, preserved_copy: pc, zip_endpoint: zip_endpoint) }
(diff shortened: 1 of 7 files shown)


spec/models/zipped_moab_version_spec.rb
```diff
@@ -45,4 +44,18 @@
...
+
+    it 'if PC is unreplicatable, returns false, does not enqueue' do
+      expect(pc).to receive(:replicatable_status?).and_return(false)
+      expect(ZipmakerJob).not_to receive(:perform_later)
+      expect(zmv.replicate!).to be(false)
+    end
+    it 'if PC is replicatable, passes druid and version to Zipmaker' do
+      expect(pc).to receive(:replicatable_status?).and_return(true)
+      expect(ZipmakerJob).to receive(:perform_later).with(pc.preserved_object.druid, pc.version)
+      zmv.replicate!
```
jmartin-sul: total nitpick, and wouldn't hold up merge over this, but the slight difference between these two tests makes me think that `ZMV#replicate!` should return true after its call to `perform_later`, just so it has a consistent return type.  unless the return value of `perform_later` would be useful (in which case i'd advocate having the guard clause return `nil`, just because i like consistent/narrow return types).

like i said, a nitpick, and not something i'd hold this up over.

_edit_:  ok, after a bit of looking, it seems that `perform_later` is supposed to return the job id, which does seem useful?  if that's the case, i'd advocate (mildly) for the guard clause returning `nil`.

jmartin-sul: apparently i also missed that the return types were listed on the method comment.

anyway, posted #1020 to deal with this nitpick, so as to get this PR merged.

PR #1020, opened 2018-07-26
have ZippedMoabVersion#replicate! guard clause return nil instead of false, for consistent return types

jmartin-sul:
just being a little OCD about return types here.

feel free to close if this feels silly or counterproductive.  seemed better to merge #1018 than to hold it up over this, and it wasn't really harder to put up a PR than it was to file a ticket describing the issue.

### DIFF
--- a/spec/models/zipped_moab_version_spec.rb
+++ b/spec/models/zipped_moab_version_spec.rb
@@ -50,7 +50,7 @@ RSpec.describe ZippedMoabVersion, type: :model do
     it 'if PC is unreplicatable, returns false, does not enqueue' do
       expect(cm).to receive(:replicatable_status?).and_return(false)
       expect(ZipmakerJob).not_to receive(:perform_later)
-      expect(zmv.replicate!).to be(false)
+      expect(zmv.replicate!).to be(nil)
     end
     it 'if PC is replicatable, passes druid and version to Zipmaker' do
       expect(cm).to receive(:replicatable_status?).and_return(true)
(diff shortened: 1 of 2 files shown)


spec/models/zipped_moab_version_spec.rb
```diff
@@ -50,7 +50,7 @@
     it 'if PC is unreplicatable, returns false, does not enqueue' do
       expect(pc).to receive(:replicatable_status?).and_return(false)
       expect(ZipmakerJob).not_to receive(:perform_later)
-      expect(zmv.replicate!).to be(false)
+      expect(zmv.replicate!).to be(nil)
```
jmartin-sul: i didn't add the analogous expectation to line 58, because we monkeypatch `ZipmakerJob.perform_later` to do nothing for most of the tests, and it didn't seem worth tweaking that or jumping through other hoops just to confirm that `ActiveJob.perform_later` returns what you'd expect.

https://github.com/sul-dlss/preservation_catalog/blob/ad124bdd0cda75ccc8e775166b10f9202dc57e26/spec/rails_helper.rb#L56-L58

atz:
Consistent with what?

jmartin-sul:
itself.  i guess i like when a method is constrained to returning a single type?  like i said, not a huge deal.

jmartin-sul:
a single type that is narrower than `Object`, if possible 🙂

atz:
OK, I'm fine w/ that if you want to rebase.  It isn't really entirely different though.  `NilClass` is itself also a separate type than `ZipmakerJob`, too.

jmartin-sul:
@atz:  rebased and re-pushed.

but, yeah, fair point about `NilClass`.  i still like this, but if you think i'm bringing too much static type baggage/bias to ruby land, feel free to nix it.  was easy quick to implement, was quick to rebase, so not worried about wasting effort.  merge or close at your discretion.

PR #1022, opened 2018-07-26
Rename PreservedCopy to CompleteMoab

tallenaz:
Fixes #980

PR #1023, opened 2018-07-26
Rename a few stray PCs to CM

tallenaz:
As recorded in comments for https://github.com/sul-dlss/preservation_catalog/pull/1022

### DIFF
--- a/app/models/preserved_object.rb
+++ b/app/models/preserved_object.rb
@@ -17,10 +17,10 @@ class PreservedObject < ApplicationRecord
 
   scope :without_complete_moabs, -> { left_outer_joins(:complete_moabs).where(complete_moabs: { id: nil }) }
 
-  # Spawn asynchronous checks of all existing archive complete_moabs.
+  # Spawn asynchronous checks of each associated ZippedMoabVersion.
   # This logic is similar to PlexerJob, for a different purpose.
-  # This should implement the start of the replication process if status is unreplicated for a ZippedMoabVersion
-  # Compare last_existence_check (from archive comp_moab) with archive TTL when checking the ZippedMoabVersion status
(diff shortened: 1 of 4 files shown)


app/models/preserved_object.rb
```diff
@@ -19,8 +19,8 @@ class PreservedObject < ApplicationRecord
 
   # Spawn asynchronous checks of all existing archive complete_moabs.
   # This logic is similar to PlexerJob, for a different purpose.
-  # This should implement the start of the replication process if status is unreplicated for an archival pres_copy
-  # Compare last_existence_check (from archive pres_copy) with archive TTL when checking the archival pres_copy status
+  # This should implement the start of the replication process if status is unreplicated for an archival comp_moab
```
SaravShah: i think archival complete moab is the zipped moab version?

SaravShah: there's a few other places that has this too

jmartin-sul: i think with the new terminology and data model, a more accurate sentence might be:  `This should implement the start of the replication process if the CompleteMoab doesn't have all of the expected ZippedMoabVersion objects in a fully replicated state.`

app/models/preserved_object.rb
```diff
@@ -19,8 +19,8 @@ class PreservedObject < ApplicationRecord
 
   # Spawn asynchronous checks of all existing archive complete_moabs.
```
jmartin-sul: not in the scope of this PR, but now that i look at this comment overall, this could probably use fixing too (maybe `Spawn asynchronous checks of each associated ZippedMoabVersion`

tallenaz:
@SaravShah @jmartin-sul this is ready for re-review. I took the liberty of porting over the changes from #1026

tallenaz:
So, this could also close #1026 if that approach is OK.

PR #1030, opened 2018-07-28
POH use more OO, especially in specs

atz:
Small revisions to app code:
- `PO#complete_moabs` gets `autosave: true`
- 2 "getter" methods go public in POH
- associations are used to find and create
- keep non-transactional code outside the transaction block

Larger revisions to tests:
- remove unnecessarily eager `let!` statements
- associations are used to find and create
- intercept POH `pres_object` and `complete_moab` directly when desirable, rather than go underneath it to intercept the model's `.find_by!`
- use `change` matcher
- use imperative enum calls (e.g. `cm.ok!`)

Rebased to make mergeable after PC renaming.

### DIFF
--- a/spec/services/preserved_object_handler_confirm_version_spec.rb
+++ b/spec/services/preserved_object_handler_confirm_version_spec.rb
@@ -224,6 +224,7 @@ RSpec.describe PreservedObjectHandler do
           allow(PreservedObject).to receive(:find_by).with(druid: druid).and_return(po)
           allow(po.complete_moabs).to receive(:find_by!).with(moab_storage_root: ms_root).and_return(cm)
         end
+
         it 'if the existing record is altered' do
           allow(po_handler).to receive(:moab_validation_errors).and_return([])
           expect(cm).to receive(:save!)
--- a/spec/services/preserved_object_handler_confirm_version_spec.rb
+++ b/spec/services/preserved_object_handler_confirm_version_spec.rb
@@ -98,6 +98,7 @@ RSpec.describe PreservedObjectHandler do
 
         context 'incoming_version > db version' do
           let(:incoming_version) { cm.version + 1 }
+
           before { allow(po_handler).to receive(:moab_validation_errors).and_return([]) }
 
           it 'had OK_STATUS, but is now UNEXPECTED_VERSION_ON_STORAGE_STATUS' do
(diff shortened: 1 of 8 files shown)


SaravShah:
Everything looks to me, not sure if you want more than 1 set of eyes.

atz:
Rebased and resolved some conflicts

spec/services/preserved_object_handler_create_spec.rb
```diff
@@ -2,49 +2,33 @@
...
   describe '#create' do
     it 'creates PreservedObject and CompleteMoab in database' do
-      expect(PreservedObject).to receive(:create!).with(po_args).and_call_original
-      expect(CompleteMoab).to receive(:create!).with(cm_args).and_call_original
       po_handler.create
+      new_po = PreservedObject.find_by(druid: druid)
+      new_cm = new_po.complete_moabs.find_by(version: incoming_version)
+      expect(new_po.current_version).to eq incoming_version
+      expect(new_cm.moab_storage_root).to eq ms_root
+      expect(new_cm.size).to eq incoming_size
```
jmartin-sul: i'd be inclined to confirm `status` is `validity_unknown` on `new_cm`.  would also prob check for `nil` validation timestamps (per the comment in the old `cm_args` hash).

spec/services/preserved_object_handler_confirm_version_spec.rb
```diff
@@ -222,59 +201,50 @@
...
+          allow(po_handler).to receive(:moab_validation_errors).and_return([]) # different po_handler now
+          expect(cm).to receive(:save!)
+          expect(po).not_to receive(:save!)
+          po_handler.confirm_version
+        end
       end
+
       it 'logs a debug message' do
-        msg = "confirm_version #{druid} called"
         allow(Rails.logger).to receive(:debug)
```
jmartin-sul: total nitpick, but this line is probably superfluous now, right?

atz: No, the point is it receives multiple calls to debug, but we only care about confirming one.  If you only set the expectation, you would get an error like `expected :x with 'val', but got :x with 'other'`

spec/services/preserved_object_handler_create_spec.rb
atz: I believe that is covered in other examples.  I think of this example being about the critical **identity** of the created objects rather than the complete examination of every attribute.  YMMV.

jmartin-sul: the next test does check those attributes, but it's for `po_handler.create(true)`, whereas this one's about `po_handler.create`, which i'd expect to create objects with differing values for status and validation times.

Issue #792, opened 2018-05-22
update DB schema diagram


jmartin-sul:
schema's in flux for a bit, so we might want to wait till things settle before doing this, so we're not churning unnecessarily.

see also #772 & #759

jmartin-sul:
document the difference between storage roots and archive endpoints, per #888 (which is closed due to inacuracy, now that the data model has been refactored and storage roots are represented by a different class/table than archive endpoints).

SaravShah:
Started working on in the db/README about the Key

PR #1036, opened 2018-07-31
DB/README

SaravShah:
Fixed and edited existing readme. Didn't change the transaction, cause I don't think that has changed? Should I remove this section, due to @atz transaction changes in #1030?

Paired w/ @jermnelson 

connects #792


atz:
Unsure, it seems like a lot of this is just "ActiveRecord 101 Tutorial" and doesn't belong as documentation of this particular app.

SaravShah:
@atz yeah thats what I'm thinking. Maybe i'll remove it. What do other devs think?

jmartin-sul:
i'd be in favor of leaving transaction advice, but i have no objections to a little ActiveRecord 101 in an app that uses ActiveRecord heavily.  ideally we'd all be super-familiar with the tools at hand when being put on a given project, but i don't think that's actually a realistic expectation given our track record as a department, so denormalizing a bit of the relevant documentation from the wider internet seems like a pragmatically good move to me.

atz:
I think we can defer the proviso about "use AR transactionality as much as you want, roll your own as little as possible" to a follow-up.

PR #1045, opened 2018-08-01
dependent option is ignored when using :through option

### LINKED ISSUE #1014: ZippedMoabVersion#preserved_object association has useless `dependent` option
https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_one
> Note that :dependent option is ignored when using :through option.

IIRC, i added that flag because rubocop complains if you specify a `has_one` without a `dependent` option.  if that's not actually the case, and rubocop didn't complain, we should just get rid of that option on that association.  otherwise, maybe we should make a rubocop exception?

but this is basically a small bit of harmless clutter as is.

atz:
See: https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_one

Closes #1014

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,9 +6,10 @@ require: rubocop-rspec
 AllCops:
   DisplayCopNames: true
   Include:
-    - Rakefile
-    - config.ru
-    - lib/**/*.rake
+    - './Rakefile'
+    - './config.ru'
+    - '**/*.rb'
+    - '**/*.rake'
   Exclude:
     - 'Gemfile.lock'
     - '**/*.md'
@@ -41,10 +42,6 @@ Layout/EmptyLinesAroundModuleBody:
 Layout/SpaceAroundEqualsInParameterDefault:
   Enabled: false
 
-Layout/SpaceInsideBrackets:
-  Exclude:
-    - 'app/services/preserved_object_handler.rb' # line 126 is clearer with the spaces
-
 # --- Lint ---
 
 Lint/AmbiguousBlockAssociation:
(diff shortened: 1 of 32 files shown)


jmartin-sul:
rubocop.

one thing i just noticed in the doc you linked:
> This method should only be used if the other class contains the foreign key.

which is not the case here.

IIRC, there's only one scope that uses that `has_one` as a convenience.  maybe it wouldn't be so bad to re-work it to use `joins` explicitly?  this thing, i think?
https://github.com/sul-dlss/preservation_catalog/blob/9ed2b70da75651df80e1c08f5d786786d0079b5a/app/models/zip_endpoint.rb#L24-L31

we'd have to join through to `complete_moabs` explicitly.  i don't think anything else needs that `has_through` (unless someone else has written a query that takes advantage of it since i added it).

maybe we should just punt on this for now and ticket the other thing and get rid of this `has_one`?  status quo seems harmless enough for the moment.

atz:
Rubocop flagging this is a bug in the old version we have:
https://github.com/rubocop-hq/rubocop/issues/4751

atz:
@jmartin-sul we looked at that when we first stumbled on the docs.  The other class *does* contain the foreign key, for the thing we are associating to.  For this, it isn't important that `ZippedMoabVersion` has the FK to `CompleteMoab`, because that is already established by `belongs_to :complete_moab`.  It is important that this class **does not** have the FK to `PreservedObject`.  I was confused by this initially too, but it makes sense to me now.  

Updated our rubocop and rubocop-rspec versions and got it to be cool w/ the 1-line change.  As a side-benefit, a bunch of extra whitespace cleanup.

jmartin-sul:
@atz:  ah, ok, i misread the docs as saying implying that the FK should point back to the thing that `has_one` (i.e., in my possible misunderstanding, i read the docs as saying that the FK in this case should be `PreservedObject` -> `CompleteMoab` -> `ZippedMoabVersion`, which is the opposite direction of what things are).

and that made sense to me, because the `has_one` otherwise seems to imply FK relationship back to the "haver", which also makes sense in calling the thing that is "had" a dependent.

but if that's not the case, i retract my above comment.

i dunno, the language feels ambiguous to me.  you've got more AR experience than me, so i'd trust your judgement on this.

jmartin-sul:
anyway, regardless of whether we should be using `has_one`, the `dependent` option gets ignored, and the other style clean up looks good, so this seems like an improvement regardless of whether i'm understanding the other `has_one` guidance correctly.  merged.

PR #1046, opened 2018-08-01
Remove double logging cronjob & add CVJob test for locking

SaravShah:


atz:
Cool, I see that passing explicit `nil` turns into a redirection to `/dev/null`, which sounds right:

https://github.com/javan/whenever/blob/master/lib/whenever/output_redirection.rb#L23

PR #1048, opened 2018-08-02
Stop nuking the DB every time rspec runs

atz:
This practice seems to have derived from having non-rails_helper'd code writing to the database, as with our old `./lib` directory.  It is unnecessary for anything using rails_helper, however, which now includes all the code that writes to the DB during tests.

We already have transactional fixtures turned on so every change during the rspec suite is atomically rolled back at the end of the example where it was made.

Also, we use the fixtures directory, but we don't use it for ActiveRecord fixtures, so that config in rails_helper is misleading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,12 +44,15 @@ Using the `psql` utility, run these two setup scripts from the command line, lik
 ```sh
 psql -f db/scripts/pres_setup.sql postgres
 psql -f db/scripts/pres_test_setup.sql postgres
+bundle exec rails db:seed
+RAILS_ENV=test bundle exec rails db:seed
 ```
 
 These scripts do the following for you:
 * create the test and dev PostgreSQL users.
 * create the test and dev databases.
 * setup the needed ownership and permissions between the DBs and the users.
+* seed data into development and test environments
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -44,12 +44,15 @@ Using the `psql` utility, run these two setup scripts from the command line, lik
...
 psql -f db/scripts/pres_test_setup.sql postgres
+bundle exec rails db:seed
+RAILS_ENV=test bundle exec rails db:seed
 ```
 
 These scripts do the following for you:
 * create the test and dev PostgreSQL users.
 * create the test and dev databases.
 * setup the needed ownership and permissions between the DBs and the users.
+* seed data into development and test environments
```
SaravShah: for development do we need to run `bundle exec rake m2c:seed_catalog`?

atz: AFAIK, no.  There's `m2c:seed_all_roots` and `m2c:seed_root`, but no `m2c:seed_catalog`.  Neither is important *for general development*.  If you want to run them for a given purpose, of course you could.  In any case, that isn't really part of the PR, which just moves `db:reset` from an every-rspec-execution scope to a "first time setup" scope.

PR #1050, opened 2018-08-03
Added tests for size and md5 in DruidVersionZipPart 

### LINKED ISSUE #969: DruidVersionZipPart model missing specs
- size method
- md5 method

jermnelson:
Spec tests for `size` and `md5` methods in DruidVersionZipPart class.

Closes #969

### DIFF
--- a/spec/models/druid_version_zip_part_spec.rb
+++ b/spec/models/druid_version_zip_part_spec.rb
@@ -25,7 +25,7 @@ describe DruidVersionZipPart do
     end
   end
 
-  context 'MD5 checksums' do
+  context 'MD5 checksums and size' do
     let(:part) { described_class.new(dvz, 'bj/102/hs/9687/bj102hs9687.v0001.zip') }
 
     before do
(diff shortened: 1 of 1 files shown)


spec/models/druid_version_zip_part_spec.rb
```diff
@@ -51,5 +51,19 @@
         expect(part.hex_to_base64('d41d8cd98f00b204e9800998ecf8427e')).to eq '1B2M2Y8AsgTpgAmY7PhCfg=='
       end
     end
+
+    describe '#md5' do
+      it 'returns MD5 value' do
+        expect(part.md5.to_s).to eq '4f98f59e877ecb84ff75ef0fab45bac5'
+        expect(part.md5).to eq part.metadata[:checksum_md5]
```
atz: Only question I have here is that you make an explicit `to_s` cast to make a string comparison in 57, but then you don't cast it in 58 when also comparing to a string.  Probably clearer to say `returns Digest::MD5 object` than *value*.

jermnelson: Thanks Joe, I made the changes you suggested and currently pushing changes up.

PR #1055, opened 2018-08-04
Write md5 and read md5 before upload

### LINKED ISSUE #1040: Additional Pre Delivery Checksumming
Check md5 of zip/part right before we send it to the cloud.

1. Can save a file w/ checksum to sdr_transfers?
or...???

SaravShah:
closes #1040

### DIFF
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -31,6 +31,9 @@ class DruidVersionZip
       combined, status = Open3.capture2e(zip_command)
       raise "zipmaker failure #{combined}" unless status.success?
     end
+    part_keys.each do |part_key|
+      DruidVersionZipPart.new(self, part_key).write_md5
+    end
   end
 
   # Ensure the directory the zip will live in exists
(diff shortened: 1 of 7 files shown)


app/models/druid_version_zip_part.rb
```diff
@@ -54,4 +54,19 @@ def metadata
   def size
     @size ||= FileTest.size(file_path)
   end
+
+  # @return [String] MD5 path
+  def md5_path
+    file_path + ".md5"
```
tallenaz: Nitpick, but how about interpolating here?

app/models/druid_version_zip.rb
```diff
@@ -82,7 +85,9 @@ def part_keys
   #  the last one will end .zip, so two parts is:  .z01, zip. (this agrees with zip utility)
   # @return [Array<Pathname>] Existing pathnames for zip parts based on glob (.zip, .z01, .z02, etc.)
   def part_paths
-    @part_paths ||= Pathname.glob(file_path.sub(/.zip\z/, '.z*'))
+    @part_paths ||= Pathname.glob(file_path.sub(/.zip\z/, '.z*')).reject do |path|
+      path.to_s =~ /.md5\z/
+    end
```
tallenaz: This could also be a one-liner, with curly braces instead of the do/end, I think?

SaravShah: rubocop complains for line length

PR #1056, opened 2018-08-06
Add zip cache cleanup cronjob and capistrano role

### LINKED ISSUE #1032: (R) implement cache expiry in zip transfers volume
can be a simple script, e.g. rake task or bash script.

schedule using `whenever` gem from pres cat?

atz:
Currently, this would run every hour and delete any files older than 1 day. So the oldest a file would be is 1 day (and 59 minutes). 

This may require further tuning depending on how we choose to enqueue backlog for `ZipmakerJob` or when we have completed the backlog and have different operational expectations for volume of work (i.e. drive less full, files could live longer).

The implication of cache purging is that the zip files must be kept longer than the latency between `ZipmakerJob`'s completion and completion of delivery job(s), inclusive of Plexer latency.  Otherwise the zip file will be deleted before it is delivered.  So any tuning should be measured against performance on those queues.

Closes #1032

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -26,3 +26,10 @@ every :sunday, at: '1am', roles: [:cv] do
   set :output, standard: nil, error: 'log/cv-err.log'
   rake "cv:all_roots"
 end
+
+every :hour, roles: [:cache_cleaner] do
+  set :output, standard: 'log/zip_cache_cleanup.log'
+  command <<-'END_OF_COMMAND'
+    find /sdr-transfers -mindepth 5 -type f -name "*.zip" -mtime +1 -exec bash -c 'TARGET="{}"; rm -v ${TARGET%ip}*' \;
+  END_OF_COMMAND
+end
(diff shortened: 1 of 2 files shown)


config/schedule.rb
```diff
@@ -26,3 +26,10 @@
   set :output, standard: nil, error: 'log/cv-err.log'
   rake "cv:all_roots"
 end
+
+every :hour, roles: [:cache_cleaner] do
+  set :output, standard: 'log/zip_cache_cleanup.log'
+  command <<-'END_OF_COMMAND'
+    find /sdr-transfers -mindepth 5 -type f -name "*.zip" -mtime +1 -exec bash -c 'TARGET="{}"; rm -v ${TARGET%ip}*' \;
```
tallenaz: I'd like to suggest setting `/sdr-transfers` via the `zip_command` setting. We did some similar work in the `preservation_robots` (https://github.com/sul-dlss/preservation_robots/blob/master/config/schedule.rb#L1-L2) to hook scheduled tasks up to settings.

tallenaz: Sorry, not `zip_command`, but `zip_storage`. I edited the above comment accordingly.

PR #1062, opened 2018-08-07
Convert C2M to worker Queue

atz:
Note: we already abandoned order when we moved order from scopes to class methods.  The service classes still called the scopes and not the class methods, so no ordering was happening.  

That's more obvious now, since moving to a queue means asynchronicity inherently.  Order enqueued does not guarantee order of completion.  Also presumably parallelization provides better enough performance and reliability that we aren't worried about some objects never making it through.  

Massive cleanup to twisty code.  No entanglement of class methods and instance methods.  A factory for fixtures!

Follow on work would include:
- ~~cleanup README~~, maybe wiki
- add new workers to resque-pool(s) **required**
- change how we schedule C2M, since the "every saturday" stuff is not particularly relevant anymore


jmartin-sul:
merged, happy to do my nitpick as a follow-on if others find it agreeable.

PR #1065, opened 2018-08-08
Convert M2C to worker queue also

atz:
This extends #1062 for M2C, using the same approach and with the same follow-on expectations. Will rebase if that is merged.


SaravShah:
should we wait for Tony before meeting, since we need to set up queues via resque pool?

PR #1066, opened 2018-08-09
Remove load fixtures

atz:
Extends #1065, will rebase when that is merged.

This excises `load_fixtures_helper` by using the fixure-aware factory when needed.

Also cleans up major swaths of the README to use rails console instead of rake tasks.  Tasks that can be removed completely are removed.


README.md
```diff
@@ -16,7 +16,6 @@ Rails application to track, audit and replicate archival artifacts associated wi
     * [Catalog to Moab](#c2m) (C2M) existence/version check
     * [Checksum Validation](#cv) (CV)
     * [Seed the catalog](#seed-the-catalog)
-    * [Audit Checks via Console](#console)
```
SaravShah: ah good catch

app/models/complete_moab.rb
```diff
@@ -26,14 +26,6 @@ class CompleteMoab < ApplicationRecord
   # NOTE: size here is approximate and not used for fixity checking
   validates :size, numericality: { only_integer: true, greater_than: 0 }, allow_nil: true
 
-  scope :by_moab_storage_root_name, lambda { |name|
-    joins(:moab_storage_root).where(moab_storage_roots: { name: name })
-  }
-
-  scope :by_storage_location, lambda { |storage_dir|
```
SaravShah: wow do we not use this anywhere?

app/models/complete_moab.rb
```diff
@@ -26,14 +26,6 @@ class CompleteMoab < ApplicationRecord
   # NOTE: size here is approximate and not used for fixity checking
   validates :size, numericality: { only_integer: true, greater_than: 0 }, allow_nil: true
 
-  scope :by_moab_storage_root_name, lambda { |name|
```
SaravShah: nice - also changed it in `lib/audit/checksum.rb` 👍

app/models/complete_moab.rb
atz: Not after my revisions.  Basically, the use case for most of our "scopes that join" should be:
1. do a `.find_by!` on the thing you are join on
2. use the association to get objects of the type where the scope used to live (which can still be chained w/ other scopes)

This reads a lot better, and if it fails (the `find_by!`) it fails in a much more obvious way.  Less custom code for us to track.

app/models/complete_moab.rb
atz: Right, same pattern: `OtherModel.find_by!(...).association`

spec/services/checksum_validator_spec.rb
```diff
@@ -567,8 +549,7 @@
     context 'transaction is rolled back' do
       before do
         # would result in a status update if the save succeeded
-        comp_moab.status = 'online_moab_not_found'
-        comp_moab.save!
+        comp_moab.online_moab_not_found!
```
tallenaz: 👍 to all of these enum-bangs

spec/services/checksum_validator_spec.rb
```diff
@@ -1,14 +1,12 @@
 require 'rails_helper'
-require_relative '../load_fixtures_helper.rb'
 
 RSpec.describe ChecksumValidator do
-  include_context 'fixture moabs in db'
```
tallenaz: 👍 to getting free from shared contexts

PR #1069, opened 2018-08-11
Use a ruboocop_todo file

atz:
The whole point of `Metrics` cops is that you run them frequently and get a number in your TODO file of how close you are to your goal.

The TODO file also identifies, effectively, "temporary exemptions".  If the file gets cleaned up or deleted, the TODO will disappear when the TODO is regenerated. If the file is renamed, the TODO will track with it automatically upon regeneration.

This is more flexible and informative than coding every exception as permanent.

Unfortunately, when the number of files that it would take to exclude to make a cop pass gets too large, sometimes the heuristic behavior of rubocop is to disable the whole cop, or adjust a setting (like Max) to add to the TODO file.  This is reasonable behavior, just something to be aware of.  It also accounts for why some individual exceptions that I flagged as FIXMEs are still in the main `.rubocop.yml` config.

Upped the MAX for `Metrics/LineLength` to 150, which seems to reflect our actual usage, leaving only 2 files as FIXME exceptions.

Replaces #733


.rubocop.yml
```diff
@@ -1,8 +1,8 @@
-inherit_from: .rubocop_todo.yml
-
 # Turn on RSpec cops
 require: rubocop-rspec
 
+inherit_from: .rubocop_todo.yml
```
SaravShah: why does it matter to move this?

PR #1070, opened 2018-08-13
MoabReplicationAuditJob

### LINKED ISSUE #900: (PO2R) Report Results of PO2R for a preserved_object ... to whom?  how? (blocked by Storytime-ish)
Need some info from Julian/Ben at next storytime about reporting decisions.

If endpoint C2A check for a specific archival pres_copy is good, then update a timestamp (last checksum validated / last audited)  on pres_copy - done in worker? (See #813)

If archival object not there, not checksum match, other error, then:
- do we need to collate pres_copies by pres_object?
- report to ??
    - status update to not found or invalid checksum or ...
    - log errors ?
    - send errors to argo via workflows

### LINKED ISSUE #920: (PO2R) for each PO, does it have complete_moab and all the zipped_moab_versions with status ok (blocked by #1008 C2A)
- for each preserved_object in the catalog, does it have an online PC with status Ok and  does it have all of the archival pres_copies with status ok, per preservation_policy.
    -  this will be a db query - the same as or similar to that used in current replication

jmartin-sul:
here's a first pass at implementing the job to audit an entire CompleteMoab to determine what, if anything, isn't replicated as expected.

still left TODO:
- [ ] query to populate the queue
- [ ] schedule queue population
  - [ ] related: as discussed F2F last week: while we're there, switch all queue population whenever jobs to use a single queue population cap role, instead of the old roles (`m2c`, `c2m`, `cv`) that were initially created for running the audits themselves, and not for populating the queues that workers will look at to perform the requested audits.
- [ ] backlog ticket: creation of a new replicationWF, ability to specify which workflow a particular audit result code goes to.
  * we'd prefer to send replication related audit results to a replication specific workflow.  but we think this is safe for now, since we don't intend to replicate `CompleteMoab`s that aren't `ok`, and as such wouldn't haven preservation audit workflow errors to step on.
- [ ] backlog ticket: spot fixity check of archive (pull down and checksum) if not already ticketed

closes #900
closes #920


app/jobs/moab_replication_audit_job.rb
```diff
@@ -0,0 +1,42 @@
+# Confirms that a CompleteMoab is fully/properly replicated to all target zip endpoints.
+# Usage info:
+# ReplicatedFileCheck.set(queue: :endpoint_check_us_west_2).perform_later(cm)
+# ReplicatedFileCheck.set(queue: :endpoint_check_us_east_1).perform_later(cm)
```
tallenaz: The class name looks wrong to me in this usage info since I see `ReplicatedFileCheck` gets deleted.

app/lib/preservation_catalog/s3/audit.rb
```diff
@@ -4,52 +4,71 @@ module S3
...
-          else
-            logger.error("Stored checksum(#{part.md5}) doesn't match the replicated checksum(#{replicated_checksum}).")
-            part.update(status: 'replicated_checksum_mismatch', last_checksum_validation: Time.zone.now)
-            next
-          end
+      private
 
-          part.ok!
+      def compare_checksum_metadata(aws_s3_object, part)
+        # NOTE: no checksum computation is happening here (neither on our side, nor on AWS's).  we're just comparing
```
atz: You can move this comment block above the `def` to score some code-coverage points.

app/jobs/moab_replication_audit_job.rb
```diff
@@ -0,0 +1,42 @@
...
+class MoabReplicationAuditJob < ApplicationJob
+  queue_as :moab_replication_audit
+  delegate :check_child_zip_part_attributes, to: Audit::CatalogToArchive
+  delegate :check_aws_replicated_zipped_moab_version, to: PreservationCatalog::S3::Audit
+  delegate :logger, to: Audit::CatalogToArchive
+
+  # @param [ZippedMoabVersion] verify that the zip exists on the endpoint
```
atz: update leftover YARD to `[CompleteMoab]`

atz:
Any alternatives to MORE CONSTANTS?

jmartin-sul:
good catch @atz on the config thing!  thanks for the reminder.

i'll refrain from being defensive of the config approach, did enough of that yesterday, and this is a good concrete example of that tactic showing some strain.  funny that we ran into the `TODO` in there about updating from configs, which seemed much more hypothetical last october  😆

Issue #780, opened 2018-05-21
plan formal handoff of audit and replication monitoring


ndushay:
Once pres phase 1 code is nearly complete, there needs to be a formal decision about who will monitor auditing errors and replication errors.

We will also need to figure out which resources are needed to address these errors, and how the resources will be obtained/allocated.

Putting this in a ticket for now, but this is not the complete agenda for handoff -- just a piece of it.

jmartin-sul:
i believe we just had this meeting at 2 this afternoon?  @SaravShah and @tallenaz have some action items (document application/postgres/resque shutdown and restart procedures, add today's slack conversation about replication queries to wiki, postgres archiving approach, etc -- they can hopefully add here or ticket that stuff elsewhere as needed).

tallenaz:
- [x] Update docs on how to shutdown the catalog cleanly
- [ ] Set up postgres backups
- [ ] Put redis on its own machine
- [ ] Tune cores and RAM on worker machines

PR #1074, opened 2018-08-16
moab replication audit scheduling, etc

### LINKED ISSUE #901: (C2A) Code to kick off C2A
Is this by storage root?   If so, would we do this via a method on Endpoint class?

To initiate C2A, it could be rails runner to call this method ... which would need to be documented, but no rake task.

Also, would we want to use whenever to set up cron jobs?   On which VM?

jmartin-sul:
as discussed F2F in the handoff meeting on monday (see #780), this adds a setting that enables/disables automatic backfilling of unreplicated moab versions when the replication audit finds such things.  that flag defaults to disabling the backfill in this PR.  it can be re-enabled when we're in more of a steady state, where we have less of a backlog, or no expected backlog.

also does some other random touchups.

closes #901

### DIFF
--- a/config/deploy/prod.rb
+++ b/config/deploy/prod.rb
@@ -1,5 +1,5 @@
-server 'preservation-catalog-prod-01.stanford.edu', user: 'pres', roles: %w[m2c c2m app db web]
-server 'preservation-catalog-prod-02.stanford.edu', user: 'pres', roles: %w[cv app resque]
+server 'preservation-catalog-prod-01.stanford.edu', user: 'pres', roles: %w[app db web]
+server 'preservation-catalog-prod-02.stanford.edu', user: 'pres', roles: %w[app resque queue_populator]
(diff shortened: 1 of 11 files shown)


config/deploy/prod.rb
```diff
@@ -1,13 +1,13 @@
-server 'preservation-catalog-prod-01.stanford.edu', user: 'pres', roles: %w[m2c c2m app db web]
-server 'preservation-catalog-prod-02.stanford.edu', user: 'pres', roles: %w[cv app resque]
+server 'preservation-catalog-prod-01.stanford.edu', user: 'pres', roles: %w[app db web]
+server 'preservation-catalog-prod-02.stanford.edu', user: 'pres', roles: %w[app resque queue_populator]
```
tallenaz: Glad to see the roles consolidate.

PR #1080, opened 2018-08-21
Tweak prescat okcomputer checks

tallenaz:
Ops doesn't need to be alerted when no workers are working
  And bump up the failure queue threshold


jmartin-sul:
ah!  got it, thanks.  all good, will merge.

PR #1084, opened 2018-08-31
Quiet moab rep audit

tallenaz:


app/jobs/moab_replication_audit_job.rb
```diff
@@ -20,7 +20,7 @@ def perform(complete_moab)
       check_aws_replicated_zipped_moab_version(zmv, results)
     end
 
-    results.report_results(logger)
+    # results.report_results(logger)
```
ndushay: could you add a comment explaining why you are commenting this out?

app/services/audit_results.rb
```diff
@@ -97,12 +97,12 @@ class AuditResults
   ].freeze
 
   HONEYBADGER_REPORT_CODES = [
```
ndushay: similarly, would it be helpful to add a comment explaining why you are commenting this out (I don't have enough context to know how these codes are used)

app/jobs/moab_replication_audit_job.rb
jmartin-sul: i think i'd somewhat prefer to leave this line as-is...

app/services/audit_results.rb
```diff
@@ -97,12 +97,12 @@ class AuditResults
   ].freeze
```
jmartin-sul: and would instead suggest commenting the replication related result codes in this list, similar to what's done w/ the HB codes.

alternatively, we could just comment out `results.report_results(logger)` (as you've done above), and leave the reporting codes in this class alone entirely.  i think i have a slight preference for commenting the result codes... but mostly i'd want to just comment result codes, or just comment the reporting invocation, but not half-n-half.

tallenaz:
@ndushay are the changes you requested OK?

PR #1089, opened 2018-09-04
Revert "checksum mismatches in CV audit seem worth a honeybadger aler…

cbeer:
…t, can undo if it gets noisy"

This reverts commit d5c72482f4e5d34ab0671847d15bbf769db38ac8.

I think we can all agree it's gotten too noisy.


tallenaz:
There's a change in https://github.com/sul-dlss/preservation_catalog/pull/1084 that addresses this, pending approval of @ndushay

tallenaz:
It does not, I misread that.

jmartin-sul:
@cbeer, i believe #1084 actually would take care of everything.  the checksum error that's flooding HB is `ZIP_PART_CHECKSUM_MISMATCH`, which does indeed get commented out by tony's PR.

i searched HB for the text of the `MOAB_FILE_CHECKSUM_MISMATCH` code, and it doesn't show up (whereas `ZIP_PART_CHECKSUM_MISMATCH` does).

i'd prefer to revert this reversion, and merge #1084.

cbeer:
It doesn't matter to me; I just want to see /something/ done to stop the errors before we get throttled for the entire month (which may have already happened...).

jmartin-sul:
fair enough, me too.  i approved 1084 on friday.  i'm just sayin, i don't think the PR that actually got merged will get you what you want.

PR #1093, opened 2018-09-17
MoabReplicationAuditJob updates archive timestamp

### LINKED ISSUE #1091: MoabReplicationAuditJob#perform should update last_archive_audit
currently, this is not getting updated, and so it looks like nothing is ever getting audited.  oversight from the implementation of replication audit.

we should also add a test to ensure the audit timestamp is getting set as expected.

tallenaz:
Fixes #1091


SaravShah:
letting john review it

PR #1097, opened 2018-10-08
ZipPart#s3_key and helper #druid_version_zip fixed to generate keys with the appropriate version

### LINKED ISSUE #1088: ZipParts using bad S3 key -- MoabReplicationAudit (correctly) marking zip_parts as checksum mismatch error
Every zip_part checked so far has been flagged with zip_parts.status = 3 (checksum mismatch).

jmartin-sul:
use the `ZMV`'s version (instead of `CompleteMoab`'s) when creating `DruidVersionZip` instance, which fixes `s3_key` always ending up with the `CompleteMoab`'s latest version, regardless of that `ZipPart`'s `ZMV` parent's specific version.

also, add test cases that speak to the original oversight.

worked w/ @tallenaz, @atz, and @SaravShah on investigation and writing of new tests.

closes #1088

### DIFF
--- a/spec/models/druid_version_zip_spec.rb
+++ b/spec/models/druid_version_zip_spec.rb
@@ -15,6 +15,14 @@ describe DruidVersionZip do
     it 'returns a tree path-based key' do
       expect(dvz.s3_key).to eq 'bj/102/hs/9687/bj102hs9687.v0001.zip'
     end
+
+    context 'version is greater than 1' do
+      let(:version) { 2 }
+
+      it 'uses the right version in the path' do
+        expect(dvz.s3_key).to eq 'bj/102/hs/9687/bj102hs9687.v0002.zip'
+      end
+    end
   end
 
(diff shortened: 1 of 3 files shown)


spec/models/druid_version_zip_spec.rb
```diff
@@ -15,6 +15,14 @@
     it 'returns a tree path-based key' do
       expect(dvz.s3_key).to eq 'bj/102/hs/9687/bj102hs9687.v0001.zip'
     end
+
+    context 'version is greater than 1' do
+      let(:version) { 2 }
+
+      it 'uses the right version in the path' do
+        expect(dvz.s3_key).to eq 'bj/102/hs/9687/bj102hs9687.v0002.zip'
```
atz: Is this just bulletproofing?  Like this test would not fail previous to the above change, right?

jmartin-sul: correct

PR #1107, opened 2018-10-11
Remove PreservationPolicy.seed_from_config

atz:
- We don't want to use `settings.yml` to manage individual database  records
- `default_policy` already has the signature needed for all of our app and tests, just do the `find_or_create!` there

This is one third of the `seed_from_config` that we should continue to excise.

### DIFF
--- a/app/models/preservation_policy.rb
+++ b/app/models/preservation_policy.rb
@@ -10,21 +10,15 @@ class PreservationPolicy < ApplicationRecord
   validates :archive_ttl, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :fixity_ttl, presence: true, numericality: { only_integer: true, greater_than: 0 }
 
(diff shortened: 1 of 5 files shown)


app/models/preservation_policy.rb
```diff
@@ -10,21 +10,15 @@ class PreservationPolicy < ApplicationRecord
...
+  def self.default_policy
+    find_or_create_by!(preservation_policy_name: default_name) do |preservation_policy|
+      preservation_policy.archive_ttl = 7_776_000 # 90 days
+      preservation_policy.fixity_ttl = 7_776_000 # 90 days
     end
   end
 
-  def self.default_policy
-    find_by!(preservation_policy_name: Settings.preservation_policies.default_policy_name)
+  def self.default_name
```
atz: This is a separate method mainly to allow me to easily test behavior with a non-`default` value, i.e. the `create` case.

PR #1111, opened 2018-10-11
Runnin the deletion through the console

SaravShah:


jmartin-sul:
deleting millions of rows in a transaction via ActiveRecord migration proved problematic, because it caused too much memory usage, and took too long.  this prevented the cap deployment from completing successfully.

for ease of maintaining migration history, we're just going to tweak the migration to be a no-op, and we'll do the data cleanup by hand from the rails console on prod.

PR #1112, opened 2018-10-11
Update dependencies

jmartin-sul:
* run bundle update
* fix tests to expect the User-Agent string for the version of Faraday we're on
* replace deprecated FactoryBot usage (static attributes) with supported usage (dynamic attributes, i.e. same static attribute wrapped in a block)
  * just use the rubocop auto-correction: `rubocop --require rubocop-rspec --only FactoryBot/AttributeDefinedStatically --auto-correct`
* manual updates to comply with updated rubocop, mostly line breaks after guard clauses, a couple other line breaks, and a couple reasonable permanent exceptions

marking this WIP because it might be slightly conflict-y with some of @atz's WIP.  i'm happy to wait for that work to get merged, and then to rebase this (i think the conflicts will be easy to resolve).  or, if joe is fine merging this first, that's fine by me too (assuming no one has issues w/ the content of this PR).


atz:
Factory updates look good.

jmartin-sul:
@atz said:
> Not feeling that rubocop empty line after guard clause(s) though

yeah, i wasn't wild about that rule either.  i kept the instances where i felt like it read better at a glance, but removed a lot of those changes, and disabled the rule in general.  i can revisit the remaining line breaks if you're still totally against them, though i slightly prefer keeping the ones that are still there.

PR #1113, opened 2018-10-11
Add honey badger notifications

### LINKED ISSUE #1085: tame replication audit error reporting, re-enable (blocked by #1088)
the messages we're sending for replication audit are too long for workflow services, and too numerous for honeybadger.  we've grudgingly disabled reporting of those result codes to those services for now, but we'd very much like this information to be reported, and so would like to figure this out ASAP.

SaravShah:
closes #1085

### DIFF
--- a/app/services/audit_results.rb
+++ b/app/services/audit_results.rb
@@ -91,7 +91,6 @@ class AuditResults
     UNABLE_TO_CHECK_STATUS
     # Temporary fix for workflow-service throwing exceptions
     # because some error reports from MoabReplicationAudit are too long
-    # also see HONEYBADGER_REPORT_CODES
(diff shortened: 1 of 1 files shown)


app/services/audit_results.rb
```diff
@@ -100,15 +100,12 @@ class AuditResults
   ].freeze
```
jmartin-sul: i think you'll also want to get rid of the corresponding `also see HONEYBADGER_REPORT_CODES` in the above list (couldn't attach the comment to that line, since it doesn't show by default in the diff context).

SaravShah: cool thanks, repushed

PR #1115, opened 2018-10-15
Part audit

### LINKED ISSUE #1077: MoabReplicationAuditJob should be queued by ZipEndpoint, and should only handle ZippedMoabVersions for that endpoint
at present, `MoabReplicationAuditJob` will check `ZMV`s for all `ZipEndpoint`s for a particular `CompleteMoab`.

However, AWS credentials are provisioned via environment variables, for a particular S3 endpoint, and are endpoint specific.  Hence, as with the delivery workers, a particular pool of replication audit workers will only be able to talk to one S3 endpoint, and so will be incapable of auditing a ZMV for any given endpoint.

Two approaches I can think of off-hand to solving this:
1) Make `MoabReplicationAuditJob` endpoint aware, and have a version for each endpoint queue (similar to what's done with delivery workers, where the logic is implemented in one class, and a very thin subclass inherits that and just changes the queue name).  The methods that the job uses to do the S3 part of the audit could then be endpoint aware, could get the endpoint info from the calling job, and could filter so that ZMVs were only checked for the specified endpoint.  Then a `MoabReplicationAuditJobEast` and a `MoabReplicationAuditJobWest` (or whatever) could be queued per `CompleteMoab` per relevant `ZipEndpoint` (instead of just one for the `CompleteMoab` overall).  And then the appropriate replication audit queue would be listed in each S3 resque pool config.
2) Leave the part of `MoabReplicationAuditJob` that checks parts count consistency etc as-is.  Remove direct invocation of the method that checks the state of the `ZippedMoabVersion`s (and their `ZipPart`s) on the cloud.  Instead, introduce a new worker class for each S3 endpoint, which is just responsible for checking the `ZipPart`s of a `ZippedMoabVersion` on the cloud (like a parts aware version of the original replication check implementation).  Instead of calling the ZMV cloud audit methods directly in `MoabReplicationAuditJob`, `MoabReplicationAuditJob` would queue invocations to the appropriate new job class for each ZMV (e.g. a call to `ZippedMoabVersionAuditJobWest.perform_later` for each ZMV on west, and a call to `ZippedMoabVersionAuditJobEast.perform_later` for each ZMV on east).  Assign appropriate counts of the new worker classes to the resque pool setup with the credentials for that worker class' endpoint.

My preference at first glance is option 2.  Less redundancy in the non-S3 parts of the check.  Downside to that: a bit more queue complexity.  But that overhead is cheap, and 2 feels like the cleaner approach architecturally, IMO.

opinions from @atz?  a usage comment he left in the original implementation of the replication audit check job (which did operate on `ZippedMoabVersion`s  instead of `CompleteMoabs`) implied that this would be an issue.  but it wasn't stated explicitly, and i remained oblivious when re-working it to be parts aware and to do the comprehensive replication audit for a `CompleteMoab`.  😿

atz:
Convert `MoabReplicationAuditJob` to use `PartReplicationAuditJob`.

Fixes #1077

### DIFF
--- a/app/jobs/moab_replication_audit_job.rb
+++ b/app/jobs/moab_replication_audit_job.rb
@@ -1,43 +1,34 @@
-# Confirms that a CompleteMoab is fully/properly replicated to all target zip endpoints.
-# Usage info:
-# MoabReplicationAuditJob.perform_later(cm)
+# Confirms a CompleteMoab has all versions/parts replicated for each of its target endpoints.
+# @note Enqueues a check per endpoint
+# @example usage
+#   MoabReplicationAuditJob.perform_later(cm)
 class MoabReplicationAuditJob < ApplicationJob
   queue_as :moab_replication_audit
-  delegate :check_child_zip_part_attributes, to: Audit::CatalogToArchive
-  delegate :check_aws_replicated_zipped_moab_version, to: PreservationCatalog::S3::Audit
-  delegate :logger, to: Audit::CatalogToArchive
 
   # @param [CompleteMoab] verify that the zip exists on the endpoint
   def perform(complete_moab)
-    druid = complete_moab.preserved_object.druid
-
-    results = AuditResults.new(druid, nil, complete_moab.moab_storage_root, "MoabReplicationAuditJob")
-
-    backfill_missing_zmvs(complete_moab, results) if Settings.replication.audit_should_backfill
-
-    complete_moab.zipped_moab_versions.each do |zmv|
-      next unless check_child_zip_part_attributes(zmv, results)
-      check_aws_replicated_zipped_moab_version(zmv, results)
-    end
-
+    backfill_missing_zmvs(complete_moab)
+    ZipEndpoint
+      .includes(:zipped_moab_versions)
+      .where(zipped_moab_versions: { complete_moab: complete_moab }).each do |endpoint|
+        PartReplicationAuditJob.perform_later(complete_moab, endpoint)
+      end
     complete_moab.update(last_archive_audit: Time.current)
-
-    results.report_results(logger)
   end
 
   private
 
-  def backfill_missing_zmvs(complete_moab, results)
-    backfilled_zmvs = complete_moab.create_zipped_moab_versions!
-    return if backfilled_zmvs.empty?
-
-    results.add_result(
-      AuditResults::ZMV_BACKFILL,
-      version_endpoint_pairs: format_backfilled_zmvs(backfilled_zmvs)
+  def backfill_missing_zmvs(complete_moab)
+    return unless Settings.replication.audit_should_backfill
+    zmvs = complete_moab.create_zipped_moab_versions!
+    return if zmvs.empty?
+    Audit::CatalogToArchive.logger.warn(
+      "#{self.class}: #{complete_moab.preserved_object.druid} backfilled #{zmvs.count} ZippedMoabVersions: #{format_zmvs(zmvs)}"
     )
   end
 
-  def format_backfilled_zmvs(backfilled_zmvs)
-    backfilled_zmvs.map { |bz| "#{bz.version} to #{bz.zip_endpoint.endpoint_name}" }.sort.join("; ")
+  # @return [String] a potentially large message
+  def format_zmvs(zmvs)
+    zmvs.map { |bz| "#{bz.version} to #{bz.zip_endpoint.endpoint_name}" }.sort.join("; ")
   end
 end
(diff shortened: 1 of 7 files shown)


app/lib/audit/catalog_to_archive.rb
```diff
@@ -5,43 +5,32 @@ def self.logger
       @logger ||= Logger.new(Rails.root.join('log', 'c2a.log'))
     end
 
-    # @return [boolean] true if we have a list of child parts to check in the cloud, false otherwise
+    # @return [Boolean] true if we have a list of child parts to check in the cloud, false otherwise
```
tallenaz: George Boole smiles.

app/jobs/moab_replication_audit_job.rb
```diff
@@ -1,43 +1,34 @@
...
 
-  def backfill_missing_zmvs(complete_moab, results)
-    backfilled_zmvs = complete_moab.create_zipped_moab_versions!
-    return if backfilled_zmvs.empty?
-
-    results.add_result(
-      AuditResults::ZMV_BACKFILL,
-      version_endpoint_pairs: format_backfilled_zmvs(backfilled_zmvs)
+  def backfill_missing_zmvs(complete_moab)
+    return unless Settings.replication.audit_should_backfill
```
tallenaz: Curious about why the choice to check for backfilling here instead of at the call of the method, as before?

spec/jobs/part_replication_audit_job_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    let(:zmv1) { cm.zipped_moab_versions.first }
+    let(:zmv2) { cm.zipped_moab_versions.second }
+    let(:results) { job.send(:new_results, cm) }
+
+    it 'only checks parts for one endpoint' do
+      other_ep = create(:zip_endpoint)
+      other_zmv = cm.zipped_moab_versions.find_by(zip_endpoint: other_ep)
+      count = cm.zipped_moab_versions.where(zip_endpoint: endpoint).count
+      expect(job).to receive(:check_child_zip_part_attributes).with(ZippedMoabVersion, AuditResults).exactly(count).times
+      expect(job).not_to receive(:check_child_zip_part_attributes).with(other_zmv, AuditResults)
```
jmartin-sul: this one, i'm a little murkier on...

so `other_ep` should auto-link to the default pres policy because of how its factory is defined... but since no backfilling is happening here (right?), i don't think i'd expect there to be any `ZMV`s linked to `other_ep` anyway at this point, and so i wouldn't expect `job` to get `check_child_zip_part_attributes` called on `other_zmv` anyway.

actually, i somewhat suspect that `other_zmv` is `nil` in this case, and that if it weren't (like if the backfill had happened), it'd be a 2 element array, since `cm` is on version 2 (and so i'd also think that a `where` would be more appropriate than a `find_by` on line 34).

spec/jobs/part_replication_audit_job_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  describe '#perform' do
+    let(:zmv1) { cm.zipped_moab_versions.first }
+    let(:zmv2) { cm.zipped_moab_versions.second }
+    let(:results) { job.send(:new_results, cm) }
+
+    it 'only checks parts for one endpoint' do
+      other_ep = create(:zip_endpoint)
+      other_zmv = cm.zipped_moab_versions.find_by(zip_endpoint: other_ep)
+      count = cm.zipped_moab_versions.where(zip_endpoint: endpoint).count
+      expect(job).to receive(:check_child_zip_part_attributes).with(ZippedMoabVersion, AuditResults).exactly(count).times
```
jmartin-sul: what about two slightly more explicit expectations, one for `zmv1` and one for `zmv2`?

jmartin-sul: talked F2F, joe tried the approach i suggested, and it introduced a weird order dependency in the rspec expectations.  we agreed this is good enough as-is.

spec/jobs/part_replication_audit_job_spec.rb
jmartin-sul: discussed F2F, this works fine because:
1.`cm` isn't instantiated till it's referred to on line 34, and then the backfill happens automatically through AR hooks (because the new zip endpoint does alredy exist, due to line 33).
2. `find_by` will just return the first result if there's more than one result, and we only need one counter-example to show that the things we expect not to be checked are in fact filtered.

app/jobs/moab_replication_audit_job.rb
jmartin-sul: i think this is probably a bit more OO than my prior approach.  just call the method, and let it determine whether there's anything to be done.

or at least, that was my read on it.

atz: It leaves less responsibility in `#perform` to test.  Now the spec for that method just has to test "does `backfill_missing_zmvs` get called"?  The behavior of the other two (logical) lines is more important to the definition of the job and the `#perform` spec can focus on that.

PR #1121, opened 2018-10-16
CV retouches

atz:
- Reduction of unnecessary constants.
- Minor reduction of shame.
- Intermediate hashes removed
- YARD docs

### DIFF
--- a/app/services/checksum_validator.rb
+++ b/app/services/checksum_validator.rb
@@ -8,11 +8,8 @@ class ChecksumValidator
   delegate :moab_storage_root, to: :complete_moab
   delegate :storage_location, to: :moab_storage_root
 
-  DATA = 'data'.freeze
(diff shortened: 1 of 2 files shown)


app/services/checksum_validator.rb
```diff
@@ -143,38 +140,30 @@ def validate_signature_catalog_entry(entry)
...
   end
 
-  # shameless green implementation
+  # @return [Array<SignatureCatalogEntry>]
   def latest_signature_catalog_entries
-    @latest_signature_catalog_entries ||= begin
-      if latest_moab_version.signature_catalog
-        latest_moab_version.signature_catalog.entries
-      else
-        absent_from_moab_data = { signature_catalog_path: latest_moab_version.signature_catalog }
```
atz: Note the hash tries to use the value it just tested in the conditional as being falsey (likely nil)

Issue #1123, opened 2018-10-29
Test non-AWS endpoints


julianmorley:
Placeholder for verifying code works on S3 endpoints not owned by AWS.
See:
https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html
Specifically, the :region and :endpoint initialization methods.

By default AWS expects :region to be set, and using AWS-secret-sauce to generate an endpoint. Non-AWS S3 providers use :endpoint but not :region. So we need to test overriding :region with :endpoint and using an IBM or Oracle S3 endpoint URL.

For example in the cli, this works:
```
aws --endpoint-url=https://s3.us-south.objectstorage.softlayer.net --profile=ibm s3 ls s3://sul-sdr-ibm-us-south-1-ia
```
So I think the issue is: how can we set an endpoint (and pick the appropriate credentials) when we instantiate an S3 client object in presCat?

julianmorley:
This code fragment works using the Ruby AWS SDK v3:
```
require 'aws-sdk'
require 'aws-sdk-s3'  # v2: require 'aws-sdk'

Aws.config.update({
   credentials: Aws::Credentials.new(' ACCESS_KEY ', 'SECRET_KEY ') })

   s3 = Aws::S3::Resource.new(
     endpoint: 'https://s3.us-south.objectstorage.softlayer.net' )

   bucket = s3.bucket(' BUCKET_NAME_IN_US_SOUTH ')
   # Show only the first 50 items
   bucket.objects.limit(10).each do |item|
     puts "Name:  #{item.key}"
   end
#
```

julianmorley:
Here's where we create an S3 resource:

https://github.com/sul-dlss/preservation_catalog/blob/2b2690a33e033f2d78e1a37545e54d3c25b7ecfa/app/lib/preservation_catalog/s3.rb#L21

And we set AWS_REGION as an envar. So I think all we need to do is add in a case statement that checks for AWS_REGION, and if it's set to our (locally-defined) IBM or Oracle value, create the resource with the appropriate endpoint. e.g., if we set AWS_REGION to 'IBM-us-south', then it creates the S3::Resource with endpoint: 'https://s3.us-south.objectstorage.softlayer.net'. Otherwise, it treats it as an AWS region and doesn't set an endpoint.

julianmorley:
Here's where we set the ENVARs, I think.
[...]

PR #1130, opened 2018-11-28
Strawperson code for non-Amazon S3 support

julianmorley:
Proposed partial fix for https://github.com/sul-dlss/preservation_catalog/issues/1123
We'd also need to do something for resque rake tasks when creating new endpoints.

Even without the rake task tweak, I think this shouldn't break anything as we're not currently setting this envar.


julianmorley:
Rightly superseded by #1136.

PR #1131, opened 2018-12-10
Update dependencies

SaravShah:


jmartin-sul:
pushed a small commit that should hopefully fix the two failing tests

PR #1134, opened 2019-01-14
Add IBM cloud provider

### LINKED ISSUE #1123: Test non-AWS endpoints

tallenaz:
- Set up a DeliveryJob class specific to IBM
  - Since the SDK needs an endpoint for IBM, have PreservationCatalog::Ibm build the client that way
  - Add ZipEndpoint configuration for the new DeliveryJob
  - Let capistrano build out and manage the new IBM worker pool
  - RSpec test that the IBM client is built correctly.

  Supersedes #1130 
  Closes #1123 

  The specs are much simpler than the parallel PreservationCatalog::S3. The only difference 
  between the two classes is that PreservationCatalog::Ibm.resource takes an argument
  so I chose to just test that difference. I'm open to fleshing specs out if the team wants that.


tallenaz:
Before merging this, we will want to set up a bucket, credentials, and capistrano settings for prod.

tallenaz:
The specs are filled out, and an IBM audit class has been added. A number of existing specs have been tweaked to handle an additional zip endpoint. The travis build has also changed to allow us to run live tests against buckets from different providers. The code is ready for review, and I'm definitely open to rebasing down to fewer commits, since the current commit history is lengthy.

tallenaz:
Closed in favor of https://github.com/sul-dlss/preservation_catalog/pull/1135, which is the same code, but rebased.

PR #1135, opened 2019-02-05
Add ibm cloud endpoint

### LINKED ISSUE #1123: Test non-AWS endpoints

tallenaz:
Adds code and specs to deliver to and audit an ibm cloud zip endpoint. Many existing specs assumed one zip endpoint, and as a result, there are numerous touch ups to account for one more. The travis build now runs a matrix of two builds, each one of which runs live tests against the staging bucket and excludes tests for the other provider's bucket.

Supersedes #1130 
Closes #1123


tallenaz:
Closed in favor of a better rebase in https://github.com/sul-dlss/preservation_catalog/pull/1136

PR #1187, opened 2019-07-29
fix rubocop errors from updating rubocop-rspec

ndushay:
rubocop-rspec 1.34 (updated from 1.33) required some tweaks

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   skip_before_action :verify_authenticity_token
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 ##
 # CatalogController allows consumers to interact with the Preservation Catalog, e.g.
 # to add an existing moab object to the catalog, or to update an entry for a moab object
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -51,15 +51,15 @@ class CatalogController < ApplicationController
   end
 
   def bare_druid
-    poh_params[:druid].split(':', 2).last if poh_params[:druid]
+    poh_params[:druid]&.split(':', 2)&.last
   end
 
   def incoming_version
-    poh_params[:incoming_version].to_i if poh_params[:incoming_version]
+    poh_params[:incoming_version]&.to_i
   end
 
   def incoming_size
-    poh_params[:incoming_size].to_i if poh_params[:incoming_size]
+    poh_params[:incoming_size]&.to_i
   end
 
   def moab_storage_root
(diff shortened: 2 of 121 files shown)


.rubocop.yml
```diff
@@ -92,6 +92,10 @@ Style/BlockDelimiters:
   Exclude:
     - 'spec/*/*' # rspec expect statements look ugly with do-end instead of {}
 
+# since this isn't coming in Ruby 3, we don't want this on by default, so we don't want this rule from rubocop-rspec
+Style/FrozenStringLiteralComment:
```
jcoyne: We have this cop enabled in all our other projects, so for consistencies sake, let's do it here.  Ruby 3 is a long way off (> 2 years)

ndushay: so we want this stupid comment in all the rspec files now?

jcoyne: All you need to do is `rubocop -a --only Style/FrozenStringLiteralComment` and it will fix whatever needs to be fixed.

ndushay: to be clearer, the complaints are coming from rspec-rubocop, not from rubocop.

jcoyne: I don't understand how, that cop is defined in rubocop, not rspec-rubocop.

jcoyne: I don't see this line being needed since the default ruby version (for rubocop 0.60) is currently 2.2.  2.2 doesn't have default string literals.

ndushay: I read what was here (before I made this PR) and it seemed like a recipe for adding noise to our files:  https://www.rubydoc.info/gems/rubocop/0.41.2/RuboCop/Cop/Style/FrozenStringLiteralComment

```
This cop is designed to help upgrade to Ruby 3.0. It will add the comment `# frozen_string_literal: true` to the top of files to enable frozen string literals. Frozen string literals will be default in Ruby 3.0. The comment will be added below a shebang and encoding comment. The frozen string literal comment is only valid in Ruby 2.3+.
```

The implication is that we are choosing to add a bit of what I'd call "rubocop noise" to each of our spec files because rubocop-rspec is anticipating Ruby 3, which is a long way off?   As opposed to turning the rule off?

I see that the choice was made to add this to at least some existing projects, and I'll go with the flow.  

Just for the record, I'm not a fan of inline comments to appease linters -- we use linters to decrease cognitive load when reading code (among other things);  adding inline rule settings or other stuff to pass the linters feels like we increase our cognitive load for reading the code so it seems the wrong direction.  But I can go with the flow.

jcoyne: The inline comment does not appease the linter, it changes the behavior of the ruby interpreter.  With the comment, inline strings are frozen, as they will be by default in Ruby 3, so making this change makes a Ruby 3 upgrade easier down the road.  By having all of our projects code behave in the same way, we reduce the cognitive complexity right now.

jcoyne: I made a PR to do the rubocop upgrade https://github.com/sul-dlss/preservation_catalog/pull/1188,  so if you rebase this PR on there, I think you'll find we only need the "rspec: use describe_class" commit.

ndushay: d'oh -- I saw the gemfile change for rubocop-spec, didn't notice I was looking at a rule for plain rubocop.

I will amend the PR.

ndushay:
@jcoyne - redid this the right way

ndushay:
apparently a PR went into master before this was ready;  it is now apparently m00t.  Thanks, Obama.

PR #1204, opened 2019-09-06
checksum object endpoint

### LINKED ISSUE #1202: API endpoint for checksum report for one (or more?) druids
(part of uber ticket sul-dlss/argo#1145)

The script (https://github.com/sul-dlss/sdr-services-app/blob/master/bin/objectFileInfor.rb) produces a checksum report used by content owners to verify that their content has been successfully and correctly preserved in the SDR;  this means the content owners can delete their local copies.

That script uses the moab-versioning gem to retrieve checksums from object manifest/signature files, and it requires access to the preservation storage roots.

This ticket:

- [ ] create a versioned ReST API endpoint that takes one (or more?) druids, and returns the results of said report.   Report currently returns a .csv file with each row starting with the druid, so the report itself can accommodate multiple druids.

Judgement calls:  

- if the list of druids is really long, or if an object has a LOT of files, it might take too long for a network call ...
  - should this be implemented with a callback URL to receive the report when it is done?
  - should we only accept one druid per API call?

I have a query out to @andrewjbtw (or @tallenaz) as to how many druids folks have asked for in one shot, and if they are for objects with lots of files ... perhaps @tallenaz knows if we can get away with plain HTTP request/response without callback ...

peetucket:
fixes #1202


ndushay:
Is it easier to return csv here (per the original checksum script) or to convert json to csv in caller?

peetucket:
> Is it easier to return csv here (per the original checksum script) or to convert json to csv in caller? The orig script makes it look easy to do it here:
> 
> [sul-dlss/sdr-services-app:bin/objectFileInfor.rb@`master`#L53](https://github.com/sul-dlss/sdr-services-app/blob/master/bin/objectFileInfor.rb#L53) [](/sul-dlss/sdr-services-app/blob/HEAD@{2019-09-09T22:15:50Z}/bin/objectFileInfor.rb#L53)

I think the caller (eventually in argo) should parse the json and convert to CSV.

spec/controllers/objects_controller_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+  let(:pres_obj) { create(:preserved_object) }
+
+  describe 'GET #show' do
+    it 'response contains the object' do
+      get :show, params: { id: "druid:#{pres_obj.druid}", format: :json }
+      expect(response.body).to include(pres_obj.to_json)
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'returns a 404 response code when object not found' do
```
jcoyne: This should be in a `context 'when object not found' do ... end` block.

spec/controllers/objects_controller_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+          md5: '93802f1a639bc9215c6336ff5575ee22',
+          sha1: '32f7129a81830004f0360424525f066972865221',
+          sha256: 'a67276820853ddd839ba614133f1acd7330ece13f1082315d40219bed10009de',
+          filesize: 905_566 }
+      ]
+      expect(response).to have_http_status(:ok)
+      expect(response.body).to eq(expected_response.to_json)
+    end
+
+    it 'returns a 404 response code when object not found' do
```
jcoyne: use a context for "when object not found"

spec/controllers/objects_controller_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      get :checksum, params: { id: "druid:xx123yy9999", format: :json }
+      expect(response).to have_http_status(:not_found)
+    end
+
+    it 'returns a 500 response code when bad parameter passed in' do
```
jcoyne: use a context for "when bad parameter passed in"

spec/controllers/objects_controller_spec.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+RSpec.describe ObjectsController, type: :controller do
```
jcoyne: Controller specs are deprecated in rails. Can you make request specs instead?

peetucket: converted to request specs and added context blocks around tests

ndushay:
I don't understand why it's "easier" to create JSON here and make the caller convert it, when it would be simple to create .csv here and give the caller what is desired.  Can you explain?

peetucket:
* rails has a facility for easily creating json from an arbitrary object (e.g. a hash, an array of hashes, etc.)
* json is a fairly standard API response type so makes it more general case than this specific report

However, we could implement a CSV version of the output for the multiple druid endpoint by adding another format option.  However, it wouldn't be a CSV file download directly to the client, but rather plain CSV formatted text in the response.  This is because I believe the caller (presumably Argo) would still need to fetch the results and re-send to the client (since our clients will not be authorized to hit this endpoint directly).

peetucket:
> I would really prefer this API endpoint be versioned right away; it could be an immediate follow on PR. I think it can be done with routes: [sul-dlss/dor-services-app:config/routes.rb@`master`](https://github.com/sul-dlss/dor-services-app/blob/master/config/routes.rb) [](/sul-dlss/dor-services-app/blob/HEAD@{2019-09-10T21:31:36Z}/config/routes.rb)
> 
> Perhaps @jcoyne has more insight there.
> 
> I'm not blocking over JSON vs .csv, but I would like a little more info. Is it because it's easier to write tests for JSON? I'm trying to understand.
> 
> This is also partly motivated by what has happened with virtual-merge: we thought we wanted one thing from dor-services app, but it turned out in the end that we wanted to send dor-services-app csv.

Would be interested in other feedback before we do versioning of the API since I'm not the routes scoping in dor-services-app is necessarily the preferred way to do API versioning (see https://paweljw.github.io/2017/07/rails-5.1-api-app-part-3-api-versioning/ for example, which has controllers being segregated into specific version namespace folders).  

I was also worried about breaking existing consumers of the other API endpoints by altering the URL structure in this PR.

app/controllers/objects_controller.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require 'csv'
```
jcoyne: This line is unnecessary, it's require by Bundler.

peetucket: The tests fail without it:

```     NameError:
       uninitialized constant CSV```

File: spec/controllers/objects_controller_spec.rb [hunk 6]
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'csv'
```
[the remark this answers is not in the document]
ndushay: yeah, it probably is unnec in the spec, even if needed in the controller ...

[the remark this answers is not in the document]
peetucket: remove from spec tests, left in controller to get it to work

peetucket:
API versioning ticket: https://github.com/sul-dlss/preservation_catalog/issues/1206

peetucket:
Going to put in WIP for now until we are sure this endpoint will work with the consumer (e.g. Argo) and we don't need to switch to a POST for the multi-druid call.

app/controllers/objects_controller.rb
jcoyne: I see, CSV is part of the standard library, not a gem, thus it's not in Gemfile and not automatically required 👍

PR #1207, opened 2019-09-12
1202 api endpoint for checksums jm touchup

jmartin-sul:
a few touchups:
* some decomposition and refactoring in `ApplicationController` and `ObjectsController` to make things slightly more idiomatic to rails, and slightly more readable/reusable (hopefully).
* conversion of what were `500` errors to `400`s (since the error cases are ones where clients are providing bad input, e.g. malformed or missing druid param values).
* more uniform normalization of druid input
  * more thorough testing of said normalization

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::API
 
   rescue_from ActiveRecord::RecordNotFound, with: :not_found
   rescue_from Moab::ObjectNotFoundException, with: :not_found
-  rescue_from InvalidSuriSyntax, with: :internal_server_error
+  rescue_from InvalidSuriSyntax, with: :bad_request
 
   protected
 
@@ -13,12 +13,12 @@ class ApplicationController < ActionController::API
     id&.split(':', 2)&.last
   end
 
-  def not_found
-    render plain: '404 Not Found', status: :not_found
+  def bad_request
+    render plain: '400 bad request', status: :bad_request
   end
 
-  def internal_server_error
-    render plain: '500 Internal Server Error', status: :internal_server_error
+  def not_found
+    render plain: '404 Not Found', status: :not_found
   end
 
   def refine_invalid_druid_error!(err)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,9 +3,27 @@
 class ApplicationController < ActionController::API
   include ActionController::MimeResponds
 
+  rescue_from ActiveRecord::RecordNotFound, with: :not_found
+  rescue_from Moab::ObjectNotFoundException, with: :not_found
+  rescue_from InvalidSuriSyntax, with: :internal_server_error
+
   protected
 
   def strip_druid(id)
     id&.split(':', 2)&.last
+  end
+
+  def not_found
+    render plain: '404 Not Found', status: :not_found
+  end
+
+  def internal_server_error
+    render plain: '500 Internal Server Error', status: :internal_server_error
+  end
+
+  def refine_invalid_druid_error!(err)
+    # make a specific moab-versioning StandardError into something more easily manageable by ApplicationController...
+    raise InvalidSuriSyntax, err.message if err.message.include?('Identifier has invalid suri syntax')
+    raise # ...but just re-raise what we got if it was something else
   end
 end
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -9,14 +9,14 @@ class ObjectsController < ApplicationController
   # return the PreservedObject model for the druid (supplied with druid: prefix)
   # GET /objects/:druid
   def show
-    object = PreservedObject.find_by!(druid: strip_druid(params[:id]))
+    object = PreservedObject.find_by!(druid: druid)
     render json: object.to_json
   end
 
   # return the checksums and filesize for a single druid (supplied with druid: prefix)
   # GET /objects/:druid/checksum
   def checksum
-    render json: checksum_for_object(params[:id]).to_json
+    render json: checksum_for_object(druid).to_json
   end
 
   # return the checksums and filesize for a list of druid (supplied with druid: prefix)
@@ -40,6 +40,15 @@ class ObjectsController < ApplicationController
 
   private
 
+  def druid
+    strip_druid(params[:id])
+  end
+
+  def druids
+    return [] unless params[:druids].present?
+    params[:druids].map { |druid| strip_druid(druid) }.sort.uniq # normalize, then sort, then de-dupe
+  end
+
   def json_checksum_list
     druids.map { |druid| { "#{druid}": checksum_for_object(druid) } }.to_json
   end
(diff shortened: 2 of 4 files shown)


app/errors/invalid_suri_syntax.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+##
+# this allows us to re-raise certain StandardErrors from moab-versioning in a more
+# specifically rescue-able way
+class InvalidSuriSyntax < StandardError
+end
```
jmartin-sul: ideally, i'd have put this in the moab-versioning gem, made the raised error more specific from there, bumped the version, and used the newly cut gem.  since this is really a prescat specific thing.  but i was trying to be a bit expedient.  totally happy to do that as follow on work if time permits.

app/controllers/application_controller.rb
```diff
@@ -3,9 +3,27 @@
 class ApplicationController < ActionController::API
   include ActionController::MimeResponds
 
+  rescue_from ActiveRecord::RecordNotFound, with: :not_found
+  rescue_from Moab::ObjectNotFoundException, with: :not_found
+  rescue_from InvalidSuriSyntax, with: :bad_request
```
jmartin-sul: my impression from prior rails work is that `rescue_from` with handler methods is more idiomatic than rescuing inline in the child controller.  it certainly seems more re-usable to me.

app/controllers/application_controller.rb
```diff
@@ -3,9 +3,27 @@
 class ApplicationController < ActionController::API
   include ActionController::MimeResponds
 
+  rescue_from ActiveRecord::RecordNotFound, with: :not_found
```
jmartin-sul: i was mildly surprised that i had to do this explicitly.

app/controllers/application_controller.rb
```diff
@@ -3,9 +3,27 @@
...
+
+  def not_found
+    render plain: '404 Not Found', status: :not_found
+  end
+
+  def refine_invalid_druid_error!(err)
+    # make a specific moab-versioning StandardError into something more easily manageable by ApplicationController...
+    raise InvalidSuriSyntax, err.message if err.message.include?('Identifier has invalid suri syntax')
+    raise # ...but just re-raise what we got if it was something else
+  end
```
jmartin-sul: see comment on exception class about how i wish i had time to throw that exception directly from the MV gem, instead of hacking around in prescat.  🚢

app/controllers/objects_controller.rb
```diff
@@ -9,57 +9,69 @@ class ObjectsController < ApplicationController
   # return the PreservedObject model for the druid (supplied with druid: prefix)
   # GET /objects/:druid
   def show
-    object = PreservedObject.find_by(druid: strip_druid(params[:id]))
-    if object
-      render json: object.to_json
-    else
-      render status: 404, json: 'Object not found'
-    end
+    render json: PreservedObject.find_by!(druid: druid).to_json
```
jmartin-sul: in general, in prescat, we've tended to bias toward `find_by!` instead of `find_by` whenever we expect the object to exist, since the former will fail fast if the object is not in the DB.

app/controllers/objects_controller.rb
```diff
@@ -9,57 +9,69 @@ class ObjectsController < ApplicationController
...
-        render plain: results
+        render plain: csv_checksum_list
       end
-      format.any  { render status: 406, plain: 'Format not acceptable' }
+      format.any  { render status: :not_acceptable, plain: 'Format not acceptable' }
     end
-  rescue Moab::ObjectNotFoundException => e
-    render status: 404, json: e.message
-  rescue StandardError => e
-    render status: 500, json: e.message
```
jmartin-sul: the reduction of nesting and inline error handling makes the controller methods, esp this one, more readable, IMHO.

app/controllers/objects_controller.rb
```diff
@@ -9,57 +9,69 @@ class ObjectsController < ApplicationController
...
   private
 
+  def druid
+    strip_druid(params[:id])
+  end
+
+  def druids
+    return [] unless params[:druids].present?
+    params[:druids].map { |druid| strip_druid(druid) }.sort.uniq # normalize, then sort, then de-dupe
+  end
```
jmartin-sul: tried to centralize all the param wrangling to make the public controller methods more readable, and more consistent in their input handling.

app/controllers/objects_controller.rb
```diff
@@ -9,57 +9,69 @@ class ObjectsController < ApplicationController
...
     content_group.path_hash.map do |file, signature|
       { filename: file, md5: signature.md5, sha1: signature.sha1, sha256: signature.sha256, filesize: signature.size }
     end
   end
+
+  def retrieve_file_group(druid)
+    Moab::StorageServices.retrieve_file_group('content', druid)
+  rescue StandardError => e
+    refine_invalid_druid_error!(e)
+  end
```
jmartin-sul: again with the thing about this not being the ideal place for this handling (compared to the gem).  but expediency.  🤷‍♂

spec/controllers/objects_controller_spec.rb
```diff
@@ -5,15 +5,22 @@
   let(:prefixed_druid) { 'druid:bj102hs9687' }
   let(:prefixed_druid2) { 'druid:bz514sm9647' }
   let(:bare_druid) { 'bj102hs9687' }
+  let(:bare_druid2) { 'bz514sm9647' }
```
jmartin-sul: it was lucky (for reducing churn in this PR) that `bare_druid` already paralleled `prefixed_druid`, and that `prefixed_druid` and `prefixed_druid2` were already in alphabetical order.

spec/controllers/objects_controller_spec.rb
```diff
@@ -53,19 +78,19 @@
     end
 
     context 'when bad parameter passed in' do
-      it 'returns a 500 response code' do
+      it 'returns a 400 response code' do
```
jmartin-sul: i'd blame the calling client (not the server) for this.  unclear to me if `400` is exactly right, but `4xx` feels more right to me than `5xx` here.

spec/controllers/objects_controller_spec.rb
```diff
@@ -110,19 +135,19 @@
     end
 
     context 'when bad parameter passed in' do
-      it 'returns a 500 response code with a bad druid passed in' do
+      it 'returns a 400 response code with a bad druid passed in' do
```
jmartin-sul: client's fault for bad input, IMHO.  unsure about exact `4xx` code to use.

spec/controllers/objects_controller_spec.rb
```diff
@@ -110,19 +135,19 @@
...
+        expect(response).to have_http_status(:bad_request)
       end
 
       it 'returns a 406 response code when an unsupported response format is provided' do
         get checksums_objects_url, params: { druids: [prefixed_druid], format: :xml }
         expect(response).to have_http_status(:not_acceptable)
       end
 
-      it 'returns a 500 response code with no druid parameters passed in' do
+      it 'returns a 400 response code with no druid parameters passed in' do
```
jmartin-sul: client's fault for bad input, IMHO.  unsure about exact `4xx` code to use.

PR #1213, opened 2019-09-19
Set up deployment tasks for 3rd endpoint

tallenaz:


lib/capistrano/tasks/capistrano-resque-pool.rake
```diff
@@ -29,6 +30,14 @@ namespace :resque do
             execute :rm, pid_path
           end
         end
+        if east_pid_file_exists?
+          pid = capture(:cat, east_pid_path)
+          if test "kill -0 #{pid} > /dev/null 2>&1"
+            execute :kill, "-s QUIT #{pid}"
+          else
+            info "Process #{pid} from #{east_pid_path} is not running, cleaning up stale PID file"
```
jmartin-sul: does this part need a `execute :rm, east_pid_path`, similar to the `execute :rm, west_pid_path` below and `execute :rm, pid_path` above?

tallenaz: Good catch! Amended.

jmartin-sul: thanks for the quick fix!

PR #1215, opened 2019-09-20
add github template for PRs

ndushay:


.github/pull_request_template.md
```diff
@@ -0,0 +1,5 @@
+## Why was this change made?
+
+
+
+## Was the API documentation (openapi.json) updated?
```
jmartin-sul: could we change this to `## Was the usage documentation (e.g. wiki, README, queue or DB specific README) updated?`  there's not yet an openapi.json file for this project, but i think adding one would be a good idea, and i'd be happy to ticket that.  should be easy while things are still small.

i do really like adding this template though.

ndushay: sure - i'll make your change;  you make the ticket.

ndushay: done.

jmartin-sul: thanks, and done:  https://github.com/sul-dlss/preservation_catalog/issues/1216

Issue #1189, opened 2019-08-01
enable audit workers for IBM zip endpoint


jmartin-sul:
there were no workers consuming this queue, and so audit jobs have piled up.  now that the endpoint has been backfilled with archive copies of our moabs, we should allocate workers to deal with that queue.

jmartin-sul:
see e.g. https://github.com/sul-dlss/preservation_catalog/blob/master/config/resque-pool.yml and the endpoint specific configs in the `config` directory in shared_configs for stage and prod.  add another pool specific config for IBM.

i believe @tallenaz and atz did the bulk of the work setting this up originally for pres cat, and @tallenaz has done a lot of the work to keep it going.  @mjgiarlo also has (iirc) some recent experience configuring resque pool (and @peetucket too?).  so maybe some subset of @tallenaz, @mjgiarlo, and @peetucket can tackle this?

jmartin-sul:
i think `resque-pool-south.yml` will get an entry for `part_audit_ibm_us_south` (with some number of workers), analogous to the entry for `part_audit_aws_s3_west_2` in `resque-pool-west.yml`.

queue name is determined by:
https://github.com/sul-dlss/preservation_catalog/blob/master/app/jobs/part_replication_audit_job.rb#L9

jmartin-sul:
deployed to prod, closing.

jmartin-sul:
connection errors between job and IBM endpoint, reopening to investigate

jmartin-sul:
working on a PR.  going to take a similar approach to what we did for configuring the delivery class.  will require shared_configs changes and a schema change, but should all be pretty straight-forward, and somewhat similar to an existing pattern in the app.

jmartin-sul:
at first blush, IBM part audit appears to be working:  the queue is draining, the error queue is empty other than a couple redis connection errors, and IBM parts are steadily getting marked as ok (see prod console output below).

```ruby
[104] pry(main)> ZipPart.joins(zipped_moab_version: [:zip_endpoint]).where(zip_endpoints: { endpoint_name: 'ibm_us_south' }).group(:status).count
=> {"ok"=>5263132, "unreplicated"=>336}
...
[108] pry(main)> ZipPart.joins(zipped_moab_version: [:zip_endpoint]).where(zip_endpoints: { endpoint_name: 'ibm_us_south' }).group(:status).count
=> {"ok"=>5263137, "unreplicated"=>337}
...
[...]

PR #1219, opened 2019-09-24
Use correct part audit class (enum implementation)

jmartin-sul:
only one of this and #1220 should be merged

## Why was this change made?

to fix the issue with connecting to the IBM endpoint to audit it, per #1189 

this PR allows the audit job to select the audit class specific to the endpoint type.

i _think_ i slightly prefer the PR without the migration (see #1220), but not as much as i thought i would when i started writing it.

what's here parallels what we did for `delivery_class`.

no, but if people like this, i will update the DB documentation as needed, either as a follow on commit in this PR, or in a separate PR (would file an issue if doing the latter).


ndushay:
at a glance, slightly prefer this over #1220;   it's easier to see the diff endpoints in the code.   I expected in #1220, it's easier to see the endpoints in Settings.

Don't have strong feelings.  Either is fine.

jmartin-sul:
closed in favor of #1220

PR #1220, opened 2019-09-24
Use correct part audit class

jmartin-sul:
only one of this and #1219 should be merged

## Why was this change made?

to fix the issue with connecting to the IBM endpoint to audit it, per #1189 

this PR allows the audit job to select the audit class specific to the endpoint type.

in implementing #1219 at the end of last week, i realized this could be done without the enum and corresponding DB migration.  and indeed, i think that means we could do the same for the `delivery_class` stuff, which i was initially trying to parallel in this work; looking at the `delivery_class` stuff again, i think we could swap that `enum` for the sort of thing done in this PR, if we like this approach better.  but that's a different can of worms, which we won't open here.  (but happy to file a tech debt ticket to that effect if folks want)

FWIW, when i realized i could do things this way, i thought i'd definitely like it more, but once i was done, it wasn't as clear a win as i expected.  i somewhat prefer the validation that comes with the `enum`, compared to the validation here.  that said, not having to do a migration is nice, and this is a bit simpler.  i also don't anticipate having to query on audit class, so i don't know how useful it is to have in the DB.

also note, this needs shared_configs changes, for which there are PRs up.

i don't think there was any documentation to update for this, but happy to be told otherwise.

### DIFF
--- a/app/jobs/part_replication_audit_job.rb
+++ b/app/jobs/part_replication_audit_job.rb
@@ -15,7 +15,7 @@ class PartReplicationAuditJob < ApplicationJob
     results = new_results(complete_moab)
     complete_moab.zipped_moab_versions.where(zip_endpoint: zip_endpoint).each do |zmv|
       next unless check_child_zip_part_attributes(zmv, results)
-      PreservationCatalog::S3::Audit.check_replicated_zipped_moab_version(zmv, results)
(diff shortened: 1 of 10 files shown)


jmartin-sul:
merging in light of the above approvals.

also, big thanks to @tallenaz for spotting the root problem this PR addresses!

PR #1223, opened 2019-10-02
Add okcomputer check to count number of workers

tallenaz:
## Why was this change made?

Our monitor didn't pick up on audit workers being down earlier this week. This check detects the currently configured number of workers in stage and prod, and though it hard codes that number, it's something.

No, since this adds a check on behavior we already assume.

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -83,18 +83,12 @@ end
 # check for the right number of workers
 class WorkerCountCheck < OkComputer::Check
   def check
-    if Resque.workers.first.to_s.include?('stage')
-      if Resque.workers.count == 117
-        mark_message '117 workers are up.'
-      else
-        mark_failure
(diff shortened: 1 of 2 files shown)


config/initializers/okcomputer.rb
```diff
@@ -80,6 +80,26 @@ def check
                                 OkComputer::SizeThresholdCheck.new(queue, 20) { Resque::Failure.count(queue) }
 end
 
+# check for the right number of workers
+class WorkerCountCheck < OkComputer::Check
+  def check
+    if Resque.workers.first.to_s.include?('stage')
```
jcoyne: Could we add the expected number of workers to settings.yml?

tallenaz: https://github.com/sul-dlss/shared_configs/pull/1084 and https://github.com/sul-dlss/shared_configs/pull/1085 to add a worker count to settings.

Issue #1229, opened 2019-10-28
ObjectsController#checksums 404 error is unhelpful -- need druid supplied.  


jmartin-sul:
Currently, will 404 the whole batch if there's one missing druid.  We need at a minimum a more specific error message (including the missing druid(s));  ideally better error handling would be nice (return results for all present druids, indicate which ones are missing?)

![image](https://user-images.githubusercontent.com/96775/67701339-fb55d000-f96c-11e9-901f-b911b1db40f4.png)

we were aware of this limitation when we implemented the endpoint, but we were fitting the work in with a bunch of other stuff during one week of maintenance, so MVP 🤷‍♂

jcoyne:
I would also argue that changing the response from 404 is warranted.  404 means "nothing here", but what we want to say is probably: "409: The request could not be completed due to a conflict with the current state of the target resource"

jmartin-sul:
huh... is 409 idiomatic as a response code for "one of the things in this batch is missing?"

for the one or more missing druids (e.g. if they were requested individually), that does seem like it'd be a 404, because from pres cat's perspective, the thing doesn't exist yet.  but i'd agree that for the batch as a whole, it's not really a 404.  not sure what the expectation would be for most clients.

ndushay:
I think 404 text is coming from app/controllers/application_controller.rb;  wondering if checksum_report_job should override that so as to provide a druid for missing object.  The current error experience is sucky.  I don't care if it's a 404 or 409, I am more concerned with getting useful information from the error.

https://github.com/sul-dlss/preservation_catalog/blob/master/app/controllers/objects_controller.rb#L26 is another opportunity to say more when throwing an error - like "expect druid list" or something like that, with an example, even.

jmartin-sul:
> I think 404 text is coming from app/controllers/application_controller.rb; wondering if checksum_report_job should override that so as to provide a druid for missing object.

so the application_controller.rb handling logic was just put there so that there's be generic handling of something that seemed 404-ish.  if we want something specific for the `ObjectsController#checksums` etc methods, i'd propose just removing the error handling in `ApplicationController`, and just doing a custom response in `ObjectsController`.  the 404 genercization didn't seem to have an effect one way or the other on the existing `CatalogController` methods, which render their own error messages.

[...]

PR #1235, opened 2019-11-15
add info to some objects_controller error messages

ndushay:
## Why was this change made?

To provide better information to us (via Honeybadger) and consumers of the API when exceptions occur.  Related to #1229, in that it gives more information about why calling checksum report fails due to missing druids.

n/a

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -20,10 +20,10 @@ class ObjectsController < ApplicationController
 
   # return the checksums and filesize for a list of druid (supplied with druid: prefix)
   # note: this is deliberately allowed to be a POST to allow for a large number of druids to be passed in
-  # GET OR POST /objects/checksums?druids[]=druida&druids[]=druidb&druids[]=druidc
+  # GET OR POST /objects/checksums?druids[]=druid1&druids[]=druid2&druids[]=druid3
   def checksums
     unless druids.present?
-      render(plain: '400 bad request', status: :bad_request)
+      render(plain: "400 bad request - druids param must be populated", status: :bad_request)
       return
     end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,18 +13,22 @@ class ApplicationController < ActionController::API
     id&.split(':', 2)&.last
   end
 
-  def bad_request
-    render plain: '400 bad request', status: :bad_request
+  def bad_request(exception)
+    msg = '400 bad request'
+    msg = "#{msg}: #{exception.message}" if exception
+    render plain: msg, status: :bad_request
   end
 
-  def not_found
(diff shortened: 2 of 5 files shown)


ndushay:
@jcoyne I found this code locally when I went back to sdr work;  since it's just tweaking err messages, not convinced a test is needed -- what would you suggest?

mjgiarlo:
> @jcoyne I found this code locally when I went back to sdr work; since it's just tweaking err messages, not convinced a test is needed -- what would you suggest?

I think adding a test around here would make sense: https://github.com/sul-dlss/preservation_catalog/blob/master/spec/requests/objects_controller_checksums_spec.rb#L167-L170

app/controllers/objects_controller.rb
```diff
@@ -20,10 +20,10 @@ def checksum
 
   # return the checksums and filesize for a list of druid (supplied with druid: prefix)
   # note: this is deliberately allowed to be a POST to allow for a large number of druids to be passed in
-  # GET OR POST /objects/checksums?druids[]=druida&druids[]=druidb&druids[]=druidc
+  # GET OR POST /objects/checksums?druids[]=druid1&druids[]=druid2&druids[]=druid3
   def checksums
     unless druids.present?
-      render(plain: '400 bad request', status: :bad_request)
+      render(plain: "400 bad request - at least one of these druids is not found: #{druids}", status: :bad_request)
```
jmartin-sul: this message strikes me as slightly misleading... i think `druids.present?` is checking to see that the caller provided a druid list, not that all of the druids are present in pres cat (hence the `400` instead of the `404`, iirc).

jmartin-sul: https://guides.rubyonrails.org/active_support_core_extensions.html#blank-questionmark-and-present-questionmark

> The following values are considered to be blank in a Rails application:
    * nil and false,
    * strings composed only of whitespace (see note below),
    * empty arrays and hashes, and
    * any other object that responds to empty? and is empty.
...
The method present? is equivalent to !blank?.

ndushay: I addressed this one - thanks for catching that.

ndushay:
Okay, so:

- application_controller  error handling is used only by objects_controller.  See https://github.com/sul-dlss/preservation_catalog/issues/1229#issuecomment-547070739 (also https://github.com/sul-dlss/preservation_catalog/pull/1207/files)
- I added some tests to objects_controller to try to show that the super modest changes I made work.  

I kinda feel like i just wasted time writing tests that wouldn't have been asked for if touching the exception handling had been rolled up with other work -- it's like the magnifying glass was put on this because it was a standalone PR.

mjgiarlo:
@jmartin-sul can you re-review this? I didn't want to merge until you had a chance to comment.

PR #1236, opened 2019-11-15
Moab vers update

ndushay:
## Why was this change made?

Removes a kludge introduced in PR #1207.   This will also clean up additional objects_controller endpoints that will be leveraging moab-versioning for information ... specifically the ones about to be written to allow us to retired the sdr-services-app.

Note that the following specs continue to pass, given moab-versioning v4.3.0:
- https://github.com/sul-dlss/preservation_catalog/blob/master/spec/requests/objects_controller_checksums_spec.rb#L173-L177
- https://github.com/sul-dlss/preservation_catalog/blob/master/spec/requests/objects_controller_checksums_spec.rb#L190-L194

n/a.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -196,7 +196,7 @@ GEM
     mini_mime (1.0.2)
     mini_portile2 (2.4.0)
     minitest (5.13.0)
-    moab-versioning (4.2.2)
+    moab-versioning (4.3.0)
       confstruct
       druid-tools (>= 1.0.0)
       json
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -388,4 +388,4 @@ DEPENDENCIES
   whenever
 
 BUNDLED WITH
-   2.0.1
+   2.0.2
```
mjgiarlo: I *believe* this change means we may need to `gem install bundler -v 2.0.2` on all the prescat stage and prod instances in advance of our next round of deployments.

ndushay: we'll find out on Monday, no doubt.

PR #1237, opened 2019-11-15
use MoabStorageService for controllers

ndushay:
## Why was this change made?

Additional API endpoints are coming in order to replace the sdr-services-app.  The MoabStorageService class introduced here will contain the code used by the objects_controller to get information about existing Moabs.   

It is possible that some of the audit job functionality could also utilize this new class for its moab lookups, but right now I am concerned only with the API endpoints.

n/a

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -72,13 +72,9 @@ class ObjectsController < ApplicationController
   end
 
   def checksum_for_object(druid)
-    content_group = retrieve_file_group(druid)
+    content_group = MoabStorageService.retrieve_content_file_group(druid)
     content_group.path_hash.map do |file, signature|
       { filename: file, md5: signature.md5, sha1: signature.sha1, sha256: signature.sha256, filesize: signature.size }
     end
-  end
-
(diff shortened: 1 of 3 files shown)


app/controllers/objects_controller.rb
```diff
@@ -72,13 +72,9 @@ def csv_checksum_list
   end
 
   def checksum_for_object(druid)
```
ndushay: one could argue this method should go in MoabStorageService - I'm open to that.

app/services/moab_storage_service.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Responsibility: interactions with Moab storage to support read-only access for ReST API calls
+class MoabStorageService
+  def self.retrieve_content_file_group(druid)
+    Moab::StorageServices.retrieve_file_group('content', druid)
+  end
```
ndushay: more methods to come!

app/controllers/objects_controller.rb
mjgiarlo: meh

PR #1241, opened 2019-11-21
update prescat to use new workflow-client methods

ndushay:
## Why was this change made?

The workflow server is no longer using the repo arg;  this uses non-deprecated methods in workflow-client.

n/a


ndushay:
closing in favor of #1242

Issue #1243, opened 2019-11-22
document API endpoints of prescat


ndushay:
whether in OpenAPI or in the README, we need to document the API endpoints of prescat.

The README of preservation-client could help, but some pres-client methods add some bells and whistles to the API calls.

PR #1244, opened 2019-11-22
Add objects/file endpoint 

ndushay:
## Why was this change made?

To replace manifest, metadata and content API endpoints in sdr-services-app.

Some small refactors to the checksums methods -- a method rename and shortening spec descriptions.

I created issue #1243 to document the endpoints.  I'm happy to do that as a separate PR.


app/controllers/objects_controller.rb
```diff
@@ -12,13 +12,28 @@ def show
     render json: PreservedObject.find_by!(druid: druid).to_json
   end
 
+  # return a specific file from the Moab
+  # GET /objects/:druid/file?type=manifest&filepath=signatureCatalog.xml
+  def file
+    err_msgs = file_params_errors(params)
```
ndushay: I know this is a janky way to validate params.  I looked for documentation on how to use ActiveRecord validations in a controller and I kept finding the rails guide admonition repeated in other places about "don't do this in controllers, it gets unwieldy and too hard to maintain" so I gave up and rolled my own.

app/controllers/objects_controller.rb
```diff
@@ -12,13 +12,28 @@ def show
...
+    if err_msgs.present?
+      render(plain: "400 Bad Request: #{err_msgs.join('; ')}", status: :bad_request)
+      return
+    end
+
+    file_content = MoabStorageService.retrieve_file(druid, params[:type], params[:filepath], params[:version])
+    render plain: file_content, status: :ok
+  rescue Moab::MoabRuntimeError => e
```
ndushay: Here is another payoff from sul-dlss/moab-versioning/pull/160 - make the superclass of errors for moab-versioning.

app/controllers/objects_controller.rb
```diff
@@ -12,13 +12,28 @@ def show
...
+    render(plain: "404 Not Found: #{e}", status: :not_found)
+  end
+
   # return the checksums and filesize for a single druid (supplied with druid: prefix)
   # GET /objects/:druid/checksum
   def checksum
-    render json: checksum_for_object(druid).to_json
+    render json: content_files_checksums(druid).to_json
```
ndushay: a much improved name, IMO.

app/controllers/objects_controller.rb
```diff
@@ -49,6 +64,21 @@ def druids
     params[:druids].map { |druid| strip_druid(druid) }.sort.uniq # normalize, then sort, then de-dupe
   end
 
+  def file_params_errors(params)
```
ndushay: see comment above re this janky param validation.

app/services/moab_storage_service.rb
```diff
@@ -2,6 +2,20 @@
...
+  # @param [String] druid
+  # @param [String] category category of desired file ('content', 'metadata', or 'manifest')
+  # @param [String] filename name of the file (path relative to base directory)
+  # @param [String] version of the file (nil = latest)
+  # @return [Pathname] Pathname object containing the full path for the specified file
+  # @raise [Moab::ObjectNotFoundException] if file not found
+  def self.retrieve_file(druid, category, filename, version=nil)
+    raise(ArgumentError, "No filename provided to MoabStorageService.retrieve_file for druid #{druid}") if filename.blank?
+    err_msg = "category arg must be 'content', 'metadata', or 'manifest' (MoabStorageService.retrieve_file for druid #{druid})"
+    raise(ArgumentError, err_msg) unless ['content', 'metadata', 'manifest'].include?(category)
```
ndushay: I wrote this before I wrote the controller method.  Do you think these lines should be removed in favor of the param validation in the objects controller?   Or leave it in both places?

config/routes.rb
```diff
@@ -8,6 +8,7 @@
   resources :objects, only: %i[show] do
     member do
       get 'checksum'
+      get 'file', format: false
```
ndushay: i believe the header supplied mime type of the returned file will be irrelevant to the caller.

spec/requests/objects_controller_checksums_spec.rb
```diff
@@ -186,22 +186,19 @@
...
+    context 'when unsupported response format' do
+      it 'returns 406 response code' do
         post checksums_objects_url, params: { druids: [prefixed_druid], format: :xml }
         expect(response).to have_http_status(:not_acceptable)
       end
-
-      it 'returns a 400 response code with no druid parameters passed in' do
-        post checksums_objects_url, params: { format: :json }
-        expect(response).to have_http_status(:bad_request)
-      end
```
ndushay: this is a duplicate of "when no param"

spec/requests/objects_controller_file_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+  let(:prefixed_druid) { 'druid:bj102hs9687' }
+  let(:bare_druid) { 'bj102hs9687' }
+  let(:pres_obj) { create(:preserved_object) }
+
+  describe 'GET #file' do
+    context 'when object exists' do
+      context 'when druid is prefixed' do
+        context 'when no version param specified' do
+          context 'when file is present' do
+            context 'manifest' do
```
ndushay: there is a lot of nesting here, I know.  Suggestions welcome.

spec/routing/routing_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require 'rails_helper'
+
+RSpec.describe 'routes', type: :routing do
+  describe 'objects/:id/file' do
+    it 'file_object_path' do
+      druid = 'druid:ab123cd4567'
+      expect(get: file_object_path(id: druid)).to route_to(controller: 'objects', action: 'file', id: druid)
+    end
+  end
```
ndushay: I wrote the routing spec when I was having trouble getting the request spec to work.  I can remove it if folks don't like it, or I can complete it, perhaps as part of issue #1243 (document API endpoints)

app/services/moab_storage_service.rb
jcoyne: I prefer the validation be in the storage service.  I don't like having it duplicated though.

app/services/moab_storage_service.rb
```diff
@@ -2,6 +2,20 @@
...
+    file_pathname = Stanford::StorageServices.retrieve_file(category, filename, druid, version)
+    File.open(file_pathname).read
```
jcoyne: This would read a whole file into memory.  Is that going to be a problem?  Do you want to stream the file instead?

spec/requests/objects_controller_file_spec.rb
jcoyne: I don't think there is a problem with the nesting, but I would change to "when the type is manifest"

app/services/moab_storage_service.rb
ndushay: could stream, sure.

ndushay: hrmm - pardon my ignorance but ... how do we stream a file to the API consumer?  Reading the file in chunks I get, but would we we serve it out in chunks to the API consumer, OR how do we create a single object to serve out without it being in memory?

spec/requests/objects_controller_file_spec.rb
ndushay: done.

app/services/moab_storage_service.rb
jcoyne: I would have this method return `File.open()` which is an IO object rather than `File.open().read` which is a String.

ndushay: great - thanks.  My ignorance has shrunk an iota.

app/controllers/objects_controller.rb
```diff
@@ -12,13 +12,28 @@ def show
...
+  # GET /objects/:druid/file?type=manifest&filepath=signatureCatalog.xml
+  def file
+    err_msgs = file_params_errors(params)
+    if err_msgs.present?
+      render(plain: "400 Bad Request: #{err_msgs.join('; ')}", status: :bad_request)
+      return
+    end
+
+    file_content = MoabStorageService.retrieve_file(druid, params[:type], params[:filepath], params[:version])
+    render plain: file_content, status: :ok
```
jcoyne: This should be a streaming response: https://guides.rubyonrails.org/action_controller_overview.html#live-streaming-of-arbitrary-data

jcoyne: It might be easier if we can use X-Sendfile: https://api.rubyonrails.org/v5.2.3/classes/ActionController/DataStreaming.html#method-i-send_file

config/routes.rb
mjgiarlo: Can you say more about adding `format: false` here?

app/controllers/objects_controller.rb
mjgiarlo: :+1:

spec/requests/objects_controller_file_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+  let(:prefixed_druid) { 'druid:bj102hs9687' }
+  let(:bare_druid) { 'bj102hs9687' }
+  let(:pres_obj) { create(:preserved_object) }
+
+  describe 'GET #file' do
+    context 'when object exists' do
+      context 'when druid is prefixed' do
+        context 'when no version param specified' do
+          context 'when file is present' do
+            context 'when type is manifest' do
```
mjgiarlo: I wonder whether you'd consider flattening this a bit. Perhaps by setting more variables above to assume happy path, and then varying one unhappy path per `context` block. That might test less interaction than you have now, but the code you're testing doesn't look too complicated, so I'm wondering whether the tests need to be this nested. You've spent more time wrapping your brain around this, though, and if you think this is a simplistic suggestion... you're probably right. :smile:

app/services/moab_storage_service.rb
mjgiarlo: Agree. The storage service cares more about the values of those params ultimately.

ndushay:
@jcoyne I'd be interested in what you feel is the minimum necessary to move this forward, and what we can ticket for later.  It would be awesome if I could get this used in production by monday's deploys, since I'm no longer first responder after Wed.

jcoyne:
@ndushay did you try using `send_file`?  that's certainly the easy thing.

jcoyne:
`send_file` is used by dor-services-app, modsulator, stacks, argo and hydra_etd, so it's likely that'll work just fine.

ndushay:
@mjgiarlo @jcoyne I have reworked this per your suggestions:

- [x] controller uses send_file
- [x] controller does not duplicate validation (but it does validate version param value, because it needs to be an integer when passed to Stanford::StorageServices in moab-versioning gem, and converting it with .to_i when it's not a postive integer was just giving 0)
- [x] got rid of super nested tests (moved a lot of them to spec for MoabStorageService)

spec/requests/objects_controller_file_spec.rb
ndushay: addressed with latest changes.

config/routes.rb
ndushay: when the caller gets the file, are they truly likely to care about the mime type in the header?  I don't think we *allow* anything other than API calls to this URL, and the calling program should know how to parse what it gets, no?  Am I missing something?  I could remove this ...

ndushay: See https://guides.rubyonrails.org/routing.html#route-globbing-and-wildcard-segments, which I may be mis-understanding.

mjgiarlo: @ndushay 
OK, so the net effect of having `format: false` here is raising a 404 if the router gets a request for `http://foo.bar/objects/druid:abc123/file.anyfileext` ? I don't see any harm in allowing the format param, since the controller action doesn't do anything with it, but I also don't see any harm in being more prescriptive about this route, so :+1:

PR #1247, opened 2019-11-25
create audit workflow if it doesn't exist

### LINKED ISSUE #1245: audit jobs now need to create WF before they update WF
We're getting honeybadger errors because audit jobs can't write their messages to WF

https://app.honeybadger.io/projects/54415/faults/57642907

The problem is that workflow-service-rails changed so it no longer allows "update_workflow_status" to create the workflow if it doesn't exist (see change here https://github.com/sul-dlss/workflow-server-rails/commit/2144529d0b245b35117a7234ccfb5315570dd079 ).

The fix is to create the workflow before updating it (somewhat similar to work done https://github.com/sul-dlss/hydrus/pull/319 )

We know that using the WF to report audit errors is janky. There are longer term plans to  create a event/provenance service, which could be a new endpoint in the current workflow.  ideally, we would log messages in there and display that in argo.  Ticket for this long range plan is/was here: https://github.com/sul-dlss/workflow-server-rails/issues/270

ndushay:
## Why was this change made?

To stop the "Failed to retrieve resource" errors for non-existent preservationAuditWF, e.g. https://app.honeybadger.io/projects/54415/faults/57642907.

NOTE:  I deployed to stage and tried to test it in situ, but the objects in prescat don't correspond to the objects in dor/argo very well so I gave up.

Closes #1245

n/a

### DIFF
--- a/app/services/audit_results.rb
+++ b/app/services/audit_results.rb
@@ -154,9 +154,9 @@ class AuditResults
       log_result(r, logger)
       if r.key?(INVALID_MOAB)
         msg = "#{workflows_msg_prefix} || #{r.values.first}"
-        WorkflowReporter.report_error(druid, 'moab-valid', msg)
+        WorkflowReporter.report_error(druid, actual_version, 'moab-valid', msg)
       elsif status_changed_to_ok?(r)
-        WorkflowReporter.report_completed(druid, 'preservation-audit')
+        WorkflowReporter.report_completed(druid, actual_version, 'preservation-audit')
       elsif WORKFLOW_REPORT_CODES.include?(r.keys.first)
         candidate_workflow_results << r
       end
@@ -187,7 +187,7 @@ class AuditResults
   def report_errors_to_workflows(candidate_workflow_results)
     return if candidate_workflow_results.empty?
     msg = "#{workflows_msg_prefix} #{candidate_workflow_results.map(&:values).flatten.join(' && ')}"
-    WorkflowReporter.report_error(druid, 'preservation-audit', msg)
+    WorkflowReporter.report_error(druid, actual_version, 'preservation-audit', msg)
   end
 
   def log_result(result, logger)
--- a/app/services/workflow_reporter.rb
+++ b/app/services/workflow_reporter.rb
@@ -5,10 +5,11 @@ class WorkflowReporter
   PRESERVATIONAUDITWF = 'preservationAuditWF'
   NO_WORKFLOW_HOOKUP = 'no workflow hookup - assume you are in test or dev environment'
   COMPLETED = 'completed'
+  MISSING_WF_REGEX = /^Failed .*#{Settings.workflow_services_url}.*preservationAuditWF.* (HTTP status 404)/.freeze
 
   # this method will always return true because of the dor-workflow-service gem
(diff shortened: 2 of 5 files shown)


app/services/audit_results.rb
```diff
@@ -154,9 +154,9 @@ def report_results(logger=Rails.logger)
       log_result(r, logger)
       if r.key?(INVALID_MOAB)
         msg = "#{workflows_msg_prefix} || #{r.values.first}"
-        WorkflowReporter.report_error(druid, 'moab-valid', msg)
+        WorkflowReporter.report_error(druid, actual_version, 'moab-valid', msg)
```
ndushay: providing a version (when avail) means workflow service doesn't need to look it up)

app/services/workflow_reporter.rb
```diff
@@ -5,10 +5,11 @@ class WorkflowReporter
   PRESERVATIONAUDITWF = 'preservationAuditWF'
   NO_WORKFLOW_HOOKUP = 'no workflow hookup - assume you are in test or dev environment'
   COMPLETED = 'completed'
+  MISSING_WF_REGEX = /^Failed .*#{Settings.workflow_services_url}.*preservationAuditWF.* (HTTP status 404)/.freeze
```
ndushay: this seemed like the best way to identify the error messages pertaining to "workflow does not yet exist" concept.

app/services/workflow_reporter.rb
```diff
@@ -17,9 +18,14 @@ def self.report_error(druid, process_name, error_message)
     else
       Rails.logger.warn(NO_WORKFLOW_HOOKUP)
     end
+  rescue Dor::WorkflowException => e
+    raise unless e.message.match?(MISSING_WF_REGEX)
+
+    create_wf(druid, version)
+    report_error(druid, version, process_name, error_message)
```
ndushay: this is the significant part of the change (and below for `report_completed`)

app/services/workflow_reporter.rb
```diff
@@ -17,9 +18,14 @@ def self.report_error(druid, process_name, error_message)
     else
       Rails.logger.warn(NO_WORKFLOW_HOOKUP)
     end
+  rescue Dor::WorkflowException => e
+    raise unless e.message.match?(MISSING_WF_REGEX)
```
jcoyne: This seems a little fragile and ideally this gets a follow up PR in the workflow client.

spec/services/workflow_reporter_spec.rb
```diff
@@ -4,44 +4,76 @@
...
+                workflow: 'preservationAuditWF',
+                process: process_name,
+                error_msg: audit_result)
+      end
+    end
 
-    it 'returns true' do
-      result = 'Invalid moab, validation error...ential version directories.'
-      # because we always get true the from the dor-workflow-service gem
-      # see issue sul-dlss/dor-workflow-client#50 for more context
```
ndushay: this issue was addressed, but since it was mocked here, we noticed no change in the specs.

app/services/workflow_reporter.rb
ndushay: See issue sul-dlss/dor-workflow-client/issues/135

PR #1250, opened 2019-12-05
add objects/content_diff endpoint

ndushay:
## Why was this change made?

To replace `cm-inv-diff` endpoint in sdr-services-app.  There will be a `content_diff` method in preservation-client to be used instead of dor-services-app calling sdr-services-app.

Some small refactors added:
- renamed a private method in objects_controller to make the code clearer
- removed unnec lines from request specs
- tweaked a few method comments

Also added more routing specs.  (because I'm compulsive?)

issue #1243 exists for documenting endpoints;  it will be part of the work of the last 2 weeks of December - it's a great one for learning more about preservation.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,6 +9,7 @@ Rails.application.routes.draw do
     member do
       get 'checksum'
       get 'file', format: false # no need to add extension to url
+      post 'content_diff'
     end
     collection do
       match 'checksums', via: %i[get post]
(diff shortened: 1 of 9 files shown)


config/routes.rb
```diff
@@ -9,6 +9,7 @@
     member do
       get 'checksum'
       get 'file', format: false # no need to add extension to url
+      post 'content_diff', via: :post
```
mjgiarlo: I don't think `via: :post` is necessary since this line already uses the `post` verb. I would expect to see the `via:` keyword arg for `match`-type routes.

ndushay: whoops!  thanks. will fix.

Issue #1087, opened 2018-09-02
Resque not correctly deleting keys from Redis after job completion


julianmorley:
Redis has almost 10 million keys in it when Resque says it has 1.1million jobs. Investigation of Redis shows many old keys still present - even from deleted queues. See: https://stackoverflow.com/questions/26057790/redis-key-not-removed-even-after-resque-job-completes-successfully for a similar situation.

Why is this a problem? Because even with an LRU deletion policy, Redis will always expand to the RAM available and always be expiring old keys, when the steady-state of the db should be consuming only 10s of MB or RAM and not require constant pruning.

tallenaz:
From conversation with @atz (feel free to correct me here if I got it wrong) in standup this morning, our sense is that redis's expected behavior is for its steady state to hover around the maxmemory setting, and for the lru eviction policy to evict old data to let new data get added in all the same. If the worry is that we want the steady state of redis to be much smaller, in order to be able to handle spikes, our sense is that since it is "just" a key value store, redis should be resilient enough, given our maxmemory and eviction policy. If we can agree on this, we should adjust the nagios check to bring it in line with current maxmemory, and close the ticket.

tallenaz:
Resque deletes keys from redis via `resque-lock`. In our `ApplicationJob` class:

```
around_perform do |job, block|
    job.class.around_perform_lock(*job.arguments, &block)
  end
```

then in `resque-lock`:

```
def around_perform_lock(*args)
  begin
    yield
  ensure
    # Always clear the lock when we're done, even if there is an
    # error.
    Resque.redis.del(lock(*args))
  end
end

# Override in your job to control the lock key. It is
# passed the same arguments as `perform`, that is, your job's
# payload.
def lock(*args)
  "lock:#{name}-#{args.to_s}"
end
```

tallenaz:
After cleaning out a lot of cruft, and while watching a lot of auditing take place, dbsize is stable at well beneath 75,000 keys, and memory usage too, at around 1GB, well under maxmemory.

tallenaz:
dbsize is down to 357 keys (no resque:locks left), and memory usage has remained stable for a few days. Closing this and we'll reopen if it becomes a problem again

justinlittman:
```
127.0.0.1:6379> info keyspace
# Keyspace
db0:keys=6565371,expires=0,avg_ttl=0
```

justinlittman:
[...]

PR #1254, opened 2019-12-10
Uses global id to construct lock key.

justinlittman:
refs #1087

## Why was this change made?
To eliminate one possibility for why locks are not being deleted from Redis, viz., objects are changing between a lock being created and the lock being deleted, resulting in the lock key also changing and therefore not being found to delete.

Note that this is at least one logical possibility for the problem. We won't know for sure until this is actually run in prod.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -11,17 +11,24 @@ require 'resque/plugins/lock'
 # @see [resque-lock] https://github.com/defunkt/resque-lock/blob/master/lib/resque/plugins/lock.rb
 # @see [redis setnx] http://redis.io/commands/setnx
 class ApplicationJob < ActiveJob::Base
-  extend Resque::Plugins::Lock
+  include Resque::Plugins::Lock
 
   before_perform do |_job|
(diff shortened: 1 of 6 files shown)


app/jobs/application_job.rb
```diff
@@ -11,17 +11,24 @@
 # @see [resque-lock] https://github.com/defunkt/resque-lock/blob/master/lib/resque/plugins/lock.rb
 # @see [redis setnx] http://redis.io/commands/setnx
 class ApplicationJob < ActiveJob::Base
-  extend Resque::Plugins::Lock
+  include Resque::Plugins::Lock
```
jcoyne: why did you change to `include`? The documentation says to `extend` https://github.com/defunkt/resque-lock#resque-lock

justinlittman: Based on my (perhaps limited understanding), extends added the functions as class methods. As such, they ignored an override of the lock function (L29), which was also suggested by https://github.com/defunkt/resque-lock#resque-lock. It seems that include is the proper way to mixin the Lock module (unless it's not).

PR #1255, opened 2019-12-10
Updated resque yaml with maxmemory and lru policy

jermnelson:
## Why was this change made?
Uses Redis built-in memory management and LRU policy to delete stale keys

n/a

For #1087

### DIFF
--- a/config/resque.yml
+++ b/config/resque.yml
@@ -1,1 +1,3 @@
 development: localhost:6379
+maxmemory: 100mb
+maxmemory_policy: allkeys-lru


jermnelson:
> Are these Resque settings or Redis settings? I would expect this to be handled by our Puppet infrastructure: https://github.com/sul-dlss/puppet/blob/production/hieradata/node/preservation-catalog-prod-05.stanford.edu.eyaml#L37-L38

Wasn't sure the best place to put this; hence the draft status for this PR. Thanks for the pointer to the Puppet location. Closing this now.

PR #1257, opened 2019-12-11
Uses global id to construct lock key.

justinlittman:
refs #1087

## Why was this change made?
To eliminate one possibility for why locks are not being deleted from Redis, viz., objects are changing between a lock being created and the lock being deleted, resulting in the lock key also changing and therefore not being found to delete.

Note that this is at least one logical possibility for the problem. We won't know for sure until this is actually run in prod.

This is a replacement for #1254.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -20,8 +20,11 @@ class ApplicationJob < ActiveJob::Base
   before_enqueue do |job|
     throw(:abort) unless job.class.before_enqueue_lock(*job.arguments)
   end
-  around_perform do |job, block|
-    job.class.around_perform_lock(*job.arguments, &block)
+
+  # Overriding so that changes in ActiveModel object don't result in new lock (which they do when just calling to_s).
+  def self.lock(*args)
+    new_args = args.map { |arg| arg.class.method_defined?(:to_global_id) ? arg.to_global_id.to_s : arg }
+    "lock:#{self.class.name}-#{new_args}"
   end
 
   # Override in subclass to tune per queue.
--- a/app/jobs/druid_version_job_base.rb
+++ b/app/jobs/druid_version_job_base.rb
@@ -10,7 +10,7 @@ class DruidVersionJobBase < ApplicationJob
   end
 
   # Does queue locking on ONLY druid and version (as first and second parameters)
-  def lock(*args)
+  def self.lock(*args)
     "lock:#{name}-#{args.slice(0..1)}"
   end
 
--- a/app/jobs/zip_part_job_base.rb
+++ b/app/jobs/zip_part_job_base.rb
@@ -10,7 +10,7 @@ class ZipPartJobBase < DruidVersionJobBase
   end
 
   # Does queue locking on ONLY druid, version and part (as first 3 parameters)
-  def lock(*args)
+  def self.lock(*args)
     "lock:#{name}-#{args.slice(0..2)}"
   end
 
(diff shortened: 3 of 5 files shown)


justinlittman:
Note:
* I did not add `on_failure_clear_lock`. I wasn't confident that the referenced code (https://github.com/defunkt/resque-lock/compare/master...hoverinc:master) correctly handled the arguments and didn't have an effective way to test. Also, I think it may be redundant of exiting around_perform.
* I left before_enqueue as is. I suspect that before_enqueue_lock is actually called twice.

Both of these can be further investigated if problem not solved.

app/jobs/application_job.rb
```diff
@@ -20,8 +20,11 @@ class ApplicationJob < ActiveJob::Base
   before_enqueue do |job|
     throw(:abort) unless job.class.before_enqueue_lock(*job.arguments)
   end
-  around_perform do |job, block|
-    job.class.around_perform_lock(*job.arguments, &block)
+
+  # Overriding so that changes in ActiveModel object don't result in new lock (which they do when just calling to_s).
+  def self.lock(*args)
+    new_args = args.map { |arg| arg.class.method_defined?(:to_global_id) ? arg.to_global_id.to_s : arg }
+    "lock:#{self.class.name}-#{new_args}"
```
mjgiarlo: What is `new_args` here? An array? I'm wondering what the lock string looks like. Not that it really matters, as long as it's functioning as expected.

jmartin-sul: i believe `new_args` is the translated version of the original `args`, for use as the lock's key, subbing in global ID for the vanilla `to_s` representation, so that object mutation between queuing and completion of the job doesn't result in attempt to purge the wrong lock (by trying to purge using a key for which a lock was never created).

jmartin-sul: `lock_args` or something might've been a slightly more precise name... top level method comment is sufficient imho, but maybe we should leave a bit more detail or use a slightly more precise name for our future selves?

mjgiarlo: so `Array#to_s` is OK in this context (vs., say, `Array#join`)?

jmartin-sul: i think so... a quick/simple test example to that effect makes the `#to_s` result seem reasonable in comparison to the `#join` result:
```ruby
[10] pry(main)> test_args = ['foo', 'bar', 'baz']
=> ["foo", "bar", "baz"]
[11] pry(main)> test_args.to_s
=> "[\"foo\", \"bar\", \"baz\"]"
[12] pry(main)> test_args.join
=> "foobarbaz"
```

i dunno if a person will ever be reading this value, but `#to_s` seems more human readable, so might even be preferable? 🤷‍♂

jmartin-sul:
@justinlittman said:
> Note:
> * I did not add `on_failure_clear_lock`. I wasn't confident that the referenced code (https://github.com/defunkt/resque-lock/compare/master...hoverinc:master) correctly handled the arguments and didn't have an effective way to test. Also, I think it may be redundant of exiting around_perform.
> * I left before_enqueue as is. I suspect that before_enqueue_lock is actually called twice.
> 
> Both of these can be further investigated if problem not solved.

sounds reasonable to me!

app/jobs/application_job.rb
mjgiarlo: because we don't really care about this value other than its uniqueness, and arrays are ordered so... agree.

mjgiarlo:
go for it, @jmartin-sul. thanks for considering!

PR #1259, opened 2019-12-12
Create missing preservationAuditWF

### LINKED ISSUE #1258: Trap errors related to missing preservationAuditWF
And create the WF instead. See honeybadger alert: https://app.honeybadger.io/projects/54415/faults/55331433

This is a small tweak to [previous work](https://github.com/sul-dlss/preservation_catalog/pull/1247) before doing work on the more complete solution described in https://github.com/sul-dlss/workflow-server-rails/issues/297.

mjgiarlo:
Fixes #1258

## Why was this change made?

The implementation was never matching the exception message due to unescaped parentheses in the regex. Fix that, and change the test to make it fail without the `WorkflowReporter::MISSING_WF_REGEX` change.

Also, fix another bug where the second call to `WorkflowReporter.report_completed` was 1) not sending the `version` argument, and 2) sending an undefined `error_message` argument (copy-pasta). I suspect this bug was being hidden by the funky test setup.

no

### DIFF
--- a/app/services/workflow_reporter.rb
+++ b/app/services/workflow_reporter.rb
@@ -5,7 +5,7 @@ class WorkflowReporter
   PRESERVATIONAUDITWF = 'preservationAuditWF'
   NO_WORKFLOW_HOOKUP = 'no workflow hookup - assume you are in test or dev environment'
   COMPLETED = 'completed'
-  MISSING_WF_REGEX = /Failed .*#{Settings.workflow_services_url}.*preservationAuditWF.* (HTTP status 404)/.freeze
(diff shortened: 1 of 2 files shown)


peetucket:
So was the workflow step never being marked as completed before due to this bug?

mjgiarlo:
@peetucket :speech_balloon: 
> So was the workflow step never being marked as completed before due to this bug?

When `preservationAuditWF` was missing, yes, because the error message *never* matched.

mjgiarlo:
Assigned @jmartin-sul as a reviewer so folks know not to merge this until he's signed off. (Trying to prevent mid-review merges.)

PR #1262, opened 2019-12-12
Trap for `Errno::ENOENT`

### LINKED ISSUE #1184: audit:  audit job failures due to object not found should update pres_cat status to not 'ok' and report preservation_audit error
spawned by https://github.com/sul-dlss/preservation_catalog/issues/1177

currently, if the CatalogToMoab audit goes to examine a moab on disk, and the moab is missing, the check will error out, and we'll get a honeybadger alert because the worker blows up with an error such as:
`Errno::ENOENT: No such file or directory @ dir_initialize - /services-disk05/sdr2objects/cm/180/ts/4203/cm180ts4203`

because the exit happens in an unexpected fashion, the catalog entry for the `CompleteMoab` in question never has its `status` field updated to be something other than `ok`.  this is, of course, misleading, and makes queries by status less accurate.

instead, we should be able to handle this sort of error gracefully, we should set the status to something appropriate, and we should proactively be triggering the honeybadger alert (instead of it happening automatically because of an unhandled exception).

mjgiarlo:
Fixes #1184

## Why was this change made?

* Trap for `Errno::ENOENT` on first call to `moab_validation_errors`
* When the trap triggers, add a `MOAB_NOT_FOUND` result and update the status to `online_moab_not_found`

Also includes opportunistic bits:
* Prefer appending arrays of values to iterating and shoveling in `MoabValidationHandler`
* Use docker-mounted volume for postgres-data instead of local folder
  * Prevents permissions errors when starting up containers
  * We do this in our other projects as well. See also: sul-dlss/dor-services-app#436 
* Fix README by adding the needed namespaces.

Yes, README fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,12 +229,12 @@ In console, start by finding the storage root.
 
 ```ruby
 msr = MoabStorageRoot.find_by!(name: name)
-MoabToCatalog.seed_catalog_for_dir(msr.storage_location)
+Audit::MoabToCatalog.seed_catalog_for_dir(msr.storage_location)
 ```
 
 Or for all roots:
 ```ruby
-MoabStorageRoot.find_each { |msr| MoabToCatalog.seed_catalog_for_dir(msr.storage_location) }
+MoabStorageRoot.find_each { |msr| Audit::MoabToCatalog.seed_catalog_for_dir(msr.storage_location) }
 ```
 
 ## Development
@@ -285,4 +285,3 @@ bundle exec cap prod deploy # for the prod servers
 
 ### Resque Pool
 The Resque Pool admin interface is available at `<hostname>/resque/overview`.
-
--- a/app/services/moab_validation_handler.rb
+++ b/app/services/moab_validation_handler.rb
@@ -36,7 +36,7 @@ module MoabValidationHandler
         if moab_errors.any?
           moab_error_msgs = []
           moab_errors.each do |error_hash|
-            error_hash.each_value { |msg| moab_error_msgs << msg }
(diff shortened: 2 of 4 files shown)


app/services/moab_validation_handler.rb
```diff
@@ -36,7 +36,7 @@ def moab_validation_errors
         if moab_errors.any?
           moab_error_msgs = []
           moab_errors.each do |error_hash|
-            error_hash.each_value { |msg| moab_error_msgs << msg }
+            moab_error_msgs += error_hash.values
```
mjgiarlo: Why loop over and shovel strings when we can merge two arrays without iterating?

README.md
```diff
@@ -229,12 +229,12 @@ In console, start by finding the storage root.
 
 ```ruby
 msr = MoabStorageRoot.find_by!(name: name)
-MoabToCatalog.seed_catalog_for_dir(msr.storage_location)
+Audit::MoabToCatalog.seed_catalog_for_dir(msr.storage_location)
```
mjgiarlo: Typo.

README.md
```diff
@@ -229,12 +229,12 @@ In console, start by finding the storage root.
...
 ```
 
 Or for all roots:
 ```ruby
-MoabStorageRoot.find_each { |msr| MoabToCatalog.seed_catalog_for_dir(msr.storage_location) }
+MoabStorageRoot.find_each { |msr| Audit::MoabToCatalog.seed_catalog_for_dir(msr.storage_location) }
```
mjgiarlo: Typo

PR #1263, opened 2019-12-13
Remove an unused initialization argument from `Audit::CatalogToMoab`

mjgiarlo:
## Why was this change made?

I am not sure why this argument was there and not used when the storage location can be grabbed from the `CompleteMoab`. Possibly used to be used for dependency injection or testing? Instead, grab the storage location from the `CompleteMoab` since we no longer appear to be using this arg which is mostly getting in the way.

no

### DIFF
--- a/app/lib/audit/catalog_to_moab.rb
+++ b/app/lib/audit/catalog_to_moab.rb
@@ -4,11 +4,10 @@ module Audit
   # Catalog to Moab existence check code
   class CatalogToMoab
     include ::MoabValidationHandler
-    attr_reader :complete_moab, :storage_dir, :druid, :results
+    attr_reader :complete_moab, :druid, :results
 
-    def initialize(complete_moab, storage_dir)
+    def initialize(complete_moab)
       @complete_moab = complete_moab
-      @storage_dir = storage_dir
       @druid = complete_moab.preserved_object.druid
       @results = AuditResults.new(druid, nil, complete_moab.moab_storage_root)
     end
@@ -45,9 +44,11 @@ module Audit
       compare_version_and_take_action
     end
 
-    alias storage_location storage_dir
-
     private
+
+    def storage_location
+      complete_moab.moab_storage_root.storage_location
+    end
 
     def online_moab_found?
       return true if moab
--- a/spec/lib/audit/catalog_to_moab_spec.rb
+++ b/spec/lib/audit/catalog_to_moab_spec.rb
@@ -70,7 +70,7 @@ RSpec.describe Audit::CatalogToMoab do
       c2m.check_catalog_version
     end
 
-    it 'calls online_moab_found?(druid, storage_location)' do
+    it 'calls online_moab_found?' do
       expect(c2m).to receive(:online_moab_found?)
       c2m.check_catalog_version
     end
--- a/spec/lib/audit/catalog_to_moab_spec.rb
+++ b/spec/lib/audit/catalog_to_moab_spec.rb
@@ -4,13 +4,13 @@ require 'rails_helper'
 
 RSpec.describe Audit::CatalogToMoab do
   let(:last_checked_version_b4_date) { (Time.now.utc - 1.day).iso8601 }
-  let(:storage_dir) { 'spec/fixtures/storage_root01/sdr2objects' }
+  let(:storage_location) { 'spec/fixtures/storage_root01/sdr2objects' }
   let(:druid) { 'bj102hs9687' }
-  let(:c2m) { described_class.new(comp_moab, storage_dir) }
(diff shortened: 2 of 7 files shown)


app/lib/audit/catalog_to_moab.rb
```diff
@@ -45,10 +44,12 @@ def check_catalog_version
       compare_version_and_take_action
     end
 
-    alias storage_location storage_dir
-
     private
 
+    def storage_location
+      complete_moab.moab_storage_root.storage_location
```
ndushay: is there any advantage in memoizing this?

also - we're sure this won't cause multiple "which storage_root am i on?" queries (seems like it won't) ... cause those can be expensive.

mjgiarlo: I... don't know! What do you think, @jmartin-sul?

jmartin-sul: my recollection is that AR will essentially do some naive caching for us.  there'll be a query for the `#moab_storage_root` call, but it turns out we're already doing that in the constructor to get an `AuditResults` object:  https://github.com/sul-dlss/preservation_catalog/blob/5398420867e2c91f4dc594643dfc698500bf9fbd/app/lib/audit/catalog_to_moab.rb#L9-L13

and then i believe we'll be accessing the same instance of `MoabStorageRoot` until we call `#reload` on either `complete_moab` or `complete_moab.moab_storage_root`.

basically my recollection is that ActiveRecord doesn't re-query for accessing a field once the data has been retrieved and the object constructed.

app/lib/audit/catalog_to_moab.rb
```diff
@@ -45,10 +44,12 @@ def check_catalog_version
...
+    end
```
jmartin-sul: hmm, so actually, nothing even appears to call `storage_location` anymore.  i think we could get rid of the method entirely.  nothing in this class seems to refer to it, afaict, and grepping through the codebase, only one thing seems to instantiate this class:
https://github.com/sul-dlss/preservation_catalog/blob/5398420867e2c91f4dc594643dfc698500bf9fbd/app/jobs/catalog_to_moab_job.rb#L14-L16

and nothing there references `storage_location` .

spec/lib/audit/catalog_to_moab_spec.rb
```diff
@@ -71,7 +70,7 @@
       c2m.check_catalog_version
     end
 
-    it 'calls online_moab_found?(druid, storage_dir)' do
+    it 'calls online_moab_found?(druid, storage_location)' do
```
jmartin-sul: looking through the code, i'm sort of guessing this parenthetical is cruft from a prior method signature.  or at least, i don't really understand what this parenthetical is communicating at present.  will leave a suggestion that gets rid of it, and you can squash it in (or not) if you'd like.

jmartin-sul:
```suggestion
    it 'calls online_moab_found?' do
```

mjgiarlo:
@jmartin-sul :speech_balloon: 
> ~assuming it's green, feel free to squash into yours and merge this PR if you agree that my follow-on makes sense.~ _retracting that PR, missed where it was called due to some indirection i didn't notice, sorry!_

Yep, `MoabValidationHandler`, right? It's that whole thing where you write a module that expects certain messages but there's no good way to enforce that contract. (This is what Joe used to say "modules not being modular.")

jmartin-sul:
> @jmartin-sul 💬
> 
> > ~assuming it's green, feel free to squash into yours and merge this PR if you agree that my follow-on makes sense.~ _retracting that PR, missed where it was called due to some indirection i didn't notice, sorry!_
> 
> Yep, `MoabValidationHandler`, right? It's that whole thing where you write a module that expects certain messages but there's no good way to enforce that contract. (This is what Joe used to say "modules not being modular.")

indeed, and that module was my idea iirc, so got what i deserved 😆

app/lib/audit/catalog_to_moab.rb
mjgiarlo: @ndushay What do you think? :point_up:

spec/lib/audit/catalog_to_moab_spec.rb
mjgiarlo: Committed!

app/lib/audit/catalog_to_moab.rb
mjgiarlo: @ndushay What say you re: :point_up: ?

ndushay: all good.  Go for it. Push the green button.

PR #1264, opened 2019-12-13
Added API docs to README.

### LINKED ISSUE #1243: document API endpoints of prescat

justinlittman:
closes #1243

## Why was this change made?
Provide documentation for the Pres Cat API.

Yup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Rails application to track, audit and replicate archival artifacts associated wi
     * [Seed the catalog](#seed-the-catalog)
 * [Development](#development)
 * [Deploying](#deploying)
+* [API](#api)
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -285,3 +286,189 @@ bundle exec cap prod deploy # for the prod servers
...
+]
+```
+
+### `POST /objects/:druid/content_diff`
+Retrieves FileInventoryDifference model from comparison of passed contentMetadata.xml with latest (or specified) version in Moab for all files (default) or a specified subset.
+
+Parameters:
+* content_metadata: contentMetadata.xml to compare.
+* subset (optional; default: all; values: all|shelve|preserve|publish): subset of files to compare.
+
```
ndushay: add optional version param here?

README.md
```diff
@@ -285,3 +286,189 @@ bundle exec cap prod deploy # for the prod servers
 
 ### Resque Pool
 The Resque Pool admin interface is available at `<hostname>/resque/overview`.
+
+## API
+### `GET /objects/:druid`
+Return the PreservedObject model for the object.
+
```
jermnelson: Not blocking but I thinking in addition to the `curl` examples, would adding an example of how the API is used with the `preservation-catalog` gem in other Ruby projects would be helpful.

PR #1266, opened 2019-12-13
Add /v1-scoped routes and redirect old URLs to new ones

### LINKED ISSUE #1206: API endpoints should be versioned
We should use versioning in the API endpoints to allow for backwards compatibility when updating.

dor-services-app is one example (using scoped routes), but there may be other better ways of doing it.

mjgiarlo:
Fixes #1206

## Why was this change made?

To begin versioning routes in prescat without breaking clients that use current routes which don't contain a path segment specifying the API version.

For now, prescat is taking the same approach as dor-services-app: using a `scope` rather than a `namespace`. A `namespace` would have the following additional consequences:

* The path helper prefixes would change, to include `v1` in them, *e.g., `v1_catalog`, `checksum_v1_object`
* The controller#action routing would change, looking for controllers in a namespaced directory, *e.g.*, `app/controllers/v1/catalog_controller.rb` (and, accordingly, all the controllers would need to be added to a new `V1` module`)

Yes, API docs iN README.

### DIFF
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -321,11 +321,4 @@ RSpec.describe CatalogController, type: :controller do
       end
     end
   end
-
-  describe 'Routing' do
-    let(:druid) { "xx000xx0000" }
-
-    it { is_expected.to route(:post, '/catalog').to(controller: :catalog, action: :create) }
-    it { is_expected.to route(:patch, "/catalog/#{druid}").to(controller: :catalog, action: :update, druid: druid) }
-  end
 end
--- a/README.md
+++ b/README.md
@@ -288,11 +288,14 @@ bundle exec cap prod deploy # for the prod servers
 The Resque Pool admin interface is available at `<hostname>/resque/overview`.
 
 ## API
-### `GET /objects/:druid`
+
+### V1
+
+#### `GET /v1/objects/:druid`
 Return the PreservedObject model for the object.
 
 ```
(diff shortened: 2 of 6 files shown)


spec/controllers/catalog_controller_spec.rb
```diff
@@ -321,11 +321,4 @@
       end
     end
   end
-
-  describe 'Routing' do
```
ndushay: is there no routing spec?  these went away (and they're not that important) from here but didn't show up anywhere else in the diff.

mjgiarlo: @ndushay converted to a request spec: https://github.com/sul-dlss/preservation_catalog/pull/1266/files#diff-c4260b347cb15fc683b99f7b34e5df04 (why? I thought it was easier to test redirects in a requet spec. I could be convinced otherwise!)

mjgiarlo: @ndushay hold off on  merging, pls. want to consider adding one bit.

ndushay: again, I've approved this, so it's not a big deal to me -- I looked a the request specs, and they aren't the same sort of thing: "this route goes to this controller" ... but I'm not a stickler for routing specs except when I'm flailing with setting up routes.

mjgiarlo: @ndushay Right you are. Routing specs are testing that given path helpers or URLs route to controllers and actions, which this PR does not change. Thanks!

PR #1269, opened 2019-12-14
a couple README touchups

jmartin-sul:
follow-on from a couple stray typos in the documentation part of https://github.com/sul-dlss/preservation_catalog/pull/1266/files

## Why was this change made?

noticed a couple things that seemed like minor oversights from #1266

it was the only thing updated


jmartin-sul:
restarted test, assume it was flapping...

PR #1270, opened 2019-12-14
Upgrade rails to 6.0

jcoyne:
## Why was this change made?

Rails 5.1 is no longer a supported release

n/a


config/environments/production.rb
```diff
@@ -44,16 +34,17 @@
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :info
+  config.log_level = :debug
```
mjgiarlo: New default? Or do we want to log more?

jcoyne: This is the new default.

PR #1271, opened 2019-12-14
[HOLD] When moab not found on disk, looks for it. If found, updates catalog.

### LINKED ISSUE #1139: Handle Moabs that are moved between storage_roots
See current manual fix here:
https://github.com/sul-dlss/preservation_catalog/wiki/A-Moab-Has-Moved

A Moab directory is moved between storage roots.

What currently happens: M2C runs, finds a Moab in the 'wrong' storage root, errors out. CM record is not updated. Further processing (new versions) of that Moab stop.

What should happen: M2C runs, finds a Moab in the 'wrong' storage root, updates CM to reflect new location, runs CV on the Moab to make sure it transferred OK.

justinlittman:
closes #1139

## Why was this change made?
To allow for the movement of moabs among store roots.

No.


app/services/preserved_object_handler.rb
```diff
@@ -73,32 +74,41 @@ def check_existence
...
-        elsif incoming_version > comp_moab.version
-          set_status_as_seen_on_disk(true) unless comp_moab.status == 'ok'
-          results.add_result(AuditResults::ACTUAL_VERS_GT_DB_OBJ, db_obj_name: 'CompleteMoab', db_obj_version: comp_moab.version)
-          if moab_validation_errors.empty?
-            comp_moab.upd_audstamps_version_size(ran_moab_validation?, incoming_version, incoming_size)
-            pres_object.current_version = incoming_version
-            pres_object.save!
-          else
-            update_status('invalid_moab')
+      # Make sure that moab exists on storage
```
jcoyne: I feel like this method is getting way too big to understand what it's doing.  Can you refactor into smaller pieces?

justinlittman: Does this help?

jcoyne:
@justinlittman I agree with Mike,  while that one method is much better, I'm having a harder time understanding the mixins.  I think making it a class would be the right solution.

justinlittman:
Fortunately, I agree with both of you.

.rubocop.yml
```diff
@@ -30,12 +30,23 @@ Layout/EmptyLineAfterGuardClause:
 Layout/SpaceAroundEqualsInParameterDefault:
   Enabled: false
 
+Metrics/AbcSize:
```
jcoyne: Typically we don't add these exceptions to the `.rubocop.yml` as we probably want to fix them in the future.  Instead, revert this change and do `rubocop --auto-gen-config` that'll put the exceptions you need into `.rubocop_todo.yml`

app/services/moab_validation_handler.rb
```diff
@@ -17,6 +17,11 @@ def moab
     @moab ||= Moab::StorageObject.new(druid, object_dir)
   end
 
+  def online_moab_found?
+    return true if moab&.exist?
```
jcoyne: Can this method just be `moab&.exist?`

app/services/preserved_object_handler.rb
```diff
@@ -9,6 +9,7 @@
 # inspired by http://www.thegreatcodeadventure.com/smarter-rails-services-with-active-record-modules/
 class PreservedObjectHandler
   include ::MoabValidationHandler
+  # include ::MoabMovedHandler
```
jcoyne: cruft?

app/services/moab_moved_handler.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+##
+# Service for handling a Moab that has moved
+class MoabMovedHandler
```
mjgiarlo: Big improvement. Thank you!

.rubocop.yml
justinlittman: Fixed.

app/services/moab_validation_handler.rb
justinlittman: Fixed.

app/services/preserved_object_handler.rb
justinlittman: Removed.

spec/services/preserved_object_handler_check_exist_spec.rb
```diff
@@ -394,13 +394,45 @@
...
+      end
     end
 
     context 'object not in db' do
-      let(:exp_po_not_exist_msg) { "PreservedObject db object does not exist" }
+      let(:exp_po_not_exist_msg) { "#<ActiveRecord::RecordNotFound: Couldn't find PreservedObject> db object does not exist" }
       let(:exp_obj_created_msg) { "added object to db as it did not exist" }
 
       context 'presume validity and test other common behavior' do
+        let(:druid) { 'bp628nk4868' }
```
jmartin-sul: not sure i understand the need for this `let`... would `druid` have been `ab123cd4567`, but now you need something real from the fixtures?

app/services/moab_moved_handler.rb
```diff
@@ -0,0 +1,68 @@
...
+      AuditResults::CM_STORAGE_ROOT_CHANGED, old_storage_root: old_storage_root.storage_location,
+                                             new_storage_root: complete_moab.moab_storage_root.storage_location
+    )
+  end
+
+  def update_moab_storage_root
+    version = nil
+    transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
+      version = complete_moab.version
+    end
```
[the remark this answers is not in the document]
justinlittman: This is required by the unit test that throws a RecordNotFound exception and tests for correct handling. I personally feel that test is a bit excessive, but this is to provide correct handling.

Note that this is all further complicated by the fact that you never know when you're making the first call to complete_moab and that RecordNotFound error might be triggered.

app/services/moab_moved_handler.rb
```diff
@@ -0,0 +1,68 @@
...
+    results.remove_db_updated_results unless transaction_ok
```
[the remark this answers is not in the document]
justinlittman: Removed.

app/services/moab_validation_handler.rb
```diff
@@ -90,4 +94,16 @@ def set_status_as_seen_on_disk(found_expected_version)
       update_status('validity_unknown')
     end
   end
+
+  def handle_missing_moab
+    transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
+      results.add_result(AuditResults::MOAB_NOT_FOUND,
+                         db_created_at: complete_moab.created_at.iso8601,
+                         db_updated_at: complete_moab.updated_at.iso8601)
```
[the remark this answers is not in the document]
justinlittman: The reason that I moved it into the transaction is because (1) if we're going to expect handling of RecordNotFound and (2) the first call to complete_moab is indeterminate, then all uses of complete_moab must be wrapped in a transaction.

Gotta say, I'm not thrilled with this pattern.

spec/services/preserved_object_handler_check_exist_spec.rb
justinlittman: Correct, I need something that matches the fixtures.

app/services/moab_moved_handler.rb
```diff
@@ -0,0 +1,68 @@
...
+    return if moved_moab.current_version_id != version
+    return if validation_errors?
+
+    # Update the disk location in the catalog
+    transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
+      update_storage_root
+      update_status_ok
+      complete_moab.save!
+    end
+    results.remove_db_updated_results unless transaction_ok
```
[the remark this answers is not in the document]
justinlittman: Worry seems reasonable to me. Setting status to `unknown` and queueing checksum validation.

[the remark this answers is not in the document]
justinlittman: Just discovered that setting status to `unknown` triggers checksum validation. You know, because "magic".

justinlittman:
The more I think about this, the more I realize that this is just one of the cases that needs to be handled to support moabs on multiple storage roots.

What I think we want to do is to collapse CatalogToMoab and MoabToCatalog / PreservedObjectHandler into a single audit. (Otherwise, the handling for fixing problems needs to be put in too many places.)

The single audit would, for a preserved object:
* For each moab (found on storage across all storage roots):
  * Make sure that there is a complete_moab. If there is not a complete_moab, create a new one with validity unknown.
* For each complete_moab:
  * Make sure that the complete_moab version matches the preserved object version.
  * Make sure that the moab exists. If the moab does not exist and there is at least one other complete_moab, then delete.
  * Make sure the moab version matches the complete_moab version.

ndushay:
How would this work?  The checks aren't doing the same thing, and they have diff starting points to get the object to be checked.  In M2C you look at the dirs of a storage root to get the objects to be checked;  in the other case you look  in the db for preserved_objects to be checked.  In the M2C check, the moab's validity is thoroughly checked with a class in the moab-versioning gem.

This may benefit from a RT discussion.

app/services/moab_moved_handler.rb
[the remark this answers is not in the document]
julianmorley: We should assume that we *do* perform a checksum validation as part of moving/copying a Moab to a new storage root, the only question is where and how do we perform it, and what would the workflow look like?

It seems that here, the proposal would be: 
- copy the Moab via filesystem utilities to the new storage root
- run the new, improved check-all-the-things audit against the moab in the new storage location
- and then this process kicks off CV.
Does that sound viable? I'm mindful that we only need to CV the Moab once on a boundary transition, and some of these objects are big and painful to do multiple times.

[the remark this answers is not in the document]
jmartin-sul: > Just discovered that setting status to `unknown` triggers checksum validation. You know, because "magic".

haha, forgot we wired up that bit of magic  ✨

justinlittman:
You would still trigger it both ways; however, in each case it runs the same checks.

If I'm reading the code correctly, both M2C and C2M invoke a Stanford::StorageObjectValidator.

Agree about warranting further discussion.

app/services/moab_moved_handler.rb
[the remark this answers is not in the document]
jmartin-sul: ah ok, my read had been that `complete_moab` was already retrieved because it seemed to be passed in via the constructor, but totally possible i have that wrong.

app/services/moab_validation_handler.rb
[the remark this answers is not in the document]
jmartin-sul: ah!  yeah, i missed that.  makes sense.

sorry again for introducing this module, i am obviously also finding it unintuitive to reason about when coming back to it.

mjgiarlo:
@justinlittman still worth keeping this guy around or?

PR #1275, opened 2019-12-16
MoabToCatalog#perform remove path argument

### LINKED ISSUE #1158: Fix MoabToCatalogJob to derive Moab_path from info in moab_storage_root
Currently:
`MoabToCatalogJob.perform_now(root, 'ft557dq2275', '/services-disk16/sdr2objects/ft/557/dq/2275/ft557dq2275')`

Propose making that 3rd argument optional; if it's nil, MoabToCatalogJob should attempt to work it out based on the info available from the MoabStorageRoot object passed in as the first argument, plus the string value for the druid passed as the second argument.

When you're manually remediating druids it's a PITA to have to construct that string every darn time. 

Something like moab_storage_root.storage_location + {druid/string/chopped/up} + druid

jermnelson:
## Why was this change made?
Request by Julian for `MoabToCatalogJob#preform`'s arguments when manually remediating druids.

n/a

Fixes #1158

### DIFF
--- a/app/jobs/moab_to_catalog_job.rb
+++ b/app/jobs/moab_to_catalog_job.rb
@@ -11,9 +11,8 @@ class MoabToCatalogJob < ApplicationJob
 
   # @param [MoabStorageRoot] root mount containing the Moab
   # @param [String] druid
-  # @param [String] optional path
-  def perform(root, druid, path=nil)
-    path = "#{root.storage_location}/#{DruidTools::Druid.new(druid).tree.join('/')}" if path.nil?
+  def perform(root, druid)
+    path = "#{root.storage_location}/#{DruidTools::Druid.new(druid).tree.join('/')}"
     moab = Moab::StorageObject.new(druid, path)
     PreservedObjectHandler.new(druid, moab.current_version_id, moab.size, root).check_existence
   end
--- a/app/jobs/moab_to_catalog_job.rb
+++ b/app/jobs/moab_to_catalog_job.rb
@@ -11,8 +11,9 @@ class MoabToCatalogJob < ApplicationJob
 
   # @param [MoabStorageRoot] root mount containing the Moab
   # @param [String] druid
-  # @param [String] path
-  def perform(root, druid, path)
+  # @param [String] optional path
+  def perform(root, druid, path=nil)
+    path = "#{root.storage_location}/#{DruidTools::Druid.new(druid).tree.join('/')}" if path.nil?
     moab = Moab::StorageObject.new(druid, path)
     PreservedObjectHandler.new(druid, moab.current_version_id, moab.size, root).check_existence
   end
--- a/app/models/moab_storage_root.rb
+++ b/app/models/moab_storage_root.rb
@@ -25,8 +25,8 @@ class MoabStorageRoot < ApplicationRecord
 
   # Use a queue to ensure each druid on this root's directory is in the catalog database
   def m2c_check!
-    Stanford::MoabStorageDirectory.find_moab_paths(storage_location) do |druid, path, _match|
-      MoabToCatalogJob.perform_later(self, druid, path)
+    Stanford::MoabStorageDirectory.find_moab_paths(storage_location) do |druid, _path, _match|
+      MoabToCatalogJob.perform_later(self, druid)
     end
   end
 
--- a/spec/jobs/moab_to_catalog_job_spec.rb
+++ b/spec/jobs/moab_to_catalog_job_spec.rb
@@ -20,4 +20,18 @@ describe MoabToCatalogJob, type: :job do
       job.perform(msr, druid, path)
     end
   end
+
+  describe '#perform without path' do
+    let(:msr) { create :moab_storage_root }
+    let(:path) { "#{msr.storage_location}/bj/102/hs/9687/bj102hs9687" }
+    let(:handler) { instance_double(PreservedObjectHandler) }
+
+    it 'builds a PreservedObjectHandler from MoabStorageRoot and druid if path is missing' do
+      expect(Moab::StorageObject).to receive(:new).with(druid, path).and_return(moab)
+      expect(PreservedObjectHandler).to receive(:new)
+        .with(druid, moab.current_version_id, moab.size, msr).and_return(handler)
+      expect(handler).to receive(:check_existence)
+      job.perform(msr, druid)
+    end
+  end
 end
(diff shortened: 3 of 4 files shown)


app/jobs/moab_to_catalog_job.rb
```diff
@@ -11,8 +11,9 @@ class MoabToCatalogJob < ApplicationJob
 
   # @param [MoabStorageRoot] root mount containing the Moab
   # @param [String] druid
-  # @param [String] path
-  def perform(root, druid, path)
+  # @param [String] optional path
+  def perform(root, druid, path=nil)
+    path = "#{root.storage_location}/#{DruidTools::Druid.new(druid).tree.join('/')}" if path.nil?
```
ndushay: question for @jmartin-sul:  do we want to warn before we do this lookup?

ndushay:
the param saves what can be an expensive call to figure out which storage root something lives on.  If using moab-versioning gem, that can go through a path pattern match on each root.

Actually, if it is NOT provided, it should do a db call to find the path, not use moab-versioning -- prescat lookup should be more efficient, overall.

Plus, we may have to stop relying on moab-versioning for determining storage location when we move prescat to shiny new VMs

app/jobs/moab_to_catalog_job.rb
jmartin-sul: i don't think that's necessary... to my knowledge, there's no appreciable difference in overhead (we've already got the storage root object, and AFAIK it should have all its fields populated by the query that instantiated it).  and even if we wanted to warn because the lack of the param implies manual remediation (per #1158), i'd argue for a flag specifically indicating that, since i'd agree w/ @mjgiarlo that the param is probably superfluous.

jcoyne:
I think that having a simpler code base (especially fewer `if` statements) makes the code easier to read and maintain.  Please remove it unless we use it at present.

jmartin-sul:
@mjgiarlo said:
> Since there's (apparently?) only one spot in the code where this is called, I wonder whether we want to keep this optional param around. Is it useful (e.g., for testing)?
> 
> Related: should we stop using the param [here](https://github.com/sul-dlss/preservation_catalog/blob/master/app/models/moab_storage_root.rb#L29)?

yeah, i would be in favor of dropping it.  we get it for free since we're already passing in the `MoabStorageRoot` AR object.  and as @jcoyne points out, the conditional adds cognitive overhead, which i hadn't really considered.  plus, having the same piece of information represented in both `MoabStorageRoot#storage_location`  and `path` gives us one more opportunity to introduce bugs by providing divergent values if we did call this elsewhere.

@jermnelson, sorry for the back and forth.  i'd almost suggested seeing whether you could get rid of the param, but figured this was progress when we got the tests to pass (and also i was sort of in a rush to catch up on other stuff after a couple hours of meetings 😬).  i think it'd be pretty easy to get rid of this param -- if you did, i think you'd just drop the param, and get rid of the conditional assignment [here](https://github.com/sul-dlss/preservation_catalog/pull/1275/files#diff-12756a6f6be3a7b86b9b71928e176cfbR16).  and while you wouldn't need the entire new test you added, i think you would still want the [`let(:path)`](let(:path)) part of your test -- if you swapped that in for what's on [line 9 of the old test](https://github.com/sul-dlss/preservation_catalog/pull/1275/files#diff-ede720cb69a000c0a91a8b53d6e2ef46R9), and altered [line 20](https://github.com/sul-dlss/preservation_catalog/pull/1275/files#diff-ede720cb69a000c0a91a8b53d6e2ef46R20) to omit the removed param, i think that'd likely be it?

jermnelson:
> @mjgiarlo said:
> 
> > Since there's (apparently?) only one spot in the code where this is called, I wonder whether we want to keep this optional param around. Is it useful (e.g., for testing)?
> > Related: should we stop using the param [here](https://github.com/sul-dlss/preservation_catalog/blob/master/app/models/moab_storage_root.rb#L29)?
> 
> yeah, i would be in favor of dropping it. we get it for free since we're already passing in the `MoabStorageRoot` AR object. and as @jcoyne points out, the conditional adds cognitive overhead, which i hadn't really considered. plus, having the same piece of information represented in both `MoabStorageRoot#storage_location` and `path` gives us one more opportunity to introduce bugs by providing divergent values if we did call this elsewhere.
> 
> @jermnelson, sorry for the back and forth. i'd almost suggested seeing whether you could get rid of the param, but figured this was progress when we got the tests to pass (and also i was sort of in a rush to catch up on other stuff after a couple hours of meetings 😬). i think it'd be pretty easy to get rid of this param -- if you did, i think you'd just drop the param, and get rid of the conditional part of the assignment [here](https://github.com/sul-dlss/preservation_catalog/pull/1275/files#diff-12756a6f6be3a7b86b9b71928e176cfbR16). and while you wouldn't need the entire new test you added, i think you would still want the `let(:path)` part of your test -- if you swapped that in for what's on [line 9 of the old test](https://github.com/sul-dlss/preservation_catalog/pull/1275/files#diff-ede720cb69a000c0a91a8b53d6e2ef46R9), and altered [line 20](https://github.com/sul-dlss/preservation_catalog/pull/1275/files#diff-ede720cb69a000c0a91a8b53d6e2ef46R20) to omit the removed param, i think that'd likely be it?
Refactored with @jmartin-sul and @mjgiarlo PR suggestions.

app/models/moab_storage_root.rb
```diff
@@ -25,8 +25,8 @@ def c2m_check!(last_checked_b4_date = Time.current)
 
   # Use a queue to ensure each druid on this root's directory is in the catalog database
   def m2c_check!
-    Stanford::MoabStorageDirectory.find_moab_paths(storage_location) do |druid, path, _match|
-      MoabToCatalogJob.perform_later(self, druid, path)
+    Stanford::MoabStorageDirectory.find_moab_paths(storage_location) do |druid, _match|
```
mjgiarlo: Won't changing the block args to an arity of two assign the former value of `path` to `_match`? On line 28, I'd go with one of two changes:

1. If the block wants arity 3, re-add the `path` block arg and rename it `_path` (the leading underscore is a convention for unused variables);
2. If the block is OK with arity < 3, go with a single block arg for `druid`.

Looking good otherwise. Thanks for considering the changes!

jermnelson: Just pushed update following option 1 that is consistent with usage at https://github.com/sul-dlss/preservation_catalog/blob/ff4cc9fc0f827b3966ecc7d82ee8f604ec868c14/spec/factories/preserved_object.rb#L18

jmartin-sul:
@jermnelson said:
> Refactored with @jmartin-sul and @mjgiarlo PR suggestions.

thanks!

PR #1292, opened 2019-12-19
Dockerize preservation_catalog to improve local development and testing

aaron-collier:
## Why was this change made?

Remaining questions:

- [x] ~Do we want to continue with the `development` environment instead of `production`? I'm not sure it makes a lot of difference, but probably introduces some configuration concerns.~
- [x] How to manage the resque config in `resque.yml`, which currently points to localhost, but should point to `redis` set in the `docker-compose.yml` setup.
- [x] Should localstack be added to the docker-compose stack in order to facilitate local testing of AWS resources? Per discussion: while it would be nice to have this, it is most appropriate to try to add it in a follow up PR.

The README has been updated to include the commands to start docker through docker-compose and seed the database.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,10 +4,18 @@ FROM ruby:2.5.3-alpine
 RUN apk --no-cache add \
   postgresql-dev \
   postgresql-client \
-  tzdata
+  tzdata \
+  libxml2-dev \
+  libxslt-dev
+
+LABEL maintainer="Aaron Collier <aaron.collier@stanford.edu>"
 
 RUN mkdir /app
 WORKDIR /app
+
+RUN gem update --system && \
+  gem install bundler && \
+  bundle config build.nokogiri --use-system-libraries
 
 COPY Gemfile Gemfile.lock ./
 
@@ -17,8 +25,6 @@ RUN apk --no-cache add --virtual build-dependencies \
   && apk del build-dependencies
 
 COPY . .
-
-LABEL maintainer="Aaron Collier <aaron.collier@stanford.edu>"
 
 ENV RAILS_ENV=production
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM ruby:2.5.3-alpine
+
+# postgresql-client is required for invoke.sh
+RUN apk --no-cache add \
+  postgresql-dev \
+  postgresql-client \
+  tzdata
+
+RUN mkdir /app
+WORKDIR /app
+
+COPY Gemfile Gemfile.lock ./
+
+RUN apk --no-cache add --virtual build-dependencies \
+  build-base \
+  && bundle install --without development test\
+  && apk del build-dependencies
+
+COPY . .
+
+LABEL maintainer="Aaron Collier <aaron.collier@stanford.edu>"
+
+ENV RAILS_ENV=production
+
+CMD ["./docker/invoke.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ COPY Gemfile Gemfile.lock ./
 
 RUN apk --no-cache add --virtual build-dependencies \
   build-base \
-  && bundle install --without development test\
+  && bundle install --without test\
   && apk del build-dependencies
 
 COPY . .
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,10 @@ LABEL maintainer="Aaron Collier <aaron.collier@stanford.edu>"
 RUN mkdir /app
 WORKDIR /app
 
+# Using argument for conditional setup in conf file  
+ARG RAILS_ENV
+ENV RAILS_ENV $RAILS_ENV
+
 RUN gem update --system && \
   gem install bundler && \
   bundle config build.nokogiri --use-system-libraries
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -24,7 +25,8 @@ Rails application to track, audit and replicate archival artifacts associated wi
 
 Use the docker-compose to start the dependencies (PostgreSQL and Redis)
 ```sh
-docker-compose up
+docker-compose up -d postgres
```
mjgiarlo: Can also do this in one line, I think: `docker-compose up -d postgres redis`

README.md
```diff
@@ -270,6 +272,47 @@ To run the tests:
...
+
+A Dockerfile is provided in order to interact with the application in development.
+
+Build the docker image:
+
+```sh
+docker-compose build app
+```
+
+Bring up the docker container and it's dependencies:
```
mjgiarlo: Typo: `s/it's/its/`

README.md
```diff
@@ -270,6 +272,47 @@ To run the tests:
...
+
+```sh
+docker-compose up -d
+```
+
+Initialize the database:
+
+```sh
+docker-compose run app bundle exec rails db:reset
```
mjgiarlo: This can be taken down to one line too, since you can pass multiple tasks to `rake`/`rails`, which will be executed in order: `docker-compose run app bundle exec rails db:reset db:seed`

README.md
```diff
@@ -270,6 +272,47 @@ To run the tests:
...
+{
+  "id":1,
+  "druid":"bj102hs9688",
+  "current_version":3,
+  "created_at":"2019-12-20T15:04:56.854Z",
+  "updated_at":"2019-12-20T15:04:56.854Z",
+  "preservation_policy_id":1
+}
+```
+
```
mjgiarlo: Should we add a sub-section about pushing the image to DockerHub?

docker-compose.yml
```diff
@@ -1,6 +1,24 @@
...
+    build:
+      context: ./
+      dockerfile: Dockerfile
+    environment:
+      - RAILS_LOG_TO_STDOUT=true
+      - POSTGRES_HOST=db
+      - SETTINGS__REDIS__HOSTNAME=redis
+      - RAILS_ENV=development
+    volumes:
+      - .:/app
```
mjgiarlo: I'd like @jcoyne or @justinlittman to comment on the `volumes:` section, if that's OK.

docker/resque.yml
```diff
@@ -0,0 +1 @@
+development: redis:6379
```
mjgiarlo: I believe you can use ERB in this file, so how about removing this file and changing `config/resque.yml` to look like:

```yaml
development: <%= ENV.fetch('REDIS_URL', 'localhost:6379') %>
```

And instead of using `volumes` for this override, set `REDIS_URL` in `docker-compose.yml` to be `redis:6379`?

docker-compose.yml
mjgiarlo: For instance, can this section handle mounting `tmp/pids` so we don't need the `.gitignore` changes above? Or do we want to share the state in this dir between local dev and the docker instance?

README.md
aaron-collier: Possibly, but I haven't yet as this is mostly for local dev, but I'm more than happy to add that.

README.md
aaron-collier: wouldn't be mine without a typo.

docker-compose.yml
aaron-collier: Good question. I can test that. When I ran into errors related to the pids, this was the suggested route, so I didn't try mounting.

README.md
jmartin-sul: i'm also OK w/ deferring that till later, but i'm also totally fine to see it added here, whatever's easiest.

docker-compose.yml
```diff
@@ -1,6 +1,24 @@
...
 
 services:
+  app:
+    build:
+      context: ./
+      dockerfile: Dockerfile
+    environment:
+      - RAILS_LOG_TO_STDOUT=true
+      - POSTGRES_HOST=db
+      - SETTINGS__REDIS__HOSTNAME=redis
```
mjgiarlo: Is this variable used in prescat?

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+  libxml2-dev \
+  libxslt-dev
+
+LABEL maintainer="Aaron Collier <aaron.collier@stanford.edu>"
+
+RUN mkdir /app
+WORKDIR /app
+
+# Using argument for conditional setup in conf file  
+ARG RAILS_ENV
```
mjgiarlo: I'm fuzzy on this, but RAILS_ENV is an ARG here but it's not listed as an arg in docker-compose.yml. Oversight or unnecessary?

README.md
julianmorley: Yeah, this has potential for prod use, minus the postgres db stuff.

Dockerfile
mjgiarlo: Unnecessary, I think. This allows to bake in RAILS_ENV later when we start pushing built images (using --build-arg). Carry on!

docker-compose.yml
aaron-collier: No, I've removed it, it was cruft.

Dockerfile
aaron-collier: Yeah, this allows it to be passed/set in docker-compose or the command line

docker/resque.yml
mjgiarlo: Crufty file now?

PR #1294, opened 2019-12-20
[Draft] Adds preserved object audit service.

### LINKED ISSUE #1159: [BLOCKED] Support same druid on multiple storage_roots
We'll need this when we start migrating content off the NetApp / onto CHAOS; the same exploded Moab will exist on two NFS filesystems simultaneously, potentially with different current_versions. Possibly related to #1139

justinlittman:
closes #1159

## Why was this change made?
* Support multiple moabs on different storage roots.
* Consolidate M2C and C2M

Not yet.

### DIFF
--- a/app/services/preserved_object_audit_service.rb
+++ b/app/services/preserved_object_audit_service.rb
@@ -1,0 +1,218 @@
+# frozen_string_literal: true
+
+# A service for auditing complete_moabs against moabs on disk (and vice versa.)
+# If a preserved_object is missing, it is created.
+# If a complete_moab is missing for a moab, it is created.
+# For each complete_moab / moab, versions are compared and validations possibly performed.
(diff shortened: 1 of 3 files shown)


app/services/preserved_object_audit_service.rb
```diff
@@ -0,0 +1,218 @@
...
+      next unless validate(moab, complete_moab) # Also adds to results and updates complete_moab
+
+      if moab.current_version_id < complete_moab.version
+        handle_unexpected_version(complete_moab)
+        next
+      end
+
+      update_status(complete_moab,'validity_unknown') # This will queue a checksum validation.
+
+      handle_bumped_version(complete_moab, moab) if moab.current_version_id > complete_moab.version
```
julianmorley: Can handle_bumped_version result in a validity_unknown status? If so, do this before update_status, so we can trigger a CV on the new version.

PR #1295, opened 2019-12-20
Remove permit params

jcoyne:
## Why was this change made?
This is only useful if you are bulk updating attributes with parameters, which we are not

n/a

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -47,31 +47,26 @@ class CatalogController < ApplicationController
 
   private
 
-  # strong params / whitelist params
-  def poh_params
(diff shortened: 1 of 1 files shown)


ndushay:
I'm fine with this but am wondering if the motivation is that this is a backend service?  Maybe we should have an api key going on, if so???

jmartin-sul:
@ndushay said:
> I'm fine with this but am wondering if the motivation is that this is a backend service? Maybe we should have an api key going on, if so???

filed https://github.com/sul-dlss/preservation_catalog/issues/1298 and added it to infra board for discussion when we come back from break.

jcoyne:
It doesn't matter if this is a front or backend service. This parameter validation is only for doing bulk parameter updates like this:

```
User.create(permitted_params)
```

otherwise it has no effect.

PR #1336, opened 2020-01-23
DB_UPDATE_FAILED message tweaked

ndushay:
## Why was this change made?

Because stack traces are most useful in Rails logs, not in messages in the workflow service, or in argo.  Honeybadger generates its own stacktrace.

n/a

### DIFF
--- a/app/lib/active_record_utils.rb
+++ b/app/lib/active_record_utils.rb
@@ -12,8 +12,9 @@ module ActiveRecordUtils
     rescue ActiveRecord::RecordNotFound => e
       audit_results.add_result(AuditResults::DB_OBJ_DOES_NOT_EXIST, e.inspect)
     rescue ActiveRecord::ActiveRecordError => e
+      Rails.logger.error("db update error; more context to follow: #{e.inspect}: #{e.backtrace.inspect}")
       audit_results.add_result(
-        AuditResults::DB_UPDATE_FAILED, "#{e.inspect} #{e.message} #{e.backtrace.inspect}"
+        AuditResults::DB_UPDATE_FAILED, e.inspect
       )
     end
     false


app/lib/active_record_utils.rb
```diff
@@ -12,8 +12,9 @@ def self.with_transaction_and_rescue(audit_results)
     rescue ActiveRecord::RecordNotFound => e
       audit_results.add_result(AuditResults::DB_OBJ_DOES_NOT_EXIST, e.inspect)
     rescue ActiveRecord::ActiveRecordError => e
+      Rails.logger.error("db update error; more context to follow: #{e.inspect}: #{e.backtrace.inspect}")
       audit_results.add_result(
-        AuditResults::DB_UPDATE_FAILED, "#{e.inspect} #{e.message} #{e.backtrace.inspect}"
+        AuditResults::DB_UPDATE_FAILED, e.inspect
```
ndushay: - `e.inspect` contains `e.message`, so no need to duplicate.
- removing backtrace here will greatly shorten messages for this audit error
- honeybadger will generate its own stack track

app/lib/active_record_utils.rb
```diff
@@ -12,8 +12,9 @@ def self.with_transaction_and_rescue(audit_results)
     rescue ActiveRecord::RecordNotFound => e
       audit_results.add_result(AuditResults::DB_OBJ_DOES_NOT_EXIST, e.inspect)
     rescue ActiveRecord::ActiveRecordError => e
+      Rails.logger.error("db update error; more context to follow: #{e.inspect}: #{e.backtrace.inspect}")
```
ndushay: we want to log the backtrace in the Rails logs;  this will do that, and then the additional contextual information (which check being run?  which storage root?)  will come from AuditResults.report_results().

Issue #1317, opened 2020-01-22
Get Druid List for Single Storage Root


ndushay:
in order to know which druids are on an old storage root  (x00,000), as there will be multiple old storage roots per new storage bring.

It would be grand to NOT have to go into rails console to do this.  Perhaps a rake task (outputs csv) and/or an API call (returns csv) and/or a GUI (sortable table)?

This is a variant of #1318 -- for this case, we believe there will be utility to just getting the druids

- sort list by druid
- to a file (csv?) #1318
- in GUI - sortable table #1320

PR #1339, opened 2020-01-24
Adds a storage root migration service and rake task.

### LINKED ISSUE #1322: Rake Task to Migrate Individual Storage Root in Database
Each storage root will be migrated individually, and there will be a point when we are ready for the prescat database for objects on a given root to stop pointing to the old root and start pointing to the new root.  

prerequisites: 
- #1321: db migration
- #1329: Determine if we want to wipe out previous validation dates as part of this rake task.  
- #1331: Do we want db to have full audit error info?

For each object on a root:
- change the root to new location in prescat db
- set status to "unknown" 
    - status unknown will automatically trigger the queueing of C2M and CV for the moab
- marks object “migrated” in our brand new boolean field (issue #1321) 
    - indicate old storage root in that field assuming it is created also (also issue #1321)
- wipe out previous validation dates ? (#1329)

questions:
- do we want to explicitly kick off C2M as part of this rake task?  As a separate rake task (can we run C2M by storage root? -- should this be spun off as a separate issue?)

Questions related to scheduling the rake task #1332:
- #1330: do we want to do MV and CV of all objects on new root BEFORE we change root w this rake task?

justinlittman:
closes #1322

## Why was this change made?
To update complete moab records when migrating storage roots.

No.

### DIFF
--- a/app/services/storage_root_migration_service.rb
+++ b/app/services/storage_root_migration_service.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+# Migrates Complete Moab records to a new Moab Storate Root.
+class StorageRootMigrationService
+  def initialize(from_name, to_name)
+    @from_name = from_name
+    @to_name = to_name
+  end
+
+  # @return [Array<String>] druids of migrated moabs
+  def migrate
+    druids = []
+    from_root.complete_moabs.find_each do |complete_moab|
+      migrate_moab(complete_moab)
+      druids << complete_moab.preserved_object.druid
+    end
+    druids
+  end
+
+  private
+
+  def from_root
+    @from_root ||= MoabStorageRoot.find_by!(name: @from_name)
+  end
+
+  def to_root
+    @to_root ||= MoabStorageRoot.find_by!(name: @to_name)
+  end
+
+  def migrate_moab(moab)
+    moab.moab_storage_root = to_root
+    moab.status = 'validity_unknown' # This will queue a CV.
+    # Fate of this to be determined by https://github.com/sul-dlss/preservation_catalog/issues/1329
+    moab.last_moab_validation = nil
+    moab.last_checksum_validation = nil
+    moab.save!
+  end
+end
--- a/lib/tasks/prescat.rake
+++ b/lib/tasks/prescat.rake
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+namespace :prescat do
+  desc 'Migrate storage root, returning druids of all migrated moabs'
+  task :migrate_storage_root, [:from, :to] => :environment do |_task, _args|
+    puts 'This will move all complete_moabs to a new storage root.'
+    puts 'WARNING: expects that "from" storage root is no longer being written to (no Moabs being created or modified)!'
+    print 'Enter YES to continue: '
+    input = $stdin.gets.chomp
+    next unless input == 'YES'
+
+    migration_service = StorageRootMigrationService.new(args[:from], args[:to])
+    migration_service.migrate.each { |druid| puts druid }
+  end
+end
--- a/spec/services/storage_root_migration_service_spec.rb
+++ b/spec/services/storage_root_migration_service_spec.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe StorageRootMigrationService do
+  let(:to_storage_root) { create(:moab_storage_root) }
+
+  let(:from_storage_root) { complete_moab1.moab_storage_root }
+
+  let(:complete_moab1) {
+    create(:complete_moab, status: 'ok',
+                           last_moab_validation: Time.now,
+                           last_checksum_validation: Time.now,
+                           last_archive_audit: Time.now)
+  }
+
+  let(:complete_moab2) { create(:complete_moab, status: 'invalid_moab', moab_storage_root: from_storage_root) }
+
+  let(:complete_moab3) { create(:complete_moab) }
+
+  it 'migrates the storage root' do
+    # Before migration
+    expect(complete_moab1.moab_storage_root).to eq(complete_moab2.moab_storage_root)
+    expect(complete_moab1.moab_storage_root).not_to eq(complete_moab3.moab_storage_root)
+
+    druids = described_class.new(from_storage_root.name, to_storage_root.name).migrate
+    expect(druids).to include(complete_moab1.preserved_object.druid, complete_moab2.preserved_object.druid)
+
+    complete_moab1.reload
+    complete_moab2.reload
+
+    expect(complete_moab1.moab_storage_root).to eq(to_storage_root)
+    expect(complete_moab2.moab_storage_root).to eq(to_storage_root)
+  end
+
+  it 'resets field values' do
+    described_class.new(from_storage_root.name, to_storage_root.name).migrate
+
+    complete_moab1.reload
+
+    expect(complete_moab1.status).to eq('validity_unknown')
+    expect(complete_moab1.last_moab_validation).to be_nil
+    expect(complete_moab1.last_checksum_validation).to be_nil
+    expect(complete_moab1.last_archive_audit).not_to be_nil
+  end
+
+  it 'does not migrate moabs on other storage roots' do
+    # Before migration
+    orig_complete_moab3_storage_root = complete_moab3.moab_storage_root
+
+    described_class.new(from_storage_root.name, to_storage_root.name).migrate
+
+    complete_moab3.reload
+
+    expect(complete_moab3.moab_storage_root).to eq(orig_complete_moab3_storage_root)
+  end
+end


justinlittman:
https://github.com/sul-dlss/dor-services-app/commit/9b0b78401435b49135a9cfdec2a8f832a1f1ff08

app/services/storage_root_migration_service.rb
```diff
@@ -0,0 +1,39 @@
...
+    @from_root ||= MoabStorageRoot.find_by(name: @from_name)
+  end
+
+  def to_root
+    @to_root ||= MoabStorageRoot.find_by(name: @to_name)
+  end
+
+  def migrate_moab(moab)
+    moab.moab_storage_root = to_root
+    moab.status = 'validity_unknown' # This will queue a CV.
```
ndushay: i believe this queues C2M *and* a CV.   We def want both a full moab validation and CV.

lib/tasks/prescat.rake
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+namespace :prescat do
```
ndushay: is there something we can/should do to show this is dangerous and should only be undertaken by experts?

Similarly, should we indicate that this rake task is temporary and expected to go away?

app/services/storage_root_migration_service.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# Migrates Complete Moab records to a new Moab Storate Root.
+class StorageRootMigrationService
```
ndushay: Should we indicate that this code is temporary and expected to go away?

app/services/storage_root_migration_service.rb
```diff
@@ -0,0 +1,39 @@
...
+    @to_name = to_name
+  end
+
+  # @return [Array<String>] druids of migrated moabs
+  def migrate
+    druids = []
+    from_root.complete_moabs.find_each do |complete_moab|
+      migrate_moab(complete_moab)
+      druids << complete_moab.preserved_object.druid
+    end
```
jmartin-sul: my recollection is that we said we'd only run this migration once
1. the contents of the original storage root had been fully copied to the destination storage root
2. writing of all new objects or versions of existing objects to the old storage root had stopped (i.e. we've paused accessioning for the weekend to do the cutover)

and i think we want to make that explicit in the relevant code and/or docs (so maybe comment here and in rake task description?).

jmartin-sul: also i played around with an alternate implementation of this method that just did a one shot `update_all` (initially thought to try because i'd forgotten that accessioning would be turned off, and worried about `find_each` batching, but after i realized that was a non-issue, it still seemed like it might be a little more concise and a little more efficient)...

...but then i realized that skips ActiveRecord validations and hooks, and so CV jobs would have to be queued explicitly, and then i think that might lose the concision benefit, which would make that approach much less worhtwhile, imho.  maybe you considered all that already when implementing this and i was _totally_ wasting my time (as opposed to _sort of_ wasting it) 😆 

branch is here, though, if you're curious:  https://github.com/sul-dlss/preservation_catalog/compare/t1322-migrate_task...t1322-migrate_task-single_update?expand=1

app/services/storage_root_migration_service.rb
```diff
@@ -0,0 +1,39 @@
...
+    druids
+  end
+
+  private
+
+  def from_root
+    @from_root ||= MoabStorageRoot.find_by(name: @from_name)
```
jmartin-sul:
```suggestion
    @from_root ||= MoabStorageRoot.find_by!(name: @from_name)
```

seems better to fail ASAP, `find_by!` will `raise` if the row doesn't exist.

app/services/storage_root_migration_service.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def to_root
+    @to_root ||= MoabStorageRoot.find_by(name: @to_name)
```
jmartin-sul:
```suggestion
    @to_root ||= MoabStorageRoot.find_by!(name: @to_name)
```

same idea as `from_root` suggestion

spec/services/storage_root_migration_service_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  let(:complete_moab2) { create(:complete_moab, status: 'invalid_moab', moab_storage_root: from_storage_root) }
+
+  let(:complete_moab3) { create(:complete_moab) }
+
+  it 'migrates the storage root' do
+    # Before migration
+    expect(complete_moab1.moab_storage_root).to eq(complete_moab2.moab_storage_root)
+    expect(complete_moab1.moab_storage_root).not_to eq(complete_moab3.moab_storage_root)
+
+    druids = described_class.new(from_storage_root.name, to_storage_root.name).migrate.map { |druid| druid }
```
jmartin-sul:
```suggestion
    druids = described_class.new(from_storage_root.name, to_storage_root.name).migrate
```

`#map` seems extraneous (and things still worked for me when running tests locally without it)

spec/services/storage_root_migration_service_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    expect(druids.length).to eq(2)
```
jmartin-sul:
```suggestion
    expect(druids).to include(complete_moab1.preserved_object.druid, complete_moab2.preserved_object.druid)
```

what about this slightly more precise expectation?

app/services/storage_root_migration_service.rb
julianmorley: druids is going to be btw 90K - 150K records in that array per root; is that OK?

jmartin-sul: i think that's fine, whether using `find_each` or `update_all`.  i expect PG to be able to update hundreds of thousands of records, esp if all other traffic is halted.  and while instantiating and chewing through 150K ActiveRecord objects (the `find_each` will instantiate a `CompleteMoab` for every row it retrieves) will prob be a little slow, i think it'll be tractable.

since we're planning on doing a dry run with prod data, i'm not too worried that we _have_ to figure this out ahead of time purely through speculation, and i'm pretty confident we can figure something out without too much trouble if our dry run ends up taking too long.  we could start with a smaller root for benchmarking.

FWIW, in working on #1317, i was able to dump the list of druids from the largest storage root in what felt like about 3-5 seconds.  that was just a select for the druids with no update, but it resulted in about 350K druids.

i would expect that the trouble would come if we tried to do an `update_all` _while_ the root was still actively being added to.  that sounds like it'd invite deadlock hell, but i don't think that's our plan here.

jmartin-sul: also to make it explicit:  the skipping of AR validations by `update_all` didn't worry me in this particular case, because i think the FK constraint on `moab_storage_root_id` and the translation of the `status` string to its int code provide all the validation we need for this very limited set of updates.  might feel differently when updating a different set of fields.

lib/tasks/prescat.rake
julianmorley: I honestly wouldn't expect it to go away, unless it's being replaced with same-druid-on-many-roots code. We're going to have to do these migrations every couple of years until that code gets sorted.

jmartin-sul: i would hope we'd get support for multiple `CompleteMoab`s done in less than a couple years from now, but also i sorta expect to be putting a laugh emoji on this comment two years from now.  i don't think supporting multiple `CompleteMoab`s is actually a crazy amount of work, though maybe it's a lot of work for the amount of time we usually have.  

but i think it'd also make this sort of migration much easier since it'd prevent having to pause accessioning, because we could copy stuff to a new place, and then remove it from the old place after we've determined it's safe in its new home.

app/services/storage_root_migration_service.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def migrate_moab(moab)
+    moab.moab_storage_root = to_root
+    moab.status = 'validity_unknown' # This will queue a CV.
+    # Fate of this to be determined by https://github.com/sul-dlss/preservation_catalog/issues/1329
+    moab.last_moab_validation = nil
+    moab.last_checksum_validation = nil
+    moab.last_archive_audit = nil
```
jmartin-sul: one more thought that made it into my branch last night, but not the review i left here:  i'd be in favor of not touching `last_archive_audit`, though i don't feel strongly about it, and could be convinced otherwise.

my reasoning is that the bits that are archived on the cloud are not moving as part of this local moab move, anymore than they are being moved around opaquely under the normal course of our cloud providers' operations.  additionally:
1. the status change doesn't actually trigger `audit_moab_version_replication!` at present anyway, so we'd have to add that hook if that's the desired behavior (not difficult, but a thing we'd need to wire up if we wanted it).
2. the cloud archive checks don't pull anything down for fixity checking or comparison to our local moab copies.  so even if the status change did trigger an audit, i'm not sure the moab move itself would have any bearing, other than that we'd be taking an opportunity to audit more than we otherwise would.

so i think i am slightly opposed to this for both semantic and pragmatic reasons, but like i said, not strongly, and i'm very open to counter-argument if i'm missing something.

spec/services/storage_root_migration_service_spec.rb
justinlittman: Yeah, it was necessary in an earlier version in which I was returning an Enumerator.

lib/tasks/prescat.rake
```diff
@@ -0,0 +1,9 @@
...
+  desc 'Migrate storage root, returning druids of all migrated moabs'
+  task :processing, [:from, :to] => :environment do |_task, _args|
+    migration_service = StorageRootMigrationService.new(args(:from), args(:to))
```
jmartin-sul: i copypasta'ed this as boilerplate for a rake task i was writing, and noticed that you probably actually want `args[:from]` and `args[:to]`, as opposed to the parens version here.  or at least, the square brackets approach worked for me, whereas the parens didn't.

justinlittman: Thanks. Fixed.

app/services/storage_root_migration_service.rb
jmartin-sul: i think maybe an issue or (slack/outlook) reminder or something to nudge us to revisit would make sense, but there doesn't seem to be a clear deprecation horizon for this.  marking resolved, but feel free to unresolve if that seems wrong.

lib/tasks/prescat.rake
jmartin-sul: > is there something we can/should do to show this is dangerous and should only be undertaken by experts?

current version now has a strong warning prompt

> Similarly, should we indicate that this rake task is temporary and expected to go away?

and this seems unsettled at present?

app/services/storage_root_migration_service.rb
jmartin-sul: if people ended up liking it (see usage/discussion in #1347), this might be worth considering for a follow on PR:  https://github.com/afair/postgresql_cursor

pros:  seems like if that gem were in place, you could swap `find_each` for `each_instance` and then you'd be able to page over results and use AR objects with no other changes, and have it all happen inside a transaction.

cons:  that'd be a long running transaction, though since accessioning should be off, there shouldn't be contention for the rows, so maybe ok? 🤷‍♂   and i'd probably want to add a test to make sure the CV jobs got queued for the moved moabs, just to make sure the gem was instantiating the AR object properly (i.e. in a way that triggers hooks properly).

anyway, just a thought, definitely would not do that here.

jmartin-sul: also maybe updates inside a cursor is a bad idea?  and PG docs specifically advise against select for update and row locking, iirc.

PR #1342, opened 2020-01-24
ApplicatonController: Honeybadger notifications should get logged too

jmartin-sul:
## Why was this change made?

we are currently being rate limited by honeybadger, and so it's useful to have the same info in the logs, in case HB doesn't capture all the notifications we'd otherwise see.

N/A

connects to sul-dlss/argo#1777 (may assist in debugging it)


jmartin-sul:
the test changes are not as DRY as could be, but they're not terrible, and i'm happy to fix them up if it's worth the time, but i figured i'd err on the side of expediency.

Issue #1332, opened 2020-01-22
Create Process for Migrating an Individual Storage Root


ndushay:
This could impact planning for which roots to migrate when, or conversely, it could imply what work to do when based on ops schedule of root migration.

e.g.
- run all audits on old storage root, one at a time ~~#1330~~
    - run one of (M2C, C2M, CV), and then run (and store) full report for the storage root object statuses (esp error statuses) and then run next check, etc.)??
    - store all error info in new db field ~~#1331~~
- fix audit errors when possible (#1324, #1325), re-run audits on fixed objects.
- ops migrates data to new brick
- ~~run all audits on data on new brick #1323 ?~~
- backup prescat database
- run rake task from #1322
- trigger M2C (and CV) from prescat db for freshly migrated objects
- proceed to process for sunsetting old root #1327

This should be a checklist somewhere that we can repeat for each storage root.  A wiki page?  A google doc?  A google sheet with columns for each root??

ndushay:
I think https://docs.google.com/spreadsheets/d/1jA9qIFuGYpn-FlR3Ko3hHGrnYgf4DkHR-_riGE7y46I/edit#gid=0 has a solid first pass at the checklist.  I will step through it on stage.

ndushay:
See https://docs.google.com/spreadsheets/d/1jA9qIFuGYpn-FlR3Ko3hHGrnYgf4DkHR-_riGE7y46I/edit#gid=0 for a spreadsheet approach.

Also for #1378, have this template PR up:  #1384

ndushay:
did CV run as expected?

    - logs from stage-01 show:
```
I, [2020-02-21T13:14:17.280906 #8098]  INFO -- honeybadger: ** [Honeybadger] Initializing Honeybadger Error Tracker for Ruby. Ship it! version=4.5.6 framework=rails level=1 pid=8098
I, [2020-02-21T13:14:20.494546 #8098]  INFO -- : [ActiveJob] Enqueued ChecksumValidationJob (Job ID: c52ece24-39b3-42ea-9a09-046e7eb93626) to Resque(checksum_validation) with arguments: #<GlobalID:0x000000000524e4d8 @uri=#<URI::GID gid://preservation-catalog/CompleteMoab/9137>>
[...]

PR #1345, opened 2020-01-29
[HOLD] rake task to dump list of druids on storage root to CSV file (select everything at once to memory)

### LINKED ISSUE #1317: Get Druid List for Single Storage Root

jmartin-sul:
## Why was this change made?

implementation for #1317, to help migrate moabs from old storage hardware to new

there is a description on the rake task.  happy to add to README if folks want, though i have a slight preference for further documenting this usage as part of process writeup for #1332.

closes #1317

### DIFF
--- a/app/models/moab_storage_root.rb
+++ b/app/models/moab_storage_root.rb
@@ -4,10 +4,24 @@
 # Metadata about a Moab storage root (a POSIX file system which contains Moab objects).
 class MoabStorageRoot < ApplicationRecord
   has_many :complete_moabs, dependent: :restrict_with_exception
+  has_many :preserved_objects, through: :complete_moabs
   has_and_belongs_to_many :preservation_policies
 
   validates :name, presence: true, uniqueness: true
   validates :storage_location, presence: true
+
+  scope :preserved_objects, lambda {
+    joins(complete_moabs: [:preserved_object])
+  }
+
+  # not a scope because it doesn't return a chainable relation.
+  # this could be a little expensive: use it judiciously.
+  # this completed in seconds on prod for root16, which had the most druids at the time of this writing (314162).  so it's not that bad.
+  def druid_list_sorted
+    preserved_objects
+      .pluck(:druid) # `#pluck` instead of e.g. `#select` keeps us from instantiating a whole AR obj when we just want this string
+      .sort # the list could be large (potentially > 500K; prefer load on app server to load on DB server)
+  end
 
   # Use a queue to validate CompleteMoab objects
   def validate_expired_checksums!
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -58,6 +58,10 @@ Naming/FileName:
     - 'Gemfile'
     - 'lib/capistrano/tasks/capistrano-resque-pool.rake'
 
+RSpec/BeforeAfterAll:
+  Exclude:
+    - 'spec/services/reporter_spec.rb' # allows inter-example state leak, & after(:all) won't auto-rollback fixtures it creates, but used here for file cleanup
+
 RSpec/ContextWording:
   Enabled: false # too dogmatic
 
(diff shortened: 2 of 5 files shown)


app/models/moab_storage_root.rb
```diff
@@ -4,11 +4,25 @@
...
+    joins(complete_moabs: [:preserved_object])
+  }
+
+  # not a scope because it doesn't return a chainable relation.
+  # this could be a little expensive: use it judiciously.
+  # this completed in seconds on prod for root16, which had the most druids at the time of this writing (314162).  so it's not that bad.
+  def druid_list_sorted
+    preserved_objects
+      .pluck(:druid) # `#pluck` instead of e.g. `#select` keeps us from instantiating a whole AR obj when we just want this string
+      .sort # the list could be large (potentially > 500K; prefer load on app server to load on DB server)
```
justinlittman: Why not sort in db? That's what dbs are for.

justinlittman: Also, this might be using a cursor for the results. However, you're obviating the value of it by forcing all results into memory.

jmartin-sul: > Why not sort in db? That's what dbs are for

in my testing on prod w/ the largest storage root (4, iirc, just under 350K druids), this returned in seconds (felt like under 5, but i didn't time it).  since it felt like this might be used during normal production activity, i thought it'd be safer to put the load on the app server than on the DB server, though i think the load is negligible either way, so i don't have strong feelings about this.  but my thinking was better to keep the DB as responsive as possible, and to keep queries running for as little time as possible, for deadlock prevention.

and i like DBs more for FK constraints and other schema enforcement and such than for sorting, but i like 'em for sorting too 😛 

> Also, this might be using a cursor for the results. However, you're obviating the value of it by forcing all results into memory.

query and in-app sort ran super quickly, so i don't think this is really an issue here at present, but again, happy to page through instead if you have a strong preference for that.  didn't seem worth the complexity based on my testing, but that was the first approach i considered.

i can see worries about this scaling as storage roots become larger, no preference as to whether we consider that now or kick the can down the road.

could instead return an enumerator here and page through in the CSV writing loop.

once nice thing about the pluck is that no AR objects get instantiated, since we don't need those here, just druid strings.

also, to be totally fair, i tested the query on prod, but not the actual CSV writing (because i could test the query on prod w/o deploying code changes).  doubt the CSV writing would make much of a difference, but i could def be wrong there.

ndushay: for the new storage roots, this list could be > 500K ...
and I'm fine with it be sorted in either location.

jmartin-sul: > Also, this might be using a cursor for the results

i got curious about this and did a bit of digging.  none of the docs seem to say it directly, but i found some circumstantial evidence indicating that `find_each` doesn't use a cursor, and that in general ActiveRecord doesn't do any automatic cursor magic.  this makes sense -- `find_each` docs are pretty up front about its primary-key-based-paging, non-orderable, multiple-query-running batching strategy.  i also got the impression that postgres will not use a cursor unless you go through the machinations of setting one up (i.e., it didn't seem to me like PG would decide to use a cursor to execute a query via a plain `select` statement).

and in general i was rusty with cursors, so i looked to see how you'd use one in rails-land, and i found:  https://github.com/afair/postgresql_cursor

seems reasonably well-maintained.  i didn't find a ton in the way of (blog, stack exchange, forum) posts about PG cursor usage (or cursors in general) for rails, but i did find a couple people recommending that gem.  it seems reasonably well written/maintained/tested, other than not showing test coverage.

so, here's an alternate implementation that uses that cursor gem:  https://github.com/sul-dlss/preservation_catalog/pull/1347

as a bonus compared to `find_each`, that branch's implementation uses a method from the postgresql_cursor gem that returns each row as a hash containing only the selected columns, instead of a fully instantiated active record object.

jmartin-sul:
@ndushay, sample output:
```
$ cat log/reports/MoabStorageRoot_fixture_sr1_druids_2020-01-29T17:41:20Z.csv
bj102hs9687
bj102hs9688
bz514sm9647
jj925bx9565
```

jmartin-sul:
i suspect we'll end up going with #1347, but if we don't, i can make the same rspec refinement here that i made there after jcoyne reviewed that one.

jmartin-sul:
superseded by #1347

PR #1347, opened 2020-01-30
rake task to dump list of druids on storage root to CSV file (page through select results w/ cursor)

### LINKED ISSUE #1317: Get Druid List for Single Storage Root

jmartin-sul:
* obtain results using DB cursor and page through them using default batch size to write CSV
  * install supporting gem for easier cursor usage

## Why was this change made?

this is an alternative implementation to #1345, prompted by PR review.

this implements #1317, to help migrate moabs from old storage hardware to new.

there is a description on the rake task. happy to add to README if folks want, though i have a slight preference for further documenting this usage as part of process writeup for #1332.

closes #1317

### DIFF
--- a/app/models/moab_storage_root.rb
+++ b/app/models/moab_storage_root.rb
@@ -4,10 +4,15 @@
 # Metadata about a Moab storage root (a POSIX file system which contains Moab objects).
 class MoabStorageRoot < ApplicationRecord
   has_many :complete_moabs, dependent: :restrict_with_exception
+  has_many :preserved_objects, through: :complete_moabs
   has_and_belongs_to_many :preservation_policies
 
   validates :name, presence: true, uniqueness: true
   validates :storage_location, presence: true
+
+  scope :preserved_objects, lambda {
+    joins(complete_moabs: [:preserved_object])
+  }
 
   # Use a queue to validate CompleteMoab objects
   def validate_expired_checksums!
--- a/spec/services/reporter_spec.rb
+++ b/spec/services/reporter_spec.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Reporter do
+  let!(:test_start_time) { DateTime.now } # useful for both output cleanup and CSV filename testing
+
+  let!(:msr_a) { create(:moab_storage_root) }
+  let!(:complete_moab_1) { create(:complete_moab, moab_storage_root: msr_a) }
+  let!(:complete_moab_2) { create(:complete_moab, moab_storage_root: msr_a) }
+  let!(:complete_moab_3) { create(:complete_moab, moab_storage_root: msr_a) }
+
+  let!(:msr_b) { create(:moab_storage_root) }
+  let!(:complete_moab_4) { create(:complete_moab, moab_storage_root: msr_b) }
+  let!(:complete_moab_5) { create(:complete_moab, moab_storage_root: msr_b) }
+  let!(:complete_moab_6) { create(:complete_moab, moab_storage_root: msr_b) }
+
+  after do
+    next unless FileTest.exist?(described_class.default_filepath)
+    Dir.each_child(described_class.default_filepath) do |filename|
+      fullpath_filename = File.join(described_class.default_filepath, filename)
+      File.unlink(fullpath_filename) if File.stat(fullpath_filename).mtime > test_start_time
+    end
+  end
+
+  describe '.moab_storage_root_druid_list_to_csv' do
+    it 'creates a file containing a list of druids from the given storage root' do
+      csv_filename = described_class.moab_storage_root_druid_list_to_csv(storage_root_name: msr_b.name)
+      expect(CSV.read(csv_filename)).to eq(
+        [complete_moab_4, complete_moab_5, complete_moab_6].map { |cm| [cm.preserved_object.druid] }
+      )
+    end
+
+    it 'names the file with the default prefix and a timestamp and writes it to the default location' do
+      # file timestamp is to the second, but test_start_time is more precise.  this sleep allows us to compare
+      # timestamp to test start time without having to round fractions of a second.
+      sleep(1.second)
+
+      csv_filename = described_class.moab_storage_root_druid_list_to_csv(storage_root_name: msr_a.name)
+      expect(csv_filename).to match(%r{^#{described_class.default_filepath}\/MoabStorageRoot_#{msr_a.name}_druids_.*\.csv$})
+      timestamp_str = /MoabStorageRoot_#{msr_a.name}_druids_(.*)\.csv$/.match(csv_filename).captures[0]
+      expect(DateTime.parse(timestamp_str)).to be >= test_start_time
+    end
+
+    it 'allows the caller to specify an alternate filename, including full path' do
+      alternate_filename = '/tmp/my_cool_druid_export.csv'
+      csv_filename = described_class.moab_storage_root_druid_list_to_csv(storage_root_name: msr_a.name, csv_filename: alternate_filename)
+      expect(csv_filename).to eq(alternate_filename)
+      expect(CSV.read(csv_filename)).to eq(
+        [complete_moab_1, complete_moab_2, complete_moab_3].map { |cm| [cm.preserved_object.druid] }
+      )
+    ensure
+      File.unlink(alternate_filename) if FileTest.exist?(alternate_filename)
+    end
+
+    it 'lets the DB error bubble up if the given storage root does not exist' do
+      expect { described_class.moab_storage_root_druid_list_to_csv(storage_root_name: 'nonexistent') }.to raise_error(ActiveRecord::RecordNotFound)
+    end
+
+    it 'raises an error if the intended file name is already in use' do
+      duplicated_filename = File.join(described_class.default_filepath, 'my_duplicated_filename.csv')
+      described_class.moab_storage_root_druid_list_to_csv(storage_root_name: msr_b.name, csv_filename: duplicated_filename)
+      expect {
+        described_class.moab_storage_root_druid_list_to_csv(storage_root_name: msr_b.name, csv_filename: duplicated_filename)
+      }.to raise_error(StandardError, "#{duplicated_filename} already exists, aborting!")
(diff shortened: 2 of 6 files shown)


jmartin-sul:
very similar to #1345:
```
$ git diff list-druids-on-MoabStorageRoot list-druids-on-MoabStorageRoot_cursor-impl 
diff --git a/Gemfile b/Gemfile
index a0f34274..95031a8d 100644
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'jwt' # for gating programmatic access to the application
 gem 'okcomputer' # ReST endpoint with upness status
 gem 'pg' # postgres database
 gem "factory_bot_rails", "~> 4.0"
+gem 'postgresql_cursor' # for paging over large result sets efficiently
 # pry is useful for debugging, even in prod
 gem 'pry-byebug' # call 'binding.pry' anywhere in the code to stop execution and get a pry-byebug console
 gem 'pry-rails' # use pry as the rails console shell instead of IRB
diff --git a/Gemfile.lock b/Gemfile.lock
index b4969d4d..a87aed50 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -246,6 +246,8 @@ GEM
     parser (2.7.0.2)
       ast (~> 2.4.0)
     pg (1.2.2)
+    postgresql_cursor (0.6.4)
+      activerecord (>= 3.1.0)
     pry (0.12.2)
       coderay (~> 1.1.0)
       method_source (~> 0.9.0)
@@ -406,6 +408,7 @@ DEPENDENCIES
   moab-versioning
   okcomputer
   pg
+  postgresql_cursor
   pry-byebug
   pry-rails
   puma (~> 3.12)
diff --git a/app/services/reporter.rb b/app/services/reporter.rb
index 53fd91c2..c6fff438 100644
--- a/app/services/reporter.rb
+++ b/app/services/reporter.rb
@@ -15,7 +15,14 @@ class Reporter
     raise "#{csv_filename} already exists, aborting!" if FileTest.exist?(csv_filename)
 
     CSV.open(csv_filename, 'w') do |csv|
-      msr.druid_list_sorted.each { |druid| csv << [druid] }
+      PreservedObject
+        .joins(:complete_moabs)
+        .where(complete_moabs: { moab_storage_root: msr })
+        .select(:druid)
+        .order(:druid)
+        .each_row do |po_hash| # #each_row is from postgresql_cursor gem
+          csv << [po_hash['druid']]
+        end
     end
 
     csv_filename
```

jmartin-sul:
example query and application output from rails console using fixture data:
```
[41] pry(main)> PreservedObject.select(:druid).joins(:complete_moabs).where(complete_moabs: { moab_storage_root: msr }).order(:druid).each_row(batch_size: 2) do |po_hash|
[41] pry(main)*   puts(po_hash['druid'])
[41] pry(main)* end  
   (2.2ms)  BEGIN
   (2.0ms)  declare cursor_2bc19c81868e44debc6fca2093f26d23 no scroll cursor for SELECT "preserved_objects"."druid" FROM "preserved_objects" INNER JOIN "complete_moabs" ON "complete_moabs"."preserved_object_id" = "preserved_objects"."id" WHERE "complete_moabs"."moab_storage_root_id" = 1 ORDER BY "preserved_objects"."druid" ASC
   (2.4ms)  fetch 1000 from cursor_2bc19c81868e44debc6fca2093f26d23
bj102hs9687
bj102hs9688
bz514sm9647
jj925bx9565
   (1.8ms)  fetch 1000 from cursor_2bc19c81868e44debc6fca2093f26d23
   (1.4ms)  close cursor_2bc19c81868e44debc6fca2093f26d23
   (1.9ms)  COMMIT
=> 4
[42] pry(main)> 
```

you can see the printing of the druids interspersed with the logged SQL for the cursor-based batching, all in a transaction ([cursors can only be used in transactions, apparently](https://www.postgresql.org/docs/9.4/plpgsql-cursors.html)).

here's a test with an artificially small batch size (just to demonstrate batching in action):
```
[44] pry(main)> PreservedObject.select(:druid).joins(:complete_moabs).where(complete_moabs: { moab_storage_root: msr }).order(:druid).each_row(block_size: 2) do |po_hash|
[44] pry(main)*   puts(po_hash['druid'])  
[44] pry(main)* end  
   (2.5ms)  BEGIN
   (3.1ms)  declare cursor_78a1eeb25803494484b76a7d32dc74f7 no scroll cursor for SELECT "preserved_objects"."druid" FROM "preserved_objects" INNER JOIN "complete_moabs" ON "complete_moabs"."preserved_object_id" = "preserved_objects"."id" WHERE "complete_moabs"."moab_storage_root_id" = 1 ORDER BY "preserved_objects"."druid" ASC
   (2.4ms)  fetch 2 from cursor_78a1eeb25803494484b76a7d32dc74f7
bj102hs9687
bj102hs9688
   (2.2ms)  fetch 2 from cursor_78a1eeb25803494484b76a7d32dc74f7
bz514sm9647
jj925bx9565
   (2.6ms)  fetch 2 from cursor_78a1eeb25803494484b76a7d32dc74f7
   (2.0ms)  close cursor_78a1eeb25803494484b76a7d32dc74f7
   (1.7ms)  COMMIT
=> 4
[45] pry(main)> 
```

jmartin-sul:
sample output:
```
$ cat log/reports/MoabStorageRoot_fixture_sr1_druids_2020-01-29T17:41:20Z.csv
bj102hs9687
bj102hs9688
bz514sm9647
jj925bx9565
```

app/models/moab_storage_root.rb
```diff
@@ -4,11 +4,16 @@
 # Metadata about a Moab storage root (a POSIX file system which contains Moab objects).
 class MoabStorageRoot < ApplicationRecord
   has_many :complete_moabs, dependent: :restrict_with_exception
+  has_many :preserved_objects, through: :complete_moabs
```
jmartin-sul: this ActiveRecord association is now a remnant of an earlier implementation.  i think i'd find it useful when doing investigations from the console (poking around e.g. stage or prod DB).  but since it was introduced in this PR but isn't actually called by any of the code in it, i'd be happy to get rid of it.

app/models/moab_storage_root.rb
```diff
@@ -4,11 +4,16 @@
...
   has_and_belongs_to_many :preservation_policies
 
   validates :name, presence: true, uniqueness: true
   validates :storage_location, presence: true
 
+  scope :preserved_objects, lambda {
+    joins(complete_moabs: [:preserved_object])
+  }
```
jmartin-sul: this scope is now a remnant of an early implementation of #1345.  i think i'd find it useful when doing investigations from the console (poking around e.g. stage or prod DB).  but since it was introduced in this PR but isn't actually called by any of the code in it, i'd be happy to get rid of it.

spec/services/reporter_spec.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Reporter do
+  TEST_START_TIME = DateTime.now # useful for both output cleanup and CSV filename testing
```
jcoyne: How about making this: `let!(:start_time) { DateTime.now }`?  and changing `after(:all)` to `after(:each)`? Would that work?

jmartin-sul: i'll give that a shot!  first thing i tried was `let(:start_time) { DateTime.now }`, and rspec complained about `let` usage in an `after`.  but it may have been specific to `after(:all)`, can't remember for sure.

jmartin-sul: that seemed to work, though it also seemed to make consistently flappy a test that was probably prone to flapping anyway.  have a (hopefully acceptable) fix for that too, will push changes shortly...

spec/services/reporter_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      csv_filename = described_class.moab_storage_root_druid_list_to_csv(storage_root_name: msr_b.name)
+      expect(CSV.read(csv_filename)).to eq(
+        [complete_moab_4, complete_moab_5, complete_moab_6].map { |cm| [cm.preserved_object.druid] }
+      )
+    end
+
+    it 'names the file with the default prefix and a timestamp and writes it to the default location' do
+      # file timestamp is to the second, but test_start_time is more precise.  this sleep allows us to compare
+      # timestamp to test start time without having to round fractions of a second.
+      sleep(1.second)
```
jmartin-sul: i can see how this test would be prone to flapping, though it hadn't happened to me before.  but it consistently happened once i made the change @jcoyne suggested in [his review](https://github.com/sul-dlss/preservation_catalog/pull/1347#discussion_r373027708).  i assume it's something like the `let!` getting instantiated later (closer to when the file is written) compared to the constant it replaced.

anyway, hopefully this approach to mitigating the problem described in this comment is acceptable -- the alternatives were to either not have the expectation about the time stamp being greater than the start time of the tests, or to round fractions of a second off the start time, and those both struck me as less appealing options than the (very short) `sleep` here.  though i do understand that this is a pattern we want to use as sparingly as possible.

jmartin-sul:
@jcoyne, ready for another look

ndushay:
remind me why this can't have a ticket for a rake task to wrap this report creation?  I am not so comfy making any but the simplest db queries in the rails console.

jmartin-sul:
i think the task is already there with documentation in description?

https://github.com/sul-dlss/preservation_catalog/pull/1347/files#diff-3c9d1cdc24cb1bd22894abce698e4d6aR16-R22

```
$ be rake -T prescat
rake prescat:migrate_storage_root[from,to]                       # Migrate storage root, returning druids of all migrated moabs
rake prescat:reports:msr_druids[storage_root_name,csv_filename]  # query for druids on storage root & dump to CSV (2nd arg optional)
```

or is something more necessary?

ndushay:
we're good - I was looking for README change, but this is fine.  Sorry to have missed it.

Issue #1249, opened 2019-12-02
instead of sending audit outcomes to workflow service, just expose status via REST endpoint


jmartin-sul:
reporting audit results to workflow server was always a hack for development expediency, because we thought it'd be a faster way (in terms of development time) to expose audit results to argo users.  however, this mechanism has been brittle in a number of unexpected ways, and it seems like it would be better to:

- [ ] expose the relevant status info for `CompleteMoab`s and `ZipPart`s (and whatever else) via REST (e.g. caller could hit pres cat with a given druid, and relevant status info etc for Moab(s) and archive copies could be returned).
- [ ] as it does with workflow service, the dor-services code that dor_indexing_app uses to build the solr doc could hit pres cat and index e.g. moab status and some sort of summary archive status (e.g. `archived`, `unreplicated`, `some_parts_not_found`, etc).
  * downside: one more thing dor_indexing_app has to call out to
  * upside: reduced fragility from the pres cat side, because there's one less thing (workflow server) that it has to call out to.
  * upside: pres cat no longer has to know about workflow service.  less circular dependency at the service level (the argo/DOR side already knows about preservation -- this would be one less argo/DOR side thing preservation has to know about)
  * upside: allows for more nuanced indexing/faceting/searching in argo

to help reduce worry about breaking the existing reporting functionality, we could implement the indexing, let the two approaches live side by side for a bit, and then get rid of the workflow reporting.

codebases that this would touch:
* pres cat: new controller method to expose the above described info
* dor-services: where the new indexing code that hits pres cat would live
* workflow services:  to get rid of the `preservationAuditWF`

doing this would obviate the need for #1240, since we'd no longer need the workflow reporting mechanism at all.

ndushay:
only index it if we are doing something more than having this display for an object at API endpoint (part of /objects/druid or separately)

ndushay:
Note that however audit outcomes are exposed, we want the DETAILS of the audits -- WHY did it fail?

This might be best handled by adding a status_details field to complete_moab and to zip_parts.  Note that field size is finite, and we've run into zip_part audits sending too much text to workflow-service, so we would want to avoid this #1240.   It might be addressed by having each zip part get its own status details.

- audit_results would update these fields when necessary
- ReST API call to prescat would get info from all these places and combine it
    - this info could be added to objects/show result?

ndushay:
I just barely started this work.  It is is the audit-status-in-db branch.

ndushay:
[...]

PR #1352, opened 2020-02-05
populate status_details for complete_moab

### LINKED ISSUE #1331: Have DB Retain Full Audit Errors For Moabs Failing Audits
Related to 2020 preservation storage migration work as well as SDR evolution/maint work to expose audit outcomes via ReST interface, rather than via workflow service (#1249)

- [x] add "status_details" field to complete_moab (PR #1349)
- [x] have audit_results update this field with result data (for errors only?  for all results?) (PR ~~#1352~~ #1355)

Bonus towards #1249
- [ ] this info could be added to objects/show result

Note: this ticket's work will *NOT*
- replace workflow-service audit results per #1249 
- add status_detail field to zip-part

-------

#### Original Ticket:
"Should DB Retain Full Audit Errors For Objects Failing Audits?"

should there be varchar column in db to hold full audit errors for objects that don't pass validation (just the most recent validation)? 

- this could reduce work to get full audit error info on a storage root before migration (#1318, #1319)
   - current full error info is at one or more of these places:
        - workflow step free text area (so in workflow db?)
        - ?argo index?  or is that display just showing info from workflow service?
        - available temporarily as part of running audits.

Relates to getting validation errors from storage root (#1318), listing info per druid for a storage root (#1319) and pre-migration db changes (#1321)

ndushay:
FIxes #1331

## Why was this change made?

To populate the status_details field for a CompleteMoab with any error results from the latest audit of the Moab.  This helps us with storage root migration, and it also positions us to provide audit results via ReST API (#1249)

QUESTION:  I assumed we only wanted the error results - true?

n/a


app/controllers/catalog_controller.rb
```diff
@@ -13,7 +13,7 @@ def create
     @poh = PreservedObjectHandler.new(bare_druid, incoming_version, incoming_size, moab_storage_root)
     poh.create(checksums_validated)
     status_code =
-      if poh.results.contains_result_code?(:created_new_object)
+      if poh.results.contains_result_code?(:created_new_object) && !poh.results.contains_result_code?(:db_update_failed)
```
ndushay: bumped into these errors when getting existing tests to pass -- we do NOT want to send a 200 status code if we are able to create a new object but a later db update fails.

it may be that this combination is impossible, but I couldn't find a good way to update existing tests without adding this check.

app/services/audit_results.rb
```diff
@@ -149,7 +149,7 @@ def remove_db_updated_results
   #   results = [result1, result2]
   #   result1 = {response_code => msg}
   #   result2 = {response_code => msg}
-  def report_results(logger=Rails.logger)
+  def report_results(logger=Rails.logger, complete_moab=nil)
```
ndushay: I felt the most elegant way to be able to write audit results to a CompleteMoab object was to do it within the AuditResults class.  However, we only want to do this when the audit pertains to a CompleteMoab (not a Zip or Part).  Sending in the CompleteMoab instance as an arg seemed to me the most elegant way to do this.

app/services/audit_results.rb
```diff
@@ -165,10 +165,23 @@ def report_results(logger=Rails.logger)
...
     report_errors_to_workflows(workflow_results)
     result_array
   end
 
+  # expects complete_moab to not be nil
+  def update_moab_status_details(complete_moab)
+    moab_error_codes = WORKFLOW_REPORT_CODES + [INVALID_MOAB]
+    moab_error_results = result_array.select { |r| moab_error_codes.include?(r.keys.first) }
+
+    # transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(moab_error_results) do
```
ndushay: I wasn't sure whether to wrap this in a transaction or not.  I leaned towards "not."  If you agree, I will remove the comments here.  If you disagree, I can uncomment what is here.

app/services/preserved_object_handler.rb
```diff
@@ -171,16 +171,22 @@ def update_version(checksums_validated = false)
...
 
-    results.report_results
+    results.report_results(@logger, comp_moab)
   end
 
   def pres_object
     @pres_object ||= PreservedObject.find_by!(druid: druid)
+  rescue ActiveRecord::RecordNotFound => e
+    results.add_result(AuditResults::DB_OBJ_DOES_NOT_EXIST, e.inspect)
+    nil
```
ndushay: these rescue blocks became necessary as we need to do more with these db objects now as part of `report_results`

spec/controllers/catalog_controller_spec.rb
```diff
@@ -233,18 +235,18 @@
 
     context 'db update failed' do
       before do
-        allow(PreservedObject).to receive(:find_by!).with(druid: bare_druid)
-                                                    .and_raise(ActiveRecord::ActiveRecordError, 'foo')
+        allow(pres_obj).to receive(:save!).and_raise(ActiveRecord::ActiveRecordError, 'save borked')
```
ndushay: `:find_by` is RecordNotFound so I had to change this test to `:save!`

spec/services/audit_results_spec.rb
```diff
@@ -159,7 +277,7 @@
         audit_results.report_results
       end
 
-      it 'multiple errors are concatenated together with || separator' do
+      it 'multiple errors are concatenated together with && separator' do
```
ndushay: oops - this was an incorrect description

spec/services/audit_results_spec.rb
```diff
@@ -256,7 +374,7 @@
       end
     end
 
-    context 'resets workflow error' do
+    context 'resets error in workflow service' do
```
ndushay: the old description confused me.

app/controllers/catalog_controller.rb
ndushay: @jmartin-sul should I make a PR with this small change to catalog_controller, or is this just an oddity from how I originally implemented it?

ndushay:
closing in favor of PR #1355 (thanks @jmartin-sul for the excellent steer!)

PR #1356, opened 2020-02-06
Add additional relationship for migrated moab storage root

### LINKED ISSUE #1321: DB Changes for Storage Migration
db changes:
- add column for migration timestamp
- add column for old storage root location  <-- can be prepopulated with current storage root

See also #1331 for adding a db column to hold validation error results.

aaron-collier:
## Why was this change made?

Fixes #1321 

This adds an additional relationship between `CompleteMoab` and `MoabStorageRoot` to record the from storage root when recording migration results.

### DIFF
--- a/db/migrate/20200206163659_add_migrated_at_to_complete_moabs.rb
+++ b/db/migrate/20200206163659_add_migrated_at_to_complete_moabs.rb
@@ -1,5 +1,0 @@
-class AddMigratedAtToCompleteMoabs < ActiveRecord::Migration[6.0]
-  def change
-    add_column :complete_moabs, :from_moab_storage_root_id, :bigint
-  end
-end
--- a/db/migrate/20200206163659_add_migrated_at_to_complete_moabs.rb
+++ b/db/migrate/20200206163659_add_migrated_at_to_complete_moabs.rb
@@ -1,0 +1,5 @@
+class AddMigratedAtToCompleteMoabs < ActiveRecord::Migration[6.0]
+  def change
+    add_column :complete_moabs, :from_moab_storage_root_id, :bigint
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrate/20200206163659_add_migrated_at_to_complete_moabs.rb
```diff
@@ -0,0 +1,5 @@
+class AddMigratedAtToCompleteMoabs < ActiveRecord::Migration[6.0]
+  def change
+    add_column :complete_moabs, :from_moab_storage_root_id, :bigint
```
jmartin-sul: can't remember off-hand what the syntax is, esp since this is a second FK relationship to `moab_storage_roots`, but could we add a DB level foreign key constraint to this column?

db/migrate/20200206163659_add_migrated_at_to_complete_moabs.rb
```diff
@@ -0,0 +1,5 @@
+class AddMigratedAtToCompleteMoabs < ActiveRecord::Migration[6.0]
```
jmartin-sul: this class name seems a bit misleading, but whatever, i don't think it matters much.  if you end up adding the FK constraint suggested in another comment and you want to rename this, great, if you don't care to bother that's also totally fine w/ me.

aaron-collier: Oh yeah, I was flipping around on  that,  I can  update.

db/migrate/20200206163659_add_migrated_at_to_complete_moabs.rb
aaron-collier: I'll  test a  FK.  I was  getting an error with it originally  but I think that was  fixed with the `required: false` so a FK is likely a good idea.

Issue #1320, opened 2020-01-22
Web Interface to Show Audit Results (per root? per druid?)


ndushay:
Ideally, there would be some sort of web interface on prescat to show current audit results - possibly rendered as a sortable table, with columns such as:

druid, storage root, status, audit errors, last M2C run, last C2M run, last CV run, whatever

related to #1318, #1319

jmartin-sul:
though we probably won't use them exactly as-is, there are some example queries in [db/README.md](https://github.com/sul-dlss/preservation_catalog/tree/master/db#some-useful-activerecord-queries) that might provide a good basis for the queries that'd be used for this implementation.

ndushay:
this is being closed in favor of #1357 and plans in #1363

PR #1360, opened 2020-02-07
Add storage root detail reports to rake tasks

### LINKED ISSUE #1318: Get Validation Errors Info for Druids from a Single Storage Root

This differs from #1319 because this is limited to objects with validation (audit) errors;  it will ignore objects without those errors.  (It could be one implementation with an errors_only argument)

We would also ideally like access not only to the error status, but to the entire error message text (See #1331 - possibly add db field to for entire error message text).

So this could be:
- look up all druids from a storage root that are  failing any audit
- get full validation error info:
    - re-run audit to get full error text for these objects OR
    - get full error text from workflow db (audit error WF message info) OR
    - argo Solr?  or does argo just go to workflow db? OR
    - add info to db from #1331

Report would have
- storage root (could be header, not every object, since all objects have same root here)
- druid (sorted this way)
- status
- error type
- full error text
- ???

In UI, the same as report, but shown in a sortable table, with sorting by druid, by status, by error type, etc.

### LINKED ISSUE #1319: List Info Per Druid for a Single Storage Root OR List of Druids
It would be grand to NOT have to go into rails console to do this. Perhaps a rake task (outputs csv) and/or an API call (returns csv) and/or a GUI (sortable table)?

This differs from #1318 because this is not limited to objects with validation (audit) errors.  (It could be one implementation with an errors_only argument), and it won't show the full error info, should that be added to db (#1331)

per list of druids (1:manymany) or from storage root (old or new) :
- What status druid had/has
- Validation (audit) status
- Migrated status
- Storage Root
- ...  

would also like to be able to select by:
- migrated status
- validation status
- old storage root
- list of druids
- current storage root

Report (csv) would have:
- all the stuff

In UI, the same as report, but shown in a sortable table, with sorting by druid every column

aaron-collier:
## Why was this change made?

Fixes #1318 
Fixes #1319 

The refactor proposed here is also in order to make the service usable for #1320 

Adds storage root detail report and storage root audit error detail report.

DRAFT until completion of some DRY refactoring and cleanup of rake arguments.

TBD

### DIFF
--- a/app/services/reporter.rb
+++ b/app/services/reporter.rb
@@ -13,6 +13,7 @@ class Reporter
   def initialize(params)
     @storage_root = MoabStorageRoot.find_by!(name: params[:storage_root_name])
     @druids = []
+    moab_storage_root_druid_list
   end
 
   # @return [Array] an array of druids on the storage root
@@ -51,7 +52,7 @@ class Reporter
   # @param [Array] values to output on each line of the csv
   # @param [String] optional filename
   # @return [String] the name of the CSV file to which the list was written
-  def write_to_csv(data, filename)
+  def write_to_csv(data, filename: nil)
     filename ||= default_filename(filename_prefix: "MoabStorageRoot_#{storage_root.name}_druids", filename_suffix: 'csv')
     raise "#{filename} already exists, aborting!" if FileTest.exist?(filename)
 
--- a/app/services/reporter.rb
+++ b/app/services/reporter.rb
@@ -22,8 +22,27 @@ class Reporter
         .select(:druid)
         .order(:druid)
         .each_row do |po_hash| # #each_row is from postgresql_cursor gem
+          byebug
           csv << [po_hash['druid']]
         end
+    end
+
+    csv_filename
+  end
+
+  def self.moab_storage_root_druid_details_to_csv(storage_root_name:, csv_filename: nil)
+    msr = MoabStorageRoot.find_by!(name: storage_root_name)
+    msr.complete_moabs.each do |cm|
+      puts "#{cm.preserved_object.druid},#{cm.status},#{cm.moab_storage_root.name},#{cm.from_moab_storage_root&.name}"
+    end
+
+    csv_filename
+  end
+
+  def self.moab_storage_root_audit_errors_to_csv(storage_root_name:, csv_filename: nil)
+    msr = MoabStorageRoot.find_by!(name: storage_root_name)
+    msr.complete_moabs.where.not(status: 'ok').each do |cm|
+      puts "#{cm.preserved_object.druid},#{cm.status},#{cm.moab_storage_root.name},#{cm.from_moab_storage_root&.name}"
     end
 
     csv_filename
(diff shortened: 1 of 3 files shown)


app/services/reporter.rb
```diff
@@ -6,38 +6,76 @@
 # run queries and produce reports from the results, for consumption
 # by preservation catalog maintainers
 class Reporter
-  # @param [String] the name of the storage root for which druids should be listed
+  attr_reader :storage_root, :druids
+
+  # @param [String] the name of the storage root to initialize for reporter
```
mjgiarlo: Should this method take `params` as a hash (in which case the YARD up here needs tweaking), or should it take the storage root name as a string param (in which case lines 13-14 need tweaking)?

app/services/reporter.rb
```diff
@@ -6,38 +6,76 @@
...
+  # @return [Reporter] the reporter
+  def initialize(params)
+    @storage_root = MoabStorageRoot.find_by!(name: params[:storage_root_name])
+    @druids = []
```
mjgiarlo: I wonder if you'd be open to `moab_storage_root_druid_list` returning the array of hashes it's generating, and assigning that to the value of `@druids` rather than setting the ivar to a blank array and then mutating it?

app/services/reporter.rb
```diff
@@ -6,38 +6,76 @@
...
+      .joins(:complete_moabs)
+      .where(complete_moabs: { moab_storage_root: storage_root })
+      .select(:druid, :status)
+      .order(:druid)
+      .each_row do |po_hash|
+        @druids << po_hash['druid']
+      end
+  end
+
+  # @param [Array] druids to output details for
```
mjgiarlo: Need to add the name of the `@param` either before or after `[Array]`

app/services/reporter.rb
```diff
@@ -6,38 +6,76 @@
...
+  # @param [Boolean] optionally only output lines with audit errors
```
mjgiarlo: Same as above re: including the param name. Would also like to see `(default: false)` in the YARD.

app/services/reporter.rb
```diff
@@ -6,38 +6,76 @@
...
+  # @return [Array] an array of hashes with details for each druid provided
+  def moab_detail_for(data, errors_only: false)
```
mjgiarlo: Open to renaming `data` to something more intention-revealing, like `druid_list` or `druids`?

app/services/reporter.rb
```diff
@@ -6,38 +6,76 @@
...
-    CSV.open(csv_filename, 'w') do |csv|
-      PreservedObject
-        .joins(:complete_moabs)
-        .where(complete_moabs: { moab_storage_root: msr })
-        .select(:druid)
-        .order(:druid)
-        .each_row do |po_hash| # #each_row is from postgresql_cursor gem
-          csv << [po_hash['druid']]
-        end
+  def write_to_csv(data, filename: nil)
```
mjgiarlo: Open to renaming `data` in this method?

PR #1365, opened 2020-02-12
Reporting fix and touchup

jmartin-sul:
## Why was this change made?

(follow on to #1360)

initially i started in on a minor bit of refactoring so that more could be done in the initial query (which was previously just used to list druids), since that query already collects almost all the data we might want to report.  when a test failed i went to do a bit of manual testing, and noticed that the rake tasks weren't actually working, so i fixed those in addition to getting rid of the need to collect the druid list separately from the loop that collates results.  then i noticed that the reports all used the same default file name no matter the type, so i fixed that by adding a report type param to the csv writing method.

i also finally realized we probably want column headers (after waffling on that when first creating the `Reporter` class), so i added those.

did some refactoring and test fixing along the way to support the above.  more detail in the commit messages (the first one's a bit of a grab bag, but i tried to make the other ones more focused).

in the last commit i finally did the last bit of the query optimization i'd originally set out to do, using the `complete_moabs` info that the main query already provided, and doing away with the need to instantiate lots of ActiveRecord objects.

two other things i'd like to do, but which i deferred because it's late:
1. rename the class to `MoabStorageRootReporter` because it's now very moab specific (i probably should've named it that from the beginning).
2. turn the `Reporter` methods that don't take advantage of internal state back into class methods (open to argument on this -- i get the impression some prefer to avoid class methods entirely? i don't have strong feelings on this, but my inclination is to not make something an instance method if it doesn't need to be)

n/a

### DIFF
--- a/app/services/reporter.rb
+++ b/app/services/reporter.rb
@@ -35,10 +35,15 @@ class Reporter
   # @param [Boolean] errors_only (default: false) - optionally only output lines with audit errors
   # @return [Array] an array of hashes with details for each druid provided
   def moab_detail_csv_list(errors_only: false)
+    query = if errors_only
+      moab_storage_root_list_preserved_objects_relation.where.not(complete_moabs: { status: 'ok' })
+    else
+      moab_storage_root_list_preserved_objects_relation
+    end
+
     detail_array = [['druid', 'from_storage_root', 'storage_root', 'last_checksum_validation', 'last_moab_validation', 'status', 'status_details']]
-    moab_storage_root_list_preserved_objects_relation.each_instance do |preserved_object|
+    query.each_instance do |preserved_object|
       preserved_object.complete_moabs.each do |cm|
-        next if errors_only && cm.status == 'ok'
         detail_array << [
           preserved_object.druid, cm.from_moab_storage_root&.name, cm.moab_storage_root.name,
           cm.last_checksum_validation, cm.last_moab_validation, cm.status, cm.status_details
--- a/app/services/reporter.rb
+++ b/app/services/reporter.rb
@@ -12,9 +12,10 @@ class Reporter
   # @return [Reporter] the reporter
   def initialize(params)
     @storage_root = MoabStorageRoot.find_by!(name: params[:storage_root_name])
+    @msr_names = {}
   end
 
-  def moab_storage_root_list_preserved_objects_relation
+  def moab_storage_root_list_pres_obj_comp_moabs_relation
     PreservedObject
       .joins(:complete_moabs)
       .where(complete_moabs: { moab_storage_root: storage_root })
@@ -24,7 +25,7 @@ class Reporter
   # @return [Array] an array of druids on the storage root
   def druid_csv_list
     druid_array = [['druid']]
-    moab_storage_root_list_preserved_objects_relation
+    moab_storage_root_list_pres_obj_comp_moabs_relation
       .select(:druid)
       .each_row do |po_hash|
         druid_array << [po_hash['druid']]
@@ -35,22 +36,36 @@ class Reporter
   # @param [Boolean] errors_only (default: false) - optionally only output lines with audit errors
   # @return [Array] an array of hashes with details for each druid provided
   def moab_detail_csv_list(errors_only: false)
-    query = if errors_only
-      moab_storage_root_list_preserved_objects_relation.where.not(complete_moabs: { status: 'ok' })
-    else
-      moab_storage_root_list_preserved_objects_relation
-    end
-
-    detail_array = [['druid', 'from_storage_root', 'storage_root', 'last_checksum_validation', 'last_moab_validation', 'status', 'status_details']]
-    query.each_instance do |preserved_object|
-      preserved_object.complete_moabs.each do |cm|
-        detail_array << [
-          preserved_object.druid, cm.from_moab_storage_root&.name, cm.moab_storage_root.name,
-          cm.last_checksum_validation, cm.last_moab_validation, cm.status, cm.status_details
-        ]
+    query =
+      if errors_only
+        moab_storage_root_list_pres_obj_comp_moabs_relation.where.not(complete_moabs: { status: 'ok' })
+      else
+        moab_storage_root_list_pres_obj_comp_moabs_relation
       end
+
+    header_row = [['druid', 'from_storage_root', 'storage_root', 'last_checksum_validation', 'last_moab_validation', 'status', 'status_details']]
+
+    # cols doesn't include storage_root name (available via storage_root ivar).  also we're not instantiating AR objects, so we have to translate
+    # the underlying status enum
+    cols = ['druid', 'from_moab_storage_root_id', 'last_checksum_validation', 'last_moab_validation', 'status AS status_code', 'status_details']
+    data_rows = query.select(cols).each_row.map do |po_cm_hash|
+      [
+        po_cm_hash['druid'], moab_storage_root_name(po_cm_hash['from_moab_storage_root_id']), storage_root.name,
+        po_cm_hash['last_checksum_validation'], po_cm_hash['last_moab_validation'], status_text_from_code(po_cm_hash['status_code']),
+        po_cm_hash['status_details']
+      ]
     end
-    detail_array
+    header_row + data_rows
+  end
+
+  def moab_storage_root_name(msr_id)
+    return nil unless msr_id.present?
+    @msr_names[msr_id] ||= MoabStorageRoot.find(msr_id).name
+  end
+
+  def status_text_from_code(status_code)
+    # statuses is a hash of the form { 'status_text' => status_code }.  this is just a reverse hash lookup.
+    CompleteMoab.statuses.to_a.find { |s| s[1] == status_code }[0]
   end
 
   # @param [Array] lines - values to output on each line of the csv
--- a/app/services/reporter.rb
+++ b/app/services/reporter.rb
@@ -23,7 +23,7 @@ class Reporter
 
   # @return [Array] an array of druids on the storage root
   def druid_csv_list
-    druid_array = []
+    druid_array = [['druid']]
     moab_storage_root_list_preserved_objects_relation
       .select(:druid)
       .each_row do |po_hash|
@@ -35,7 +35,7 @@ class Reporter
   # @param [Boolean] errors_only (default: false) - optionally only output lines with audit errors
   # @return [Array] an array of hashes with details for each druid provided
   def moab_detail_csv_list(errors_only: false)
-    detail_array = []
+    detail_array = [['druid', 'from_storage_root', 'storage_root', 'last_checksum_validation', 'last_moab_validation', 'status', 'status_details']]
     moab_storage_root_list_preserved_objects_relation.each_instance do |preserved_object|
       preserved_object.complete_moabs.each do |cm|
         next if errors_only && cm.status == 'ok'
--- a/app/services/reporter.rb
+++ b/app/services/reporter.rb
@@ -6,45 +6,43 @@ require 'csv'
 # run queries and produce reports from the results, for consumption
 # by preservation catalog maintainers
 class Reporter
-  attr_reader :storage_root, :druids
+  attr_reader :storage_root
 
   # @params [Hash] params used to initialize the Reporter service
   # @return [Reporter] the reporter
   def initialize(params)
     @storage_root = MoabStorageRoot.find_by!(name: params[:storage_root_name])
-    @druids = moab_storage_root_druid_list
   end
 
-  # @return [Array] an array of druids on the storage root
-  def moab_storage_root_druid_list
-    druid_array = []
+  def moab_storage_root_list_preserved_objects_relation
     PreservedObject
       .joins(:complete_moabs)
       .where(complete_moabs: { moab_storage_root: storage_root })
-      .select(:druid)
       .order(:druid)
+  end
+
+  # @return [Array] an array of druids on the storage root
+  def druid_csv_list
+    druid_array = []
+    moab_storage_root_list_preserved_objects_relation
+      .select(:druid)
       .each_row do |po_hash|
-        druid_array << po_hash['druid']
+        druid_array << [po_hash['druid']]
       end
     druid_array
   end
 
-  # @param [Array] druids - list of druids to output details for
   # @param [Boolean] errors_only (default: false) - optionally only output lines with audit errors
   # @return [Array] an array of hashes with details for each druid provided
-  def moab_detail_for(druids, errors_only: false)
+  def moab_detail_csv_list(errors_only: false)
     detail_array = []
-    druids.each do |druid|
-      preserved_object = PreservedObject.find_by(druid: druid)
+    moab_storage_root_list_preserved_objects_relation.each_instance do |preserved_object|
       preserved_object.complete_moabs.each do |cm|
         next if errors_only && cm.status == 'ok'
-        detail_array << { druid: druid,
-                          status: cm.status,
-                          status_details: cm.status_details,
-                          last_moab_validation: cm.last_moab_validation,
-                          last_checksum_validation: cm.last_checksum_validation,
-                          storage_root: cm.moab_storage_root.name,
-                          from_storage_root: cm.from_moab_storage_root&.name }
+        detail_array << [
+          preserved_object.druid, cm.from_moab_storage_root&.name, cm.moab_storage_root.name,
+          cm.last_checksum_validation, cm.last_moab_validation, cm.status, cm.status_details
+        ]
       end
     end
     detail_array
@@ -53,14 +51,15 @@ class Reporter
   # @param [Array] lines - values to output on each line of the csv
   # @param [String] filename - optional filename to override the default
   # @return [String] the name of the CSV file to which the list was written
-  def write_to_csv(lines, filename: nil)
-    filename ||= default_filename(filename_prefix: "MoabStorageRoot_#{storage_root.name}_druids", filename_suffix: 'csv')
+  def write_to_csv(lines, report_type: nil, filename: nil)
+    raise ArgumentError, 'Must specify at least one of report_type or filename' unless report_type.present? || filename.present?
+
+    filename ||= default_filename(filename_prefix: "MoabStorageRoot_#{storage_root.name}_#{report_type}", filename_suffix: 'csv')
     raise "#{filename} already exists, aborting!" if FileTest.exist?(filename)
 
     ensure_containing_dir(filename)
     CSV.open(filename, 'w') do |csv|
       lines.each do |line|
-        line = line.values if line.is_a? Hash
         csv << line
       end
     end
--- a/app/services/reporter.rb
+++ b/app/services/reporter.rb
@@ -15,17 +15,10 @@ class Reporter
     @msr_names = {}
   end
 
-  def moab_storage_root_list_pres_obj_comp_moabs_relation
-    PreservedObject
-      .joins(:complete_moabs)
-      .where(complete_moabs: { moab_storage_root: storage_root })
-      .order(:druid)
-  end
-
-  # @return [Array] an array of druids on the storage root
+  # @return [Array<Array>] an array of arrays; each inner array represents a CSV row w/ a druid from the storage root. includes a header row.
   def druid_csv_list
     druid_array = [['druid']]
-    moab_storage_root_list_pres_obj_comp_moabs_relation
+    complete_moabs_and_preserved_objects_in_storage_root
       .select(:druid)
       .each_row do |po_hash|
         druid_array << [po_hash['druid']]
@@ -34,16 +27,18 @@ class Reporter
   end
 
   # @param [Boolean] errors_only (default: false) - optionally only output lines with audit errors
-  # @return [Array] an array of hashes with details for each druid provided
+  # @return [Array<Array>] an array of arrays; each inner array represents a CSV row w/ details for each druid in query result.  includes header row.
   def moab_detail_csv_list(errors_only: false)
     query =
       if errors_only
-        moab_storage_root_list_pres_obj_comp_moabs_relation.where.not(complete_moabs: { status: 'ok' })
+        complete_moabs_and_preserved_objects_in_storage_root.where.not(complete_moabs: { status: 'ok' })
       else
-        moab_storage_root_list_pres_obj_comp_moabs_relation
+        complete_moabs_and_preserved_objects_in_storage_root
       end
 
-    header_row = [['druid', 'from_storage_root', 'storage_root', 'last_checksum_validation', 'last_moab_validation', 'status', 'status_details']]
+    header_row = [
+      ['druid', 'previous storage root', 'current storage root', 'last checksum validation', 'last moab validation', 'status', 'status details']
+    ]
 
     # cols doesn't include storage_root name (available via storage_root ivar).  also we're not instantiating AR objects, so we have to translate
     # the underlying status enum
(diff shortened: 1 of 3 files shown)


justinlittman:
If I'm reading correctly, these reports read all of the data into memory before writing to CSV. I don't know the size of the reports. Is this problematic?

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
   def initialize(params)
     @storage_root = MoabStorageRoot.find_by!(name: params[:storage_root_name])
-    @druids = moab_storage_root_druid_list
+    @msr_names = {}
   end
 
-  # @return [Array] an array of druids on the storage root
-  def moab_storage_root_druid_list
-    druid_array = []
+  def moab_storage_root_list_pres_obj_comp_moabs_relation
```
mjgiarlo: I appreciate renaming this method to be more descriptive of what it's doing. If I'm grokking (and I may not be), this method is primarily operating on PreservedObjects, using CompleteMoabs only to filter the POs in the given storage root. If that is true, I wonder about a tweak to the name like `preserved_objects_in_storage_root`? (The name currently is almost Germanic in its love for compound nouns, which is impressive. :smile:)

mjgiarlo: And I might also like to see some YARD documentation to show that it returns an AR relation. (Would rather see that in YARD than in the method name, but I'm open to being overruled there! For me, I think having `preserved_objects` in the method name already implies it's an AR relation but maybe that's just me.)

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
-      .select(:druid)
       .order(:druid)
+  end
+
+  # @return [Array] an array of druids on the storage root
+  def druid_csv_list
+    druid_array = [['druid']]
+    moab_storage_root_list_pres_obj_comp_moabs_relation
+      .select(:druid)
       .each_row do |po_hash|
```
mjgiarlo: I'm not that familiar with the `#each_row` implementation. Will that loop over rows in the DB without loading it all into memory? Maybe this is also what @justinlittman was asking.

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
       .where(complete_moabs: { moab_storage_root: storage_root })
-      .select(:druid)
       .order(:druid)
+  end
+
+  # @return [Array] an array of druids on the storage root
+  def druid_csv_list
+    druid_array = [['druid']]
+    moab_storage_root_list_pres_obj_comp_moabs_relation
+      .select(:druid)
```
mjgiarlo: Is there any chance we will get duplicate druids in this list? And, if we do, do we care about druids being in the CSV list multiple times?

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
-        next if errors_only && cm.status == 'ok'
-        detail_array << { druid: druid,
-                          status: cm.status,
-                          status_details: cm.status_details,
-                          last_moab_validation: cm.last_moab_validation,
-                          last_checksum_validation: cm.last_checksum_validation,
-                          storage_root: cm.moab_storage_root.name,
-                          from_storage_root: cm.from_moab_storage_root&.name }
+  def moab_detail_csv_list(errors_only: false)
+    query =
```
mjgiarlo: Dig this change, and I'm happy to see relations passed around for chaining just like this.

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
+      if errors_only
+        moab_storage_root_list_pres_obj_comp_moabs_relation.where.not(complete_moabs: { status: 'ok' })
+      else
+        moab_storage_root_list_pres_obj_comp_moabs_relation
       end
+
+    header_row = [['druid', 'from_storage_root', 'storage_root', 'last_checksum_validation', 'last_moab_validation', 'status', 'status_details']]
```
mjgiarlo: This might could be a constant or its own method or some such, but I don't think it's too worth fussing over.

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
       .each_row do |po_hash|
-        druid_array << po_hash['druid']
+        druid_array << [po_hash['druid']]
       end
     druid_array
   end
 
-  # @param [Array] druids - list of druids to output details for
   # @param [Boolean] errors_only (default: false) - optionally only output lines with audit errors
   # @return [Array] an array of hashes with details for each druid provided
```
mjgiarlo: Is this now returning an array of arrays? How about `# @return [Array<Array>] an array of arrays, with each inner array representing a CSV row (including a header row)` ?

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
+        po_cm_hash['last_checksum_validation'], po_cm_hash['last_moab_validation'], status_text_from_code(po_cm_hash['status_code']),
+        po_cm_hash['status_details']
+      ]
     end
-    detail_array
+    header_row + data_rows
+  end
+
+  def moab_storage_root_name(msr_id)
+    return nil unless msr_id.present?
```
mjgiarlo: In this case, I have a slight preference for `return nil if msr_id.blank?` for being just a little easier (for me?) to grok. But it's no biggie.

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
+    @msr_names[msr_id] ||= MoabStorageRoot.find(msr_id).name
+  end
+
+  def status_text_from_code(status_code)
+    # statuses is a hash of the form { 'status_text' => status_code }.  this is just a reverse hash lookup.
+    CompleteMoab.statuses.to_a.find { |s| s[1] == status_code }[0]
```
mjgiarlo: I think you can shorten this to `CompleteMoab.statuses.key(status_code)`.

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
   end
 
   # @param [Array] lines - values to output on each line of the csv
   # @param [String] filename - optional filename to override the default
```
mjgiarlo: Pls add `@param` documentation of the new `report_type` kwarg.

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
   # @return [String] the name of the CSV file to which the list was written
```
mjgiarlo: Pls add `@raise` documentation for the new `ArgumentError` condition.

mjgiarlo: Would you also consider adding `@raise` doc for the `RuntimeError` on line 78?

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
-  def write_to_csv(lines, filename: nil)
-    filename ||= default_filename(filename_prefix: "MoabStorageRoot_#{storage_root.name}_druids", filename_suffix: 'csv')
+  def write_to_csv(lines, report_type: nil, filename: nil)
+    raise ArgumentError, 'Must specify at least one of report_type or filename' unless report_type.present? || filename.present?
```
mjgiarlo: As above, I might have a slight preference for `if report_type.blank? && filename.blank?`

jmartin-sul:
@justinlittman said:
> If I'm reading correctly, these reports read all of the data into memory before writing to CSV. I don't know the size of the reports. Is this problematic?

that's correct.  that wasn't the case before the refactoring in #1360, but the DRYing there made it so that everything was collected first and then passed to the CSV writer method.

i don't personally have any objection to that at this point -- that was how i'd initially approached this reporting work, though i did end up writing from within the query result paging in the reporting PR that initially got merged (#1347).  i'm happy to move back to something that pages through from within in the cursor iteration loop.  fwiw, i think we'd be getting at most a few hundred thousand rows with a few hundred characters of data each.  that feels probably fine to me, esp for a manually triggered report?

either way, i think what's in this PR is an improvement over what's in master at the moment, and doesn't change the loading everything into memory aspect of things either way.  i'd be happy to re-visit not loading everything into memory, but at this point, i think i'd prefer to do that in a separate PR.  i think i have an idea of how i'd do that, but i think i'd want to pair with someone to do it i a readable and idiomatically rails-y way.  my first thought along those lines is something like

app/services/reporter.rb
jmartin-sul: `#each_row` is from this gem:  https://github.com/afair/postgresql_cursor

it allows the result set to be returned in pages using a real database cursor, without the jankiness of `find_each`.  in the #1347 implementation, it _was_ used to address the concern of loading everything into memory before writing, but that was lost in the DRYing of #1360.  i could go either way on whether we should spend effort on getting back to writing the CSV out within the query paging instead of loading the full result set into memory.

app/services/reporter.rb
jmartin-sul: > The name currently is almost Germanic in its love for compound nouns, which is impressive

ha, glad you like it, i think my tendency for that drives @ndushay crazy 😆 

> this method is primarily operating on PreservedObjects, using CompleteMoabs only to filter the POs in the given storage root

sorta.  i found it more useful to think of this in SQL than in ActiveRecord objects.

this method is really about getting all the `CompleteMoabs` on a storage root, and all their relevant details.  `PreservedObject` is the master record for an object in the abstract, tying together its uncompressed moabs on storage roots and the zip copies in the cloud (and the preservation policies that say what goes where).  as such, the `druid` lives on the `PreservedObject`.  that's the main point of tying in the `PreservedObject`.

so really i want all the info for `CompleteMoab`s on the storage root, plus some info that lives on the related `PreservedObject` and `MoabStorageRoot` rows.  might've been better to have the query be rooted in `PreservedObject`, joining to `CompleteMoab`?  i don't have any strong feelings on that, esp since all i really want to do is get the query results as a list of column values, without bothering to instantiate ActiveRecord objects.

i think this is just a natural artifact of the tendency for OOP to clash with relational modeling at the seams where the two meet.  since this is a query for a large number of things that we want to run correctly and efficiently, i erred on the side of writing good SQL (in the most maintainable rails way i could imagine) rather than erring on the side of writing things in the rails-iest way.  but i also wanted to use AR relation composability instead of just writing straight SQL or something.  very open to feedback on how to check all those boxes in a cleaner way.

app/services/reporter.rb
jmartin-sul: it is possible in that the schema allows for multiple `CompleteMoab`s for a given `PreservedObject`.  in practice, we never got around to supporting this in the rest of the code, and some of our shameless green implementations elsewhere assume one `CompleteMoab` per `PreservedObject`, iirc.  also, in practice, i don't think we'd put the same moab on the same storage root multiple times.  so i think that in practice, the druids will be unique in this list, both because druids are currently unique across `CompleteMoab`s, and because even in our more robust future, they'd likely still be unique for all `CompleteMoab`s on a given storage root.

app/services/reporter.rb
jmartin-sul: this composability is literally the thing that softened my general dislike for heavy/magical ORMs like ActiveRecord 🙂

app/services/reporter.rb
jmartin-sul: oh yeah, hadn't thought of that!  happy to do that if you'd prefer.  i also sorta like having it right by the row that's being constructed, but i don't feel strongly about that (and that wasn't intentional, that's just how i ended up writing it).

app/services/reporter.rb
jmartin-sul: ah, yes!  good catch.

app/services/reporter.rb
jmartin-sul: fin by me, will do

app/services/reporter.rb
jmartin-sul: ah, nice, this felt circuitous when i was writing it, will fix

app/services/reporter.rb
jmartin-sul: will do

app/services/reporter.rb
jmartin-sul: will do

app/services/reporter.rb
jmartin-sul: works for me

jmartin-sul:
@mjgiarlo good catch on the YARD stuff.  i realized i missed those updates (and adding YARD docs for the new methods in this PR) after i woke up this morning.  will fix up and ping you once that's done.

app/services/reporter.rb
jmartin-sul: > And I might also like to see some YARD documentation to show that it returns an AR relation. (Would rather see that in YARD than in the method name, but I'm open to being overruled there! For me, I think having preserved_objects in the method name already implies it's an AR relation but maybe that's just me.)

works for me.  will try to get to a terser name in light of YARD doc for return type.  if you have suggestions on naming in light of the data i'm trying to get back out of the relation, i'm all ears.

jmartin-sul:
oh also, sample rake task output:

```
$ be rake prescat:reports:msr_druids['fixture_sr1']
druids for fixture_sr1 written to /preservation_catalog/log/reports/MoabStorageRoot_fixture_sr1_druids_2020-02-12T09:45:09Z.csv
$
$ cat "/preservation_catalog/log/reports/MoabStorageRoot_fixture_sr1_druids_2020-02-12T09:45:09Z.csv"
druid
bj102hs9687
bz514sm9647
jj925bx9565
```

```
$ be rake prescat:reports:msr_druid_detail['fixture_sr1']
druid details for fixture_sr1 written to /preservation_catalog/log/reports/MoabStorageRoot_fixture_sr1_druid_detail_2020-02-12T09:46:34Z.csv
$
$ cat "/preservation_catalog/log/reports/MoabStorageRoot_fixture_sr1_druid_detail_2020-02-12T09:46:34Z.csv"
druid,from_storage_root,storage_root,last_checksum_validation,last_moab_validation,status,status_details
bj102hs9687,,fixture_sr1,2020-02-12 05:29:37 UTC,2020-02-12 05:29:36 UTC,invalid_checksum,validate_checksums (actual location: fixture_sr1; ) Moab file spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0001/data/metadata/versionMetadata.xml was not found in Moab signature catalog spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0003/manifests/signatureCatalog.xml && CompleteMoab status changed from validity_unknown to invalid_checksum
bz514sm9647,,fixture_sr1,2020-02-12 05:29:37 UTC,2020-02-12 05:29:37 UTC,ok,validate_checksums (actual location: fixture_sr1; ) checksum(s) match && CompleteMoab status changed from validity_unknown to ok
jj925bx9565,,fixture_sr1,2020-02-12 05:29:37 UTC,2020-02-12 05:29:37 UTC,ok,validate_checksums (actual location: fixture_sr1; ) checksum(s) match && CompleteMoab status changed from validity_unknown to ok
```

```
$ be rake prescat:reports:msr_audit_errors['fixture_sr1']
druids with errors details for fixture_sr1 written to /preservation_catalog/log/reports/MoabStorageRoot_fixture_sr1_audit_errors_2020-02-12T09:47:12Z.csv
$
$ cat "/preservation_catalog/log/reports/MoabStorageRoot_fixture_sr1_audit_errors_2020-02-12T09:47:12Z.csv"
druid,from_storage_root,storage_root,last_checksum_validation,last_moab_validation,status,status_details
bj102hs9687,,fixture_sr1,2020-02-12 05:29:37 UTC,2020-02-12 05:29:36 UTC,invalid_checksum,validate_checksums (actual location: fixture_sr1; ) Moab file spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0001/data/metadata/versionMetadata.xml was not found in Moab signature catalog spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0003/manifests/signatureCatalog.xml && CompleteMoab status changed from validity_unknown to invalid_checksum
```

app/services/reporter.rb
mjgiarlo: Aha! Now that I look at how this relation is used and cross-reference it with the database schema, I **now** see that the PreservedObject is primarily for the druid, and that all of the other fields come from the CompleteMoab. How about a name like `complete_moabs_and_preserved_objects_in_storage_root` or some such? This is me saying that I am OK with verbose method names and that I also prefer in-domain language over abbreviations, which is :100: subjective.

app/services/reporter.rb
mjgiarlo: As long as you think this is unlikely to cause heartache or confusion, this is all good. Otherwise, we could do a distinct query but that would open up the can of worms of which row to  prefer over the others.

app/services/reporter.rb
mjgiarlo: I am totally fine with going forward with this and optimizing for performance/robustness later on if it is determined that this is *actually* problematic.

app/services/reporter.rb
mjgiarlo: Do not feel strongly about this.

app/services/reporter.rb
ndushay: I'm thinking this method should be private.  Does that ring true for you?

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
+  end
+
+  # @return [Array] an array of druids on the storage root
+  def druid_csv_list
+    druid_array = [['druid']]
+    moab_storage_root_list_pres_obj_comp_moabs_relation
+      .select(:druid)
       .each_row do |po_hash|
-        druid_array << po_hash['druid']
+        druid_array << [po_hash['druid']]
```
ndushay: add comment  ```# csv needs elements to be arrays``` ?  I didn't find that intuitive, and apparently neither did Aaron.

app/services/reporter.rb
ndushay: "from_storage_root" sucks.  How about "previous storage root"?   And column headers wouldn't need the underscores -- just make's 'em harder to read.

ndushay: I started my own branch with some of these fixes, and I just added the header row directly without assigning it a variable name.  I don't have strong feelings.  <feelings ... nothing more than feelings ...>

app/services/reporter.rb
ndushay: we could add a comment stating `there will only be uniq druid values for a single storage root`

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
+
+    # cols doesn't include storage_root name (available via storage_root ivar).  also we're not instantiating AR objects, so we have to translate
+    # the underlying status enum
+    cols = ['druid', 'from_moab_storage_root_id', 'last_checksum_validation', 'last_moab_validation', 'status AS status_code', 'status_details']
+    data_rows = query.select(cols).each_row.map do |po_cm_hash|
+      [
+        po_cm_hash['druid'], moab_storage_root_name(po_cm_hash['from_moab_storage_root_id']), storage_root.name,
+        po_cm_hash['last_checksum_validation'], po_cm_hash['last_moab_validation'], status_text_from_code(po_cm_hash['status_code']),
+        po_cm_hash['status_details']
+      ]
```
ndushay: if this won't all fit on one line, my preference is to do each one as its own line.  Possibly same for header row for easy visual analogy.

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
+  def write_to_csv(lines, report_type: nil, filename: nil)
+    raise ArgumentError, 'Must specify at least one of report_type or filename' unless report_type.present? || filename.present?
+
+    filename ||= default_filename(filename_prefix: "MoabStorageRoot_#{storage_root.name}_#{report_type}", filename_suffix: 'csv')
     raise "#{filename} already exists, aborting!" if FileTest.exist?(filename)
 
     ensure_containing_dir(filename)
     CSV.open(filename, 'w') do |csv|
       lines.each do |line|
-        line = line.values if line.is_a? Hash
```
ndushay: yes, nice to get rid of this line;  it was confusing.

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
-      preserved_object.complete_moabs.each do |cm|
-        next if errors_only && cm.status == 'ok'
-        detail_array << { druid: druid,
-                          status: cm.status,
-                          status_details: cm.status_details,
-                          last_moab_validation: cm.last_moab_validation,
-                          last_checksum_validation: cm.last_checksum_validation,
-                          storage_root: cm.moab_storage_root.name,
-                          from_storage_root: cm.from_moab_storage_root&.name }
+  def moab_detail_csv_list(errors_only: false)
```
ndushay: i'm not a huge fan of named params (params hash - whatever it's called) when there are a small number and the method is unlikely to be used outside the single coded caller ...

app/services/reporter.rb
```diff
@@ -6,61 +6,80 @@
...
+    # statuses is a hash of the form { 'status_text' => status_code }.  this is just a reverse hash lookup.
+    CompleteMoab.statuses.to_a.find { |s| s[1] == status_code }[0]
   end
 
   # @param [Array] lines - values to output on each line of the csv
   # @param [String] filename - optional filename to override the default
   # @return [String] the name of the CSV file to which the list was written
-  def write_to_csv(lines, filename: nil)
-    filename ||= default_filename(filename_prefix: "MoabStorageRoot_#{storage_root.name}_druids", filename_suffix: 'csv')
+  def write_to_csv(lines, report_type: nil, filename: nil)
```
ndushay: same comment about params hash;  I think positional params are fine for the context of this method.

app/services/reporter.rb
mjgiarlo: @ndushay I think that makes good sense.

mjgiarlo: @ndushay adding to that, I think only these methods should be public:

* `#initialize`
* `#write_to_csv`
* `#druid_csv_list`
* `#moab_detail_csv_list`

The rest could/should be private to constrain the interface of the reporter and hide its implementation details.

app/services/reporter.rb
mjgiarlo: @ndushay :laughing: welllllll, I'm a huge fan of preferring keyword arguments over standard arguments for two reasons:

1. They are not ordered so a dependency is removed: needing to know what order the args are passed in. We still do have the dependency of needing to know what args are required, but that's true of both styles);
1. They are more intention-revealing than standard arguments. How many times have you seen code like `do_the_thing(42, true)` and wondered WTH `42` and `true` mean? So then you have to go trawling in another codebase, potentially, for enlightenment. Compare with `do_the_thing(max_records: 42, skip_accessioned: true)`. This also means we need to be vigilant about coming up with good names, avoiding names like `process` and `data` and such, but that (again) is true in both styles.

app/services/reporter.rb
jmartin-sul: > How about a name like `complete_moabs_and_preserved_objects_in_storage_root` or some such?

i like it, will go w/ that

app/services/reporter.rb
jmartin-sul: did this by adding YARD doc for return type of `#*_csv_list` methods and param type expected by `#write_to_csv`

app/services/reporter.rb
jmartin-sul: i took all the suggestions for privatization, except for `#default_filepath`:  it was easier to leave that public in this follow up because it kept me from having to change the tests at all, and it seems defensible to leave it in the public interface (but that's me rationalizing convenience, and i'm happy to privatize that too, but would prefer to do so in a follow-on PR).

mjgiarlo: I'm not seeing the changes yet, @jmartin-sul. did you push?

app/services/reporter.rb
jmartin-sul: i also prefer keyword args for the reasons @mjgiarlo gives.

app/services/reporter.rb
jmartin-sul: will do this in a follow on PR, if that's ok (sorry, just saw this when combing through to see whether i'd missed any comments -- turns out i did)

jmartin-sul:
@mjgiarlo and @ndushay, i believe i've addressed all of your comments, except for the one i missed from naomi about multi-line lists.

if y'all don't mind, i'll deal with that and the class renaming in a follow on PR.

ndushay:
@jmartin-sul i think you can remove "filename_suffix" from the args passed to default_filename, because we don't ever input one;  it's always ```.csv```.   Another thing for the follow on PR.

app/services/reporter.rb
jmartin-sul: `#default_filepath` will be made private as part of addressing #1367

app/services/reporter.rb
jmartin-sul: addressed in fix for #1367

jmartin-sul:
> i think you can remove "filename_suffix" from the args passed to `default_filename`

will be fixed as part of addressing #1367

PR #1379, opened 2020-02-17
Update dependencies

sul-devops-team:


spec/requests/objects_controller_file_spec.rb
```diff
@@ -65,14 +65,6 @@
...
         end
       end
-
-      context 'when MoabRuntimeError from MoabStorageService' do
-        it 'returns 404 response code; body has additional information' do
-          get file_object_url(id: prefixed_druid), params: { category: 'manifest', filepath: 'foobar' }, headers: valid_auth_header
-          expect(response).to have_http_status(:not_found)
-          expect(response.body).to eq '404 Not Found: manifest file foobar not found for bj102hs9687 - 3'
-        end
-      end
```
jmartin-sul: per a rubocop complaint, this test was entirely duplicative of the one on line 27 of this file:  https://github.com/sul-dlss/preservation_catalog/pull/1379/files#diff-11800c8a982e2c6f9a8eb7c170e7fdfdR27-R33

i liked the description of the other test better in relation to what was being tested, so i kept that one.

there are actually two separate code paths that return a `404` response in the method being tested.  i think it'd be a good idea to test both, but i would suggest ticketing that work for a follow-on PR, since that seems outside the scope of this dependency update.  at first glance, i _think_ it's the `rescue Moab::MoabRuntimeError` at the end of the method, but i'm not 100% sure without tracing more closely through the moab-versioning code underneath.

i don't think this fix actually shrinks test coverage, it just shrinks the range of scenarios we purport to test.

ndushay: follow ticket ok, if it's even worth it.  maybe check code coverage report?

PR #1383, opened 2020-02-19
Remove OKComputerController logger bespoke configuration

### LINKED ISSUE #1371: don't autoload okcomputer during initialization - it's deprecated.
getting this in logs when running rake:

```
DEPRECATION WARNING: Initialization autoloaded the constant OkComputer::OkComputerController.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload OkComputer::OkComputerController, for example,
the expected changes won't be reflected in that stale Class object.

`config.autoloader` is set to `classic`. This autoloaded constant would have been unloaded if `config.autoloader` had been set to `:zeitwerk`.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <top (required)> at /Users/ndushay/sul-dlss-github/preservation_catalog/config/environment.rb:7)
```

jermnelson:
Should fix #1371. Thanks @mjgiarlo for pointing to this [stackoverflow](https://stackoverflow.com/questions/56402093/how-can-i-preload-concerns-in-a-rails-initializer-using-rails-6-zeitwerk/59971682#59971682)  article.

## Why was this change made?
The current method generates a depreciation warning.

n/a

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -5,7 +5,9 @@ require Rails.root.join('config', 'initializers', 'resque.rb').to_s
 
 OkComputer.mount_at = 'status' # use /status or /status/all or /status/<name-of-check>
 OkComputer.check_in_parallel = true
-OkComputer::OkComputerController.logger = ActiveSupport::Logger.new(File.join('log', 'ok_computer_controller.log'))
+Rails.configuration.to_prepare do
+  OkComputer::OkComputerController.logger = ActiveSupport::Logger.new(File.join('log', 'ok_computer_controller.log'))
+end
 
 # check models to see if at least they have some data
 class TablesHaveDataCheck < OkComputer::Check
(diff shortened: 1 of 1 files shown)


ndushay:
cool!  If it works, any reason to keep this a draft PR?

ndushay:
Let's triple check this by testing on pres-cat stage, once we can deploy there again;  looks great to me.

(Just worried about nagios checks in prod)

jcoyne:
May I ask why `OkComputer::OkComputerController` needs it's own logger?

justinlittman:
I'd agree that this should be consistent with every other project.

mjgiarlo:
@jcoyne @justinlittman This is why: https://github.com/sul-dlss/preservation_catalog/pull/1042

And I don't know if we need it, tbh. It's easy enough to `grep`. I too support removing this unless @ndushay or @jmartin-sul have a recollection or preference that might override the call for consistency.

ndushay:
I do not recall.

jmartin-sul:
no objection to removing.  would also not oppose fixing what was previously intended.  since the consensus seems to be removal, i amended the title accordingly, assuming that the removal version is what will get merged.

jmartin-sul:
discussed F2F w/ team at standup just now, merging...

mjgiarlo:
Thanks for being open to the changes, @jermnelson

Issue #1378, opened 2020-02-14
explore using github template for cutover checklist


ndushay:
Do a partial mockup of github template for cutover weekend checklist;  see if we like it better than using spreadsheet.

mjgiarlo:
@ndushay Hi! Anything left to do for the exploration?

ndushay:
@jmartin-sul - I prefer the github template, personally.  Shall we close this ticket?

jmartin-sul:
> @jmartin-sul - I prefer the github template, personally. Shall we close this ticket?

yup, i like the GH issue template, let's go with that, closing this issue...

PR #1384, opened 2020-02-20
Storage migration template

ndushay:
## Why was this change made?

We are experimenting with using a github issue template for storage migration weekend checklists.

Connected to #1378

n/a


ndushay:
@jmartin-sul I will do a follow-on PR, looking carefully at each of your suggestions.  I knew this wasn't perfect but I just wanted to get it in master so we can see how it comes out with github issues ...

PR #1386, opened 2020-02-20
add bug_report template

ndushay:
## Why was this change made?

To have more than the storage-migration-checklist template;  want to make sure the repo doesn't make it hard for non-migration issues to be reported during the next few months.

n/a


ndushay:
very vaguely connected with #1378

PR #1389, opened 2020-02-21
[HOLD] Add OpenAPI spec and validate requests against it

mjgiarlo:
Fixes #1216

## Why was this change made?

Includes:
* Add OpenAPI specification (in YAML format)
* Use Committee gem to validate requests (but not yet responses)
* Validate the OpenAPI specification in CI
* Align ObjectsController documentation with implementation
* Adjust specs now that committee is doing some error handling

This PR could go farther by validating responses in addition to requests, and also by removing some error-checking/validation code paths in the controllers that may now effectively be dead by virtue of the work Committee is doing for us, but I opted to get this PR submitted instead due to its size and complexity.

Note that the [drop in coverage](https://coveralls.io/builds/28888801) is related to :point_up:. I'd like to hear what folks think about this.

## Was the usage documentation (e.g. wiki, README, queue or DB specific README) updated?

yes

PR #1408, opened 2020-02-28
Correct code comment in ObjectsController

mjgiarlo:
Spotted by @ndushay here: https://github.com/sul-dlss/preservation_catalog/pull/1389#discussion_r384666534

## Why was this change made?

Inaccurate documentation is misleading.

Yes, a code comment.

No.

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -6,7 +6,7 @@ require 'csv'
 # ObjectsController allows consumers to interact with preserved objects
 #  (Note: methods will eventually be ported from sdr-services-app)
 class ObjectsController < ApplicationController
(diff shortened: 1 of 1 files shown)


app/controllers/objects_controller.rb
```diff
@@ -6,7 +6,7 @@
 # ObjectsController allows consumers to interact with preserved objects
 #  (Note: methods will eventually be ported from sdr-services-app)
 class ObjectsController < ApplicationController
-  # return the PreservedObject model for the druid (supplied with druid: prefix)
+  # return the PreservedObject model for the druid (supplied *without* `druid:` prefix)
```
ndushay: actually, i believe it works either way.

PR #1415, opened 2020-03-03
AuditResults#events_client now uses namespaced druid

### LINKED ISSUE #1414: report prefixed druids to dor-services-client after audit
**Describe the bug**
Running checksum audits from the rails console in prescat is producing this error:
```
Dor::Services::Client::UnexpectedResponse: Bad Request: 400 ({"errors":
[{"status":"bad_request","detail":"#/components/schemas/Druid pattern 
^druid:[b-df-hjkmnp-tv-z]{2}[0-9]{3}[b-df-hjkmnp-tv-z]{2}[0-9]{4}$ does not match value:
vq518cf9512, example: druid:bc123df4567"}]}) for vq518cf9512
from /opt/app/pres/preservation_catalog/shared/bundle/ruby/2.5.0/gems/dor-services-client-4.16.0/lib/dor/services/client/versioned_service.rb:23:in `raise_exception_based_on_response!'
```

**User Impact**
This may be preventing audits from updating the preservationAuditWF

Run an audit on a druid and you will probably see this error. I've been running 

```Audit::Checksum.validate_druid('vq518cf9512')``` 

and seeing this for a druid I'm remediating.

### PR DESCRIPTION
**Pull request by jmartin-sul**
also refine specs so they'd have caught that and the equivalent issue in WorkflowReporter

see #1414, #1388

## Why was this change made?

fixes #1414 (same issue as #1388, but in the one other spot that sort of call is made)

n/a

If so, please confirm:
- [x] change was tested on stage    and/or
- [ ] ~~test added to sul-dlss/infrastructure-integration-test~~

### DIFF
--- a/app/services/audit_results.rb
+++ b/app/services/audit_results.rb
@@ -222,7 +222,11 @@ class AuditResults
   end
 
   def events_client
-    Dor::Services::Client.object(druid).events
+    Dor::Services::Client.object(namespaced_druid).events
+  end
+
+  def namespaced_druid
+    druid.start_with?('druid:') ? druid : "druid:#{druid}"
   end
 
   def result_code_msg(code, addl=nil)
--- a/spec/services/audit_results_spec.rb
+++ b/spec/services/audit_results_spec.rb
@@ -7,12 +7,15 @@ RSpec.describe AuditResults do
     allow(WorkflowReporter).to receive(:report_error)
     allow(WorkflowReporter).to receive(:report_completed)
     allow(Socket).to receive(:gethostname).and_return('fakehost')
+    allow(Dor::Services::Client).to receive(:object).with("druid:#{druid}").and_return(
+      instance_double(Dor::Services::Client::Object, events: events_client)
+    )
   end
 
   let(:actual_version) { 6 }
   let(:audit_results) { described_class.new(druid, actual_version, ms_root) }
   let(:druid) { 'ab123cd4567' }
-  let(:events_client) { audit_results.send(:events_client) }
+  let(:events_client) { instance_double(Dor::Services::Client::Events) }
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: app/services/audit_results.rb [hunk 1]
```diff
@@ -222,7 +222,11 @@ def send_honeybadger_notification(result)
   end
 
   def events_client
-    Dor::Services::Client.object(druid).events
+    Dor::Services::Client.object(namespaced_druid).events
+  end
+
+  def namespaced_druid
+    druid.start_with?('druid:') ? druid : "druid:#{druid}"
```
jcoyne: Do we need to have this contitional?  Can we just define the interface as accepting one or the other?

jmartin-sul: so would the preference be to modify the dor-services-client code to accept both, and then bump the version of DSC that pres cat uses?

i'd be fine with that.  though i'd also have a slight preference for merging this in the meantime and doing that as follow on work, since this is done, and fixes a production bug.

jcoyne: No, I'm suggesting that `AuditResults` should pick a way and use it and not accept both.

jmartin-sul: ah, gotcha, apologies for misunderstanding.

yeah, so now that you mention it, preservation catalog does consistently use the unprefixed version of the druid internally, throughout the code and in the DB.  so the conditional is actually superfluous safeguarding.  i wasn't thinking when i suggested that approach in #1388, and then i copypasta'd my own bad suggestion.  happy to remove.

thanks for the merge despite the rough edge!  follow on coming shortly...

jmartin-sul: addressed in https://github.com/sul-dlss/preservation_catalog/pull/1418

PR #1436, opened 2020-03-10
fix openapi for checksum

### LINKED ISSUE #1434: Checksum report fails on accessioned objects, error from preservation catalog API
**Describe the bug**
When running a checksum report on a set of accessioned objects, the report will error out.

**User Impact**
Users are unable to do checksum comparisons on large sets of objects in the SDR against those on staging servers.

n/a

Steps to Reproduce, if Known:

Take an example set of objects, i.e. https://argo.stanford.edu/catalog?utf8=%E2%9C%93&q=%22JIRA+%3A+DIGREQ-1585%22

Bulk action --> new bulk action --> checksum report

**Expected behavior**

On the results page, the total objects and success numbers should be identical. Right now, all objects are failing.

**Screenshots**
2020-03-10 16:40:36 UTC Starting ChecksumReportJob for BulkAction 8594
2020-03-10 16:40:36 UTC ChecksumReportJob got error from Preservation Catalog API (Preservation::Client::UnexpectedResponseError): Preservation::Client.http_response got 400 from Preservation at v1/objects/checksums: {"errors":[{"status":"bad_request","detail":"\"Content-Type\" request header must be set to \"application/x-www-form-urlencoded\"."}]}
2020-03-10 16:40:36 UTC Finished ChecksumReportJob for BulkAction 8594

**Additional context**
n/a

peetucket:
## Why was this change made?

to fix #1434 (checksum report failing) likely due to new enforcement in openapi 

no

no

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -216,7 +216,7 @@ paths:
       requestBody:
         required: true
         content:
-          application/x-www-form-urlencoded:
+          application/json:
             schema:
               type: object
               properties:
--- a/spec/requests/objects_controller_checksums_spec.rb
+++ b/spec/requests/objects_controller_checksums_spec.rb
@@ -6,12 +6,12 @@ RSpec.describe ObjectsController, type: :request do
   let(:prefixed_druid2) { 'druid:bz514sm9647' }
   let(:bare_druid) { 'bj102hs9687' }
   let(:bare_druid2) { 'bz514sm9647' }
-  let(:headers) { valid_auth_header.merge('Content-Type' => 'application/json') }
+  let(:post_headers) { valid_auth_header.merge('Content-Type' => 'application/json') }
 
   describe 'GET #checksum' do
     context 'when object found' do
       it 'response contains the object checksum when given a prefixed druid' do
-        get checksum_object_url(prefixed_druid, format: :json), headers: headers
+        get checksum_object_url(prefixed_druid, format: :json), headers: valid_auth_header
         expected_response = [
           { filename: 'eric-smith-dissertation.pdf',
             md5: 'aead2f6f734355c59af2d5b2689e4fb3',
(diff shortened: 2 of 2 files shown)


spec/requests/objects_controller_checksums_spec.rb
```diff
@@ -199,16 +200,16 @@
     context 'when no druids param' do
       context 'when param is empty' do
         it 'body has additional information from the exception if available' do
-          post checksums_objects_url, params: { druids: [], format: :json }, headers: valid_auth_header
+          post checksums_objects_url, params: { druids: [], format: :json }.to_json, headers: post_headers
           expect(response).to have_http_status(:bad_request)
           error_response = JSON.parse(response.body)['errors'].first
-          expect(error_response['detail']).to include('does not match value: , example: druid:bc123df4567')
+          expect(error_response['detail']).to include('druids [] contains fewer than min items')
```
mjgiarlo: Hmmm, curious about this change.

peetucket: came about because openapi is enforcing a minimum number of druids, which is triggering a different response

mjgiarlo: I guess I don't get how this wasn't failing before. I wouldn't expect changing the content-type of the request body to make this start failing.

peetucket: i think there wasn't any enforcement of openapi before at all, so there two failures occurring (content type and druid missing) now

openapi.yml
```diff
@@ -216,7 +216,7 @@ paths:
       requestBody:
         required: true
         content:
-          application/x-www-form-urlencoded:
+          application/json:
```
ndushay: @mjgiarlo !!!!

mjgiarlo: @ndushay yessss? This was another instance where the prescat test suite tested different behavior from the implementation in pres-client, if this is like the other related change. I trusted the test suite. 😁

Issue #1435, opened 2020-03-10
Test effect of turning off accessioning robots on Hydrus (stage/test)


andrewjbtw:
One component of the preservation storage migration will be temporarily turning off the accessioning robots. We should test what effect this may have on the Hydrus interface, which will help us understand the potential service impact for the migration.

To test:

- [ ] Verify that hydrus-test is hooked up all the way through to prescat stage (Andrew to do this)

Once verified that this will be a useful test:

- [ ] stop google books stage from automatically depositing
- [ ] turn off prescat stage
- [ ] turn off preservation robots stage
- [ ] turn off accessioning robots stage
- [ ] Andrew to try to create, edit items in Hydrus

andrewjbtw:
Assigning to myself while I do the initial hydrus testing.

peetucket:
May not be needed given new migration strategy

ndushay:
hydrus is dead.

Issue #1400, opened 2020-02-22
how should removal of preservation catalog entry for a storage root happen?


jmartin-sul:
due to some mis-communication (and an assumption on my part that this would get captured in the manual operations described in https://github.com/sul-dlss/preservation_catalog/blob/master/.github/ISSUE_TEMPLATE/storage-migration-checklist.md), this has fallen through the cracks until now.

**NOTE** this issue is just about deleting the preservation catalog database entry for a storage root from which content has been migrated to new storage.  it is _not_ about actually sunsetting the physical storage or deleting the files from the old file system:  for that discussion, see #1327.

since pres cat does not currently fully support having the same druid multiple times among the moabs it tracks (`CompleteMoab` objects/`complete_moabs` rows), and since we're not immediately deleting content from the storage root from which content is being migrated, we must keep pres cat from knowing about that storage root anymore.  that way, it doesn't e.g. try to run M2C, and re-discover an old copy of a moab that's been migrated to a new storage root.

i had been assuming that this would be a manual process:  the person doing the migration (running through the above linked checklist) would:
* run the migration rake task
* make sure everything had been moved successfully
* make sure nothing still referenced the old storage root
* delete the database entry for the old `MoabStorageRoot`, as well as the corresponding shared configs entry.  note that db seeding is entirely additive, so removing an entry from shared_configs changes will not automatically remove it from the database.

the other option would be to have the rake task which migrates moabs from an old storage root also delete that old storage root's entry from preservation catalog's database.  i am wary of this option because:
* i _think_ i'd want the rake task to do the move in one transaction instead of a `#find_each`
* i'd definitely want the rake task to do some other integrity checking to make sure that everything had been moved, and that things were present on the file system as expected.  if the rake task weren't made more transactional, i think it'd have to query to see if it had successfully moved everything, but that sounds prone to error and the need for continually re-polling.
  * if we used the facilities of the `postgresql_cursor` gem that was leveraged for reporting, we probably could replace the `#find_each` with something that was entirely transactional but still performant.

still, for this migration, i think it'd be less effort to have a human use a checklist to do those things manually than it would be to code that into the rake task.

but i think this merits some discussion with more team members.  i'm certainly open to arguments that run counter to my current preference.

jmartin-sul:
I ended up testing a different option when playing the part of ops for our dry run in #1420/#1377.

[...]

PR #1438, opened 2020-03-11
update storage migration checklist

### LINKED ISSUE #1417: storage migration checklist needs to include updates to technical-metadata-service storage roots
.github/ISSUE_TEMPLATE/storage-migration-checklist.md

in "After Migration Success" section,

after updating shared_configs for preservation-catalog and preservation-robots, we will also need to update shared_configs for technical-metadata-service, which also reads data from our storage roots.

please add the appropriate info.

ndushay:
## Why was this change made?

Fixes #1417, more or less.

Questions (could become new tickets or could make changes here):
- should we be turning off all accessioning workers?
- should we be turning off technical-metadata-service workers? (reads preservation storage roots)
- should we *not* do PRs for shared_configs for migrations?
- should we consider a single checklist for multiple roots, instead of assuming a separate one for each root?
- should we consider a much terser checklist with an accompanying document?  This is overwhelming and hard to edit.

na

na

### DIFF
--- a/.github/ISSUE_TEMPLATE/storage-migration-checklist.md
+++ b/.github/ISSUE_TEMPLATE/storage-migration-checklist.md
@@ -66,13 +66,13 @@ Run all validation checks on Moabs and generate reports.  Note that error detail
   - preservation-catalog: https://app.honeybadger.io/projects/54415/faults?q=-is%3Aresolved+-is%3Aignored
   - preservation_robots: https://app.honeybadger.io/projects/55564/faults?q=-is%3Aresolved+-is%3Aignored
 
-### Negotiate with Andrew/Julian as to how early on Friday you can do the Following:
+### Negotiate with Andrew/Julian as to how early on (Friday) you can do the Following:
 
 - [ ] get list of druids on storage root ```RAILS_ENV=production bundle exec rake prescat:reports:msr_druids[storage_root_name,druids_b4]```
 
 - [ ] (Ops) turn off nagios alerts for preservation_robots prod boxes
 
-- Shut Down Preservation Robots Workers
+- **Shut Down Preservation Robots Workers**
   - [ ] look for running preservation robots workers in the resque GUI:  https://robot-console-prod.stanford.edu/workers
   - [ ] shut off the workers manually by killing (only) the `resque-pool-master` process (`hotswap` interferes with other cap resque commands)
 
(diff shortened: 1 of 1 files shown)


jmartin-sul:
> Questions (could become new tickets or could make changes here):
> 
>     * should we be turning off all accessioning workers?
> 

i think so.  it felt like we were leaning that way after yesterday's meeting, and that seems to be implied [here](https://github.com/sul-dlss/preservation_catalog/issues/1435):  "One component of the preservation storage migration will be temporarily turning off the accessioning robots."

can @andrewjbtw give a final confirmation on that?  see #1439 for the question and the actual checklist editing if the answer to the question is "yes."

>     * should we be turning off technical-metadata-service workers? (reads preservation storage roots)
> 

i think it's safe to leave them on, because:
* the moabs aren't disappearing, and the NFS shares that hold the old copies shouldn't be destroyed immediately, just unmounted from pres cat so it doesn't see the same thing in multiple locations at once (see https://github.com/sul-dlss/preservation_catalog/issues/1400#issuecomment-596034278).
* if the migration is successful, the same content should be available in the new location.
* presumably, for the tech MD workers to see the new storage location after it's added to the tech MD service configs, the tech MD service will be re-deployed with the updated config, and thus restarted.  so even the tech MD workers shouldn't have access to the same moab in two diff locations simultaneously.
* tech MD workers also don't ask preservation catalog for storage location, they just use the moab-versioning gem to walk the storage roots.  so there should be no issue with pres cat giving a soon-to-be-stale answer right before an individual moab's entry is migrated in the catalog.

all of that together makes me think that tech MD workers are fine to leave on, and should not see things in an inconsistent state at any point while running.

>     * should we _not_ do PRs for shared_configs for migrations?
> 

i think it'd be nice to get a second set of eyes on them if possible, but i wouldn't necessarily propose making that second set of eyes work over the weekend when they otherwise wouldn't have.  i think our rule of thumb from last week's infra dev meeting should hold, about it being ok in to auto-merge shared_configs PRs or commit straight to the relevant branch, when dealing with exigent circumstances.  PR to this effect coming shortly.

>     * should we consider a single checklist for multiple roots, instead of assuming a separate one for each root?
> 

i had thought about that briefly yesterday, but i think i'm in favor of keeping this checklist per root format.  reasons:
* this is now a pretty decently tested document, and re-working to pluralize risks introducing regressions, i think.  and i'd prefer to not re-test the whole thing.  _on the other hand_:  it hasn't been tested with multiple instances for multiple roots at once...
* _but_, i suspect that it'd be easy enough to spin up a checklist for each root, and if a task only has to be done once, it ends up getting checked off as a no-op on the lists for roots 2 through N.  _on the other hand_:  maybe what only has to be done won't be obvious?  i suspect it will, but maybe that warrants a read-through looking for that.
* a document that could accommodate multiple roots would require more editing of the checklist (each spot that dealt with a single root might now deal with N roots).  that seems error prone, and i think goes against the guiding principle of keeping it as simple as possible for the people sheparding the migration on the weekend.

i dunno, i have some uncertainty, and do not feel strongly about my conclusion, but i currently vote for keeping the wording singular, and spawning a ticket for each root.

i would be happy to leave this as-is, but if you think further investigation/discussion is warranted, we can ticket a bit of investigation work.  i'm happy to pick that up.

>     * should we consider a much terser checklist with an accompanying document?  This is overwhelming and hard to edit.

i like this format as-is.  understood about it getting overwhelming, but i'd rather push through that (and it's fine if that means more of the writing/editing falls to me).  per your original preferences, i think it should be as easy as possible to go through the process on the migration weekends, and i think multiple linked documents would make it harder to see progress and tasks completed easily in one place.  i'd rather have the document writing be the high cognitive load activity (as opposed to the migration).

andrewjbtw:
Yes, at our Monday storage meeting we thought turning off the accessioning workers made sense, as they interact with preservation to do things like get current version, content diff for content metadata and a few other things.

PR #1445, opened 2020-03-16
Add cap stage for storage migration instance

tallenaz:
## Why was this change made?

To deploy to the instance of prescat we will be using in the storage migration

n/a -- this is a transient instance that will go away once the migration is complete

no

### DIFF
--- a/config/deploy/migrate.rb
+++ b/config/deploy/migrate.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+server 'preservation-catalog-migrate.stanford.edu', user: 'pres', roles: %w[app db web resque queue_populator]
+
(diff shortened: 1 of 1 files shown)


config/deploy/migrate.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+server 'preservation-catalog-migrate.stanford.edu', user: 'pres', roles: %w[app db web resque queue_populator]
```
jmartin-sul: this pres cat instance will have its own db and redis instances, right?

jmartin-sul: (i.e. it won't share with prod or stage?)

tallenaz: That's correct, the vm has its own db and redis instance already configured.

PR #1446, opened 2020-03-17
tweak okcomputer check for number of workers

ndushay:
## Why was this change made?

Because hot_swap for resque workers can temporarily allow extra workers, most often due to long running jobs.

This change does imply we won't get nagios failures if prescat Resque has dead workers ... how often does that happen?

na

only in that fewer spurious nagios alerts are likely to be raised.

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -82,12 +82,19 @@ end
 # check for the right number of workers
 class WorkerCountCheck < OkComputer::Check
   def check
-    count = Settings.total_worker_count
-    if Resque.workers.count == count
-      mark_message "#{count} workers are up."
+    expected_count = Settings.total_worker_count
+    actual_count = Resque.workers.count
+    message = "#{actual_count} out of #{expected_count} expected workers are up."
+    if actual_count == expected_count
+      mark_message message
+    elsif actual_count > expected_count
+      # this can happen briefly with hot_swap when deploying,
+      #   but if it persists, there is likely a problem (or a big file being uploaded to cloud)
+      mark_failure
+      mark_message "TOO MANY WORKERS: #{message}"
     else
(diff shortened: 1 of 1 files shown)


config/initializers/okcomputer.rb
```diff
@@ -82,12 +82,14 @@ def check
 # check for the right number of workers
 class WorkerCountCheck < OkComputer::Check
   def check
-    count = Settings.total_worker_count
-    if Resque.workers.count == count
-      mark_message "#{count} workers are up."
+    expected_count = Settings.total_worker_count
+    actual_count = Resque.workers.count
+    message = "#{actual_count} out of #{expected_count} expected workers are up."
+    if actual_count >= expected_count # this can happen with hot_swap when deploying
```
jmartin-sul: i would somewhat prefer `mark_failure` in this code path too, esp since @julianmorley said he was ok with some alerts right after deployment when we expect this to happen.  i mean what i really want is something like "there have been too many workers for more than a day", but that seems like much more work to implement.  but i'd hate to not get alerted if we had too many workers lingering for more than a few hours, because that'd be unusual, and having stale workers has bitten us (in hard to debug ways) in the past.

i think julian was more interested in differentiated messages than in not alerting in the case at all, because then it's at least easy to decide whether to ignore the alert at a glance (totally fine to ignore such an alert at 3 pm on a monday -- less so at 3 pm on a thursday, when we're not right after a dep update deployment, and all the jobs from that time should've wound down).

julianmorley: Yeah, exactly that. 7 workers where we expect 5 implies stale workers after a deploy; 3 workers where we expect 5 indicates something else.

ndushay:
@jmartin-sul @julianmorley I have revised the check per your desires.  :-)

PR #1448, opened 2020-03-18
make moab migration method public, document usage

### LINKED ISSUE #1447: moved Moabs cause preservationIngestWF and preservationAuditWF errors
**Describe the situation**
As part of the migration of data from SDR1 to current preservation storage (the SDR services disks), Moabs are being moved from storage roots that don't have available space to roots that do have available space. This is being done this way because

1. The druids were created many years ago and are already in use for PURLs displaying JP2s
2. The preservation master TIFs were not migrated at that time
3. In the intervening years the storage roots have filled to a point where the TIFs cannot all be added to the druids in their original locations

The migration workflow was set up last year to follow these steps:

1. Move an existing Moab to a storage root with enough space
2. Accession the preservation files (usually one big tar.gz) to that Moab
3. Update prescat with the new location

However, these new accessions are hitting errors

1. at `update-catalog` in the `preservationIngestWF`, probably because prescat needs to know the new location before this step can be completed
2. at `preservation-audit` in the `preservationAuditWF`, probably again because prescat needs to know about the new location for an audit to pass

In order for the migration to move forward, we need to:

1. Work out how to update prescat so that in-progress items can keep moving through the workflows
2. Determine whether to change the order of the SDR1 migration steps given changes in the preservation system in the time since the SDR1 migration workflow was set up.

**Important additional context**

Items caught in the preservationIngestWF remain on the `/dor` workspace without being cleaned up until accessioning is complete. This puts a higher priority on at least getting items past that workflow step because there is a risk of filling up the `/dor` workspace with in-process migrated content.

jmartin-sul:
## Why was this change made?

to support manual migration of individual moabs between storage roots.

closes #1447 (per https://github.com/sul-dlss/preservation_catalog/issues/1447#issuecomment-600355906)

yes

~~If so, please confirm change was tested on stage and/or test added to sul-dlss/infrastructure-integration-test.~~

n/a, just makes it easier to leverage functionality from rails console that was previously a private method for a manually run rake task.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Rails application to track, audit and replicate archival artifacts associated wi
     * [Catalog to Moab](#c2m) (C2M) existence/version check
     * [Checksum Validation](#cv) (CV)
     * [Seed the catalog](#seed-the-catalog-with-data-about-the-moabs-on-the-storage-roots-the-catalog-tracks----presumes-rake-dbseed-already-performed)
+    * [Update the Catalog Because a Moab Moved](#migrate_moab_manually)
 * [Development](#development)
     * [Dockerized Development](#docker)
 * [Deploying](#deploying)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -290,6 +291,26 @@ Or for all roots:
 MoabStorageRoot.find_each { |msr| Audit::MoabToCatalog.seed_catalog_for_dir(msr.storage_location) }
 ```
 
+## <a name="migrate_moab_manually"> Update the catalog when moving a Moab to a different storage root
+
+Sometimes it's necessary to move Moabs from one storage root to another, either in bulk as part of a storage hardware migration, or manually for a small number, as when an existing Moab might get too much additional content for the storage root it's currently on.
+
+There are rake tasks and documentation to suppor the bulk migration scenario.  See the [migration issue template](.github/ISSUE_TEMPLATE/storage-migration-checklist.md).
```
ndushay: suppor is typo.

jmartin-sul: suppor catch , thanks! 😄

jmartin-sul: just pushed a fix, should be the only change:
```
$ git diff origin/move-moab-migration-method-avoid-alliteration-assiduously 
diff --git a/README.md b/README.md
index 13875e00..941a3623 100644
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ MoabStorageRoot.find_each { |msr| Audit::MoabToCatalog.seed_catalog_for_dir(msr.
 
 Sometimes it's necessary to move Moabs from one storage root to another, either in bulk as part of a storage hardware migration, or manually for a small number, as when an existing Moab might get too much additional content for the storage root it's currently on.
 
-There are rake tasks and documentation to suppor the bulk migration scenario.  See the [migration issue template](.github/ISSUE_TEMPLATE/storage-migration-checklist.md).
+There are rake tasks and documentation to support the bulk migration scenario.  See the [migration issue template](.github/ISSUE_TEMPLATE/storage-migration-checklist.md).
 
 When the need for moving a single Moab arises, the repository manager or a developer should:
 1. [on the file system] Move the Moab to a storage root with enough space
```

jmartin-sul:
merging w/ the typo fix, and @andrewjbtw can give the instructions a look later and file an issue if he thinks something needs tweaking.

Issue #1439, opened 2020-03-11
migration checklist:  to shut down accessioning robots or not?


jmartin-sul:
assuming we do want to shut down the accessioning robots, flesh out the instructions here with actual step by step instructions:  https://github.com/sul-dlss/preservation_catalog/pull/1438/files#diff-a9539480162943c784e6e055fb34c87eR95-R97

assigning to @andrewjbtw for confirmation.  if it turns out the answer's no, feel free to close.  otherwise, you can assign it to me, and i will either pick it up if free, or move it back to the ready column if i'm not yet able to pick it up.

jmartin-sul:
@andrewjbtw confirmed here:  https://github.com/sul-dlss/preservation_catalog/pull/1438#issuecomment-598290915

un-assigning andrew, i can write instructions.

jmartin-sul:
after much discussion with @andrewjbtw and @ndushay over the last couple weeks, and some discussion w/ the wider infra team, we feel that it's safe to leave the accessioning robots on, while leaving the preservation robots shut down.  the preservation robots should be the only service outside of pres cat that asks pres cat to update itself, and should also be the only service that manipulates moabs in their storage roots -- all other service communication to pres cat should be read-only, which should be safe, since those reads are for things like getting current version, getting preserved file content, etc.  during the period where moabs location is in flux from pres cat's perspective, serving preserved file content might be flakey, but reads to get current version should be safe.

remaining work for this ticket:
- [ ] amend pres cat's REST route disabling migration branch so that it doesn't disable routes for reading from pres cat (but writes for updating the catalog will be left disabled in that branch).
- [ ] update migration checklist accordingly.

during our dry run of the migration in early april, i'd suggest that we try a small accessioning job while pres cat is in this weird read-only state, to prove to ourselves that this will work out ok.  we expect/hope that things will just harmlessly pile up at the `preservationIngestWF`, waiting for pres robots to be re-enabled.  we don't expect any other robots-related errors, but would hope that any we encounter can be resolved by retrying the errored step.  we shall see during the test run.

either way, @andrewjbtw will message to users that accessioning should be halted for the migration weekend, to prevent large jobs from coming through and filling up the DOR workspace.  we feel that we have less control over e.g. self-deposit activity, and are worried about what might happen if people try to use that while accessioning is fully shut off.

corrections welcome if any of that seems wrong.

jmartin-sul:
[...]

PR #1452, opened 2020-03-24
remove reference to disabling accessioning workers for storage migrations

jmartin-sul:
## Why was this change made?

turns out we don't think we need to do that, see https://github.com/sul-dlss/preservation_catalog/issues/1439#issuecomment-600759173

just to be extra safe, i'd appreciate it if someone familiar with how the robots pick up work would review what i found in this comment before merging this, to make sure my understanding is correct:  https://github.com/sul-dlss/preservation_catalog/issues/1439#issuecomment-603078033

connects to #1439

documentation only

~~If so, please confirm change was tested on stage and/or test added to sul-dlss/infrastructure-integration-test.~~

n/a

### DIFF
--- a/.github/ISSUE_TEMPLATE/storage-migration-checklist.md
+++ b/.github/ISSUE_TEMPLATE/storage-migration-checklist.md
@@ -94,12 +94,6 @@ Run all validation checks on Moabs and generate reports.  Note that error detail
     ```
   - [ ] ensure there are no running preservation robots workers in the resque GUI:  https://robot-console-prod.stanford.edu/workers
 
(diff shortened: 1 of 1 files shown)


ndushay:
this looks fine, but Is there another spot in the template about starting them up again?

jmartin-sul:
> this looks fine, but Is there another spot in the template about starting them up again?

not that i saw.  i searched for `accession`, and the only matches were in the removed text.  did a read through and didn't notice any other relevant sections.  i think since it was just a placeholder to be fleshed out, there wasn't much to remove.

Issue #1420, opened 2020-03-04
[STAGE TEST] Old storage root services-disk-stage1 to New storage root at /services-disk-stage/store-100/


jmartin-sul:
**Goal: Migrate This Storage Root Without Introducing New Errors**

- Use this template for a single storage root.
- If multiple roots will be migrated in one weekend, create a separate github issue from this template for each root being migrated.
  - provide a link here to the github issue for the other root's migration.

Note that many old storage roots will migrate to each of the new storage bricks, so not all weekend migrations will be introducing a new storage brick into the ecosystem.

## Week Before Cutover Weekend

Ops will be ensuring all the data on the old root has been migrated to the new brick.

Run all validation checks on Moabs and generate reports.  Note that error details change with different checks, so we want to run these checks sequentially, not simultaneously, and we want error reports generated after each check completes, before the next one starts.

#### M2C
- [x] run on storage root: ```RAILS_ENV=production bundle exec rake prescat:audit:m2c[stor_root_name]```
  - [x] requeue failed jobs / ensure no jobs failed via resque GUI https://preservation-catalog-prod-01.stanford.edu/resque/overview
- [x] after M2C finishes, generate report for objects with error status ```RAILS_ENV=production bundle exec rake prescat:reports:msr_moab_audit_errors[stor_root_name,m2c_b4]```
- [x] note druids for each existing M2C error (from report in /opt/app/pres/preservation_catalog/current/log/reports)
  - [x] Ensure there is a github issue in preservation_catalog for each error/object (mult objects with same error can be in same github issue.)

#### C2M
- [x] run on storage root ```RAILS_ENV=production bundle exec rake prescat:audit:c2m[stor_root_name]```
  - [x] requeue failed jobs / ensure no jobs failed via resque GUI https://preservation-catalog-prod-01.stanford.edu/resque/overview
- [x] after C2M finishes, generate report for objects with error status ```RAILS_ENV=production bundle exec rake prescat:reports:msr_moab_audit_errors[stor_root_name,c2m_b4]```
- [x] note druids for each existing C2M error (from report in /opt/app/pres/preservation_catalog/current/log/reports)
  - [x] Ensure there is a github issue in preservation_catalog for each error/object (mult objects with same error can be in same github issue.)

#### CV
- [x] run on storage root ```RAILS_ENV=production bundle exec rake prescat:audit:cv[stor_root_name]```
[...]

PR #1453, opened 2020-03-24
re-enable read-only routes

jmartin-sul:
## Why was this change made?

per https://github.com/sul-dlss/preservation_catalog/issues/1439#issuecomment-600759173 we won't shut down accessioning robots for migration, which means that we'll need to enable the routes that allow accessioning robots to talk to pres cat for their read-only requests.

we'll leave disabled the create and update endpoints, but only the preservation robots should use those (and the preservation robots will still be disabled).

connects to #1439

yes, see #1452

we didn't shut down the accessioning robots in the [run through on stage](#1420) (the instructions removed here were added after that).  and we plan to test this again in a dry run the week of april 3.  so this has been tested and should get tested again soon.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,18 +7,18 @@ Rails.application.routes.draw do
         at: '/resque',
         constraints: ->(req) { Settings.resque_dashboard_hostnames.include?(req.host) }
 
-  # scope 'v1' do
-  #   resources :catalog, param: :druid, only: %i[create update]
-  #
-  #   resources :objects, only: %i[show] do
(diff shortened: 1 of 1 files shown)


jmartin-sul:
test failure is expected since we're disabling the write routes without disabling corresponding tests.  note that this isn't being merged to master, this is just a temp branch for use on migration weekends.

jmartin-sul:
merging this into the migration weekend branch...

ndushay:
yup - LGTM

PR #1455, opened 2020-03-24
Remove id and preservation_policy_id from object response

### LINKED ISSUE #1412: filter internal IDs out of ObjectsController#show responses
currently we just `render json: PreservedObject.find_by!(druid: druid).to_json` in that method, which means the `PreservedObject`'s `id` and `preservation_policy_id` fields get returned.  since we don't let clients retrieve objects by their internal primary keys (and probably have no reason to in the future), we should stop leaking this info.  likely harmless, but also better not to, i think.

And don't forget to update `openapi.yml`.

aaron-collier:
## Why was this change made?

Fixes #1412 by removing the id and preservation_policy_id from the show response.

N/A

No.

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -9,7 +9,10 @@ class ObjectsController < ApplicationController
   # return the PreservedObject model for the druid (supplied *without* `druid:` prefix)
   # GET /v1/objects/:id.json
   def show
-    render json: PreservedObject.find_by!(druid: druid).to_json
+    obj = JSON.parse(PreservedObject.find_by!(druid: druid).to_json)
+    obj.delete("id")
+    obj.delete("preservation_policy_id")
+    render json: obj.to_json
   end
 
   # return a specific file from the Moab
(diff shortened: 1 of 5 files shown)


app/controllers/objects_controller.rb
```diff
@@ -9,7 +9,10 @@ class ObjectsController < ApplicationController
   # return the PreservedObject model for the druid (supplied *without* `druid:` prefix)
   # GET /v1/objects/:id.json
   def show
-    render json: PreservedObject.find_by!(druid: druid).to_json
+    obj = JSON.parse(PreservedObject.find_by!(druid: druid).to_json)
```
mjgiarlo: How would you feel about moving the logic in these lines into the `PreservedObject` model? So, something like:

```ruby
  def as_json(*)
    super.except('id', 'preservation_policy_id')
  end
```

mjgiarlo:
@aaron-collier uh, of course now I think... should this PR also remove these attrs from `openapi.yml`? Open to a follow-on?

jmartin-sul:
> should this PR also remove these attrs from `openapi.yml`?

good call, imo

PR #1481, opened 2020-04-10
ChecksumValidator#validate_checksums should check existence of moab dir first

jmartin-sul:
## Why was this change made?

to make handling of wholly missing moab directories more robust when CV encounters druids like that (so we get an appropriate `status` change on the `CompleteMoab` instead of just a honeybadger error).

when first making the test changes (without the app changes in this PR), i was able to reproduce the backtrace seen in #1480.

when testing this app code change on pres cat's migrate instance, the `ChecksumValidator#validate_checksums` method now correctly updates `CompleteMoab#status` to `online_moab_not_found` if the combo of `druid` (via related `PreservedObject`) and `storage_location` (via related `MoabStorageRoot`) points to a moab directory that doesn't exist on the file system.  previously, this scenario would result in the uncaught error noted in #1480.

n/a

n/a

connects with #1480

### DIFF
--- a/app/services/checksum_validator.rb
+++ b/app/services/checksum_validator.rb
@@ -24,6 +24,16 @@ class ChecksumValidator
   end
 
   def validate_checksums
+    # check first thing to make sure the moab is present on disk, otherwise weird errors later
+    unless File.exist?(object_dir)
+      transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
+        mark_moab_not_found
+        complete_moab.save!
+      end
+      results.remove_db_updated_results unless transaction_ok
+      return results.report_results(Audit::Checksum.logger)
+    end
+
     validate_manifest_inventories
     validate_signature_catalog
 
--- a/app/services/checksum_validator.rb
+++ b/app/services/checksum_validator.rb
@@ -25,14 +25,7 @@ class ChecksumValidator
 
   def validate_checksums
     # check first thing to make sure the moab is present on disk, otherwise weird errors later
-    unless File.exist?(object_dir)
-      transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
-        mark_moab_not_found
-        complete_moab.save!
-      end
-      results.remove_db_updated_results unless transaction_ok
-      return results.report_results(Audit::Checksum.logger)
-    end
+    return results.result_array unless check_file_existence!
 
     validate_manifest_inventories
     validate_signature_catalog
@@ -58,6 +51,21 @@ class ChecksumValidator
     results.remove_db_updated_results unless transaction_ok
 
     results.report_results(Audit::Checksum.logger)
+  end
+
+  # return true and exit if the moab exists.  otherwise update status to missing on complete_moab, report results, and return false.
+  # @return [Boolean] true and exit if the moab exists, false otherwise
+  def check_file_existence!
+    return true if File.exist?(object_dir)
+
+    transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
+      mark_moab_not_found
+      complete_moab.save!
+    end
+    results.remove_db_updated_results unless transaction_ok
+    results.report_results(Audit::Checksum.logger)
+
+    false
   end
 
   def validate_manifest_inventories
--- a/app/services/moab_validation_handler.rb
+++ b/app/services/moab_validation_handler.rb
@@ -64,6 +64,13 @@ module MoabValidationHandler
     complete_moab.status_details = results.results_as_string(results.result_array)
   end
 
+  def mark_moab_not_found
+    results.add_result(AuditResults::MOAB_NOT_FOUND,
+                       db_created_at: complete_moab.created_at.iso8601,
+                       db_updated_at: complete_moab.updated_at.iso8601)
+    update_status('online_moab_not_found')
+  end
+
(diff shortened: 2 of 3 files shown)


jmartin-sul:
build failure is due to rubocop complaint about method length, will look at a bit of light refactoring on the overlong method.

spec/services/checksum_validator_spec.rb
```diff
@@ -207,6 +207,24 @@
   end
 
   describe '#validate_checksums' do
+    context 'moab is missing from storage' do
+      before do
+        # fake a moab gone missing by updating the preserved object to use a non-existent druid
+        comp_moab.preserved_object.update(druid: 'tr808sl1200')
```
julianmorley, line 213: Not a valid druid, spec should fail on invalid format. :-)

jmartin-sul: heh, good catch! (and also 🤦 that my music nerd easter egg wasn't a proper druid afterall -- twas late at night when i wrote this test)

um, will investigate and maybe ticket.  i thought we validated this?  https://github.com/sul-dlss/preservation_catalog/blob/master/app/models/preserved_object.rb#L16

maybe my expectations about AR hooks are off.

jmartin-sul: @julianmorley actually, after looking back at other druids and this one, i tentatively retract my facepalm =)

what's invalid about this one?  2 letters, 3 numbers, 2 letters, 4 numbers.  no vowels.  what's the salient difference from e.g. `bz514sm9647` below?

jmartin-sul: for a moment i was thinking the third section is supposed to be 3 letters, but it is supposed to be 2, right?  the canonical `DruidTools::Druid.pattern` is what's used to validate.

ndushay: no L

ndushay: strict druids don't allow vowels or L

ndushay: https://github.com/sul-dlss/druid-tools/blob/master/lib/druid_tools/druid.rb#L14-L16

ndushay:
I would be ok with allowing rubocop  auto-config do its thing and making a ticket to refactor, if there is no obvious refactor suggesting itself.

app/services/checksum_validator.rb
```diff
@@ -24,6 +24,16 @@ def initialize(complete_moab)
...
+    # check first thing to make sure the moab is present on disk, otherwise weird errors later
+    unless File.exist?(object_dir)
+      transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
+        mark_moab_not_found
+        complete_moab.save!
+      end
+      results.remove_db_updated_results unless transaction_ok
+      return results.report_results(Audit::Checksum.logger)
+    end
+
```
ndushay, line 36: I guess this could be a method?

app/services/moab_validation_handler.rb
```diff
@@ -64,6 +64,13 @@ def update_status(new_status)
     complete_moab.status_details = results.results_as_string(results.result_array)
   end
 
+  def mark_moab_not_found
+    results.add_result(AuditResults::MOAB_NOT_FOUND,
+                       db_created_at: complete_moab.created_at.iso8601,
+                       db_updated_at: complete_moab.updated_at.iso8601)
+    update_status('online_moab_not_found')
+  end
```
ndushay, line 72: lovin' this refactor

spec/services/checksum_validator_spec.rb
jmartin-sul: ah, didn't realize `L` is disallowed!  thanks.  will change the example here, and will do a bit of investigation into why the validation isn't throwing an error as expected.  will timebox and ticket that if it looks like a rabbit hole.

app/services/moab_validation_handler.rb
jmartin-sul: i was a bit hesitant to further embrace a module pattern we're not wild about in retrospect, but this did feel more right than copypasta-ing the same code into a new place, since refactoring away from the module isn't on the table atm.

spec/services/checksum_validator_spec.rb
ndushay: you have to tell it to be strict.  had to do it that way for backwards compatibility.

got an incredible amount of flack from a former developer about "where did these druid requirements come from".

julianmorley: Well tbf it's not like we could point to an actual spec. :)

jmartin-sul: ok, i believe i fixed this example to conform to the strict version of the druid pattern, and i left it at that since we purposefully validate with the looser version for backwards compatibility.

app/services/checksum_validator.rb
jmartin-sul: yup, went with that approach

jmartin-sul:
@ndushay 
> I'm marking approve, pending green.

merging since the refactoring i made was pretty small.  was thinking similarly to you on the possible refactor to make rubocop happy, and that worked out nicely.

happy to circle back and make changes if you have any qualms about the touchup commit:  https://github.com/sul-dlss/preservation_catalog/pull/1481/commits/eb2e00ac87b8431bb06d8c759868db95eede9696

PR #1486, opened 2020-04-15
update short storage migration checklist re: when to run CV and C2M

jmartin-sul:
in light of the fact that checksum validation of all migrated moabs will take weeks, adjust the instructions to:
* remove it from the "week before" procedure
* move it to the end of the migration checks, in such a way that it can wind down in the weeks following the migration
* replace its immediate use before and after the migration with the much speedier C2M check (which just ensures that the moab directories exists at the expected locations, without checksum validating their contents).

also notes about why this change is not as scary as it seems at first glance.

### DIFF
--- a/.github/ISSUE_TEMPLATE/storage-migration-checklist-short.md
+++ b/.github/ISSUE_TEMPLATE/storage-migration-checklist-short.md
@@ -38,14 +38,6 @@ Run all validation checks on Moabs and generate reports.  Note that error detail
 - [ ] note druids for each existing C2M error (from report in /opt/app/pres/preservation_catalog/current/log/reports)
   - [ ] ? Ensure there is a github issue in preservation_catalog for each error/object (mult objects with same error can be in same github issue.)
 
-### CV (checksum validation of all cataloged moabs)
-- [ ] run on storage root ```RAILS_ENV=production bundle exec rake prescat:audit:cv[stor_root_name]```
-  - [ ] requeue failed jobs / ensure no jobs failed via resque GUI https://preservation-catalog-prod-01.stanford.edu/resque/overview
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/storage-migration-checklist-short.md
```diff
@@ -38,13 +38,9 @@ Run all validation checks on Moabs and generate reports.  Note that error detail
...
-- [ ] note druids for each existing CV error (from report in /opt/app/pres/preservation_catalog/current/log/reports)
-  - [ ] ? Ensure there is a github issue in preservation_catalog for each error/object (mult objects with same error can be in same github issue.)
+### CV -- we won't run this anew before the migration
+
+We had originally planned to run all objects to be migrated through checksum validation the week before the cutover.  Testing indicates that this won't be possible in the time we have with the computational and storage resourcing we have.  Additionally, checksum validation has been running as expected on production (as of 2020-04-15, our least recently checksum validated object on production was checked at 2020-01-13 11:03:05 UTC).  We have other safeguards and assurances in the migration process and in our normal workflow that should give peace of mind re: integrity of preserved objects (C2M before and after the move, FS block level integrity checking built into the `rsync` invocation that copies content from old to new storage, FS block level integrity checking on the originating NetApp filer).
```
ndushay, line 43: this is meant to be a very checklist oriented, without a lot of text blobs.  Maybe move this to the wiki?

jmartin-sul: https://github.com/sul-dlss/preservation_catalog/wiki/Storage-Migration---Additional-Information#cv----why-we-wont-run-this-anew-right-before-the-migration

PR #1522, opened 2020-05-06
send current moab version in CV audit results, test adjustment/cleanup

jmartin-sul:
also

* fix one test to expect moab version to get passed to workflow reporter (for passing along if/when creating workflow, which is the point of this fix)
* take two other tests broken by app code change, consolidate into one test that's more behavior focused and less focused on underlying implementation

## Why was this change made?

connects to #1515 (and hopefully resolves the vast majority of occurrences of that issue)

note there are still some situations where audit won't send a version along
* C2M detects that the `PreservedObject` and `CompleteMoab` record different versions
* no moab is found on disk
* any time `PartReplicationAudit` detects an error that it reports to workflows, though this is rare, and `PartReplicationAudit` doesn't currently concern itself with the on premises moab, so i am hesitant to figure out logic for that now, since we hope to further disentangle from workflows for pres cat reporting anyway.

but i think this should be an improvement over what we have on master now, and should resolve the acute issue of eating up our honeybadger quota with all the workflow server deprecation warnings.

N/A

not really... it does send `version` to workflows for a call where we weren't previously sending it, but that's a thing we already did in other audits, and this usage feels very similar and safe.  that said, happy to test in stage anyway, and/or happy to try a small handful of CV audits on prod before re-queuing the entire remainder.

### DIFF
--- a/app/services/checksum_validator.rb
+++ b/app/services/checksum_validator.rb
@@ -20,7 +20,7 @@ class ChecksumValidator
   def initialize(complete_moab)
     @complete_moab = complete_moab
     @bare_druid = complete_moab.preserved_object.druid
-    @results = AuditResults.new(bare_druid, nil, moab_storage_root, 'validate_checksums')
+    @results = AuditResults.new(bare_druid, moab.current_version_id, moab_storage_root, 'validate_checksums')
   end
 
   def validate_checksums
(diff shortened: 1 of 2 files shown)


jmartin-sul:
hmm... i'm not sure where the `total-coverage` drop is coming from, esp in light of the 100% `diff-coverage`.  i am inclined to ignore, but happy to look into it if others would prefer no drop at all.

PR #1523, opened 2020-05-06
Unique constraint for MoabStorageRoot's storage_root

### LINKED ISSUE #1497: moab_storage_roots.storage_location should be unique (add a DB constraint)
we shouldn't have two different `moab_storage_roots` entries that actually point to the same `storage_location`.  we have an index on that field already, but the index should enforce uniqueness.  lack of this constraint was likely an oversight during the first pres cat work cycle.

jermnelson:
## Why was this change made?
Fixes #1497 

n/a

### DIFF
--- a/spec/models/moab_storage_root_spec.rb
+++ b/spec/models/moab_storage_root_spec.rb
@@ -19,30 +19,38 @@ RSpec.describe MoabStorageRoot, type: :model do
     expect(moab_storage_root).to be_valid
   end
 
-  describe 'enforces unique constraint on name' do
-    it 'at model level' do
-      msg = 'Validation failed: Name has already been taken'
-      expect do
-        described_class.create!(name: 'storage-root-01', storage_location: '/storage_root02')
-      end.to raise_error(ActiveRecord::RecordInvalid, msg)
+  describe 'name attribute' do
+    context 'at model level' do
+      it 'must be unique' do
+        msg = 'Validation failed: Name has already been taken'
+        expect do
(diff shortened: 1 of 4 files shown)


spec/models/moab_storage_root_spec.rb
```diff
@@ -19,14 +19,31 @@
     expect(moab_storage_root).to be_valid
   end
 
-  it 'enforces unique constraint on name (model level)' do
-    expect do
-      described_class.create!(moab_storage_root.attributes.slice('name', 'storage_location'))
-    end.to raise_error(ActiveRecord::RecordInvalid)
+  describe 'enforces unique constraint on name' do
```
mjgiarlo, line 22: What would you think about re-organizing the tests to look more like the following:

```ruby
  describe 'name attribute' do
    context 'at model level' do
      it 'must be unique' do
        msg = 'Validation failed: Name has already been taken'
        expect do
          described_class.create!(name: 'storage-root-01', storage_location: '/storage_root02')
        end.to raise_error(ActiveRecord::RecordInvalid, msg)
      end
    end

    context 'at db level' do
      it 'must be unique' do
        expect { moab_storage_root.dup.save(validate: false) }.to raise_error(ActiveRecord::RecordNotUnique)
      end
    end
  end
```

I'm thinking mostly about the readability conventions around the strings that go with `describe`, `context`, and `it` blocks. Not a big concern, admittedly.

jermnelson: Done; just pushed up commit with suggested changes.

PR #1524, opened 2020-05-06
Prefer `chdir:` arg to Open3 method over `Dir.chdir`

### LINKED ISSUE #1519: switch from Dir.chdir in Zipmaker, to using chdir param in Open3.capture2e
per https://github.com/sul-dlss/preservation_catalog/issues/1487#issuecomment-621309175:

>     Make the following, tiny patch (HT @justinlittman: [sul-dlss/google-books#529](https://github.com/sul-dlss/google-books/pull/529)). From:
> 
> 
> ```ruby
>     Dir.chdir(work_dir.to_s) do
>       combined, status = Open3.capture2e(zip_command)
>       raise "zipmaker failure #{combined}" unless status.success?
>     end
> ```
> 
> to:
> 
> ```ruby
>     combined, status = Open3.capture2e(zip_command, chdir: work_dir.to_s)
>     raise "zipmaker failure #{combined}" unless status.success?
> ```

this should be the only instance of this usage in pres cat.  it's not threadsafe, but the alternate approach @mjgiarlo describes is.  switching to that approach will help pave the way for switching over to sidekiq from resque (see the top level description of the issue linked above).

question:  is there a good way to document that we want to avoid `Dir.chdir` going forward?  a thread safety section in the readme if we switch from resque to sidekiq?

mjgiarlo:
Fixes #1519

I briefly looked into testing this change but I don't think we get much value out of essentially testing a dependency. I suspect the test suite as currently written suffices as this is a refactoring.

## Why was this change made?

Why? Because `Dir.chdir` is not thread-safe, and that will be a problem when we switch to using Sidekiq for background jobs.

No.

No.

### DIFF
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -29,10 +29,9 @@ class DruidVersionZip
   # not 'ab/123/cd/4567/ab123cd4567/...'
   def create_zip!
     ensure_zip_directory!
-    Dir.chdir(work_dir.to_s) do
-      combined, status = Open3.capture2e(zip_command)
-      raise "zipmaker failure #{combined}" unless status.success?
-    end
+    combined, status = Open3.capture2e(zip_command, chdir: work_dir.to_s)
+    raise "zipmaker failure #{combined}" unless status.success?
+
     part_keys.each do |part_key|
       DruidVersionZipPart.new(self, part_key).write_md5
     end


mjgiarlo:
See related PR: https://github.com/sul-dlss/DeveloperPlaybook/pull/119

Issue #366, opened 2017-12-04
[BLOCKED] try to make CompleteMoabHandler less confusing


ndushay:
Per discussion on #1487, start small.  E.g., are there knotty nested conditionals in long methods, where some of the nested logic could be extracted into a nicely named helper method?  If we chip away at that sort of thing, hopefully the class becomes more approachable, and easier to test and refactor.

A larger sort of refactor might be splitting the class into multiple classes for different categories of work (e.g. validating a moab we know is in the catalog, vs adding a moab we haven't seen before to the catalog vs updating a moab we know is in the catalog when it gets versioned, etc).

It might be easiest to spawn smaller more concrete tickets from this, leaving this open as a catch all, as this will probably require a few iterations to get things into a satisfactory state.

~~It would probably be best to do the `MoabValidationHandler` refactoring first (whichever we keep of #1276 and #1494, since those are dupes).  Refactoring `CompleteMoabHandler` and `MoabValidationHandler` concurrently sounds like a recipe for rebase hell.~~

As a start, maybe look at the class and note where things get hard to follow, or where intent/purpose is hard to discern.  E.g. when your eyes glaze over on a super long method or a 3 levels of nested conditionals, or whatever.

jmartin-sul:
https://github.com/sul-dlss/preservation_catalog/pull/1294 might be able to provide some inspiration

ndushay:
@jmartin-sul I'm closing this ticket as it is 5 years old and I think the refactors done recently have helped a lot.

Issue #1530, opened 2020-05-12
ZippedMoabVersion should switch to referring to PreservedObject, rather than CompleteMoab


jmartin-sul:
#### background

Currently, `zipped_moab_versions` (`ZippedMoabVersion`) has a pointer to `complete_moabs`, with a non-null foreign key constraint.  As pointed out [here](https://github.com/sul-dlss/preservation_catalog/issues/1159#issuecomment-571377163), this makes it impossible to delete a `CompleteMoab` for which ZMVs have been created (i.e., any moab we've shipped off to cloud archives).

Per a later comment on that issue, the purpose of the existing field is to indicate from which moab an archival zip was created.  It was a tossup at the time we initially wrote that between pointing `zipped_moab_versions` to `complete_moabs` or `preserved_objects`.  Since pointing to `complete_moabs` provided just a bit of extra info, while still making it easy enough to associate the zip to the druid, we went with `complete_moabs`.  But to easily allow for more regular deletion of old moabs that have been migrated/copied to new storage, while keeping the schema sane and tightly constrained, we're going to switch to pointing to `preserved_objects` (there will _always_ be a `PreservedObject` for a druid:  an abstract parent record for all the physical instantiations is the point of `preserved_objects`).

#### TODO
- [ ] write the rails DB migration
  - [ ] add a `zipped_moab_versions.preserved_object_id` column to the schema.  it should have a foreign key constraint to `preserved_objects.id`.
  - [ ] populate the column with the appropriate values.  in sql, this is something like `UPDATE zipped_moab_versions zmv SET zmv.preserved_object_id = (SELECT preserved_object_id FROM complete_moabs cm WHERE cm.id = zmv.complete_moab_id)`.  i'm not sure off-hand if there's a more ActiveRecord-y way of doing that that still does it in one shot.  i'm also not sure if doing it in one shot will be performant enough.  if we do it by paging and updating in batches, i'd strongly favor using functionality from the `postgresql_cursor` gem (already in use in the project) so that we can keep it transactional.
  - [ ] add a non-null constraint to the new `zipped_moab_versions.preserved_object_id` column.
  - [ ] drop the old `zipped_moab_versions.complete_moab_id` col.
  - [ ] move the `last_archive_audit` column from `complete_moabs` to `preserved_objects`.
- [ ] update all app and test code that uses `ZippedMoabVersion.complete_moab[_id]` to instead use `ZippedMoabVersion.preserved_object[_id]`
[...]

PR #1539, opened 2020-05-20
Refactor MoabValidationHandler

### LINKED ISSUE #1276: Refactor `MoabValidationHandler`
The `MoabValidationHandler` mixin and its interactions with the classes in which it is `include`d are long and complex and, arguably, messy.

1) The mixin provides a bunch of public methods, so there are numerous interaction points and, thus, different "flows" through the mixin, which means there may be time-sensitive data points;
2) The mixin is not self-contained, requiring its containing classes to respond to certain messages, and there is no good way to enforce that contract;
3) Containing classes use only bits and pieces of the mixin; and
4) One containing class overrides the behavior of the mixin.

Overall, this suggests `MoabValidationHandler` is a "kitchen sink" abstraction, and given the length and complexity of this module and its containing classes and the sheer amount of shameless :green_apple: code, makes it difficult to reason about and difficult to work with.

I have attempted (twice) to extract a `MoabValidator` class out of the mixin  and each time walked it back due to not having enough time to sufficiently disentangle the code and touch/update all affected tests. In _lieu_ of trying to refactor it again, I have instead done a small bit of analysis which I will share here (knowing, of course, that it will age poorly):

# `MoabValidationHandler` public methods

## `object_dir`

Depends on `druid` and `storage_location` (the latter of which depends on `complete_moab`).

## `moab`

Depends on `druid` and `object_dir`.

## `can_validate_checksums?`

No dependencies.

## `can_validate_current_comp_moab_status?`

Depends on `can_validate_checksums?`, `complete_moab`, and `results`

## `moab_validation_errors`

Depends on `moab` and `ran_moab_validation!`

## `ran_moab_validation!`

No dependencies.

## `ran_moab_validation?`

No dependencies.

## `update_status`

Depends on `complete_moab` and `results`

## `set_status_as_seen_on_disk`

Depends on `update_status`, `moab_validation_errors`, `results`, and `complete_moab`

# `PreservedObjectHandler`

Uses the following methods from `MoabValidationHandler`:

* `can_validate_current_comp_moab_status?`
* `moab_validation_errors`
* `ran_moab_validation?`
* `ran_moab_validation!`
* `update_status`
* `set_status_as_seen_on_disk`

# `CatalogToMoab`

Uses the following methods from `MoabValidationHandler`:

* `moab`
* `can_validate_current_comp_moab_status?`
* `ran_moab_validation?`
* `update_status`
* `set_status_as_seen_on_disk`

# `ChecksumValidator`

Uses the following methods from `MoabValidationHandler`:

* `object_dir`
* `moab`
* `can_validate_checksums?`
* `update_status`
* `set_status_as_seen_on_disk`

jmartin-sul:
It'd be great to get a few reviewers, since the point of this is to start making the code more approachable, which is somewhat subjective.  This change at least removes the circular dependency between the old `MoabValidationHandler` module and the classes that included it, and since the lack of clarity about those interactions was a big complaint, I'm hoping this paves the way for further class reorganization and method decomposition.  see #366

## Why was this change made?

To start making it easier to understand and reason about the work that `MoabValidator` and its consumers do.

closes #1276

unit tests.  we can deploy to stage and run the validations on objects there to further put it through its paces, and along those lines, could also create/version an SDR object.

I attempted to:
* keep the intent of the existing tests
* minimize code churn and to break into somewhat logical commits to reduce the pain of reviewing (also tried to provide useful commit messages)
* make some easy touchups in places where i was already working, including trying to make some test setup less fragile

N/A


spec/services/complete_moab_handler_check_exist_spec.rb
```diff
@@ -379,18 +380,22 @@
...
       it 'calls CompleteMoab.save! (but not PreservedObject.save!) if the existing record is NOT altered' do
         druid = 'zy987xw6543'
-        po = create(:preserved_object, druid: druid)
-        cm = create(:complete_moab, preserved_object: po)
-        allow(PreservedObject).to receive(:find_by!).with(druid: druid).and_return(po)
-        allow(po.complete_moabs).to receive(:find_by!).with(moab_storage_root: ms_root).and_return(cm)
-        expect(cm).to receive(:save!)
-        expect(po).not_to receive(:save!)
-        described_class.new(druid, 1, 1, ms_root).check_existence
+        # po = create(:preserved_object, druid: druid)
```
jermnelson, line 383: Should this commented-out code be removed?

jmartin-sul: good catch, will fix!

app/services/moab_validator.rb
```diff
@@ -1,28 +1,45 @@
...
+class MoabValidator
+  attr_reader :druid, :storage_location, :results, :caller_validates_checksums
+
   # #druid - String (the "bare" druid, e.g. 'ab123cd4567', sans 'druid:' prefix)
   # #storage_location - String - the root directory holding the druid tree (the storage root path)
   # #results - AuditResults - the instance the including class is using to track findings of interest
-  # #complete_moab - CompleteMoab - instance of the complete moab being validated
+  # #complete_moab - CompleteMoab - optional - instance of the complete moab being validated, if already available from the caller
+  # #caller_validates_checksums - was this called by code that re-computes checksums to confirm that they match the values listed in the manifests?
+  def initialize(druid:, storage_location:, results:, complete_moab: nil, caller_validates_checksums: false)
```
mjgiarlo, line 13: Can this object consistently get the druid from the `AuditResults` instance? If so, we could drop a param.

app/services/moab_validator.rb
```diff
@@ -1,28 +1,45 @@
...
+    @caller_validates_checksums = caller_validates_checksums
+  end
 
   def object_dir
     @object_dir ||= "#{storage_location}/#{DruidTools::Druid.new(druid).tree.join('/')}"
   end
 
-  def moab
-    @moab ||= Moab::StorageObject.new(druid, object_dir)
+  def complete_moab
```
mjgiarlo, line 25: I wonder if you are open to renaming this method to disentangle it from the possibly injected `complete_moab`. For instance, could line 17 be `@complete_moab = complete_moab || default_complete_moab` with this method renamed to `#default_complete_moab`? (I don't love the name; it's a placeholder for demonstrating my point.) Or is it not possible to settle on the value of the complete_moab at initialization time?

.rubocop_todo.yml
```diff
@@ -32,7 +33,14 @@ Layout/MultilineMethodCallIndentation:
   Exclude:
     - 'spec/lib/preservation_catalog/s3/audit_spec.rb'
 
-# Offense count: 31
+# Offense count: 2
+# Cop supports --auto-correct.
```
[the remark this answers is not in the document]
jmartin-sul: will try to do this and will push up an extra commit on this PR if successful (or, will do it in a follow on PR if it'd make this a pain to review).

app/services/moab_validator.rb
jmartin-sul: ah yeah... some callers already have the `CompleteMoab` instance available, and allowing them to provide it allowed less code churn and more incremental refactoring (as well as seeming more efficient in terms of reducing DB lookups).  but i do like the idea of disentangling this a bit more.

slight worry that changing when the moab is looked up will result in more test churn, but if that ends up exposing overly brittle tests, and isn't too much of a pain to fix, i'll go ahead and fix that too (though also maybe in a separate PR, to keep things easier to review, will see how it plays out).

mjgiarlo: @jmartin-sul OK, sounds good to me.

.rubocop_todo.yml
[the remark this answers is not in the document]
jmartin-sul: i started on this in #1547, but would like to get #1545 merged before seeing that the rest of the way through, to avoid merge conflicts.

app/services/moab_validator.rb
jmartin-sul: i believe it can, but it breaks a bunch of tests to do this.  the reason is, we spy on `AuditResults` interactions a bunch, and so we're often passing around instance doubles of it that aren't fully fleshed out, and which don't necessarily have a druid on hand the way they're set up at present.

but if you think that change would be a significant improvement, we could ticket follow on work to get rid of it.  would rather not hold up this PR for that, if that's ok.

fwiw, i also:
* felt like this might be a extra weird entanglement to introduce between `MoabValidator` and `AuditResults`, but it is nice having one param fewer, and one less place for possible divergence of values that shouldn't diverge.
* considered doing the same for the `storage_location` param, which `results` should have via `moab_storage_root`, which i believe it always gets at present.  and for the cases where we're using `MoabValidator`, we always should have that param.  though looking into this also made me realize that `AuditResults` should get the `ZipEndpoint`(s) instead of the `MoabStorageRoot` for cloud audits, so i noted that on #1530 (since that change will likely switch `PartReplicationAuditJob` from getting a `CompleteMoab`, which references `MoabStorageRoot`, to a `PreservedObject`, which will have possibly many associated `MoabStorageRoot`s once we support multiple `CompleteMoab`s per druid).

app/services/moab_validator.rb
jmartin-sul: ok, so i tried this, and similar story to getting `druid` from `results`:  it breaks a bunch of existing tests to look up the `CompleteMoab` earlier.  many of these are likely more brittle than they need to be to accurately test what they're looking for.  but i'd prefer to ticket this suggestion as follow on work, if that's ok.

jmartin-sul:
@mjgiarlo and @jermnelson, i think this is ready for another look.

@mjgiarlo, if you're ok with follow-on tickets for your suggestions, i'm happy to write those if you want.

PR #1543, opened 2020-05-21
[DRAFT] Add a rake task to manually audit zip part replication

aaron-collier:
As of the first commit, the is very much a WIP. Pushing this branch and PR up for visibility and recommendations on approach.

### DIFF
--- a/lib/tasks/prescat.rake
+++ b/lib/tasks/prescat.rake
@@ -68,4 +68,44 @@ namespace :prescat do
       MoabStorageRoot.find_by!(name: args[:storage_root_name]).complete_moabs.find_each(&:validate_checksums!)
     end
   end
+
+  namespace :druid do
+    desc "verify zip parts on cloud storage roots"
+    task :parts, [:druid, :version] => [:environment] do |_task, args|
+      druid = args[:druid]
+      versions(druid)
+      version = args[:version]
+      cm = CompleteMoab.by_druid(druid).first
+      next if cm.nil?
+
+      MoabReplicationAuditJob.perform_now(cm)
+      cm.reload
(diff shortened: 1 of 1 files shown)


jmartin-sul:
one thought:  if the current part audit code were modified so that the skipping of `unreplicated` parts was optional, you could just run that job synchronously, without having to reproduce so much of what it does internally inside of this rake task.

or, maybe we should just not filter out `unreplicated` parts from audits in general.  i'm not sure how frequently audit runs on an object that isn't fully replicated, and how much spurious noise it'd add if we didn't skip parts in that state.

jmartin-sul:
i'm thinking specifically of this `:unreplicated` filter in this method:  https://github.com/sul-dlss/preservation_catalog/blob/adf8b59d23be5d9101154abc615b64c55cc9f59a/app/lib/preservation_catalog/s3_audit.rb#L20-L28

jmartin-sul:
if we followed through on the change in this branch:  https://github.com/sul-dlss/preservation_catalog/compare/check-unreplicated-zip-parts

i believe a lot of the rake task here (and the wiki entry/gist before it) could be replaced by the contents of the body of `PartReplicationAuditJob#perform`, since a lot of the gist/wiki that's being ported was actually just recreating that code, but removing the where clause removed by that branch.

aaron-collier:
@jmartin-sul - yeah, I was thinking the same thing. After the above work, this is likely unnecessary (or can be cleaned up to call the proper code). There really isn't any reason this shouldn't be part of audit, and allow some automated clean up (i.e. we can determine if the endpoints look correct and the DB just needs to be updated like we have done manually...)

ndushay:
@jmartin-sul 
- do we now have this capability?  
- If we need this, does it warrant a ticket?
- is this PR worth keeping, or have all the refactors made it no longer useful?

PR #1544, opened 2020-05-21
Rerun rubocop --auto-gen-config

aaron-collier:
## Why was this change made?

I was getting a few specific rubocop failures on master.

Ran rubocop

N/A


mjgiarlo:
This might cause a conflict with #1539 which also re-gens rubocop todos. Cc: @jmartin-sul

aaron-collier:
@mjgiarlo oh good, did not realize that was a part of it. As long as it's being addressed, happy to close.

PR #1460, opened 2020-04-06
Update dependencies

sul-devops-team:

PR #1545, opened 2020-05-21
Moves duplicative code in audit class to parent

jermnelson:
## Why was this change made?
Part of the refactoring mentioned in #1460 

Unit tests

n/a


app/lib/preservation_catalog/aws/audit.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module PreservationCatalog
+  module AWS
+    # Methods for auditing the state of a ZippedMoabVersion on an AWS S3 endpoint.
+    class Audit < PreservationCatalog::S3Audit
+      delegate :bucket_name, to: ::PreservationCatalog::S3
+
+      private
+
+      def bucket
```
jcoyne, line 11: Can this bucket method move to the superclass too?

app/lib/preservation_catalog/s3_audit.rb
```diff
@@ -0,0 +1,72 @@
...
+      @zmv = zmv
+      @results = results
+    end
+
+    # convenience method for instantiating the audit class and running the check in one call
+    def self.check_replicated_zipped_moab_version(zmv, results)
+      new(zmv, results).check_aws_replicated_zipped_moab_version
+    end
+
+    def check_aws_replicated_zipped_moab_version
```
jcoyne, line 20: If this is given a generic name, then `self.check_replicated_zipped_moab_version` could move to the superclass.

app/lib/preservation_catalog/aws/audit.rb
jmartin-sul: i like that idea, but the methods are slightly different at present, and while i'm sure they're consolidate-able, the exact way to do so doesn't jump out at me immediately.  since this PR as a whole seems like an improvement, could we do that move as follow-on work, and get this merged?

jmartin-sul: alternatively, if the way to move this to the superclass is clear to you and i'm missing it, could you give just a bit more detail on how you'd abstract the differences?

app/lib/preservation_catalog/s3_audit.rb
jmartin-sul: i think it's already in the superclass?

but agreed, the name could be genericized, just looked closely by playing around with it in a text editor, and other than the method names and some minor spacing and var name differences, `check_aws_replicated_zipped_moab_version` is exactly the same (in terms of functionality) as `check_ibm_replicated_zipped_moab_version`.

jmartin-sul: so i think this method is essentially always being called, and i think maybe `check_ibm_replicated_zipped_moab_version` isn't being called at all, but it doesn't matter, because both do the exact same thing.  that might also explain the coverage drop?  `check_ibm_replicated_zipped_moab_version` is essentially no longer being exercised?

PR #1547, opened 2020-05-22
Rubocop auto correct

jmartin-sul:
## Why was this change made?

prompted by [comment](https://github.com/sul-dlss/preservation_catalog/pull/1539#discussion_r428132804) from @mjgiarlo on #1539.

unit tests

n/a


jmartin-sul:
there are a couple files that still have potential auto-corrections that i'd like to include in this:
* spec/models/druid_version_zip_part_spec.rb
* spec/lib/preservation_catalog/s3/audit_spec.rb

but i was wary of conflicts with #1545, so i'd like to get that merged first, then will finish this up.

jmartin-sul:
from the commit message:

#### lots of easy/safe rubocop cleanup

* remove from .rubocop.yml anything we actually want to fix (for auto-correction, or relocation to .rubocop_todo.yml if not auto-correctable).
* switch to ruby 2.7 for target version: rebuilt pres cat stage VM and CI are both already on 2.7 with no issue, prod will soon be rebuilt on 2.7.
* run rubocop --safe-auto-correct
* run rubocop --auto-gen-config
* modify LineLength and MultipleExpectations to list specific files so the limits aren't just blanket increased across the codebase

PR #1549, opened 2020-05-22
Add uniqueness database constraint for ZipParts

### LINKED ISSUE #1546: unexpected duplicate zip parts for zipped_moab_versions rows
A `ZippedMoabVersion` represents a zipped version directory from a moab, for a specific druid, on a specific cloud endpoint.  Since each cloud endpoint only stores one logical copy of a moab-version directory for a given druid, we should not see more than one `ZipPart` of a given suffix for a given `ZippedMoabVersion`.

Ultimately, we should add a unique constraint to zip_parts on combo of zipped_moab_version_id and suffix.  But...

Maybe I'm constructing the wrong query for this, but there appears to be a disconcerting amount of this sort of thing to clean up on prod before we can apply such a constraint:

```ruby
pry(main)> zmvs_with_dupe_suffixes = ZipPart.group(:zipped_moab_version_id, :suffix).having('count(id) > 1').pluck(:zipped_moab_version_id, :suffix)
=> [[1, ".zip"],
 [3, ".zip"],
 [5, ".zip"],
...
 [19656334, ".zip"]]
pry(main)> zmvs_with_dupe_suffixes.size
=> 5904
pry(main)> ZippedMoabVersion.count
=> 16727610
pry(main)> ZipPart.group(:zipped_moab_version_id, :suffix).having('count(id) > 1').select(:zipped_moab_version_id, :suffix).to_sql
=> "SELECT \"zip_parts\".\"zipped_moab_version_id\", \"zip_parts\".\"suffix\" FROM \"zip_parts\" GROUP BY \"zip_parts\".\"zipped_moab_version_id\", \"zip_parts\".\"suffix\" HAVING (count(id) > 1)"
```

the sql looks right to me, but a second set of eyes would be good:
```sql
SELECT zip_parts.zipped_moab_version_id, zip_parts.suffix
FROM zip_parts
GROUP BY zip_parts.zipped_moab_version_id, zip_parts.suffix
HAVING (count(id) > 1)
```

5904 out of 16727610 isn't insane, but it's still significant, esp since we're not sure how this happens.

the only automatable remediation i can think of to start would be to:
* iterate over the results of the above query
  * for each combination, if there's an `ok` version of the `ZipPart` for the given suffix/ZMV combo, check the cloud endpoint to see that it's actually still present.
    * if it is, keep the `ok` row and destroy the redundant `ZipPart`s

as a start, maybe we should add an ActiveRecord validation to prevent more occurrences of this before remediating?

i was worried that this would be blocked by #1530, but i don't think that schema change will cause problems here, since the `preserved_objects` and `complete_moabs` never come into play directly.

discovered while investigating #1410.

I could imagine splitting this ticket into three (sequenced) sub tickets:
- [ ] add an ActiveRecord validation to prevent further occurrences
- [ ] remediate the existing bad data (could be done before the validation was added, but risks having more bad data accrue till we figure the issue out and stop it from growing)
- [ ] add the unique constraint (blocked by remediating bad data)

mjgiarlo:
Fixes #1546

## Why was this change made?

This prevents duplicate rows from appearing in the database, which is not expected to happen. This was initially discovered in #1410, though not *that* many rows are affected by this.

Unit tests.

None.


mjgiarlo:
My take on coverage: it is likely not worth the time to write a test for the rescue/retry logic. I am happy to reconsider, though!

app/jobs/plexer_job.rb
```diff
@@ -52,6 +52,10 @@ def find_or_create_unreplicated_part(zmv, part_s3_key, metadata)
       size: metadata[:size],
       suffix: File.extname(part_s3_key)
     ) { |part| part.unreplicated! }
+  rescue ActiveRecord::RecordNotUnique
+    # This catches an exception triggered by a race condition because find_or_create_by is not atomic.
+
+    retry
```
jmartin-sul, line 58: i'm a bit worried this ends up in an infinite loop in the rare case where it might get triggered, but i haven't really used `retry` myself, so i don't have much intuition for it.  but from what i was reading just now, it seems like it would try the same thing over from the top of the `find_or_create_unreplicated_part` method with the same `metadata` (because that seems like the implicit `begin` for the `retry`'s `rescue`?).  and if there's already a part for that ZMV with that suffix, i'd expect the constraint to get violated again, and then the same error/retry.

or,  am i misunderstanding, and what's retried is the next iteration of the `zmvs.each` loop?

i'm a little rusty on it since i haven't really touched it during this maintenance WC, but there are delivery pipeline integration tests in pres cat, so i think we could simulate this scenario, though not sure off-hand how much of a pain it'd be.

mjgiarlo: @jmartin-sul the exception would be raised by `find_or_create_by`, and that method first tries a `find_by` before issuing a `create`. Presumably subsequent retries would not fail because the `find_by` returns a record (it does not raise `ActiveRecord::RecordNotUnique`). So if there is already a part for that ZMV with that suffix, it ought to get returned. Does that assuage your infinite loop concerns? It should only get triggered when the momentary race condition is encountered.

jmartin-sul: ah!  yes.  that makes sense.  thanks!

PR #1560, opened 2020-05-28
S3Audit#check_replicated_zipped_moab_version checks all parts

jmartin-sul:
needs tests

## Why was this change made?

to make it easier to audit parts that might have a spurious `unreplicated` status.

unit tests as they are pass (to my slight surprise), but i think we should add a new test or two (e.g. does an `unreplicated` part entry without a corresponding part entry in the cloud stay in the `unreplicated` state and trigger appropriate alerting/logging, and does an `unreplicated` part entry that actually has a corresponding physical part in the cloud with a matching checksum get set to `ok`).

n/a

### DIFF
--- a/app/lib/preservation_catalog/s3_audit.rb
+++ b/app/lib/preservation_catalog/s3_audit.rb
@@ -5,22 +5,24 @@ module PreservationCatalog
   class S3Audit
     delegate :bucket, :bucket_name, to: :s3_provider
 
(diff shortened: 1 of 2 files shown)


jmartin-sul:
might help simplify:
* implementation of #1543
* investigation into #1546
* https://github.com/sul-dlss/preservation_catalog/wiki/Investigating-a-druid-with-replication-errors

PR #1572, opened 2020-06-11
Refactors reporting to workflow service, event service, honeybadger, …

### LINKED ISSUE #1416: Clean up workflow and dor event reporting code
The related code has accreted some functionality in a piecemeal way that has left some unnecessary duplication and some misleading names (in service of adding the functionality expediently).

Some obvious things to look at:
* `WorkflowReporter` has a misleading name, as it now reports to the DOR events service too (by way of `Dor::Services::Client.object(namespaced_druid).events`).
  * maybe a better name would be something like `DorEventReporter`?  is that too specific/misleading in terms of tying it to the event service?  that seems defensible (the class handles reporting events to workflow service and the event service).  but it was just the best i could come up with without further discussion.  only other thought was `ExternalServicesReporter`, but that feels a little too generic to me, maybe.
* There is some duplicated event service reporting code in `AuditResults` and in `WorkflowReporter`.  not much, but it'd be nice to consolidate if possible.

This all happened because the obvious places to put the DOR event service reporting code (introduced after Honeybadger and Workflow reporting) were the spots where we already reported to honeybadger and the workflow service.  This is why we just stuck that new code along side the old code.

I spent a bit of time on this when working on #1415, but while I saw an obvious expedient fix for that issue, I didn't see a totally obvious path for sanding the rough edges described here, and wanted to time box the initial refactoring attempt to an hour or so, since we're already a little pressed for time.

justinlittman:
…and logger.

closes #1416

## Why was this change made?
Audit_results tangled together reporting to the workflow service, event service, HB, and logger.

Unit.

No.


app/services/reporters/event_service_reporter.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    def handled_single_codes
+      [
+        AuditResults::INVALID_MOAB
+      ]
+    end
+
+    def handled_merge_codes
+      [
+        AuditResults::ACTUAL_VERS_LT_DB_OBJ,
```
ndushay, line 16: honestly, I'm not sure these need to be in AuditResults namespace, but certainly wouldn't hold this up for that.

ndushay:
- I'm gonna override the code-climate errors.
- I just reduced the codeclimate threshold to 95% fwiw

PR #1573, opened 2020-06-11
Moab by storage root scope follow on

jmartin-sul:
## Why was this change made?

a couple small touchups related to #1568

unit tests

n/a


.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
-# `rubocop --auto-gen-config`
-# on 2020-05-22 16:54:44 -0700 using RuboCop version 0.73.0.
+# `rubocop --auto-gen-config --auto-gen-only-exclude`
```
mjgiarlo, line 2: Oooh, `--auto-gen-only-exclude`, nice find! Much better.

jmartin-sul: yeah, just discovered this option from `rubocop --help`!  i personally prefer the specific exceptions to the global maximum increases, so i'm likely to use this approach for `.rubocop_todo.yml` generation by default going forward.

looked into this because i also [ran into 8 rubocop errors](https://github.com/sul-dlss/preservation_catalog/pull/1568#discussion_r438945296), like @jermnelson did after re-generating the todo file and re-running rubocop, and i noticed that 7 were for line length, and 1 was for max examples in a spec.  those are both things we specifically override in `.rubocop.yml`.  previously, the exceptions in the todo file (that were overwritten by the config re-gen) were hand edits i made to exclude specific files instead of upping global maximums.  so i _suspect_ that rubocop has issues with double-specification of max limits, but not with a mix of max and file specific exclusion.  i was happy to see an option for getting that from auto-gen, since hand editing `.rubocop_todo.yml` (as i've occasionally done) seems to be frowned upon.

Issue #1528, opened 2020-05-11
add notion of a "primary" CompleteMoab for a druid


jmartin-sul:
_**first thing to do:  pair w/ @jmartin-sul to discuss the two implementation options and pick one**_

For every druid in Preservation Catalog, there should be at least one `CompleteMoab` among our `MoabStorageRoot`s (exactly one at present, but adding proper support for more than one is the point of the epic that contains this ticket).  Further, there should be _exactly_ one `CompleteMoab` that is seen as the "primary".

The main purpose of this notion of "primary"-ness is to indicate which copy the Preservation Robots should update when an existing object is being versioned.

#### possible implementation approaches
_edit: went with a join table after some discussion, see #1577 for implementation, max one primary per druid is enforced by unique `preserved_object_id` col on the join table (in addition to its own primary key id)_

To whoever picks this up, I think some pairing and discussion about those implementation options is warranted, unless I'm missing something obvious that'd be easier than what I described there (which would be welcome).

#### rationale

In discussing how we'd implement multiple copies of a moab for a given druid, we settled on an approach where Preservation Robots would only update one copy (to keep down the required time and potential for error when robots are writing a large new version, e.g. a version that has a ton of video or something).  Since the main use case we're targeting is moving moabs from old storage to new storage, by way of making a copy and cutting over to the newer copy as the primary at some point, we expect that keeping the non-primary copy in sync will be the responsibility of some process outside of Preservation Catalog or Preservation Robots.  For example, an automated `rsync` of old storage root contents to new storage hardware, a manual copy by a person moving a moab to a storage root with more room for growth, or eventually, the ongoing `rsync` that backs up our primary storage brick's contents to our secondary storage brick.

#### next steps (for consideration in e.g. #1536)

We may also want to differentiate when audits discover unexpected inconsistencies.  E.g., if the "primary" moab is of a lower version than what's expected (lower than what we see on disk, lower than what `preserved_objects.current_version` indicates), our audits should alert loudly.  But if a non-primary copy is behind in version, we may want to log something and skip, e.g., the Honeybadger alert.

jmartin-sul:
i'm marking this blocked by #366 and #1276.  i don't think this is impossible without first doing those, but i think this will be easier if we can make progress on those (also, i can imagine merge conflicts if all of those are being worked simultaneously).

[...]

PR #1577, opened 2020-06-16
add preserved_objects_primary_moab table

ndushay:
## Why was this change made?

Part of #1528 (John - perhaps this is all of what you intended for #1528 ?)

To allow multiple CompleteMoab objects to be associated with a PreservedObject, we need for a single CompleteMoab to be "primary."   This PR adds a thin join table to accomplish this.

### DIFF
--- a/app/models/preserved_object_primary_moab.rb
+++ b/app/models/preserved_object_primary_moab.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+# thin join table to contain a primary CompleteMoab for each PreservedObject
+class PreservedObjectPrimaryMoab < ApplicationRecord
+  belongs_to :preserved_object, inverse_of: :preserved_objects_primary_moabs
+  belongs_to :complete_moab, inverse_of: :preserved_objects_primary_moabs
+end
--- a/db/migrate/20200616220804_create_preserved_objects_primary_moabs.rb
+++ b/db/migrate/20200616220804_create_preserved_objects_primary_moabs.rb
@@ -1,0 +1,9 @@
+class CreatePreservedObjectsPrimaryMoabs < ActiveRecord::Migration[6.0]
+  def change
+    create_table :preserved_objects_primary_moabs do |t|
+      t.references :preserved_object, foreign_key: true, null: false, index: { unique: true }
+      t.references :complete_moab, foreign_key: true, null: false, index: { unique: true }
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_05_22_144007) do
+ActiveRecord::Schema.define(version: 2020_06_16_220804) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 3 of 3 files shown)


db/migrate/20200616220804_create_join_table_preserved_objects_complete_moabs.rb
```diff
@@ -0,0 +1,9 @@
+class CreateJoinTablePreservedObjectsCompleteMoabs < ActiveRecord::Migration[6.0]
```
mjgiarlo, line 1: I might expect the class name to match the table name. Are you open to naming the class `CreatePreservedObjectsPrimaryMoabs` and the table `preserved_objects_primary_moabs`? Or some such. Cc: @jmartin-sul

ndushay: Hi Mike - I was trying to use rails generator migration magic to create the table;  it didn't work for me -- I'm absolutely fine changing the class name.

jmartin-sul: no strong feelings from me, but yeah, seems reasonable to align if it's not a pain

app/models/preserved_object_primary_moab.rb
```diff
@@ -0,0 +1,7 @@
+class PreservedObjectPrimaryMoab < ApplicationRecord
+  belongs_to :preserved_object, inverse_of: :preserved_objects_primary_moabs
+  belongs_to :complete_moab, inverse_of: :preserved_objects_primary_moabs
+
+  validates :preserved_object, presence: { strict: true }, uniqueness: { strict: true }
```
mjgiarlo, line 5: I might be in favor of leaving the uniqueness restriction in the database layer rather than in the application. We've made similar moves (towards DB constraints and away from model validations related to uniqueness) in a couple of other codebases recently to help with race conditions, and to avoid having to rescue/deal with both `ActiveRecord::RecordInvalid` and `ActiveRecord::RecordNotUnique`.

jcoyne: Agreed, this is belt and suspenders situation.

mjgiarlo: For that matter, I'm not sure the presence validations are useful here either for the same reason.

ndushay: removed.  @jmartin-sul and I discussed that there might be pushback on this point.  John prefers the error messages from Rails, I don't care myself.   I'm actually wondering if this model is going to be useful, but trust that John has a use for it in the future (and he did ask for it).

mjgiarlo: I also discussed w/ @jcoyne and he pointed out the value of the model is the ability to access the created_at/updated_at values in the join table.

app/models/preserved_object_primary_moab.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# thin join table to contain a primary CompleteMoab for each PreservedObject
```
ndushay, line 3: Humor:   replace comment with "uh, John Martin said we would be needing this ..."

ndushay: serious answer:  this will give us access to created_at and updated_at fields should we need them.

app/models/preserved_object_primary_moab.rb
jmartin-sul: avoiding the race condition is a compelling argument, hadn't considered that.  tbh i was mostly in the habit of using both because validations seemed like the idiomatic/rails-y way to do things, but the DB constraint always felt like the must-have.  definitely happy to just use DB constraints going forward.

would also be totally fine to file a tech debt ticket (prob backlog?) to remove the duplicative rails uniqueness and presence validations since this project has been pretty good about DB constraints overall afaict, other than a few multi-column uniqueness constraints we've missed the opportunity to add.

app/models/preserved_object_primary_moab.rb
jmartin-sul: haha, @jcoyne and @mjgiarlo provide a good justification in the other thread w/ the easy timestamp field access.  but i also had vague memories of a preservation policies related query where it was a little harder to write a good idiomatic active record query that also resulted in a nice single/atomic SQL query, because there was less ActiveRecord magic available without the model definition.  couldn't remember specifics, but happy to see other benefits i hadn't thought of too.

Issue #1535, opened 2020-05-13
fix code that assumes one CompleteMoab per druid/PreservedObject


jmartin-sul:
luckily, we do have a helpful constraint:  only one `CompleteMoab` per `PreservedObject`/`MoabStorageRoot` permutation (because a given druid can only map to one path under a given storage root).  see #1498 and #1525.

so, hopefully, anything that currently assumes only one `CompleteMoab` for a druid can add `MoabStorageRoot` as a qualifier to the query if that's already known, in which case the code can continue assuming at most one `CompleteMoab`.

for example, in `CompleteMoabHandler`, `#create_after_validation` and `#create` should check for the combo of druid and storage root, not just that the `PreservedObject` already exists for the druid
* `#check_existence` is trickier -- or maybe it's not?  maybe it also just checks whether there's a CM for the druid/root combo?
* what about `#update_version_after_validation`?
* what about `#create_db_objects`?

we should be checking at least these things for correctness
* code
  * code that queries for a `CompleteMoab`
  * code that creates a `PreservedObject`
  * more?
* documentation
  * https://github.com/sul-dlss/preservation_catalog/wiki/Investigating-a-druid-with-replication-errors
  * https://github.com/sul-dlss/preservation_catalog#migrate_moab_manually
  * db/README.md
  * other wiki pages?
  * more?

per #366, `CompleteMoabHandler` is one of the more confusing parts of the codebase.  since it's very likely you'll touch that class as part of this work, feel free to do some refactoring groundwork there, if you feel that'd make it easier to directly address this ticket.

jmartin-sul:
reopening since i'd like to add tests that exercise the use of multiple moabs for a druid before considering this done.

jmartin-sul:
remaining tasks:
- [ ] add tests that exercise more than one moab per druid
  - [ ] at least a little something that uses on disk fixtures would be nice (could copy existing moab from one fixture storage root to another and test around taht)
  - [ ] more thorough testing around the database can just use FactoryBot to create DB fixture instances
- [ ] documentation updates

jmartin-sul:
closing in favor of #1594, #1588, #1595, and #1591.

jmartin-sul:
since we are removing multi-moab support (since it is unused complexity), i'm closing this as moot.

see also #1892 and #1891

PR #1587, opened 2020-06-22
PreservedObjectsPrimaryMoab populated in CompleteMoabHandler

### LINKED ISSUE #1583: if a CompleteMoab is the first one for its druid (PreservedObject), make it the primary automatically
off-hand, the two ways i can think of to do this are:
1. in `CompleteMoabHandler`, which i believe creates all `CompleteMoabs` (though i'd have to double-check).
2. in an `after_create` hook on `CompleteMoab`.  since, iirc, hooks operate in transactions that also contain the operations happening to the associated object, and since `CompleteMoabHandler` should do all its creating and updating in explicit transactions anyway, this should end up being atomic with the creation of the `CompleteMoab` (which i think we want).

i think i prefer approach 2, but happy to hear other ideas or arguments for 1.

ndushay:
## Why was this change made?

When we create a new CompleteMoab, populate the PreservedObjectsPrimaryMoab join table

Closes #1583

unit tests - which sufficiently kicked my behind

n/a

### DIFF
--- a/app/services/complete_moab_handler.rb
+++ b/app/services/complete_moab_handler.rb
@@ -186,6 +186,8 @@ class CompleteMoabHandler
     @moab_validator ||= MoabValidator.new(druid: druid, storage_location: storage_location, results: results)
   end
 
+  # Note that this may be called by running M2C on a storage root and discovering a second copy of a Moab,
+  #   or maybe by calling #create_after_validation directly after copying a Moab
   def create_db_objects(status, checksums_validated = false)
     cm_attrs = {
       version: incoming_version,
@@ -203,11 +205,14 @@ class CompleteMoabHandler
 
     # TODO: remove tests' dependence on 2 "create!" calls, use single built-in AR transactionality
     transaction_ok = with_active_record_transaction_and_rescue do
-      PreservedObject
-        .find_or_create_by!(druid: druid) do |po|
-          po.current_version = incoming_version
-          po.preservation_policy_id = ppid
-        end.complete_moabs.create!(cm_attrs)
+      this_po = PreservedObject
+                .find_or_create_by!(druid: druid) do |po|
+                  po.current_version = incoming_version
+                  po.preservation_policy_id = ppid
+                end
+      this_cm = this_po.complete_moabs.create!(cm_attrs)
+      # add to join table unless there is already a primary moab
+      PreservedObjectsPrimaryMoab.find_or_create_by!(preserved_object: this_po, complete_moab: this_cm)
     end
     results.add_result(AuditResults::CREATED_NEW_OBJECT) if transaction_ok
   end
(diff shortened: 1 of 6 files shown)


app/services/complete_moab_handler.rb
```diff
@@ -203,11 +204,14 @@ def create_db_objects(status, checksums_validated = false)
...
-          po.preservation_policy_id = ppid
-        end.complete_moabs.create!(cm_attrs)
+      this_po = PreservedObject
+                .find_or_create_by!(druid: druid) do |po|
+                  po.current_version = incoming_version
+                  po.preservation_policy_id = ppid
+                end
+      this_cm = this_po.complete_moabs.create!(cm_attrs)
+      # add to join table
+      PreservedObjectsPrimaryMoab.create!(preserved_object: this_po, complete_moab: this_cm)
```
jmartin-sul, line 214: i think you want something like an `if this_po.complete_moabs.count == 1` guard clause, so that you're not accidentally marking a second moab primary if one gets picked up for the same druid/`PreservedObject` on a different storage root (now that `CompleteMoabHandler` can differentiate moabs for the same druid by storage root).

would probably be good to add a test that exercised that behavior.

jmartin-sul: another option, if you want to mark the moab primary whenever there's not already a primary (as opposed to only marking it primary when its the first moab for the druid):  you could do a `find_or_create_by!` on line 214 instead of a `create!`.  that should be no less safe than the guard clause approach, since this entire section is wrapped in an explicit transaction, and already uses `find_or_create_by!` (which is not itself transactional).

i have a very slight preference for only making the moab the primary if its the first moab for the druid, because that seems a bit more deterministic, and a bit easier to test well.  but that's a _very_ slight preference, and i'm open to `find_or_create_by!` (or other approaches i haven't thought of).

app/services/complete_moab_handler.rb
```diff
@@ -186,6 +186,7 @@ def moab_validator
     @moab_validator ||= MoabValidator.new(druid: druid, storage_location: storage_location, results: results)
   end
 
+  # only call this method if a CompleteMoab does not already exist for the druid
```
jmartin-sul, line 189: i don't think this will be true, necessarily.  otherwise, we wouldn't have an easy path for getting a second moab for a druid into the catalog (hence the switch to `PreservedObject.find_or_create_by!` instead of always assuming `.create!` is ok).

unfortunately there wasn't a good test in place to break for this already -- will ticket that as specific follow on work to #1535 (and have a failing WIP example test i can push up on a branch).

app/services/complete_moab_handler.rb
ndushay: I actually checked the code and added the comment that the method is ONLY called when the complete_moab does not exist.  AND the db constraint will error if the record is already there.

app/services/complete_moab_handler.rb
ndushay: It is currently empirically true, based on the code as I read it.

ndushay:
John and I spoke in RL;  the use case for multiple Moabs is "it got copied outside of prescat, now we want to put it into prescat via a direct call to `CompleteMoabHandler.create_after_validation` or an M2C audit check on a storage root.  

Pushed the change John suggested.

jmartin-sul:
d'oh, i suggested an overly specific find condition when we discussed.  i just posed a small follow on PR to address that (#1593).

Issue #1588, opened 2020-06-23
update db/README.md (including ER diagram) for recent data model changes


jmartin-sul:
there are some recent and soon-to-be-merged data model changes that will require updates to the db/README and its schema diagram:
- [x] adding the notion of a "primary" moab (#1528, #1577)
- [x] `zipped_moab_versions` points to `preserved_objects` instead of `complete_moabs` (#1530, #1589)
- [x] `CompleteMoab` should be queried for by druid _and_ storage root (not just druid), #1535 (technically this is the app code taking better advantage of the data model, not a change to the underlying schema itself)

jmartin-sul:
from db/README.md, how to generate the ER diagram:

> Diagram generated from lucidchart.com (File > Import Data > Entity Relationship > SQL Import > PostgreSQL; follow instructions for schema dump/import, arrange tables given on left-hand toolbar)

unfortunately, i couldn't find an easily usable open source tool for generating a proper ER diagram (with the indicators for 1:1/1:many/etc relationships, FK constraints, etc).  but stanford has a license for lucidchart, so it should be free for us to use.

if this ticket is picked up before #1589 is merged, please use that branch's schema to generate the chart.

in addition to updating the ER diagram, please give db/README.md a read through, and correct table descriptions and example queries as needed.

jmartin-sul:
fixed by https://github.com/sul-dlss/preservation_catalog/pull/1673

Issue #1558, opened 2020-05-27
[UBER] preservation PROD storage migration #2


jmartin-sul:
#### storage roots to be migrated
* services-disk13
* services-disk14
* services-disk15
* services-disk16
* services-disk18
* services-disk19
* services-disk20

####  here is the same list as a ruby string array, for easier rails console query usage:
```ruby
%w[services-disk13 services-disk14 services-disk15 services-disk16 services-disk18 services-disk19 services-disk20]
```

jmartin-sul:
drawn from https://github.com/sul-dlss/preservation_catalog/issues/1465#issuecomment-611193512

## Week Before Cutover

Run M2C and C2M validation checks on Moabs and generate reports, for each storage root being migrated; details in the linked individual migration checklist issue for each root.  Different roots can be run in parallel for a given check, but the reports for each type of check should be run in sequence (i.e. all M2C before all C2M).

* _**TODO**_ notes from M2C run, including date/time of start/end, and summary of errors (if any).
* _**TODO**_ notes from C2M run, including date/time of start/end, and summary of errors (if any).

---

* services-diskN
    - [ ] M2C (run all objects, get report of all errors, not druids ...)
      - _log file output location_
    - [ ] C2M ditto
      - _log file output location_
* ...

jmartin-sul:
drawn from https://github.com/sul-dlss/preservation_catalog/issues/1465#issuecomment-611195392

## Cutover Weekend - Shut Down Services Before Final Reports

NOTE:  ask Andrew/Julian how early (on Friday) you can shut down
- Preservation Robot Workers
- All Preservation Catalog Workers (we don't want validation jobs picked up before the storage is visible)
- ReST API for `/catalog` route
- weekend crons

SDR Manager:
   - [ ] is ready for accessioning to stop (including turning off Google Books accessioning)

Ops:
   - [ ] is ready to do the final backup of prescat database (but the actual snapshot will be taken after the shutdowns described in the rest of this comment)

### Shut Down Preservation Robots Workers

- [ ] (Ops) turn off nagios alerts for preservation_robots prod boxes
- [ ] look for running preservation robots workers in the resque GUI:  https://robot-console-prod.stanford.edu/workers
- [ ] shut off the workers manually by killing (only) the `resque-pool-master` process (`hotswap` interferes with other cap resque commands)

    ```sh
    # from the preservation_robots project on your laptop:
    $ bundle exec cap prod ssh

[...]

PR #1589, opened 2020-06-23
Move zipped moab version

### LINKED ISSUE #1530: ZippedMoabVersion should switch to referring to PreservedObject, rather than CompleteMoab

jmartin-sul:
## Why was this change made?

to update the data model, so that the `zipped_moab_versions` table (`ZippedMoabVersion` model) points to the `preserved_objects` table (`PreservedObject`) for the associated parent record, instead of the specific moab (`complete_moabs`/`CompleteMoab`).

the data model change allows easier support of multiple moabs for a single druid.  since we'll mostly be using that feature to move moabs between storage locations (which usually involves re-writing, via copy and delete, the bits on storage), it becomes apparent that tying the cloud archive copy to the abstract record for the druid as a whole (the `PreservedObject`) makes more sense than tying it to the slightly more ephemeral record for the specific moab on disk from which the archive copy was created (the `CompleteMoab`).

closes #1530

- [x] unit test updates
- [x] will deploy to stage to
  - [x] confirm the migration works
  - [x] confirm that replication still works after the update
  - [x] confirm that cloud archive audit still works after the update
- [x] will deploy to a clone of production (preservation-catalog-tester) to get a sense of timing for how long the migration will take on prod

ticketed separately, see #1588


app/jobs/moab_replication_audit_job.rb
```diff
@@ -8,24 +8,24 @@ class MoabReplicationAuditJob < ApplicationJob
   queue_as :moab_replication_audit
 
   # @param [CompleteMoab] verify that the zip exists on the endpoint
```
mjgiarlo, line 10: s/completemoab/preservedobject/

app/jobs/zipmaker_job.rb
```diff
@@ -4,13 +4,25 @@
...
+  # Does queue locking on ONLY druid and version (as first and second parameters)
+  def self.lock(*args)
+    "lock:#{name}-#{args.slice(0..1)}"
+  end
+
   # @param [String] druid
   # @param [Integer] version
-  def perform(druid, version)
+  # @param [String] moab_replication_path The path containing the druid tree from which the zipped version should be created
+  def perform(druid, version, moab_replication_path) # rubocop:disable Lint/UnusedMethodArgument
```
mjgiarlo, line 25: Please clarify in the job that the `moab_replication_path` argument *is* in fact used (on line 14 as `job.arguments.third`).

app/models/preserved_object.rb
```diff
@@ -7,8 +7,10 @@
 # those instances.
 class PreservedObject < ApplicationRecord
   PREFIX_RE = /druid:/i.freeze
+
```
mjgiarlo, line 10: :beers:

app/models/preserved_object.rb
```diff
@@ -21,7 +23,51 @@ class PreservedObject < ApplicationRecord
...
+  # Queue a job that will check to see whether this PreservedObject has been
+  # fully replicated to all target ZipEndpoints
+  def audit_moab_version_replication!
+    MoabReplicationAuditJob.perform_later(self)
+  end
+
+  # We need a specific copy of a moab from which to create the zip file(s) to send to the cloud.
+  # Of those eligible for replication, use whichever was checksum validated most recently.
+  # @return [String, nil] The storage location (storage_root/storage_trunk) where the Moab that should be replicated lives.
+  def moab_replication_storage_location
```
mjgiarlo, line 63: Consider moving to private method?

app/models/zip_endpoint.rb
```diff
@@ -65,6 +65,10 @@ def audit_class
     raise "Failed to return audit class based on setting for #{endpoint_name}.  Check setting string for accuracy."
   end
 
+  def to_s
+    endpoint_name
```
mjgiarlo, line 69: Add test coverage? Or CodeClimate may be mistaken.

jmartin-sul: i think codeclimate was right, added a test that specifically exercised this via the `AuditResults` `storage_area` generalization that prompted it, and diff coverage went up to 100%.

jmartin-sul:
@mjgiarlo, i think i've implemented all the suggested feedback.  i noticed a couple other comments that needed updating, so i got those too.  also, in fixing the small test coverage drop related to the `AuditResults` `storage_area` generalization, i noticed an opportunity to simplify an `AuditResults` call in `MoabValidator#update_status` by eliminating some redundant parameter passing, so i threw that change on as one more commit.

jmartin-sul:
whelp, the migration took quite a while on a production clone when i did a test deployment last night:
```
00:33 deploy:migrating
      01 bundle exec rake db:migrate
      01 == 20200527171011 AddPreservedObjectRefToZippedMoabVersions: migrating ========
      01 -- add_reference(:zipped_moab_versions, :preserved_object, {:foreign_key=>true, :null=>true})
      01    -> 8.9116s
      01 -- execute("UPDATE zipped_moab_versions AS zmv SET preserved_object_id = (SELECT preserved_object_id FROM complete_…
      01    -> 515.3899s
      01 -- change_column_null(:zipped_moab_versions, :preserved_object_id, false)
      01    -> 3.2088s
      01 -- remove_foreign_key(:zipped_moab_versions, :complete_moabs)
      01    -> 0.0613s
      01 -- remove_column(:zipped_moab_versions, :complete_moab_id)
      01    -> 0.0195s
      01 == 20200527171011 AddPreservedObjectRefToZippedMoabVersions: migrated (527.5929s)
      01
      01 == 20200528161829 MoveLastArchiveAuditToPreservedObjects: migrating ===========
      01 -- add_column(:preserved_objects, :last_archive_audit, :datetime)
      01    -> 39182.6025s
      01 -- execute("UPDATE preserved_objects AS po SET last_archive_audit = (SELECT cm.last_archive_audit FROM complete_moa…
      01    -> 66.7607s
      01 -- add_index(:preserved_objects, :last_archive_audit)
      01    -> 2.0120s
      01 -- remove_column(:complete_moabs, :last_archive_audit)
      01    -> 0.0044s
      01 == 20200528161829 MoveLastArchiveAuditToPreservedObjects: migrated (39251.3799s)
      01
    ✔ 01 pres@preservation-catalog-tester.stanford.edu 39783.541s
663:36 db_seed
      01 bundle exec rake db:seed
      01 seeded database.  state of seeded object types after seeding:
      01 > PreservationPolicy.all: [#<PreservationPolicy id: 1, preservation_policy_name: "default", archive_ttl: 7776000, f…
      01 > MoabStorageRoot.all: [#<MoabStorageRoot id: 12, name: "services-disk13", created_at: "2018-01-18 18:55:35", updat…
      01 > ZipEndpoint.all: [#<ZipEndpoint id: 1, endpoint_name: "aws_s3_west_2", delivery_class: S3WestDeliveryJob, endpoin…
    ✔ 01 pres@preservation-catalog-tester.stanford.edu 4.402s
663:41 whenever:update_crontab
```

the really perplexing part to me is that it takes ~40k seconds just to add a date column to `preserved_objects`.  but only about 60s to populate it using an `UPDATE` with a subquery, which afaik is not the speediest thing in database-land.
i'm assuming we'd like to avoid taking 11 hours of pres robots downtime to run a migration (and hoping that a terminal session stays connected for an 11 hr cap deploy also seems like tempting fate).  i'm also assuming we want to turn pres robots off for this migration, because we don't want stuff being updated while it's running.

paired w/ @jermnelson to bounce around ideas for a bit.  leads so far:
* is AR's `add_column` doing what we'd expect (i.e. just an `ALTER TABLE preserved_objects ADD COLUMN last_archive_audit TIMESTAMP` (or whatever the type is that AR uses for what it calls `datetime`)?
* would it help to do this particular column add separately, maybe outside a transaction, since that's one thing that could be separated and done as prep for the rest of the migration?

jmartin-sul:
in better news, the migration _did_ seem to do the right thing with the data.  i dumped a mapping of all `zipped_moab_versions.id` values to their respective druids, both before and after the migration, and the output was the same when the migration was run on a clone of prod DB.  so that's great, at least.

jmartin-sul:
julian agreed with the transactionality suspicion, so i did some testing on that front, modifying the second part of the migration to not be entirely wrapped in a transaction.  i put the column add outside the transaction and wrapped just the `UPDATE` and the `add_index` in a transaction.  diff of this branch vs the one with the original migration (`6b0414a3` last time it was pushed):

```diff
diff --git a/db/migrate/20200528161829_move_last_archive_audit_to_preserved_objects.rb b/db/migrate/20200528161829_move_last_archive_audit_to_preserved_objects.rb
index 268c0b6e..465f1836 100644
--- a/db/migrate/20200528161829_move_last_archive_audit_to_preserved_objects.rb
+++ b/db/migrate/20200528161829_move_last_archive_audit_to_preserved_objects.rb
@@ -1,10 +1,17 @@
 class MoveLastArchiveAuditToPreservedObjects < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction! # note: this actually disables the wrapping transaction for the entire migration
+
   def change
     add_column :preserved_objects, :last_archive_audit, :datetime
-    execute <<-SQL
-UPDATE preserved_objects AS po SET last_archive_audit = (SELECT cm.last_archive_audit FROM complete_moabs AS cm WHERE cm.preserved_object_id = po.id);
-    SQL
-    add_index :preserved_objects, :last_archive_audit
+
+    # and then we explicitly make this update transactional
+    ApplicationRecord.transaction do
+      execute <<-SQL
+        UPDATE preserved_objects AS po SET last_archive_audit = (SELECT cm.last_archive_audit FROM complete_moabs AS cm WHERE cm.preserved_object_id = po.id);
+      SQL
+      add_index :preserved_objects, :last_archive_audit
+    end
+
     remove_column :complete_moabs, :last_archive_audit
   end
 end
```

and that seemed to do the trick!  tony z refreshed `preservation-catalog-tester` with the same DB snapshot, to which i first deployed the current master branch, then this branch (as currently pushed w/ the modified migration).  on that deploy, the entire migration took just under 10 min:
```console
00:34 deploy:migrating
      01 bundle exec rake db:migrate
      01 == 20200527171011 AddPreservedObjectRefToZippedMoabVersions: migrating ========
      01 -- add_reference(:zipped_moab_versions, :preserved_object, {:foreign_key=>true, :null=>true})
      01    -> 8.9045s
      01 -- execute("UPDATE zipped_moab_versions AS zmv SET preserved_object_id = (SELECT preserved_object_id FROM complete_…
      01    -> 506.9832s
      01 -- change_column_null(:zipped_moab_versions, :preserved_object_id, false)
      01    -> 3.2617s
      01 -- remove_foreign_key(:zipped_moab_versions, :complete_moabs)
      01    -> 0.2826s
      01 -- remove_column(:zipped_moab_versions, :complete_moab_id)
      01    -> 0.0020s
      01 == 20200527171011 AddPreservedObjectRefToZippedMoabVersions: migrated (519.4342s)
      01
      01 == 20200528161829 MoveLastArchiveAuditToPreservedObjects: migrating ===========
      01 -- add_column(:preserved_objects, :last_archive_audit, :datetime)
      01    -> 0.0014s
      01 -- execute("        UPDATE preserved_objects AS po SET last_archive_audit = (SELECT cm.last_archive_audit FROM comp…
      01    -> 64.8179s
      01 -- add_index(:preserved_objects, :last_archive_audit)
      01    -> 1.8957s
      01 -- remove_column(:complete_moabs, :last_archive_audit)
      01    -> 0.0634s
      01 == 20200528161829 MoveLastArchiveAuditToPreservedObjects: migrated (66.7985s) =
      01
    ✔ 01 pres@preservation-catalog-tester.stanford.edu 590.663s
10:25 db_seed
      01 bundle exec rake db:seed
      01 seeded database.  state of seeded object types after seeding:
      01 > PreservationPolicy.all: [#<PreservationPolicy id: 1, preservation_policy_name: "default", archive_ttl: 7776000, f…
      01 > MoabStorageRoot.all: [#<MoabStorageRoot id: 12, name: "services-disk13", created_at: "2018-01-18 18:55:35", updat…
      01 > ZipEndpoint.all: [#<ZipEndpoint id: 1, endpoint_name: "aws_s3_west_2", delivery_class: S3WestDeliveryJob, endpoin…
    ✔ 01 pres@preservation-catalog-tester.stanford.edu 5.357s
10:31 whenever:update_crontab
```

mjgiarlo:
:clap: :beer:

jmartin-sul:
#### the rest of the deployment plan
- [ ] kill pres robots and pres cat workers on stage
- [ ] wait for everything that's in progress to finish
- [ ] run the before part of the sanity check queries against stage and save the results (see below)
- [ ] deploy to stage and test
  - [ ] run the after part of the sanity check queries against stage and save the results (see below)
  - [ ] do some manual sanity checks, examining relationships between random `ZippedMoabVersion`s and `PreservedObject`s.  just a first, or a random grab (`.order('random()').limit(1)`).
    - [ ] start pres robots and pres cat workers back up
  - [ ] accession something new
  - [ ] version something that exists (could cover both with integration tests?)
  - [ ] make sure both got replicated
- [ ] if all goes well on stage
  - [ ] kill pres robots, turn pres cat worker count to zero (see #1558 for instructions on both).
    - [ ] let in process stuff finish (both pres robots and pres cat workers)
    - [ ] run the before part of the sanity check queries against prod and save the results (see below)
    - [ ] snapshot of preservation-catalog-prod DB to be safe
    - [ ] merge this branch and deploy to prod, wait for migration to finish
    - [ ] run the after part of the sanity check queries against prod and save the results (see below)
    - [ ] do some manual sanity checks, examining relationships between random `ZippedMoabVersion`s and `PreservedObject`s.  just a first, or a random grab (`.order('random()').limit(1)`).
    - [ ] if sanity checks look off, bail and rollback to snapshot
    - [ ] if good, re-enable pres cat workers, re-start pres robots
    - [ ] accession something new
    - [ ] version something that exists
    - [ ] make sure both get replicated
    - [ ] keep an eye on things for a few days

#### sanity check queries:
###### before migration
```ruby
# query *before* data model change
druid_to_zmv_id_list = ZippedMoabVersion.joins(complete_moab: [:preserved_object]).order(:druid, 'zipped_moab_versions.id').pluck(:druid, 'zipped_moab_versions.id') ; nil
filename = "log/druid_to_zmv_id_list_before_sorted_#{DateTime.now.utc.iso8601.gsub(':', '')}.csv"
CSV.open(filename, 'w') { |csv| druid_to_zmv_id_list.each { |line| csv << line } } ; nil
```

###### after migration
```ruby
# query *after* data model change
druid_to_zmv_id_list = ZippedMoabVersion.joins(:preserved_object).order(:druid, 'zipped_moab_versions.id').pluck(:druid, 'zipped_moab_versions.id') ; nil
filename = "log/druid_to_zmv_id_list_after_sorted_#{DateTime.now.utc.iso8601.gsub(':', '')}.csv"
CSV.open(filename, 'w') { |csv| druid_to_zmv_id_list.each { |line| csv << line } } ; nil
```

once you have the second output file, diff it against the first.  if everything that was in progress had settled before the first query, and no robots or workers were restarted till after the second query, the lists should have no difference.

i'm disappointed to ask for it with downtime already planned for both the CentOS 7 upgrade and for the storage migration, but something like 30 minutes of expected accessioning pause might be a helpful thing here, mostly so that we can test that things worked by versioning something ourselves before letting things resume full speed ahead.

jmartin-sul:
tested on stage, and replication appears to work correctly when it's triggered (database looks good, cloud copies look good).  however, when i ran the integration test that creates a DOR object with two versions, the second version wasn't replicated automatically.  pres cat's database record for the local moab as a whole looked good, the first moab version automatically replicated, and the first version's cloud copy looked good.  however, i had to manually call `#create_zipped_moab_versions!` on the preserved object from rails console after the second version was accessioned to get the version to replicate.  once i did that, the cloud copies looked fine for both versions.

will investigate further on monday.

in the meantime, pres cat stage DB has been restored to a snapshot from earlier this afternoon (taken before this branch was deployed for testing), since the data model change isn't backwards compatible with the code that's currently on master.

jmartin-sul:
fix for replication bug in this PR was tested on stage, fix seemed to work, notes on that PR (which was just merged into this one).

working w/ ops now to coordinate deployment to prod.

jmartin-sul:
this has been deployed to production.  will keep an eye on things.  initial manual testing and observation of accessioning that was happening anyway looks good.

Issue #1591, opened 2020-06-23
test CompleteMoabHandler.create with multiple moabs for the same druid


jmartin-sul:
i noticed that one spot we missed in testing for the #1535 related changes was complete_moab_handler_create_spec.rb.

i created a branch that takes a super simple pass at adding a test along these lines, and it fails at the moment.

the current application behavior is that if a `CompleteMoab` exists for a `PreservedObject`, and the catalog adds another `CompleteMoab` for the same `PreservedObject` (for a copy of the moab on another storage root for the same druid), the `PreservedObject` will not have its `current_version` bumped to match the higher of the two.

i think this behavior is defensible, but we should probably get @andrewjbtw to weigh in on the desired behavior.  in which case, my branch could be fixed by updating the test code, leaving the app code as-is, and using the test to illustrate this behavior explicitly.

the other approach would be that creating a record for a `CompleteMoab` with a higher version than we'd seen for that druid would automatically bump `current_version` on the `PreservedObject`.  if we choose this second approach, i think we'd update `CompleteMoabHandler#create_db_objects` accordingly (since that's what's used underneath by both `CompleteMoabHandler#create` and `CompleteMoabHandler#check_existence`, the latter of which creates catalog entries for moabs found by the `MoabToCatalogJob`).

realistically, i think the usage we're envisioning is that we call `CompleteMoabHandler#create` or `CompleteMoabHandler#check_existence` on a moab that we're adding as a secondary copy, maybe in preparation for switching to it later as the primary (e.g. on new storage hardware).  as such, the methods in question will likely be called on moabs of the same version, or possibly a lower version (as copies that lag the main copy).  so encountering a moab with a new highest version would be unexpected, i think (@andrewjbtw?).

if we do leave the behavior as-is, we should create a ticket to update phrasing in the app and documentation that refers to `PreservedObject#current_version` as indicating the highest version seen of the moab.  not a super high priority, but that will no longer technically be true if we choose the current behavior.

jmartin-sul:
see https://github.com/sul-dlss/preservation_catalog/pull/1592 for failing test case.

jmartin-sul:
assigning to @andrewjbtw and i while desired behavior is figured out, will kick back to ready or re-ticket once we know what we want to implement.

jmartin-sul:
since we are removing multi-moab support (since it is unused complexity), i'm closing this as moot.

PR #1592, opened 2020-06-23
[WIP] add a test (currently failing) exercising CompleteMoabHandler#c…

jmartin-sul:
## Why was this change made?

connects to #1591

test only change (shores up unit tests to exercise support of more than one moab for a given druid).

TBD based on desired behavior, see #1591

### DIFF
--- a/spec/services/complete_moab_handler_create_spec.rb
+++ b/spec/services/complete_moab_handler_create_spec.rb
@@ -53,6 +53,23 @@ RSpec.describe CompleteMoabHandler do
       expect(results).to include(a_hash_including(code => a_string_matching('CompleteMoab db object already exists')))
     end
 
+    context 'a moab (for the same druid) on a different storage root already exists' do
+      before do
+        po = create(:preserved_object, druid: druid)
+        other_storage_root = create(:moab_storage_root)
+        create(:complete_moab, preserved_object: po, moab_storage_root: other_storage_root)
+      end
+
+      it 'creates the new CompleteMoab in database' do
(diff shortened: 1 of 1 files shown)


spec/services/complete_moab_handler_create_spec.rb
```diff
@@ -52,6 +52,23 @@
...
+        po = create(:preserved_object, druid: druid)
+        other_storage_root = create(:moab_storage_root)
+        create(:complete_moab, preserved_object: po, moab_storage_root: other_storage_root)
+      end
+
+      it 'creates the new CompleteMoab in database' do
+        complete_moab_handler.create
+        existing_po = PreservedObject.find_by(druid: druid)
+        new_cm = existing_po.complete_moabs.find_by(moab_storage_root: ms_root)
+        expect(existing_po.current_version).to eq incoming_version
```
jmartin-sul, line 66: i believe this test would pass if this line were changed to be

```ruby
expect(existing_po.current_version).to eq existing_cm.version
```

and if the `before` block were turned into 3 `let` statements (or whatever similar approach):
```ruby
let(:po) { create(:preserved_object, druid: druid) }
let(:other_storage_root) { create(:moab_storage_root) }
let(:existing_cm) { create(:complete_moab, preserved_object: po, moab_storage_root: other_storage_root) }
```

jmartin-sul:
since we are removing multi-moab support (since it is unused complexity), i'm closing this as moot.

see also #1892 and #1891

Issue #1585, opened 2020-06-22
initial population of primary moab table


jmartin-sul:
currently, we only have one moab per druid.  before we start using multiple-moabs-per-druid functionality, insert rows into the primary moab tracking table to mark the moabs we have as primaries.

blocked by #1583, so we don't have to do this repeatedly.  as such, this migration should tolerate cases where there's already a primary CompleteMoab for a PreservedObject.

PR #1597, opened 2020-06-24
Populates preserved_objects_primary_moabs table.

justinlittman:
closes https://github.com/sul-dlss/preservation_catalog/issues/1585

## Why was this change made?
Provide initial population of primary moabs.

Local.

No.


jmartin-sul:
discussed post-standup, @jcoyne says all good, and that the worry was specifically with e.g. model code that might change or disappear as the codebase shifts (which could render the old migration broken in the future).  not specifically a worry here, since this migration is a plain SQL `INSERT`.

PR #1606, opened 2020-06-30
PreservedObjectPrimaryMoab custom validation

### LINKED ISSUE #1601: AR validation: complete_moab matches preserved_object
the `#complete_moab` for a `PreservedObjectsPrimaryMoab` should belong to the `#preserved_object`, since this model represents a join table that associates a `PreservedObject` (abstract unifying record for a druid) with its primary `CompleteMoab` (a copy on disk of a moab for a given druid).

jermnelson:
## Why was this change made?
Fixes #1601

Added unit test to suite

n/a

### DIFF
--- a/app/models/preserved_objects_primary_moab.rb
+++ b/app/models/preserved_objects_primary_moab.rb
@@ -5,4 +5,10 @@
 class PreservedObjectsPrimaryMoab < ApplicationRecord
   belongs_to :preserved_object, inverse_of: :preserved_objects_primary_moab
   belongs_to :complete_moab, inverse_of: :preserved_objects_primary_moab
+
+  validate :complete_moab_matches_preserved_object
+
+  def complete_moab_matches_preserved_object
+    raise "Complete Moab #{complete_moab.preserved_object_id} does not match" unless complete_moab.preserved_object == preserved_object
+  end
 end
--- a/app/models/preserved_objects_primary_moab.rb
+++ b/app/models/preserved_objects_primary_moab.rb
@@ -9,6 +9,7 @@ class PreservedObjectsPrimaryMoab < ApplicationRecord
   validate :complete_moab_matches_preserved_object
 
   def complete_moab_matches_preserved_object
-    raise "Complete Moab #{complete_moab.preserved_object_id} does not match" unless complete_moab.preserved_object == preserved_object
+    err_msg = 'must match the preserved object associated with the complete moab'
+    errors.add(:preserved_object, err_msg) unless complete_moab.preserved_object == preserved_object
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/preserved_objects_primary_moab.rb
```diff
@@ -5,4 +5,10 @@
 class PreservedObjectsPrimaryMoab < ApplicationRecord
   belongs_to :preserved_object, inverse_of: :preserved_objects_primary_moab
   belongs_to :complete_moab, inverse_of: :preserved_objects_primary_moab
+
+  validate :complete_moab_matches_preserved_object
+
+  def complete_moab_matches_preserved_object
+    raise "Complete Moab #{complete_moab.preserved_object_id} does not match" unless complete_moab.preserved_object == preserved_object
```
jermnelson, line 12: This currently raises a `RuntimeError` open to suggestions about more specific error that should be raised instead.

mjgiarlo: I haven't read the issue linked to this PR, but I wonder if we want to raise an exception here or populate the validation errors list, e.g.:

```ruby
def complete_moab_matches_preserved_object
  if complete_moab.preserved_object != preserved_object
    errors.add(:preserved_object, 'must match the preserved object associated with the complete moab')
  end
end
```

I might be surprised by a validation raising an error instead of conforming to the validation interface.

spec/models/preserved_objects_primary_moab_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  let(:cm) { create(:complete_moab) }
+  let(:po) { cm.preserved_object }
+  let(:alt_po) { create(:preserved_object) }
+
+  it 'is valid when primary moab preserved objects agree' do
+    expect(described_class.new(complete_moab: cm, preserved_object: po)).to be_valid
+  end
+
+  it 'raises error if complete moab preserved object differs from primary moab' do
+    expect { described_class.create(complete_moab: cm, preserved_object: alt_po) }.to raise_error(RuntimeError, /does not match/)
```
mjgiarlo, line 15: Per above, I might expect this test to check `.not_to be_valid` and to ensure the errors list is populated with the expected error (rather than raising).

app/models/preserved_objects_primary_moab.rb
jermnelson: Thanks @mjgiarlo, my first implementation was using the validation error list so I can easily revert back.

mjgiarlo: haha! It happens. :smiley:  Thanks, @jermnelson

mjgiarlo:
@jermnelson Thanks for taking my suggestions!

PR #1608, opened 2020-06-30
Add column to PreservedObject table to flag whether robot versioning is allowed

### LINKED ISSUE #1598: add a preserved_objects.robot_versioning_allowed column
Proposed approach:
- [x] Add a boolean column to `preserved_objects.robot_versioning_allowed`.  It should disallow `NULL`, and it should default to `true`.
- [x] Modify `ObjectsController#primary_moab_location` to respond with an error code if `robot_versioning_allowed` is false for the `PreservedObject` for the druid.  I'm not sure what response code to use for this.  409 Conflict?  423 Locked?  I lean very slightly towards 423.  It seems like most semantically appropriate ("The resource that is being accessed is locked"), though wiki indicates it's a WebDAV code.  409 isn't wrong ("indicates... request could not be processed because of conflict in the current state of the resource"), exactly, and it is a response that's in general use.  423 just seems to describe the actual situation better, and we're the only consumers of this API.  Either way, provide an informative error message.
  - `PreservedObject` info like this new field should already be available in the query `#primary_moab_location` uses, so maybe just assign the result to a var instead of calling it in the `render`, examine the `robot_versioning_allowed` field, and return the `storage_location` or an error code as appropriate.

see also https://github.com/sul-dlss/preservation_robots/issues/242

### why do this?

This is meant to help manage (but not entirely prevent by itself) a race condition such as the following:
1. A pres robots worker gets the location for the primary moab for a druid, and starts writing a new version that takes a while to finish (e.g. because the new version has many GB of video content).
2. One of the catalog maintainers, unaware this write has started, sets a different copy of the moab to be the primary.
3. The new primary, not having had a final sync from the old primary, is now behind what it should be by one version.

This also explains why I don't think this flag should block calls to e.g. `CatalogController#create` or `CatalogController#update`.  Those only get called by a pres robots worker after the moab's files are already updated, so disallowing those writes would likely only serve to keep the catalog from accurately reflecting the state of our content.

### caveats

Any updates from pres robots that were in progress when `robot_versioning_allowed` is set to false will have to finish if versioning is being temporarily disabled to allow a final sync to a moab copy that isn't yet primary (e.g. when migrating to new storage).  Thus, a bit of resque dashboard watching will still be necessary.  But, by programmatically blocking further writes by pres robots to specific druids, small to medium sized batches of moabs could be set to disallow robot versioning without announcing a pause in accessioning (or relying on people and processes to honor the pause, or needing to kill pres robots entirely).  For larger migrations, an intentional pause in accessioning is probably wise, to prevent failed robot jobs from piling up.

### alternate approachs

Don't automate this part, just allow room for the cutover by way of accessioning pauses and pres robot shut downs.

see also https://github.com/sul-dlss/preservation_catalog/issues/1600

mjgiarlo:
Fixes #1598 

## Why was this change made?

Includes:
* Add robot_versioning_allowed boolean field to PreservedObject table
* Return HTTP 423 from primary_moab_location endpoint when robot versioning not allowed for a preserved object
* Update OpenAPI spec to include 423 response for primary_moab_location endpoint

Opportunistic additions:
* Add an initialization arg to Committee in preparation for new version (and to silence a warning)
* Regenerate Rubocop TODOs

Unit tests

OpenAPI specification.


spec/requests/objects_controller_primary_moab_spec.rb
```diff
@@ -1,32 +1,48 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+
 RSpec.describe ObjectsController, type: :request do
+  let(:pres_obj) { primary_moab.preserved_object }
   let(:primary_moab) do
-    create(:complete_moab).tap { |cm| PreservedObjectsPrimaryMoab.create(complete_moab: cm, preserved_object: cm.preserved_object) }
+    create(:complete_moab) do |cm|
```
mjgiarlo, line 8: `#tap` not needed here since FactoryBot's `create` method takes an optional block.

spec/requests/objects_controller_primary_moab_spec.rb
```diff
@@ -1,32 +1,48 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+
 RSpec.describe ObjectsController, type: :request do
+  let(:pres_obj) { primary_moab.preserved_object }
```
mjgiarlo, line 6: Organized `let` statements alphabetically.

spec/requests/objects_controller_primary_moab_spec.rb
```diff
@@ -1,32 +1,48 @@
...
-    create(:complete_moab).tap { |cm| PreservedObjectsPrimaryMoab.create(complete_moab: cm, preserved_object: cm.preserved_object) }
+    create(:complete_moab) do |cm|
+      PreservedObjectsPrimaryMoab.create(complete_moab: cm, preserved_object: cm.preserved_object)
+    end
   end
-  let(:pres_obj) { primary_moab.preserved_object }
 
   describe 'GET #primary_moab_location' do
-    context 'when object found' do
-      it 'response contains the locaiton of primary moab of the given prefixed druid' do
```
mjgiarlo, line 12: tpyo

mjgiarlo:
@jmartin-sul :speech_balloon: 
> this LGTM, but let's hold for a bit, till pres cat CentOS migration is over.
> 
> actually, slight worry this migration will take a while, since a col is being added to `preserved_objects` (worried about this running into the same thing as the first version of the #1589 migration ran into). maybe we should make the migration non-transactional, since it's just a (seemingly safe) column add?

Did I get this right? https://github.com/sul-dlss/preservation_catalog/pull/1608/files#diff-373068d233f24094bce58ea42771bb4bR2-R5

jmartin-sul:
> @jmartin-sul 💬
> 
> > this LGTM, but let's hold for a bit, till pres cat CentOS migration is over.
> > actually, slight worry this migration will take a while, since a col is being added to `preserved_objects` (worried about this running into the same thing as the first version of the #1589 migration ran into). maybe we should make the migration non-transactional, since it's just a (seemingly safe) column add?
> 
> Did I get this right? https://github.com/sul-dlss/preservation_catalog/pull/1608/files#diff-373068d233f24094bce58ea42771bb4bR2-R5

looks right to me, @mjgiarlo!

jmartin-sul:
i'm not sure i understand code climate's gripe

mjgiarlo:
@jmartin-sul beats me. worth trying to convince codeclimate otherwise, or do you think this is approve/merge-ready?

jmartin-sul:
> @jmartin-sul beats me. worth trying to convince codeclimate otherwise, or do you think this is approve/merge-ready?

nah, not too worried about it, i'll approve, and then we can override the check failure when the CentOS upgrades are done and we're ready to merge/deploy.

Issue #1600, opened 2020-06-26
[BLOCKED] document storage migration procedure once we have support in place for more than one moab per druid


jmartin-sul:
we need to see how much we get done in the last week of this WC in terms of writing code that automates/manages allowing time for final sync to target storage, but at a very high level, this is:

1. moab on old storage is synced to counterpart moab on new storage
2. after sync finishes, and before any more writes are made to the copy on old storage by pres robots, switch the primary for the `PreservedObject` to point to the `CompleteMoab` on the target storage.  thus, further updates from pres robots will flow to the new primary on new storage, no more syncing can be needed, and once the old storage no longer has any primary moabs, it can be decommissioned.

for example, the assurance that no new writes were made could be provided in at least a couple ways:
* shut down pres robots entirely, wait for all write activity to finish for versioning work that's already been picked up.  cut over some moabs, bring pres robots back up.
* instead of shutting down pres robots entirely, ask for accessioning to stop on some aggregation of druids (or pause a more limited scope of automated processing that results in versioning, e.g. google books).  make sure that all versioning jobs picked up for the last updates to those druids has settled, make sure they're synced, cut over primary, give the all clear to resume work on those druids
* as described in #1598, add a flag that returns an error code when pres robots ask (via REST call to pres cat) for the location to update for a new druid version, resulting in pres robots not being able to make the update.  wait for versioning that's already been started on any of those druids to settle (since the flag may have been set, e.g., right after a version with a couple GB of video gets ingested).  cut over primary on those druids.
  * could be combined with intentional accessioning pause approach to reduce count of failed workflow steps if a significant amount of content were to be cut over to new primary storage all at once.

the approaches that don't involve entirely killing robots allow for somewhat less disruptive cut overs for moderately sized batches of druids, say a few hundred to a few thousand at a time, which might allow for more incremental and less disruptive storage migrations going forward.

jmartin-sul:
since we are removing multi-moab support (since it is unused complexity), i'm closing this as moot.

see also #1892 and #1891

Issue #1595, opened 2020-06-24
update wiki for recent data model changes


jmartin-sul:
* adding the notion of a "primary" moab (#1528, #1577)
* `zipped_moab_versions` points to `preserved_objects` instead of `complete_moabs` (#1530, #1589)
* `CompleteMoab` should be queried for by druid _and_ storage root (not just druid) (technically this is the app code taking better advantage of the data model, not a change to the underlying schema itself)

browse the wiki, and look for pages that have to change in light of the above.  for example, i'd expect the `zipped_moab_versions` change to necessitate a change to:  https://github.com/sul-dlss/preservation_catalog/wiki/Investigating-a-druid-with-replication-errors

jmartin-sul:
here's what i think the updated investigation code looks like for https://github.com/sul-dlss/preservation_catalog/wiki/Investigating-a-druid-with-replication-errors -- tested while testing https://github.com/sul-dlss/preservation_catalog/pull/1614 and https://github.com/sul-dlss/preservation_catalog/pull/1589, and it seemed to work:

```ruby
druid = 'bc123df4567'

po = PreservedObject.find_by!(druid: druid)
​
po.zipped_moab_versions

# see https://github.com/sul-dlss/preservation_catalog/blob/master/db/README.md#view-the-zip-parts-for-a-given-druid
ZipPart.joins(
  zipped_moab_version: [:preserved_object, :zip_endpoint]
).where(
  preserved_objects: { druid: druid }
).order(
  :druid, :zip_version, :endpoint_name, :suffix # you could also provide an array of druids instead of `druid`, in which case this order will make for more readable results
).pluck(
  :druid, 'current_version AS highest_version', 'zipped_moab_versions.version AS zip_version', :endpoint_name, :status, :suffix, :parts_count, :size, :updated_at
)
​
# returns e.g.
# [["bc123df4567", 1, 1, "aws_s3_west_2", "unreplicated", ".z01", 2, 10737418240],
#  ["bc123df4567", 1, 1, "aws_s3_west_2", "ok", ".zip", 2, 10667110264],
#  ["bc123df4567", 1, 1, "aws_s3_east_1", "ok", ".z01", 2, 10737418240],
#  ["bc123df4567", 1, 1, "aws_s3_east_1", "ok", ".zip", 2, 10667110264],
#  ["bc123df4567", 1, 1, "ibm_us_south", "ok", ".zip", 2, 10667110264],
[...]

PR #1616, opened 2020-07-15
Only update PreservedObject#current version if CompleteMoab being checked/updated is primary

jmartin-sul:
## Why was this change made?

I didn't get around to ticketing this, but I discussed this with @andrewjbtw at the beginning of last week.

Our expectation is that pres robots will keep the primary updated, and the non-primary copy will be synced out of band from pres cat and pres robots to keep it up to date, using the primary copy as the syncing source, e.g. in preparation for migrating moab content to new storage hardware (allowing content to be synced and validated gradually from within the catalog, without needing to shut down pres cat and migrate DB entries while it refrains from doing anything with storage, so as to only see one copy of a moab in one known location at a time).  Alternatively, a non-primary copy might be created when retrieving a cloud copy, e.g. for use in remediating a damaged local copy.

In both of those scenarios, it would be surprising for a non-primary copy to have a higher version than the primary, and in general the "primary" copy is expected to be on the correct version, so we'll only update `current_version` on the parent `PreservedObject` when we see an increase in the `version` for the primary `CompleteMoab` for a druid.  Previously, we generally described `current_version` as the highest version we'd seen so far for the moab.  Now we're just explicitly qualifying that to apply to the primary copy only.

unit tests, but i tried to do much of it through integration-y tests that dealt with actual moab fixtures on disk, at least for the new tests that i added.

for some of the existing tests, i just explicitly marked the `CompleteMoab` on which expectations were set as the primary, since that is the role those test objects have been playing so far.

documentation is covered under other tickets (#1600, #1588, #1595).  configuration n/a here.

### DIFF
--- a/app/services/complete_moab_handler.rb
+++ b/app/services/complete_moab_handler.rb
@@ -228,6 +228,8 @@ class CompleteMoabHandler
       raise_rollback_if_cm_po_version_mismatch
 
       # FIXME: what if there is more than one associated complete_moab?
+      # TODO: is the complete_moab.matches_po_current_version? even useful, since we should've raised an error with
+      # the call to raise_rollback_if_cm_po_version_mismatch if that condition weren't met?
       if incoming_version > complete_moab.version && complete_moab.matches_po_current_version?
         # add results without db updates
         code = AuditResults::ACTUAL_VERS_GT_DB_OBJ
--- a/app/services/complete_moab_handler.rb
+++ b/app/services/complete_moab_handler.rb
@@ -186,6 +186,10 @@ class CompleteMoabHandler
     @moab_validator ||= MoabValidator.new(druid: druid, storage_location: storage_location, results: results)
   end
 
+  def primary_moab?
+    @primary_moab ||= complete_moab.primary?
+  end
+
   # Note that this may be called by running M2C on a storage root and discovering a second copy of a Moab,
   #   or maybe by calling #create_after_validation directly after copying a Moab
   def create_db_objects(status, checksums_validated = false)
@@ -207,7 +211,7 @@ class CompleteMoabHandler
     transaction_ok = with_active_record_transaction_and_rescue do
       this_po = PreservedObject
                 .find_or_create_by!(druid: druid) do |po|
-                  po.current_version = incoming_version
+                  po.current_version = incoming_version # init to match version of the first moab for the druid
                   po.preservation_policy_id = ppid
                 end
       this_cm = this_po.complete_moabs.create!(cm_attrs)
@@ -233,7 +237,7 @@ class CompleteMoabHandler
         complete_moab.last_checksum_validation = Time.current if checksums_validated && complete_moab.last_checksum_validation
         update_status(status) if status
         complete_moab.save!
-        pres_object.current_version = incoming_version
+        pres_object.current_version = incoming_version if primary_moab? # we only want to track highest seen version based on primary
         pres_object.save!
       else
         status = 'unexpected_version_on_storage' if set_status_to_unexp_version
@@ -247,7 +251,7 @@ class CompleteMoabHandler
   def update_cm_po_set_status
     if moab_validation_errors.empty?
       complete_moab.upd_audstamps_version_size(ran_moab_validation?, incoming_version, incoming_size)
-      pres_object.current_version = incoming_version
+      pres_object.current_version = incoming_version if primary_moab? # we only want to track highest seen version based on primary
       pres_object.save!
     else
       update_status('invalid_moab')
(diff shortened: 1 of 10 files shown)


app/services/complete_moab_handler.rb
```diff
@@ -223,8 +227,7 @@ def update_online_version(status = nil, set_status_to_unexp_version = false, che
     transaction_ok = with_active_record_transaction_and_rescue do
       raise_rollback_if_cm_po_version_mismatch
 
-      # FIXME: what if there is more than one associated complete_moab?
-      if incoming_version > complete_moab.version && complete_moab.matches_po_current_version?
```
jmartin-sul, line 227: per the commit message for this change, second half of the condition seemed unnecessary in light of the `raise_rollback_if_cm_po_version_mismatch` call above.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -151,13 +162,32 @@
...
       end
+
+      context 'updating a non-primary' do
+        let!(:pres_obj) { create(:preserved_object, druid: bare_druid, current_version: 1) } # as if the one on srA was always primary
+        let!(:comp_moab) { create(:complete_moab, preserved_object: pres_obj, version: 3) } # create fixture_sr1 record, not created w/ PO this case
+        let(:primary_moab) { comp_moab_sr_a } # but we're still doing PATCH on the fixture_sr1 moab
+
+        it 'updates CompleteMoab#version' do
+          pending('this is known to fail, because CMH does not update CompleteMoab if its version does not match parent PO#current_version')
+          expect(comp_moab.reload.version).to eq upd_version
```
jmartin-sul, line 184: will ticket this as a bug (the failure is that the `CompleteMoab` record for this druid on this storage root doesn't get its `#version` updated to 3 as it should).  not ideal, but also we don't really have an expected use case where a non-primary is of a higher version than the primary, as is the case here.  and in the opposite case (as seen in other tests), the primary gets its `#version` updated properly, because it should stay in sync with the parent `PreservedObject` (i _think_ maybe the lack of sync between the non primary CM and its parent PO is causing the trouble here).

app/services/complete_moab_handler.rb
```diff
@@ -247,14 +250,16 @@ def update_online_version(status = nil, set_status_to_unexp_version = false, che
...
-      pres_object.current_version = incoming_version
+      pres_object.current_version = incoming_version if checking_primary_moab? # we only want to track highest seen version based on primary
       pres_object.save!
     else
       update_status('invalid_moab')
     end
   end
 
   def raise_rollback_if_cm_po_version_mismatch
+    return unless checking_primary_moab?
```
jmartin-sul, line 261: we're not so worried if a non-primary is not in sync with the parent `PreservedObject`, since only the primary is expected to be kept up to date automatically by pres robots.

spec/services/complete_moab_handler_check_exist_spec.rb
```diff
@@ -429,6 +429,46 @@
         end
       end
 
+      context 'there are two moabs for the druid' do
```
jmartin-sul, line 432: this is a bit redundant in light of the `spec/jobs/moab_to_catalog_job_spec.rb` addition.  i actually wrote this one first, then thought it'd be slightly better if it were just a bit more end-to-end, but for some reason i was also hesitant to remove this, since this is the spec file for the meat of the implementation.

happy to trim one or the other if people would prefer.  i think i prefer the other slightly, if forced to choose.

app/services/complete_moab_handler.rb
```diff
@@ -186,6 +186,10 @@ def moab_validator
     @moab_validator ||= MoabValidator.new(druid: druid, storage_location: storage_location, results: results)
   end
 
+  def checking_primary_moab?
```
ndushay, line 189: would this be better named `is_primary_moab?`  the method name is throwing me, because I don't think of "checking" being an appropriate verb for updates.   Other ideas: `moab_is_primary` ?  `handling_primary_moab`?  (this is the complete_moab_HANDLER) `primary_moab(complete_moab)` ?  Or roll your own.

ndushay: How about a `.primary?` method in the model itself?   so `complete_moab.primary?` and in complete_moab_handler, that could be memoized ...

jmartin-sul: @ndushay, done, thanks for the suggestion!  mind having another look when you get a chance?

app/services/complete_moab_handler.rb
```diff
@@ -186,6 +186,10 @@ def moab_validator
     @moab_validator ||= MoabValidator.new(druid: druid, storage_location: storage_location, results: results)
   end
 
+  def primary_moab?
```
ndushay, line 189: 👍

PR #1631, opened 2020-09-30
configurable artificial delay before making zip for replication

jmartin-sul:
## Why was this change made?

in hopes of ameliorating problems due to lag between when pres robots worker finishes writing a file, and when pres cat is able to read it

* unit tests cover the code change for basic functionality
* as a basic smoke test will deploy to stage and run infra integration tests and confirm preservation works, before deploying to prod.
  * will babysit in prod for a while after deployment

added a couple settings, left explanation in code comments, since that seems like the most visible/useful place for the necessary info.

### DIFF
--- a/app/jobs/zipmaker_job.rb
+++ b/app/jobs/zipmaker_job.rb
@@ -24,6 +24,8 @@ class ZipmakerJob < ApplicationJob
   # @param [String] moab_replication_path The path containing the druid tree from which the zipped version should be
   #   created.  used via job.arguments in before_perform setup.
   def perform(druid, version, moab_replication_path) # rubocop:disable Lint/UnusedMethodArgument
+    wait_as_needed
+
     if File.exist?(file_path)
       FileUtils.touch(file_path)
     else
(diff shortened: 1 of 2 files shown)


justinlittman (approved):

app/jobs/zipmaker_job.rb
```diff
@@ -33,4 +35,14 @@ def perform(druid, version, moab_replication_path) # rubocop:disable Lint/Unused
       PlexerJob.perform_later(druid, version, part_key, DruidVersionZipPart.new(zip, part_key).metadata)
     end
   end
+
+  # sleep for a configured amount of time, including a random additional delay to help ensure staggering.
+  #
+  # BUT WHY? to prevent duplication and possible explanation drift, see Robots::SdrRepo::PreservationIngest::UpdateCatalog#wait_as_needed
+  # for more detailed explanation.  the short version is, this helps alleviate Ceph MDS write/read sync and/or contention issues that
+  # we don't yet fully understand.
```
ndushay, line 43: I think the fuller explanation should be in a github issue (or even wiki).   

Staggering the zipmaker times ... again, we didn't used to need it -- if the real problem is the ceph access, then the stagger is not needed.  Over-engineered?

jmartin-sul: yeah i was sort of cargo culting from what i've seen other delay mechanisms do.  happy to excise.

Issue #1637, opened 2020-10-05
zipmaker job: delete zips that fail file size check


ndushay:
TO DO: if the zip size check determines that the zip is bad (because it's not large enough to contain all the expected content), delete the bad zip. 

WHY: This will make it easier to retry the failed zipmaker job without either manually cleaning up busted zip files, or waiting for the regular zip cache cleaning process to clean the zip up (zip file currently expires when it hasn't been accessed in a week, i.e. a time more than 7 days old, iirc).

jmartin-sul:
i've done part of this work -- i've implemented the app code change in https://github.com/sul-dlss/preservation_catalog/pull/1644, but the work still needs a new test.  happy to pair if someone else wants to pick this up to finish it off.

Issue #1633, opened 2020-09-30
preservation services are having trouble interacting reliably with Ceph backed storage


jmartin-sul:
**Describe the bug**

The clearest symptom is that a preservation_robots worker will write a new version of a Moab, and some time after it finishes, another process will have trouble reading content or file system metadata from one or more files in the Moab version.  By far the most trouble has cropped up with the `ZipmakerJob` -- either the zip binary will throw an error when it's unable to add a file from the Moab version to the zip it's creating for replication, or a sanity check on size after zip creation (see #1302, #1540) will fail with an access denied error.  Occasionally, an API call to pres cat from pres robots will fail because pres cat won't be able to fully read the contents of the Moab it's being asked about.  Very occasionally, we've seen pres robots be unable to read the contents of a Moab version.  Will show example errors in the comments below.

@julianmorley and @andrewjbtw have also scripted up some automated checks that proactively look for files on the Ceph backed preservation mounts, to look for issues when listing file system attributes, since we've noticed inability to resolve file system attributes or permissions issues reading the files at all for files that trigger pres cat and pres robots errors.  Examples of that output below also.

Filing this in pres cat repo since that's the codebase that experiences this error the most.  But it seems to involve the interplay of preservation_robots writing Moab content (possibly due to our particular usage of hardlinking), preservation_catalog and preservation_robots reading Moab content, and something to do with Ceph and its MDS (metadata service) configuration.

**User Impact**

We experience unpredictable failures in `ZipmakerJob`, which requires babysitting and manual intervention to assure that all our content is replicated to all endpoints, and when preservation_robots run into errors or have to be throttled to prevent errors from cropping up, accessioning can slow down in general, and/or stall for specific objects that encounter errors.  This problem has required pausing some automated accessioning, and has required us to ask for more pauses in human accessioning than initially asked for in prep for the recent storage migration weekend (see #1559).

**Object(s) In This Error State**

There are 1420 `zipmaker_failed` entries at the moment.  In a comment below, I'll post a ruby snippet for programmatically capturing the relevant info (druid, version, etc) so as to not need to copy it all out of the paginated list in Resque's web console.

**Steps to Reproduce, if Known**

Not sure exactly yet.  We just know that pres robots will write a Moab version, and subsequent processes that attempt to read the content are sometimes unexpectedly unable to.  We suspect stale MDS cache or stale locks, further explanation in comments below.

**Expected behavior**

[...]

PR #1640, opened 2020-10-06
Check moab readability

### LINKED ISSUE #1639: zipmaker job: run "stat" on Moab version directory to address possible MDS failures
From #1633:

WHAT: Recursively `stat` the contents of the Moab version dir before calling `create_zip!` in `ZipmakerJob`.

WHY: we think/hope this will avoid zipmaker _silently creating a zip that's missing files it's supposed to have_.  

We currently discover this problem _AFTER_ creating the zip via a size sanity check comparing the newly created (uncompressed) zip, against the Moab version dir.   (We know the uncompressed zip should be slightly larger than the total dir size being zipped, in terms of logical size, due to of a bit of metadata about the zipped content.)

Somewhat related to #1637, which is to remove zips that fail the file size check.

jmartin-sul:
## Why was this change made?

To confirm that the contents of a Moab directory are readable before attempting to create a zip file from it, and to maybe kick Ceph MDS into making the files readable if MDS still sees the files as strays.

Per our hypothesizing in #1633, preservation_robots' implementation of similar logic (see https://github.com/sul-dlss/preservation_robots/pull/273) should (we hope) mitigate issues reading the new Moab version content.  But in the event there are still read issues, this will at least make it easier to retry the `zipmaker` failure without having to clean up a bad zip file from `/sdr-transfers`, by failing the job early (#1637 will also help keep retrying easy, if for some reason read issues crop up after the check added by this PR -- we haven't seen that in practice, but we're just getting a handle on the problem).

closes #1639 

unit tests.  will also deploy to stage and run infra integration tests, and will confirm that content from tests was preserved.

comments left where appropriate in code, w/ links to github issues.

### DIFF
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -151,7 +151,7 @@ class DruidVersionZip
   end
 
   def zip_size_ok?
-    total_part_size >= moab_version_size
+    total_part_size > moab_version_size
   end
 
   def total_part_size
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -32,10 +32,13 @@ class DruidVersionZip
   # the archival directory structure, just the object, e.g. starting at 'ab123cd4567/...' directory,
   # not 'ab/123/cd/4567/ab123cd4567/...'
   def create_zip!
+    check_moab_version_readability!
     ensure_zip_directory!
     combined, status = Open3.capture2e(zip_command, chdir: work_dir.to_s)
     raise "zipmaker failure #{combined}" unless status.success?
-    raise "zip size (#{total_part_size}) is smaller than the moab size (#{moab_size})! zipmaker failure #{combined}" unless zip_size_ok?
+    unless zip_size_ok?
+      raise "zip size (#{total_part_size}) is smaller than the moab version size (#{moab_version_size})! zipmaker failure #{combined}"
+    end
 
     part_keys.each do |part_key|
       DruidVersionZipPart.new(self, part_key).write_md5
@@ -137,22 +140,32 @@ class DruidVersionZip
 
   private
 
+  # Throws an error if any of the files in the moab are not yet readable.  For example due to
+  # Ceph MDS instance for a pres cat worker VM thinking that a file is a stray as a result of our
+  # particular use of hardlinking in preservation ingest.  Allows for a quick directory walk before
+  # attempting to create the zip file(s).  See https://github.com/sul-dlss/preservation_catalog/issues/1633
+  # @raise [StandardError] if storage_location is not available (should have been provided in constructor)
+  # @raise [Errno::EACCES, Errno::EIO, Errno::ENOENT, Errno::ESTALE, ?] if it is not possible to stat one or more files in the Moab
+  def check_moab_version_readability!
+    moab_version_files.map { |f| File.stat(f) }
+  end
+
   def zip_size_ok?
-    total_part_size >= moab_size
+    total_part_size >= moab_version_size
   end
 
   def total_part_size
-    part_paths
-      .map { |part_path| File.size(part_path) }
-      .sum
+    part_paths.sum { |part_path| File.size(part_path) }
   end
 
-  def moab_size
+  def moab_version_size
+    moab_version_files.sum { |file_path| File.size(file_path) }
+  end
+
+  def moab_version_files
     Dir
       .glob("#{moab_version_path}/**/*")
       .select { |path| File.file?(path) }
-      .map { |file_path| File.size(file_path) }
-      .sum
   end
 
   # @return [String] e.g. 'Zip 3.0 (July 5th 2008)' or 'Zip 3.0.1'
--- a/spec/models/druid_version_zip_spec.rb
+++ b/spec/models/druid_version_zip_spec.rb
@@ -3,6 +3,9 @@
 require 'rails_helper'
 
 describe DruidVersionZip do
+  # Some tests below will use the constructor that takes storage_location, because they are
+  # testing behavior where a zip is being created from a Moab on disk.  For tests that don't
+  # exercise that behavior, the parameter is omitted to test that it isn't needed.
   let(:dvz) { described_class.new(druid, version) }
   let(:druid) { 'bj102hs9687' }
   let(:version) { 1 }
@@ -96,6 +99,24 @@ describe DruidVersionZip do
         expect(status).to be_success
         _, status = Open3.capture2e("unzip -lq #{zip_path} bj/102/hs/9687/#{techmd}")
         expect(status).not_to be_success
+      end
+
+      it 'calls check_moab_version_readability! before attempting to create the zip' do
+        # This checks that we only attempt zip creation after checking to see that all files are readable.
+        expect(dvz).to receive(:check_moab_version_readability!).ordered.and_call_original
+        expect(Open3).to receive(:capture2e).with(dvz.zip_command, chdir: dvz.work_dir.to_s).ordered.and_call_original
+        expect { dvz.create_zip! }.not_to raise_error
+        expect(File).to exist(zip_path)
+      end
+    end
+
+    context 'when the moab version directory to be zipped has unreadable files' do
+      let(:unreadable_filename) { 'spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0003/manifests/versionInventory.xml' }
+
+      it 'allows the error to bubble up when stat is unsuccessfully called on a file in the moab version' do
+        allow(File).to receive(:stat).and_call_original
+        expect(File).to receive(:stat).with(unreadable_filename).and_raise(Errno::EACCES, 'no file for you')
+        expect { dvz.create_zip! }.to raise_error(Errno::EACCES, /no file for you/)
       end
     end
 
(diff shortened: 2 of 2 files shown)


app/models/druid_version_zip.rb
```diff
@@ -32,10 +32,11 @@ def base_key
   # the archival directory structure, just the object, e.g. starting at 'ab123cd4567/...' directory,
   # not 'ab/123/cd/4567/ab123cd4567/...'
   def create_zip!
+    check_moab_version_readability!
     ensure_zip_directory!
     combined, status = Open3.capture2e(zip_command, chdir: work_dir.to_s)
     raise "zipmaker failure #{combined}" unless status.success?
-    raise "zip size (#{total_part_size}) is smaller than the moab size (#{moab_size})! zipmaker failure #{combined}" unless zip_size_ok?
+    raise "zip size (#{total_part_size}) is smaller than the moab size (#{moab_version_size})! zipmaker failure #{combined}" unless zip_size_ok?
```
ndushay, line 39: love the method name change - how about adding version to the text here as well?

ndushay (approved):
one strong suggestion, o.w. LGTM

ndushay (changes requested):
actually, i will block the merge for my request ...

app/models/druid_version_zip.rb
julianmorley: Also maybe say that it the error message? "moab version size" instead of "moab size"?

jmartin-sul: good catch, fixed!

jmartin-sul: ah, yeah, @julianmorley that was what i interpreted @ndushay's suggestion to mean (sorry, GH didn't auto-refresh the page, and i didn't see this till i'd already pushed a change -- let me know if i missed something from either of your suggestions)

app/models/druid_version_zip.rb
```diff
@@ -137,8 +140,18 @@ def zip_storage
 
   private
 
+  # should throw an error if any of the files in the moab are not yet readable, for example due to
```
jcoyne, line 143:
```suggestion
  # Throws an error if any of the files in the moab are not yet readable.  For example due to
```

app/models/druid_version_zip.rb
```diff
@@ -137,8 +140,18 @@ def zip_storage
...
+  # Ceph MDS instance for a pres cat worker VM thinking that a file is a stray as a result of our
+  # particular use of hardlinking in preservation ingest.  allows for a quick directory walk before
```
jcoyne, line 145:
```suggestion
  # particular use of hardlinking in preservation ingest.  Allows for a quick directory walk before
```

app/models/druid_version_zip.rb
```diff
@@ -137,8 +140,18 @@ def zip_storage
...
+  # attempting to create the zip file(s).  see https://github.com/sul-dlss/preservation_catalog/issues/1633
```
jcoyne, line 146:
```suggestion
  # attempting to create the zip file(s).  See https://github.com/sul-dlss/preservation_catalog/issues/1633
```

app/models/druid_version_zip.rb
```diff
@@ -137,8 +140,18 @@ def zip_storage
...
+  # @raise [StandardError] if storage_location is not available (should have been provided in constructor)
+  # @raise [Errno::EACCES, ?] if it is not possible to stat one or more files in the Moab
```
jcoyne, line 148:
```suggestion
  # @raise [Errno::EACCES] if it is not possible to stat one or more files in the Moab
```

app/models/druid_version_zip.rb
```diff
@@ -147,12 +160,14 @@ def total_part_size
       .sum
   end
 
-  def moab_size
+  def moab_version_size
+    moab_version_files.map { |file_path| File.size(file_path) }.sum
```
jcoyne, line 164:
```suggestion
    moab_version_files.sum { |file_path| File.size(file_path) }
```

spec/models/druid_version_zip_spec.rb
```diff
@@ -3,6 +3,9 @@
 require 'rails_helper'
 
 describe DruidVersionZip do
+  # some tests below will use the version of the constructor that takes storage_location, because they
+  # are testing behavior where a zip is being created from a moab on disk.  for tests that don't exercise
```
jcoyne, line 7:
```suggestion
  # are testing behavior where a zip is being created from a Moab on disk.  For tests that don't exercise
```

spec/models/druid_version_zip_spec.rb
```diff
@@ -3,6 +3,9 @@
 require 'rails_helper'
 
 describe DruidVersionZip do
+  # some tests below will use the version of the constructor that takes storage_location, because they
```
jcoyne, line 6:
```suggestion
   # Some tests below will use the constructor that takes storage_location, because they
```

spec/models/druid_version_zip_spec.rb
```diff
@@ -97,6 +100,26 @@
         _, status = Open3.capture2e("unzip -lq #{zip_path} bj/102/hs/9687/#{techmd}")
         expect(status).not_to be_success
       end
+
+      it 'calls check_moab_version_readability! before attempting to create the zip' do
+        # normally checking call order is to be avoided, but we want to make sure that we only attempt zip
+        # creation after proactively checking to see that all files to be zipped are readable.  readability
+        # checking logic is tested elsewhere in this spec.
```
jcoyne, line 107:
```suggestion
        # This checks that we only attempt zip creation after checking to see that all files are readable.
```

spec/models/druid_version_zip_spec.rb
```diff
@@ -295,4 +318,35 @@
...
+        spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0001/data/metadata/versionMetadata.xml
+        spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0001/manifests/fileInventoryDifference.xml
+        spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0001/manifests/manifestInventory.xml
+        spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0001/manifests/signatureCatalog.xml
+        spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0001/manifests/versionAdditions.xml
+        spec/fixtures/storage_root01/sdr2objects/bj/102/hs/9687/bj102hs9687/v0001/manifests/versionInventory.xml
+      ]
+    end
+
+    # note: if this test is dropped or altered, revisit the #create_zip! tests in this spec w.r.t.
```
jcoyne, line 344: Why would this test be dropped or altered?

jcoyne (changes requested):

ndushay (approved):

app/models/druid_version_zip.rb
```diff
@@ -137,8 +140,18 @@ def zip_storage
...
+  # attempting to create the zip file(s).  see https://github.com/sul-dlss/preservation_catalog/issues/1633
+  # @raise [StandardError] if storage_location is not available (should have been provided in constructor)
+  # @raise [Errno::EACCES, ?] if it is not possible to stat one or more files in the Moab
+  def check_moab_version_readability!
+    moab_version_files.map { |f| File.stat(f) }
+  end
+
   def zip_size_ok?
-    total_part_size >= moab_size
+    total_part_size >= moab_version_size
```
julianmorley, line 154: Really should be `>` here; the zip overhead dictates that the sum of all parts *must* be larger than moab_version_size on disk.

app/models/druid_version_zip.rb
jmartin-sul: So in this case, I was not entirely sure what the full list of plausible errors might be.  Admittedly, I didn't look back through all of the HB alerts when I first did this, but I wanted to call out that any number of ruby file read errors could be raised.

Looking back through Honeybadger alerts now, I also see some others that've occurred when reading Moab contents, which I'll added to the list, resulting in:  `Errno::EACCES, Errno::EIO, Errno::ENOENT, Errno::ESTALE`.

Still, looking over all the possibilities for error codes in `Errno.constants`, it's clear that most are not relevant to this method, but I suspect that some which we haven't seen are also plausible error codes.  For example, `EBUSY`, `ENOMEM`, and `EPERM`.

My bad for not being more thorough in listing the errors we've already encountered, but I would like to indicate that there are errors which might be raised, but which we're unsure about.  Though if you know for sure that the error codes I mentioned speculatively can't come up, happy to leave off the `?`.

i realize that the `?` may not technically be valid YARD syntax, but to my knowledge we don't auto-generate documentation off the YARD comments, and i though there was more value in expressing the uncertainty about possible errors than there was in producing strictly compliant YARD comments.

(I also just spent about 10 minutes, which was 10 minutes more than i should've, reading a 2003 Slashdot thread about some controversy with the `EDOOFUS` error code.  Same as it ever was 😆 )

app/models/druid_version_zip.rb
jmartin-sul: good catch, and fixed on the prior method definition too

spec/models/druid_version_zip_spec.rb
jmartin-sul: not totally sure, and i was a bit on the fence about including this comment (and the corresponding one on line 105).

i don't think this test is strictly necessary in terms of code coverage results, and so i can imagine it getting deleted accidentally, or intentionally as cruft/overkill, and then not getting noticed if the PR review is cursory.  i wanted to avoid that.  it's also testing some behavior that seems unnecessary at first glance, so it was easy to imagine aggressive code pruning removing it.

if that scenario seems unlikely to you, and you think this adds too much noise, i can remove the comment.

app/models/druid_version_zip.rb
jmartin-sul: good call, done

jmartin-sul:
@jcoyne and @julianmorley, i think i've addressed your comments, mind giving another look?

@jcoyne i implemented yours on my local instance and squashed them into the existing commits just to keep from having a messy history or squashing everything, but thanks for using suggestions, i do usually like that mechanism.  also there was one change that i didn't implement exactly as suggested, but happy to revisit if you think it needs more fixing.

spec/models/druid_version_zip_spec.rb
jcoyne: I think this is confusing. I think this comment could apply to almost every test case so I'm not clear why we need to call this out in particular.

jmartin-sul: fair enough, got rid of it

jmartin-sul:
@jcoyne i _think_ i've addressed all of your feedback.  since we saw a small batch of `zipmaker` errors yesterday that this would likely help catch just a bit earlier, i wanted to get this merged and deployed this morning.  apologies if this was presumptuous.

app/models/druid_version_zip.rb
jmartin-sul: feel free to file an issue if you want me to circle back on this one again.

PR #1644, opened 2020-10-09
Cleanup bad archive zips

### LINKED ISSUE #1637: zipmaker job: delete zips that fail file size check

jmartin-sul:
## Why was this change made?

closes #1637

TODO:
- [x] existing tests pass (show this change didn't introduce regressions)
- [x] write a new test that shows bad zip gets cleaned up by this code change
- [ ] run infra integration tests against branch on stage and make sure DOR objects that are created get preserved.

n/a


justinlittman (approved):

jmartin-sul:
tested in stage using infra integration tests.  xg760gr4474 is a two version druid in stage, and appears to have been replicated properly.

deployed to prod.

PR #1646, opened 2020-10-13
raise if trying to delete a ZippedMoabVersion for which ZipParts still exist

jmartin-sul:
## Why was this change made?

This will make it easier to safely remediate Moab replications that got wedged part way through the pipeline, by allowing more confidence that data about replicated parts won't be accidentally deleted when trying to clean up hanging `ZippedMoabVersion` records.

(discussed briefly on slack last week)

unit tests.  in the normal course of app operation, `ZippedMoabVersion`s aren't deleted.  that's only done manually, to remediate replication problems.

n/a

### DIFF
--- a/app/models/zipped_moab_version.rb
+++ b/app/models/zipped_moab_version.rb
@@ -10,7 +10,7 @@
 class ZippedMoabVersion < ApplicationRecord
   belongs_to :preserved_object, inverse_of: :zipped_moab_versions
   belongs_to :zip_endpoint, inverse_of: :zipped_moab_versions
-  has_many :zip_parts, dependent: :destroy, inverse_of: :zipped_moab_version
+  has_many :zip_parts, dependent: :restrict_with_exception, inverse_of: :zipped_moab_version
 
   validates :preserved_object, :version, :zip_endpoint, presence: true
 


ndushay (approved):
to confirm:  this isn't directly related to the issues filed last week about problems with zips -- it's something extra that wasn't ticketed?

conditional approval on tests being green.

jmartin-sul:
> to confirm: this isn't directly related to the issues filed last week about problems with zips -- it's something extra that wasn't ticketed?
> 
> conditional approval on tests being green.

yup.

for posterity, the comment from slack where i proposed doing this (as a wild guess related to an unexpectedly inefficient set of queries, and as a proposal to make things a little safer regardless of whether that hypothesis was right):

@jmartin-sul [said](https://stanfordlib.slack.com/archives/G6K8R3VBK/p1602199145174800):
> i have a wild guess as to what the problem _[see below]_ was:  as plain SQL has a notion of cascading deletes for foreign key relationships (or at least some implementations?), ActiveRecord allows you to implement something like that with callbacks, where an object with a FK relationship will be destroyed if the object to which it points gets destroy called on it.  in general, as a safety measure, we don't allow that in pres cat, and all dependent relationships will allow the DB to raise an error if you delete something that's still referenced by another row w/ a non-null FK constraint (`dependent: :restrict_with_exception`).  however, one of the two relationships where we allow the child to be destroyed if the parent gets destroyed is... `ZippedMoabVersion` to `ZipPart`, which is `dependent: :destroy`.  i was inclined to flip that over to `dependent: :restrict_with_exception` for safety's sake anyway.  maybe it will also avoid this `destroy_all` behavior.
either way, i was going to touchup the wiki entry for cleaning up after zipmaker failures, and i'll avoid `destroy_all` in that example code, in favor of e.g. find_each with a targeted destroy on each individual ZMV.

\* The problem was a large number of `zip_parts` queries (thousands per second), causing PG to fall over.  We're really not sure this was caused by the `dependent: :destroy` configuration, but the change in this PR shouldn't hurt on that front, and it makes the app a bit safer in terms of guarding against accidental deletion of metadata around replicated Moab versions

PR #1649, opened 2020-10-17
Zipmaker raise on undersized extant zip

### LINKED ISSUE #1642: ZipmakerJob should check size of zip if it finds one already made
If `ZipmakerJob` finds a zip file already made in the zip temp space, it should check the size of it just as it would do right after making the zip file itself.  If the zip file it finds is too small, `ZipmakerJob` should... delete it and re-create?  Just `raise`?  The latter seems safest.  Either would prevent archiving an incomplete moab version, as might happen when retrying without cleaning up after a `ZipmakerJob` that failed upon discovering the zip file it created was too small (see #1633 for explanation of why this zip problem occurs sometimes).

I think this looks something like:
* making `DruidVersionZip#zip_size_ok?` public
* delegating `:zip_size_ok?` to `zip` in `ZipmakerJob` (as is already done for a bunch of other methods)
* using `#zip_size_ok?` when `File.exist?(file_path)` is `true` to either `raise` or delete as described above:

https://github.com/sul-dlss/preservation_catalog/blob/d3ebaf65ef61e6ab83e45f797365b2b08bc4d8eb/app/jobs/zipmaker_job.rb#L26-L37

and of course a new test that shows that a mis-created zip file will no longer be delivered.

jmartin-sul:
## Why was this change made?

closes #1642

builds on #1644 (hence the non-master base branch, since that's already been approved)

First commit does some test improvement without touching the app code, hopefully making the modified tests more illustrative, more realistic, and less likely to break in the face of an implementation change that doesn't introduce regressions. (https://github.com/sul-dlss/preservation_catalog/commit/c7f064780040eaf41587a7993107388d60b9fac4 -- to make this easier to review, i pushed this commit as a separate branch to show that it passed CI on its own)

Second commit is the change that the first commit paves the way for, and is the functionality asked for by #1642 (I went with the "just raise" approach).  Third commit has two new tests to show that the new check does what it's supposed to, and hopefully the lack of change to existing tests (compared to the first commit) gives some confidence that no regressions were introduced.

Once this is merged, we can delete [zipmaker-test-improvements](https://github.com/sul-dlss/preservation_catalog/compare/cleanup-bad-archive-zips...zipmaker-test-improvements?expand=1), since this PR includes that branch's commit.

- [x] unit test changes as described above 
- [ ] will deploy to stage and run infra integration tests

if this gets merged, we can simplify https://github.com/sul-dlss/preservation_catalog/wiki/Investigating-a-failed-zipmaker-job/6f2d65706ec1c963ecd62ee8e82ec25f54c65999 to indicate that any failed `zipmaker` job is retriable without worry, and if it fails again with a size check error up front, use the advice that's already there for cleanup.

### DIFF
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -31,7 +31,7 @@ class DruidVersionZip
   # file to refresh atime and mtime, so the zip cache cleaning cron job doesn't see it as stale.  If it doesn't,
   # create it.
   # @raise [StandardError] if there's a zip file for this druid-version, but it looks too small to be complete.
-  def ensure_zip!
+  def find_or_create_zip!
     if File.exist?(file_path)
       raise "zip already exists, but size (#{total_part_size}) is smaller than the moab version size (#{moab_version_size})!" unless zip_size_ok?
       FileUtils.touch(file_path)
--- a/app/models/druid_version_zip.rb
+++ b/app/models/druid_version_zip.rb
@@ -27,6 +27,19 @@ class DruidVersionZip
     @base_key ||= s3_key.sub(/.zip\z/, '')
   end
 
+  # Checks to see whether a zip file already exists for this druid-version.  If it does, just touch the
+  # file to refresh atime and mtime, so the zip cache cleaning cron job doesn't see it as stale.  If it doesn't,
+  # create it.
+  # @raise [StandardError] if there's a zip file for this druid-version, but it looks too small to be complete.
+  def ensure_zip!
+    if File.exist?(file_path)
+      raise "zip already exists, but size (#{total_part_size}) is smaller than the moab version size (#{moab_version_size})!" unless zip_size_ok?
+      FileUtils.touch(file_path)
+    else
+      create_zip!
+    end
+  end
+
   # Creates a zip of Druid-Version content.
   # Changes directory so that the storage root (and druid tree) are not part of
   # the archival directory structure, just the object, e.g. starting at 'ab123cd4567/...' directory,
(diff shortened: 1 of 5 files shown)


app/models/druid_version_zip.rb
```diff
@@ -27,6 +27,19 @@ def base_key
     @base_key ||= s3_key.sub(/.zip\z/, '')
   end
 
+  # Checks to see whether a zip file already exists for this druid-version.  If it does, just touch the
+  # file to refresh atime and mtime, so the zip cache cleaning cron job doesn't see it as stale.  If it doesn't,
+  # create it.
+  # @raise [StandardError] if there's a zip file for this druid-version, but it looks too small to be complete.
+  def ensure_zip!
```
jcoyne, line 34: I think `find_or_create_zip` might be a more clear name for this.

jcoyne (changes requested):

app/models/druid_version_zip.rb
jmartin-sul: ok, i took this suggestion.  pushing the fix exposed a pretty consistently [flappy](https://app.circleci.com/pipelines/github/sul-dlss/preservation_catalog/402/workflows/dd88079d-0b07-4da9-91e2-02712931ec19/jobs/628) test ([this one](https://github.com/sul-dlss/preservation_catalog/pull/1649/files#diff-0a92f4583841e265e22a1dbb203489bf0238d083d14fd749f73c1e0129687934R78-R87)).  per the comment i left, i wasn't wild about it, but i added a 1/10th of a second delay to that test so the `touch` would be sure to leave `atime` and `mtime` modified.  i liked that a bit more than testing that `File.touch` got called, since it tests effect more than implementation, but happy to revisit if you're more grossed out by that `sleep` call than i was (fwiw i tried 0.001 s, and still got the same test failure).

jmartin-sul:
@jcoyne, suggestion taken, and made another small change to address intermittent test failure, mind having another look?

(i'm not sure why that test didn't fail consistently friday evening -- maybe they've just got more compute power and/or IO provisioned on monday morning, leading to consistently faster test runs? 🤷)

jcoyne (approved):

jmartin-sul:
tested in stage using infra integration tests.  xg760gr4474 is a two version druid in stage, and appears to have been replicated properly.

deployed to prod.

PR #1652, opened 2020-10-27
Use dlss-capistrano to manage resque-pool tasks 

mjgiarlo:
## Why was this change made?

This replaces `capistrano-resque-pool` and works as expected with a hot-swappable pool.

Also includes:

* Store resque-pool lockfile in shared dir
  * this allows the `resque:pool:hot_swap` capistrano task to do what we expect it to do.

in prescat-qa

None

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -13,8 +13,8 @@ require "capistrano/rails/migrations"
 require "capistrano/passenger"
 require "capistrano/honeybadger"
 require "dlss/capistrano"
+require "dlss/capistrano/resque_pool"
 require 'whenever/capistrano'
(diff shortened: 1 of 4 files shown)


ndushay:
I would be more comfy if all the touches had already been done?   Or you promise to follow up on that?   basically a bit nervous on the async between this pr and deployments.

mjgiarlo:
@ndushay :speech_balloon: 
> I would be more comfy if all the touches had already been done? Or you promise to follow up on that? basically a bit nervous on the async between this pr and deployments.

converted to draft until we work this out

ndushay (approved):

PR #1662, opened 2021-01-05
update to rails 6.1

peetucket:
## Why was this change made?

Update to rails 6.1

Unit tests

N/A


app/models/preserved_object.rb
```diff
@@ -24,7 +24,7 @@ class PreservedObject < ApplicationRecord
             length: { is: 11 },
             format: { with: /(?!#{PREFIX_RE})#{DruidTools::Druid.pattern}/ } # ?! group is a *negative* match
   validates :current_version, presence: true, numericality: { only_integer: true, greater_than: 0 }
-  validates :preservation_policy, null: false
+  validates :preservation_policy, presence: true
```
peetucket, line 27: Changes this because the `null: false` causes `ArgumentError:
  Unknown validator: 'NullValidator'` otherwise.  I believe this validation has the same affect.

justinlittman (approved):

PR #1673, opened 2021-03-04
documentation updates, mostly database related, and dependency culling

jmartin-sul:
_note_: 505 of the new lines are in the updated ER diagram, since it's now an SVG instead of a PNG, so this isn't as terrible as the diff summary makes it look (or maybe it is, because it's proofreading documentation 😆).

## Why was this change made?

lingering documentation updates...

* update table descriptions and example queries for as-yet uncaptured (summer 2020) data model changes
* easier approach for ER diagram generation, updated ER diagram and key
* drop the hirb gem from dependencies -- i don't think we really use it, and it doesn't seem actively maintained
* but wait, there's more!  act now and we'll freshen up the front page README

* new ER diagram generation was approach was used for this PR
* it's all doc changes otherwise, so some proof-reading and some playing with the updated queries would be welcome (i recall testing them all when i started the DB readme updates last summer, but it'd be good for others to give them a spin)

database readme and front page readme


jmartin-sul:
* https://github.com/sul-dlss/preservation_catalog/tree/db-readme-at-least#readme
* https://github.com/sul-dlss/preservation_catalog/tree/db-readme-at-least/db#schema-explanation

ndushay (approved):

PR #1703, opened 2021-05-25
handle Nokogiri::XML::SyntaxError for signatureCatalog.xml

### LINKED ISSUE #1696: ChecksumValidator should handle Nokogiri parse errors gracefully
⚠️ _warning_ ⚠️ -- it doesn't matter whether this happens before or after #1697, just make sure both aren't being worked at the same time, because then one or more developers will likely be in for a painful rebase.

**Describe the bug**
If Nokogiri runs into an unexpected parsing error, say an empty XML file, it will crash `ChecksumValidator` in such a way that `ChecksumValidator` will not update the `CompleteMoab#status` to `invalid_checksum`.  In the usual situation, where `CheckumValidator` is being run by `ChecksumValidationJob`, the job will fail, land will and in the `checksum_validation_failed` queue.  Even if the Moab fails checksum validation, the job should succeed from the perspective of Resque's job management, and the error should be reported by marking `status` and sending a honeybadger alert, as happens with most situations where `ChecksumValidator` detects an error in the Moab.

https://app.honeybadger.io/projects/54415/faults/79310268
https://github.com/sul-dlss/preservation_catalog/blob/80bcd90c54d8d40edbf14c177f4c53d88b89ca9e/app/services/checksum_validator.rb#L187

```
[PROJECT_ROOT]/app/services/checksum_validator.rb:187 :in `latest_signature_catalog_entries`
  # @return [Array<SignatureCatalogEntry>]
  def latest_signature_catalog_entries
    @latest_signature_catalog_entries ||= latest_moab_version.signature_catalog.entries
  rescue Errno::ENOENT, NoMethodError # e.g. latest_moab_version.signature_catalog is nil (signatureCatalog.xml does not exist)
    results.add_result(AuditResults::SIGNATURE_CATALOG_NOT_IN_MOAB, signature_catalog_path: latest_signature_catalog_path)
[PROJECT_ROOT]/app/services/checksum_validator.rb:194 :in `paths_from_signature_catalog`
[PROJECT_ROOT]/app/services/checksum_validator.rb:208 :in `signature_catalog_has_file?`
[PROJECT_ROOT]/app/services/checksum_validator.rb:204 :in `validate_against_signature_catalog`
[PROJECT_ROOT]/app/services/checksum_validator.rb:104 :in `block in flag_unexpected_data_files`
[PROJECT_ROOT]/app/services/checksum_validator.rb:104 :in `each`
[PROJECT_ROOT]/app/services/checksum_validator.rb:104 :in `flag_unexpected_data_files`
[PROJECT_ROOT]/app/services/checksum_validator.rb:81 :in `validate_signature_catalog`
[PROJECT_ROOT]/app/services/checksum_validator.rb:33 :in `validate_checksums`
[PROJECT_ROOT]/app/jobs/checksum_validation_job.rb:14 :in `perform` 
```

**User Impact**

Some Moabs that are corrupt will have incorrect status, usually `ok`, and the issue with the Moab will not be easily queryable.  The only indication of a problem is a cryptic entry in the Resque failure queue.  This might lead to a false sense of security about the integrity of some of our preservation objects.

at present:
* yq896mc2480
* yg880zm4762
* cq580gn5234
* ym971gq9293

run `ChecksumValidator` on a `CompleteMoab` record for a Moab in preservation storage that has an empty or otherwise unparsable XML manifest file (e.g. a zero-length `signatureCatalog.xml`).

For example, as of 2021-05-19 on prod:
```ruby
ChecksumValidator.new(CompleteMoab.by_druid('yq896mc2480').take).validate_checksums
```

will reproduce the bad behavior.

**Expected behavior**

`ChecksumValidator` marks the `CompleteMoab#status` as `invalid_checksum`, as it would with any other validation failure.  It does not result in the method call crashing, or a failed Resque job if invoked from `ChecksumValidationJob`.

**Screenshots**

Screen Shot 2021-05-19 at 3 23 14 PM

**Additional context**

see #1695 for more background info and possibly useful scripting.

ndushay:
## Why was this change made?

Avoid crashing ChecksumValidator for Nokogiri::XML::SyntaxError on parsing empty signatureCatalog.xml

Fixes #1696 

unit tests


app/services/checksum_validator.rb
```diff
@@ -188,6 +188,9 @@ def latest_signature_catalog_entries
   rescue Errno::ENOENT, NoMethodError # e.g. latest_moab_version.signature_catalog is nil (signatureCatalog.xml does not exist)
     results.add_result(AuditResults::SIGNATURE_CATALOG_NOT_IN_MOAB, signature_catalog_path: latest_signature_catalog_path)
     []
+  rescue Nokogiri::XML::SyntaxError => e
+    results.add_result(AuditResults::INVALID_MANIFEST, manifest_file_path: latest_signature_catalog_path, addl: e.inspect)
```
ndushay, line 192: INVALID_MANIFEST seemed a perfectly good error code, so I used it.

spec/services/checksum_validator_spec.rb
```diff
@@ -59,7 +59,6 @@
 
     context 'file checksums in manifestInventory.xml do not match' do
       let(:druid) { 'zz925bx9565' }
-      let(:results) { instance_double(AuditResults, report_results: nil, :actual_version= => nil) }
```
ndushay, line 62: unnecessary duplication ...

ndushay (commented):
Technically I could have used the object with the bad signatureCatalog, but it seemed prudent to ensure we also trap for empty signatureCatalog.xml.

I took the large content files out of the new object.

jmartin-sul (approved):

PR #1704, opened 2021-05-27
readme touchups

jmartin-sul:
## Why was this change made?

noticed some easy touchups when using these queries to look at test objects on stage.

* fix a query that had fallen slightly out of date w/ current ActiveRecord sanitization expectations, add advice if other queries throw that error
* collapse some sections by default for slightly better readability

thanks to this gist for the collapsing sections example and gotchas: https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab

* ran the modified query
* previewed the README for the layout changes

db/README.md only


jmartin-sul:
examples of the new collapsed sections:

Screen Shot 2021-05-27 at 12 03 20 PM
Screen Shot 2021-05-27 at 12 02 49 PM
Screen Shot 2021-05-27 at 12 02 55 PM
Screen Shot 2021-05-27 at 12 03 12 PM

ndushay (approved):
wow - we have a lot of details in the 369 line readme.  Probably too many.  But now is not the time to fix.

PR #1706, opened 2021-05-27
checksum_validator_spec: fix typos

ndushay:
## Why was this change made?

because I'm a little OCD?


justinlittman (approved):
I think it is safe to say that the entire team supports proper grammar.

But not failing rubocop ...

jmartin-sul (approved):

PR #1713, opened 2021-06-03
add endpoint for new MoabValidation job

peetucket:
## Why was this change made?

Adds a new endpoint to start a MoabValidation job which can later be called from a new preservationWF robot

Added new tests

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -134,8 +134,6 @@ paths:
           description: Job queued
         '400':
           description: Bad request (e.g., malformed druid)
-        '404':
-          description: Object not found
       parameters:
         - name: id
           in: path
--- a/openapi.yml
+++ b/openapi.yml
@@ -124,6 +124,25 @@ paths:
           example: 'druid:bc123df4567.json'
           schema:
             $ref: '#/components/schemas/DruidWithOptionalFormat'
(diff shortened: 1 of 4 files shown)


ndushay (approved):
LGTM, but I touched it.   @jmartin-sul ?

openapi.yml
```diff
@@ -124,6 +124,25 @@ paths:
...
+      tags:
+        - objects
+      summary: Starts a validate moab job, which does a validation of the on disk Moab, without reference to the preservation-catalog basis
+      responses:
+        '200':
+          description: Job queued
+        '400':
+          description: Bad request (e.g., malformed druid)
+        '404':
+          description: Object not found
```
jmartin-sul, line 138: i'd be inclined to drop the 404, as i don't think this will ever be a response from the current controller method -- the controller method doesn't check for presence itself, and the job can't report that back to the caller since the job is run async.

if no dir can be found for the Moab, the job'll report failure and then a person can retry the workflow step (once that's all wired up by subsequent tickets).

jmartin-sul (changes requested):
one little suggestion for trimming the API spec, otherwise LGTM!

jmartin-sul (approved):

PR #1714, opened 2021-06-03
add more rigorous okcomputer checks for pres disks

### LINKED ISSUE #1712: more thorough OKComputer check for whether storage roots are properly mounted
**Describe the bug**
The okcomputer check for storage root readability is too cursory.

**User Impact**
We can run into situations where the mountpoint directories for the storage roots are available, but the contents of the storage roots are not.  This leads to a situation where preservation storage is unreachable, but the okcomputer check passes, leaving us with no proactive Nagios alert to warn us of the problem.

All three Ceph backed preservation storage roots on prod.

`¯\_(ツ)_/¯`

**Expected behavior**

If the content on the storage root content isn't genuinely readable, the okcomputer check should fail, so that Nagios can warn us of the problem.

I can think of two possible solutions, the first slightly more cursory, neither too much work:
* Confirm that the `deposits` and `sdr2objects` directories (i.e. the "deposit trunk" and "storage trunk" subdirectories) are readable by preservation catalog for each configured storage root.
* Pick a random druid from each storage root, filtering out all but the smallest Moabs (e.g. a < 2 MB Moab at random from each storage root).  Have the okcomputer check pass if all three Moabs pass synchronous checksum validation.  On such small Moabs, CV should be speedy enough to finish very comfortably within a synchronous network request (but yes, synchronously doing CV on arbitrary Moabs is a bad idea).
  * Since this requires reading all the content in one Moab from each storage root, this should be a pretty good smoke test of the storage roots at least being mounted.

**Screenshots**
https://preservation-catalog-prod-01.stanford.edu/status/all

    Default Collection
      ...snip...
      feature-pres-01: PASSED '/pres-01' is a reachable directory (0.000s)
      feature-pres-02: PASSED '/pres-02' is a reachable directory (0.000s)
      feature-pres-03: PASSED '/pres-03' is a reachable directory (0.010s)
      ...snip...

**Additional context**

this was spurred by an actual outage, see https://stanfordlib.slack.com/archives/G6K8R3VBK/p1622673998105800

peetucket:
## Why was this change made?

Fixes #1712 -- add more thorough okcomputer checks for pres-cat disks

This now verifies:
1. That both the `sdr2objects` and `deposits` directories exist for each pres-cat disk and that they are directories
2. That the `sdr2objects` directory has a minimum number of subfolders contained in it (configurable in settings, but should be set at a threshold below the current numbers for all disks, which is currently 402 in production).  This could be a problem when a new disk is added since the count would then be lower for just the last disk.  At this point we may want to configure it per disk instead by adding some new settings.

Updated the existing unit test.
Deployed to stage to validate

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -68,10 +68,8 @@ end
 # rubocop:enable Metrics/AbcSize
 
 Settings.storage_root_map.default.each do |name, location|
-  deposit_location = "#{location}/deposit"
-  sdrobjects_locations = "#{location}/sdr2objects"
-  OkComputer::Registry.register "feature-#{name}-deposit", DirectoryExistsCheck.new(deposit_location)
-  OkComputer::Registry.register "feature-#{name}-sdr2objects", DirectoryExistsCheck.new(sdrobjects_locations, Settings.minimum_subfolder_count)
+  sdrobjects_location = "#{location}/#{Settings.moab.storage_trunk}"
+  OkComputer::Registry.register "feature-#{name}-sdr2objects", DirectoryExistsCheck.new(sdrobjects_location, Settings.minimum_subfolder_count)
 end
 
 OkComputer::Registry.register 'ruby_version', OkComputer::RubyVersionCheck.new
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -35,31 +35,43 @@ class TablesHaveDataCheck < OkComputer::Check
 end
 OkComputer::Registry.register 'feature-tables-have-data', TablesHaveDataCheck.new
 
+# rubocop:disable Metrics/AbcSize
 # check that directory is accessible without consideration for writability
 class DirectoryExistsCheck < OkComputer::Check
-  attr_accessor :directory
+  attr_accessor :directory, :min_subfolder_count
 
-  def initialize(directory)
+  def initialize(directory, min_subfolder_count = nil)
     self.directory = directory
+    self.min_subfolder_count = min_subfolder_count
   end
 
   def check
-    stat = File.stat(directory) if File.exist?(directory)
-    if stat
-      if stat.directory?
-        mark_message "'#{directory}' is a reachable directory"
-      else
-        mark_message "'#{directory}' is not a directory."
-        mark_failure
-      end
-    else
+    unless File.exist? directory
       mark_message "Directory '#{directory}' does not exist."
+      mark_failure
+    end
+
+    unless File.directory? directory
+      mark_message "'#{directory}' is not a directory."
+      mark_failure
+    end
+
+    mark_message "'#{directory}' is a reachable directory"
+    if min_subfolder_count && Dir.entries(directory).size > min_subfolder_count
+      mark_message "'#{directory}' has the required minimum number of subfolders (#{min_subfolder_count})"
+    elsif min_subfolder_count
+      mark_message "'#{directory}' does not have the required minimum number of subfolders (#{min_subfolder_count})"
       mark_failure
     end
   end
 end
+# rubocop:enable Metrics/AbcSize
+
 Settings.storage_root_map.default.each do |name, location|
-  OkComputer::Registry.register "feature-#{name}", DirectoryExistsCheck.new(location)
+  deposit_location = "#{location}/deposit"
+  sdrobjects_locations = "#{location}/sdr2objects"
+  OkComputer::Registry.register "feature-#{name}-deposit", DirectoryExistsCheck.new(deposit_location)
+  OkComputer::Registry.register "feature-#{name}-sdr2objects", DirectoryExistsCheck.new(sdrobjects_locations, Settings.minimum_subfolder_count)
 end
 
 OkComputer::Registry.register 'ruby_version', OkComputer::RubyVersionCheck.new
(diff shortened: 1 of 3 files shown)


ndushay (approved):
LGTM but would like @jmartin-sul to weigh in.

peetucket:
Note that we will want to follow up with a shared_configs PR to set a reasonable value for the minimum subfolder checks in prod.  But this will work just fine without that updated config.

peetucket:
Also note that the configure roots in stage, where I tried this branch, include some folders that don't include both required subfolders, resulting in some failures.  But this seems ok in stage?

```
  feature-fixture_sr1-deposit: FAILED 'spec/fixtures/storage_root01/deposit' is a reachable directory (0.009s)
  feature-fixture_sr1-sdr2objects: PASSED 'spec/fixtures/storage_root01/sdr2objects' is a reachable directory (0.010s)
  feature-fixture_sr2-deposit: FAILED 'spec/fixtures/storage_root02/deposit' is a reachable directory (0.009s)
  feature-fixture_sr2-sdr2objects: PASSED 'spec/fixtures/storage_root02/sdr2objects' is a reachable directory (0.009s)
  feature-fixture_sr3-deposit: FAILED 'spec/fixtures/checksum_root01/deposit' is a reachable directory (0.000s)
  feature-fixture_sr3-sdr2objects: PASSED 'spec/fixtures/checksum_root01/sdr2objects' is a reachable directory (0.000s)
```

config/initializers/okcomputer.rb
```diff
@@ -35,31 +35,43 @@ def table_check(klass)
...
+      mark_failure
+    end
   end
 end
+# rubocop:enable Metrics/AbcSize
+
 Settings.storage_root_map.default.each do |name, location|
-  OkComputer::Registry.register "feature-#{name}", DirectoryExistsCheck.new(location)
+  deposit_location = "#{location}/deposit"
+  sdrobjects_locations = "#{location}/sdr2objects"
```
jmartin-sul, line 72: unlikely to make a material difference in the near future, and will alert loudly/harmlessly if things do change... but just for consistency, would you mind doing:
```ruby
sdrobjects_locations = "#{location}/#{Settings.moab.storage_trunk}"
```

i was going to suggest similar for the `deposit_location` init, but it turns out we don't configure that in this app, AFAICT, since pres cat never directly deals with deposits.  i dunno, maybe that indicates that we should just get rid of the deposit check part (but otherwise keep the logic in this PR the same).  or maybe it indicates this comment is too nitpicky.  your call on whether to change this 🙂

config/settings.yml
```diff
@@ -43,6 +43,8 @@ replication:
   audit_should_backfill: false
 
 total_worker_count: 117 # for okcomputer endpoint
+minimum_subfolder_count: null # for okcomputer pres-cat mount check, verifies the storage_trunk has at least this many subfolders
```
jmartin-sul, line 46: what about a default of 1 in the base settings (and then we can increase the threshold in shared_configs for qa/stage/prod after this is merged)?  i believe all 3 deployed envs can be expected to have at least some Moabs, so a non-null threshold seems preferable to me, at first glance.

jmartin-sul (approved):
i have two minor suggestions, but i would also be happy to merge this as-is if you don't think they're preferable and/or worth the effort.

config/initializers/okcomputer.rb
peetucket: if pres_cat doesn't care about the "deposits" directory, then I would argue the okcomputer check herer shouldn't care either  - and I have removed that explicit check

config/settings.yml
peetucket: sure, set to 1

config/initializers/okcomputer.rb
jmartin-sul: makes sense, apologies for the misleading suggestion in the ticket

config/settings.yml
jmartin-sul: thanks!

jmartin-sul (approved):

jmartin-sul:
> Also note that the configure roots in stage, where I tried this branch, include some folders that don't include both required subfolders, resulting in some failures. But this seems ok in stage?
> 
> ```
>   feature-fixture_sr1-deposit: FAILED 'spec/fixtures/storage_root01/deposit' is a reachable directory (0.009s)
>   feature-fixture_sr1-sdr2objects: PASSED 'spec/fixtures/storage_root01/sdr2objects' is a reachable directory (0.010s)
>   feature-fixture_sr2-deposit: FAILED 'spec/fixtures/storage_root02/deposit' is a reachable directory (0.009s)
>   feature-fixture_sr2-sdr2objects: PASSED 'spec/fixtures/storage_root02/sdr2objects' is a reachable directory (0.009s)
>   feature-fixture_sr3-deposit: FAILED 'spec/fixtures/checksum_root01/deposit' is a reachable directory (0.000s)
>   feature-fixture_sr3-sdr2objects: PASSED 'spec/fixtures/checksum_root01/sdr2objects' is a reachable directory (0.000s)
> ```

sorry, i'd missed this comment when i initially left my review, but now it's moot w/ the most recent tweak

peetucket:
And the associated shared_configs (overriding the subfolder count for production only at this point: https://github.com/sul-dlss/shared_configs/pull/1660)

PR #1717, opened 2021-06-15
Set the query_hash_key for committee

jcoyne:
## Why was this change made?

This avoids a deprecation notice.  Copied from the example at https://github.com/interagent/committee/blob/master/examples/openapi3_rails/config/application.rb#L44


ndushay (approved):

mjgiarlo:
@ndushay @jcoyne :pray:

PR #1728, opened 2021-06-30
ValidateMoabJob only does data file checksums for latest version

### LINKED ISSUE #1720: ValidateMoabJob should only verify checksums for the most recently added content files, not ALL content files
In `ValidateMoabJob`, the validate method currently does checksum validation on all files of an object -- all manifest files and all content files for all versions.

For objects with very large files, we don't want to perform unnecessary checksum validation as it is too time consuming.  Compromise position that should still fulfill the purpose of this new workflow step:

For each version, run:
- verify_signature_catalog (ensure all files in signatureCatalog.xml exist)
- verify_manifest_inventory (checksum validation of all manifest files)
- verify_version_inventory

For ***only the most recent version***, run:
- verify_version_additions

ndushay:
## Why was this change made?

Fixes #1720

We only want to perform checksums for Moab content files for the latest version, not for all versions.

unit tests

### DIFF
--- a/spec/jobs/validate_moab_job_spec.rb
+++ b/spec/jobs/validate_moab_job_spec.rb
@@ -55,6 +55,8 @@ describe ValidateMoabJob, type: :job do
     end
 
     it 'reports failure to workflow server when there are validation errors' do
+      # ensure test object bj102hs9687 has expected errors
+      allow(job).to receive(:validate).and_return(job.send(:verification_errors, moab.version_list.first.verify_version_storage))
       job.perform(druid) # test object bj102hs9687 has errors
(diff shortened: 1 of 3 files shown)


spec/jobs/validate_moab_job_spec.rb
```diff
@@ -67,70 +69,151 @@
     context 'when validation runs' do
       let(:expected_validation_err_regex) { /^Problem with Moab validation run on .*#{error_regex_str}.*/ }
 
-      before do
-        job.perform(druid)
-      end
-
```
ndushay, line 73: with the mocking, it was easier to put this into each spec.

justinlittman (approved):

PR #1729, opened 2021-07-01
inline the logic provided by the resque-lock gem, drop the dependency

### LINKED ISSUE #1726: resque locks should be deleted upon job completion, success or failure
**Describe the bug**

Based on testing https://github.com/sul-dlss/shared_configs/pull/1670 and https://github.com/sul-dlss/workflow-server-rails/pull/470 using `create_preassembly_image_spec.rb` in the integration test suite, it appears that pres cat jobs are not cleaning up the locks they create to prevent duplicate jobs from running when a job with a given set of params is already enqueued.  This prevents e.g. `ValidateMoabJob` from being enqueued twice in short succession (because it will decline to enqueue until the prior run's lock expires, which takes an hour).

**User Impact**

Because this prevents running the same job with the same parameters twice in rapid succession, this prevents `ValidateMoabJob` from running on the same druid twice in less than an hour.  If we don't fix this, we can't merge https://github.com/sul-dlss/workflow-server-rails/pull/470 to enable the new `validate-moab` integrity check (see #1691), because then versioning something twice in less than an hour would result in the `validate-moab` step silently hanging.

some druids that ran into this on stage, that i fixed manually last night (by cleaning up stale resque locks and queuing `ValdiateMoabJob`):
https://argo-stage.stanford.edu/view/druid:bb843md0579
https://argo-stage.stanford.edu/view/druid:ct246zw6526
https://argo-stage.stanford.edu/view/druid:dr564nd1378
https://argo-stage.stanford.edu/view/druid:hg653gq1927
https://argo-stage.stanford.edu/view/kk341tv7892
https://argo-stage.stanford.edu/view/bt821jk7040

Steps to Reproduce, if Known:
* deploy https://github.com/sul-dlss/workflow-server-rails/pull/470
* run create_preassembly_image_spec.rb

**Expected behavior**

The lock that prevents a pres cat job with a specific set of parameters from being enqueued twice at the same time gets cleaned up when said job finishes running.

**Screenshots**

n/a

**Additional context**

We use the resque-lock gem to prevent callers from simultaneously enqueuing multiple instances of the same job before the workers have a chance to pick it up.  resque-lock is supposed to set a mutex value that other workers will respect, but it also has a hook to unset that mutex value when the job finishes, regardless of success, so that the same job with the same parameters can then run again.  It appears that the cleanup isn't happening, and that the lock will always last until its expiry time (1 hour).

This hasn't hasn't been a serious issue for us so far, though Julian did suspect something like this was happening, and I had run into trouble queuing the same checksum validation job repeatedly, even when I didn't attempt to re-queue until after the first run finished.  We did spend a bit of time investigating this, but since the main thing that got us to look at it was Redis memory usage, and since we ended up being comfortable with the amount of memory Redis was using, we ended up dropping it.  

* https://github.com/defunkt/resque-lock/blob/master/lib/resque/plugins/lock.rb#L78
  * https://redis.io/commands/setnx
  * https://redis.io/commands/getset
* https://github.com/resque/resque/blob/master/docs/HOOKS.md
  * https://github.com/resque/resque/blob/7f6b88404dd18698f6f4023e18fdc1ae8318e5e5/test/plugin_test.rb
  * https://github.com/resque/resque/blob/7f6b88404dd18698f6f4023e18fdc1ae8318e5e5/test/job_hooks_test.rb

jmartin-sul:
## Why was this change made?

* wire up lock cleanup so it's actually invoked (fixes https://github.com/sul-dlss/preservation_catalog/issues/1726)
* comments explaining locking behavior and providing necessary license info

closes #1726

- [x] existing unit tests to help detect regressions
- [x] manual testing on stage (by enqueuing multiple copies of the same checksum validation job, some overlapping, some not)
- [x] the infra integration test suite, particularly `create_preassembly_image_spec.rb` (since its failure is what caused me to open #1726)
- [x] i'd like to add a unit test to `application_job_spec.rb` showing that the lock gets cleaned up by the newly inlined code when the job runs successfully
- [x] i'd like to add a unit test to `application_job_spec.rb` showing that the lock gets cleaned up by the newly inlined code even when the job errors out (had a little trouble with this because the error gets raised in the test code when the job is run, which is not what i'd expect for an async job, and not what happens when the job errors IRL)
- [x] a lock timeout related test would be nice
- [ ] a test to address the uncovered conditional that's resulting in the code coverage drop would be nice (to be fair, this is likely a very unusual race condition)

WIP until testing's all sorted out

bonus: have `ObjectsController#validate_moab` return an informative error to the caller in the unlikely event that it can't queue the job.

comments in relevant code

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -15,8 +15,18 @@ class ObjectsController < ApplicationController
   # queue a ValidateMoab job for a specific druid, typically called by a preservationIngestWF robot
   # GET /v1/objects/:id/validate_moab
   def validate_moab
-    ValidateMoabJob.perform_later(druid)
-    render(plain: 'ok', status: :ok)
+    # ActiveJob::Base.perform_later will return an instance of the job if it was added to the
+    # queue successfully. It will return false if the job was not enqueued.
+    # The most likely cause is a temporary queue lock for the job and its specific payload, so
+    # it is up to the caller to determine whether retrying later is appropriate.
+    if ValidateMoabJob.perform_later(druid)
+      render(plain: 'ok', status: :ok)
+    else
+      err_msg = "Failed to enqueue ValidateMoabJob for #{druid}. "\
+                'The most likely cause is that the job was already enqueued and is waiting to be picked up. '\
+                'Retry later if appropriate.'
+      render(plain: err_msg, status: :locked)
+    end
   end
 
   # return a specific file from the Moab
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -51,7 +51,7 @@ class ApplicationJob < ActiveJob::Base
     # return true if we successfully acquired the lock
     # "Set key to hold string value if key does not exist" (otherwise no-op) -- https://redis.io/commands/setnx
     if Resque.redis.setnx(key, timeout)
-      Rails.logger.debug("acquired lock on #{key} (none existed)")
+      Rails.logger.info("acquired lock on #{key} (none existed)")
       return true
     end
 
@@ -59,7 +59,7 @@ class ApplicationJob < ActiveJob::Base
     # (we cannot acquire the lock during the timeout period)
     key_expires_at = Resque.redis.get(key).to_i
     if now <= key_expires_at
-      Rails.logger.debug("failed to acquire lock on #{key}, because it has not expired (#{now} <= #{key_expires_at})")
+      Rails.logger.info("failed to acquire lock on #{key}, because it has not expired (#{now} <= #{key_expires_at})")
       return false
     end
 
@@ -68,19 +68,19 @@ class ApplicationJob < ActiveJob::Base
     # "Atomically sets key to value and returns the old value stored at key." -- https://redis.io/commands/getset
     key_expires_at = Resque.redis.getset(key, timeout).to_i
     if now > key_expires_at
-      Rails.logger.debug("acquired lock on #{key} (old lock expired, #{now} > #{key_expires_at})")
+      Rails.logger.info("acquired lock on #{key} (old lock expired, #{now} > #{key_expires_at})")
       true
     else
-      Rails.logger.debug("failed to acquired lock on #{key} but updated expiry time to #{timeout} (#{now} <= #{key_expires_at})")
+      Rails.logger.info("failed to acquired lock on #{key} but updated expiry time to #{timeout} (#{now} <= #{key_expires_at})")
       false
     end
   end
 
   def self.clear_lock(*args)
     key = queue_lock_key(*args)
-    Rails.logger.debug("clearing lock for #{key}...")
+    Rails.logger.info("clearing lock for #{key}...")
     Resque.redis.del(key).tap do |del_result|
-      Rails.logger.debug("...cleared lock for #{key} (del_result=#{del_result})")
+      Rails.logger.info("...cleared lock for #{key} (del_result=#{del_result})")
     end
   end
 
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -46,16 +46,16 @@ class ApplicationJob < ActiveJob::Base
   def self.before_enqueue_lock(*args) # rubocop:disable Metrics/AbcSize
     key = queue_lock_key(*args)
     now = Time.now.to_i
-    timeout = now + lock_timeout + 1
+    new_expiry_time = now + lock_timeout + 1
 
     # return true if we successfully acquired the lock
     # "Set key to hold string value if key does not exist" (otherwise no-op) -- https://redis.io/commands/setnx
-    if Resque.redis.setnx(key, timeout)
+    if Resque.redis.setnx(key, new_expiry_time)
       Rails.logger.info("acquired lock on #{key} (none existed)")
       return true
     end
 
-    # see if the existing timeout is still valid and return false if it is
+    # see if the existing lock is still valid and return false if it is
     # (we cannot acquire the lock during the timeout period)
     key_expires_at = Resque.redis.get(key).to_i
     if now <= key_expires_at
@@ -63,15 +63,15 @@ class ApplicationJob < ActiveJob::Base
       return false
     end
 
-    # otherwise set the timeout and ensure that no other worker has
+    # otherwise set the new_expiry_time and ensure that no other worker has
     # acquired the lock, possibly pushing out the expiry time further
     # "Atomically sets key to value and returns the old value stored at key." -- https://redis.io/commands/getset
-    key_expires_at = Resque.redis.getset(key, timeout).to_i
+    key_expires_at = Resque.redis.getset(key, new_expiry_time).to_i
     if now > key_expires_at
       Rails.logger.info("acquired lock on #{key} (old lock expired, #{now} > #{key_expires_at})")
       true
     else
-      Rails.logger.info("failed to acquired lock on #{key} but updated expiry time to #{timeout} (#{now} <= #{key_expires_at})")
+      Rails.logger.info("failed to acquire lock on #{key} but updated expiry time to #{new_expiry_time} (#{now} <= #{key_expires_at})")
       false
     end
   end
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,18 +1,17 @@
 # frozen_string_literal: true
 
-require 'resque/plugins/lock'
-
 # Base job for this Application
-# @note We do queue locking via 2 methods from resque-lock in callbacks.
-# We abort on lock collision, preventing duplicate Jobs from enqueuing.
-# The key is specified by a 3rd `lock` method, by default based on job name and
-# serialized arguments (which works well for druid/version), but can be overridden
-# per job class, if needed.  Locks are stored in Redis w/ TTL date integers.
-# @see [resque-lock] https://github.com/defunkt/resque-lock/blob/master/lib/resque/plugins/lock.rb
-# @see [redis setnx] http://redis.io/commands/setnx
+#
+# @note We abort on lock collision, preventing duplicate Jobs from enqueuing.
+# The key is specified by the `queue_lock_key` class method, by default based on job name
+# and serialized arguments (which works well for druid/version), but can be overridden
+# per job class, if needed.  Locks are stored in Redis w/ expiry dates as integers.
+# @see [redis setnx] https://redis.io/commands/setnx (explains the reasoning behind the pattern)
+#
+# @note The queue locking mechanism was lifted from resque-lock. See end of file for
+# obligatory license info and rationale for inlining the gem's logic.
+# @see [resque-lock] https://github.com/defunkt/resque-lock/blob/e06fc2bd26f96f4f3fe893caa49c1cd42c0c9423/lib/resque/plugins/lock.rb
 class ApplicationJob < ActiveJob::Base
-  extend Resque::Plugins::Lock
-
   before_perform do |_job|
     ActiveRecord::Base.clear_active_connections!
   end
@@ -21,16 +20,69 @@ class ApplicationJob < ActiveJob::Base
     throw(:abort) unless job.class.before_enqueue_lock(*job.arguments)
   end
 
-  # Overriding so that changes in ActiveModel object don't result in new lock (which they do when just calling to_s).
-  def self.lock(*args)
-    new_args = args.map { |arg| arg.class.method_defined?(:to_global_id) ? arg.to_global_id.to_s : arg }
-    "lock:#{self.class.name}-#{new_args}"
+  around_perform do |job, block|
+    block.call
+  ensure
+    # Always clear the lock when we're done, even if there was an error.
+    job.class.clear_lock(*job.arguments)
   end
 
   # Override in subclass to tune per queue.
-  # def lock_timeout
-  #   3600
-  # end
+  # @return [Integer] lock_timeout in seconds, after which lock is no longer valid,
+  # regardless of whether the job that created the lock is still in queue or working.
+  def self.lock_timeout
+    # 1 hour is a reasonable default, because most jobs in this app will
+    # cause only minor trouble if re-executed too aggressively.
+    3600
+  end
+
+  # @return [String] the key for locking this job/payload combination, e.g. 'lock:MySpecificJob-bt821jk7040;1'
+  def self.queue_lock_key(*args)
+    # Changes in ActiveModel object don't result in new lock (which they do when just calling to_s).
+    queue_lock_args = args.map { |arg| arg.class.method_defined?(:to_global_id) ? arg.to_global_id.to_s : arg }
+    "lock:#{name}-#{queue_lock_args.join(';')}"
+  end
+
+  def self.before_enqueue_lock(*args) # rubocop:disable Metrics/AbcSize
+    key = queue_lock_key(*args)
+    now = Time.now.to_i
+    timeout = now + lock_timeout + 1
+
+    # return true if we successfully acquired the lock
+    # "Set key to hold string value if key does not exist" (otherwise no-op) -- https://redis.io/commands/setnx
+    if Resque.redis.setnx(key, timeout)
+      Rails.logger.debug("acquired lock on #{key} (none existed)")
+      return true
+    end
+
+    # see if the existing timeout is still valid and return false if it is
+    # (we cannot acquire the lock during the timeout period)
+    key_expires_at = Resque.redis.get(key).to_i
+    if now <= key_expires_at
+      Rails.logger.debug("failed to acquire lock on #{key}, because it has not expired (#{now} <= #{key_expires_at})")
+      return false
+    end
+
+    # otherwise set the timeout and ensure that no other worker has
+    # acquired the lock, possibly pushing out the expiry time further
+    # "Atomically sets key to value and returns the old value stored at key." -- https://redis.io/commands/getset
+    key_expires_at = Resque.redis.getset(key, timeout).to_i
+    if now > key_expires_at
+      Rails.logger.debug("acquired lock on #{key} (old lock expired, #{now} > #{key_expires_at})")
+      true
+    else
+      Rails.logger.debug("failed to acquired lock on #{key} but updated expiry time to #{timeout} (#{now} <= #{key_expires_at})")
+      false
+    end
+  end
+
+  def self.clear_lock(*args)
+    key = queue_lock_key(*args)
+    Rails.logger.debug("clearing lock for #{key}...")
+    Resque.redis.del(key).tap do |del_result|
+      Rails.logger.debug("...cleared lock for #{key} (del_result=#{del_result})")
+    end
+  end
 
   # Automatically retry jobs that encountered a deadlock
   # retry_on ActiveRecord::Deadlocked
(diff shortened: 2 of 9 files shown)


app/jobs/application_job.rb
```diff
@@ -21,16 +20,69 @@ class ApplicationJob < ActiveJob::Base
...
+    # otherwise set the new_expiry_time and ensure that no other worker has
+    # acquired the lock, possibly pushing out the expiry time further
+    # "Atomically sets key to value and returns the old value stored at key." -- https://redis.io/commands/getset
+    key_expires_at = Resque.redis.getset(key, new_expiry_time).to_i
+    if now > key_expires_at
+      Rails.logger.info("acquired lock on #{key} (old lock expired, #{now} > #{key_expires_at})")
+      true
+    else
+      Rails.logger.info("failed to acquired lock on #{key} but updated expiry time to #{new_expiry_time} (#{now} <= #{key_expires_at})")
+      false
```
jmartin-sul, line 75: this `else` is the only part of the PR that's uncovered by tests, and that coverage deficit is what's failing the build.

maybe i'm missing something, but i think this is a hard case to test for, because it requires that a second worker pick up the lock in between the `get` from Redis on line 60 and the `getset` to Redis on line 69.  i do think this condition is worth covering, though, and i've handled it in the same way as the original resque-lock gem, albeit with a bit more logging detail.  but the logic should be equivalent (or that was the intent).

app/jobs/application_job.rb
```diff
@@ -53,3 +105,35 @@ def zip_info_check!(metadata)
...
+# THE SOFTWARE IS PROVIDED AS IS, WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+# https://github.com/defunkt/resque-lock/blob/e06fc2bd26f96f4f3fe893caa49c1cd42c0c9423/LICENSE
+#
+# This additional copyright statement applies solely to the code that was taken from resque-lock.
```
jmartin-sul, line 139: this is the first time i've re-used code in this way, i think, so definitely interested in feedback on whether i got this licensing stuff right.

app/jobs/application_job.rb
```diff
@@ -1,18 +1,17 @@
...
+#
+# @note We abort on lock collision, preventing duplicate Jobs from enqueuing.
+# The key is specified by the `queue_lock_key` class method, by default based on job name
+# and serialized arguments (which works well for druid/version), but can be overridden
+# per job class, if needed.  Locks are stored in Redis w/ expiry dates as integers.
+# @see [redis setnx] https://redis.io/commands/setnx (explains the reasoning behind the pattern)
+#
+# @note The queue locking mechanism was lifted from resque-lock. See end of file for
+# obligatory license info and rationale for inlining the gem's logic.
+# @see [resque-lock] https://github.com/defunkt/resque-lock/blob/e06fc2bd26f96f4f3fe893caa49c1cd42c0c9423/lib/resque/plugins/lock.rb
```
jmartin-sul, line 13: i renamed some of the original methods for clarity.

app/jobs/application_job.rb
```diff
@@ -21,16 +20,69 @@ class ApplicationJob < ActiveJob::Base
...
+  end
+
+  # @return [String] the key for locking this job/payload combination, e.g. 'lock:MySpecificJob-bt821jk7040;1'
+  def self.queue_lock_key(*args)
+    # Changes in ActiveModel object don't result in new lock (which they do when just calling to_s).
+    queue_lock_args = args.map { |arg| arg.class.method_defined?(:to_global_id) ? arg.to_global_id.to_s : arg }
+    "lock:#{name}-#{queue_lock_args.join(';')}"
+  end
+
+  def self.before_enqueue_lock(*args) # rubocop:disable Metrics/AbcSize
```
jmartin-sul, line 46: this method contains the trickiest and most important part of the resque-lock logic.  i tried to keep it basically the same as the original in terms of its use of redis and the conditions it checks, but i did rename some variables and add some logging for clarity.

jmartin-sul:
The newly added logging statements are at `info` level right now.  I'm inclined to leave them that way for a few weeks while this is new to prod, and to file a ticket to turn them down to `debug` later.

spec/jobs/application_job_spec.rb
```diff
@@ -2,34 +2,78 @@
...
   it 'queues start empty' do
-    expect(Resque.info).to include(pending: 0)
+    expect(enqueued_jobs.size).to eq 0
   end
 
   context 'a subclass with message(s) queued' do
-    before do
-      allow(ZipmakerJob).to receive(:perform_later).and_call_original # undo rails_helper block
-      ZipmakerJob.perform_later('1234abc', 1)
-    end
```
jmartin-sul, line 21: i thought this was clearer inline in the individual tests, and no need to undo zipmaker monkeypatching from rails_helper with the new test class

spec/jobs/application_job_spec.rb
```diff
@@ -2,34 +2,78 @@
...
+      expect { RegularParameterJob.perform_later('4321cba', 1, false) }
+        .to change(enqueued_jobs, :size).from(0).to(1)
+    end
+
+    it 'cleans up its lock after failing so that the same job with the same params can be queued again immediately' do
+      RegularParameterJob.perform_later('0987zyx', 1, true)
+
+      begin
+        perform_enqueued_jobs
+      rescue # rubocop:disable Style/RescueStandardError terrible pattern in almost all cases, but acceptable here
```
jmartin-sul, line 61: now that i think about it, i guess i could raise and trap for a custom error, to reduce the chance of a test passing but hiding another issue if code is changed later.  happy to do that if others think it's worth it.

spec/jobs/application_job_spec.rb
```diff
@@ -2,34 +2,78 @@
...
+        # the error is bubbling up to here despite coming from an async job.  IRL, this would not happen.
+        # no idea why ActiveJob is behaving differently in the tests.
```
jmartin-sul, line 63: also, now that i think about this, maybe this comment is off-base.  app code that uses ActiveJob really only ever queues jobs, and doesn't pick them up for execution, and whatever adapter actually runs the job probably does have to deal with this sort of error.  will fix...

spec/jobs/application_job_spec.rb
```diff
@@ -2,34 +2,78 @@
...
+      end
+
+      expect { RegularParameterJob.perform_later('0987zyx', 1, true) }
+        .to change(enqueued_jobs, :size).from(0).to(1)
+    end
+
+    it 'enqueues the job if it detects an existing lock that has expired' do
+      RegularParameterJob.perform_later('4321cba', 1, false)
+
+      sleep(RegularParameterJob.lock_timeout + 3) # lock timeout is very short, this should go well past it
```
jmartin-sul, line 73: i always feel bad `sleep`ing in tests, but i looked and there's actually only one other `sleep` in the tests, for 0.1 seconds.  totally open to non-`sleep` suggestions for testing timeout.

spec/jobs/application_job_spec.rb
```diff
@@ -38,24 +82,23 @@
     let(:cm2) { create :complete_moab }
 
     before do
-      allow(CatalogToMoabJob).to receive(:perform_later).and_call_original # undo rails_helper block
```
jmartin-sul, line 41: turns out this wasn't needed because rails_helper doesn't intercept `CatalogToMoabJob.perform_later`.  maybe cruft.

spec/requests/objects_controller_validate_moab_spec.rb
```diff
@@ -9,7 +9,7 @@
   describe 'GET #validate_moab' do
     context 'when valid druid passed in' do
       before do
-        allow(ValidateMoabJob).to receive(:perform_later).and_return(true)
+        allow(ValidateMoabJob).to receive(:perform_later).and_return(ValidateMoabJob.new)
```
jmartin-sul, line 12: tweaked to give slightly more realistic mocking since i was revising related code and adding tests anyway

app/controllers/objects_controller.rb
```diff
@@ -15,8 +15,18 @@ def show
...
+    # The most likely cause is a temporary queue lock for the job and its specific payload, so
+    # it is up to the caller to determine whether retrying later is appropriate.
+    if ValidateMoabJob.perform_later(druid)
+      render(plain: 'ok', status: :ok)
+    else
+      err_msg = "Failed to enqueue ValidateMoabJob for #{druid}. "\
+                'The most likely cause is that the job was already enqueued and is waiting to be picked up. '\
+                'Retry later if appropriate.'
+      render(plain: err_msg, status: :locked)
+    end
```
ndushay, line 29: This looks awesome ... wonder if this affects ... other robots?   Other queuing systems besides resque?  Etc.  If it's just about resque ... maybe it should say? 

And how sure are you about this ?

justinlittman (approved):

app/jobs/application_job.rb
```diff
@@ -21,16 +20,69 @@ class ApplicationJob < ActiveJob::Base
...
+    key = queue_lock_key(*args)
+    now = Time.now.to_i
+    new_expiry_time = now + lock_timeout + 1
+
+    # return true if we successfully acquired the lock
+    # "Set key to hold string value if key does not exist" (otherwise no-op) -- https://redis.io/commands/setnx
+    if Resque.redis.setnx(key, new_expiry_time)
+      Rails.logger.info("acquired lock on #{key} (none existed)")
```
ndushay, line 54: debug level rather than info?

app/jobs/application_job.rb
```diff
@@ -21,16 +20,69 @@ class ApplicationJob < ActiveJob::Base
...
+      return true
+    end
+
+    # see if the existing lock is still valid and return false if it is
+    # (we cannot acquire the lock during the timeout period)
+    key_expires_at = Resque.redis.get(key).to_i
+    if now <= key_expires_at
+      Rails.logger.info("failed to acquire lock on #{key}, because it has not expired (#{now} <= #{key_expires_at})")
```
ndushay, line 62: ditto - debug level rather than info?

app/jobs/application_job.rb
```diff
@@ -21,16 +20,69 @@ class ApplicationJob < ActiveJob::Base
...
+      return false
+    end
+
+    # otherwise set the new_expiry_time and ensure that no other worker has
+    # acquired the lock, possibly pushing out the expiry time further
+    # "Atomically sets key to value and returns the old value stored at key." -- https://redis.io/commands/getset
+    key_expires_at = Resque.redis.getset(key, new_expiry_time).to_i
+    if now > key_expires_at
+      Rails.logger.info("acquired lock on #{key} (old lock expired, #{now} > #{key_expires_at})")
```
ndushay, line 71: ditto

app/jobs/application_job.rb
```diff
@@ -21,16 +20,69 @@ class ApplicationJob < ActiveJob::Base
...
+      true
+    else
+      Rails.logger.info("failed to acquire lock on #{key} but updated expiry time to #{new_expiry_time} (#{now} <= #{key_expires_at})")
```
ndushay, line 74: warn? no idea.

app/jobs/application_job.rb
ndushay: i'm fine with this - this isn't the only else clause untested in the portfolio.

app/jobs/application_job.rb
```diff
@@ -21,16 +20,69 @@ class ApplicationJob < ActiveJob::Base
...
+      false
+    end
+  end
+
+  def self.clear_lock(*args)
+    key = queue_lock_key(*args)
+    Rails.logger.info("clearing lock for #{key}...")
+    Resque.redis.del(key).tap do |del_result|
+      Rails.logger.info("...cleared lock for #{key} (del_result=#{del_result})")
```
ndushay, line 83: both of these as debug?

app/jobs/application_job.rb
ndushay: This is definitely a good faith attempt;  no idea what is correct, but I think this is sufficient.

app/jobs/application_job.rb
```diff
@@ -53,3 +105,35 @@ def zip_info_check!(metadata)
     end
   end
 end
+
+# Q: Why copy the logic out of resque-lock instead of using the gem?
+# A: It's a small amount of (MIT licensed) code, the gem is and unmaintained (last updated in 2012),
+# and the lock cleanup hook wasn't getting invoked automatically in our jobs.  Inlining the code
+# reduces indirection and dependency (on both resque-lock and on Resque itself, since we use ActiveJob
+# hooks here instead of Resque hooks).
```
ndushay, line 113: I think in the future, no one will remember there was a resque-lock gem.  I think adding "gem" after resque-lock below might be sufficient, and leave lines 109-113 out.   If you're attached to the above, that's fine too.

app/jobs/zip_part_job_base.rb
```diff
@@ -11,8 +11,8 @@ class ZipPartJobBase < ApplicationJob
   end
 
   # Does queue locking on ONLY druid, version and part (as first 3 parameters)
-  def self.lock(*args)
-    "lock:#{name}-#{args.slice(0..2)}"
+  def self.queue_lock_key(*args)
```
ndushay, line 14: is this doing the locking, or getting a key?   the name is a bit confusing.  I'd sooner see "queue_lock" with a comment about it returning the key, or about it returning the key as a side effect ... or something like that.

ndushay (approved):
I have nothing blocking this PR and would suggest merging and deploying immediately so it gets most of today to shake out any unexpected (tautology) problems.

My thoughts:
- level of logging messages might want to be debug and a warn for the else case
- wonder if the lock thing would make sense as a separate class ... it's sort of indirect as a part of application_job.rb.   If this isn't a poor idea, I suggest merging this and doing a separate PR, or making it an issue ... it's not urgent.
- `queue_lock_key` to me implies you're getting a key, not doing the locking.  if it's doing both, I care more about the locking happening, and the key being returned as maybe a side effect?

app/controllers/objects_controller.rb
jmartin-sul: > wonder if this affects ... other robots?

I don't think so... since this method (AFAIK) is only invoked by the new `validate-moab` step's robot.

>  Other queuing systems besides resque?

This uses ActiveJob behavior, and so shouldn't be specific to resque.  I mean, practically, I think we'd have trouble switching from e.g. resque to sidekiq because moab-versioning caches things in class vars (iirc), which isn't thread safe (and so isn't sidekiq safe).  But this, at least, shouldn't block switching 🙂

> And how sure are you about this ?

Very sure that this behavior will work in rails 6.1.x (based on lots of testing in rails console in the course of other testing for this PR).  And after some research yesterday, also fairly certain this will work in rails 6.2+...

The `new_framework_defaults_6_0.rb` file has a config entry `Rails.application.config.active_job.return_false_on_aborted_enqueue = true`.  This actually generates a deprecation warning, `DEPRECATION WARNING: return_false_on_aborted_enqueue= is deprecated and will be removed from Rails 6.2`.  this was worrisome, but after some research, what i found indicated that the old default was `false`, and that the old behavior was to return an instance of the job regardless of whether it was successfully queued.  since we had not previously inspected or acted on the output of our `ActiveJob::Base.perform_later` calls, we were able to switch what will be the default behavior in rails 6.2 onward, when we upgraded to rails 6 in dec 2019.

see:
* https://guides.rubyonrails.org/v6.0.3/configuring.html#configuring-active-job (end of section)
* https://github.com/rails/rails/blob/v6.0.4/activejob/CHANGELOG.md#rails-600beta1-january-18-2019
* https://guides.rubyonrails.org/v6.2.0/configuring.html#configuring-active-job (no longer mentions the config field, also note this redirects to the `edgeguides.` URL since 6.2.x is the current latest)

i believe the deprecation warning is because you'll no longer be able to configure the old behavior in 6.2+.

might be good to get confirmation from e.g. @jcoyne on this, since he did the upgrade to rails 6 on this app (iirc), and since he's generally more up on this sort of rails behavioral detail than i am.

jmartin-sul: example from random testing, where i attempt to re-queue a checksum validation job for the same moab before the first job for that moab finished working:

```
[18] pry(main)> ChecksumValidationJob.perform_later(cm)
acquired lock on lock:ChecksumValidationJob-gid://preservation-catalog/CompleteMoab/25315 (none existed)
Enqueued ChecksumValidationJob (Job ID: 37bb8216-40a2-4d6d-97ce-45dac35dabb0) to Resque(checksum_validation) with arguments: #<GlobalID:0x0000000005a45ce0 @uri=#<URI::GID gid://preservation-catalog/CompleteMoab/25315>>
=> #<ChecksumValidationJob:0x0000000005ac8550
 @arguments=
  [#<CompleteMoab:0x00000000062ba178
    id: 25315,
    version: 2,
    preserved_object_id: 25301,
    moab_storage_root_id: 18,
    created_at: Wed, 11 Mar 2020 16:04:37.670398000 UTC +00:00,
    updated_at: Thu, 30 Apr 2020 16:56:11.688153000 UTC +00:00,
    last_moab_validation: Wed, 11 Mar 2020 16:04:37.664436000 UTC +00:00,
    last_checksum_validation: Thu, 30 Apr 2020 16:56:11.686353000 UTC +00:00,
    size: 609026338,
    status: "ok",
    last_version_audit: Thu, 30 Apr 2020 16:56:11.686309000 UTC +00:00,
    status_details: nil,
    from_moab_storage_root_id: nil>],
 @exception_executions={},
 @executions=0,
 @job_id="37bb8216-40a2-4d6d-97ce-45dac35dabb0",
 @priority=nil,
 @queue_name="checksum_validation",
 @timezone="UTC">
[19] pry(main)> ChecksumValidationJob.perform_later(cm)
failed to acquire lock on lock:ChecksumValidationJob-gid://preservation-catalog/CompleteMoab/25315, because it has not expired (1625188536 <= 1625192135)
Failed enqueuing ChecksumValidationJob to Resque(checksum_validation), a before_enqueue callback halted the enqueuing execution.
=> false
```

app/jobs/zip_part_job_base.rb
jmartin-sul: this only computes the key for the lock, and is used by other methods (via the superclass in this case) both to try to set the lock, and to clear it.  i'd wanted to avoid just `lock` or `queue_lock` as a name, because both felt ambiguous as to whether they were setting the lock, or returning a lock object (though the method does neither of those things).

i think this ends up being clearer in the base class, with the surrounding context of the other callers and stuff... maybe a better comment here would suffice?  also happy to consider other names if the name itself is what seems really off-base.

jmartin-sul:
thanks for the quick reviews @justinlittman and @ndushay!

since this was approved, i'm going to merge, and address naomi's suggestions in a follow-on PR.

app/controllers/objects_controller.rb
ndushay: I was wondering about common-accessioning, which definitely uses a lot of perform-later and also uses resque.   Less so about within prescat.

app/jobs/zip_part_job_base.rb
ndushay: maybe, if it's not actually doing the locking, just change the comment here (and in other similar spots) to something like "key will ensure we lock on a combination of druid, version, and part (first 3 parameters)" or something like that.   Makes it clear it's just a key, and this method isn't performing the lock.  :-)

spec/jobs/application_job_spec.rb
ndushay: Seems the way around this would be to look for an existing lock, but if that code isn't already available to us (maybe it's private methods in resque or something), then meh.   I do hate sleeps, and I hate tests taking longer than they should ... but this can be a separate ticket.

jcoyne:
I don't understand why we need or want to in-line a library.

PR #1738, opened 2021-08-03
allow checksum job to return for multiple checksums when some not found

peetucket:
## Why was this change made?

Fixes sul-dlss/argo#2784 - instead of failing the entire checksum job if some druids are not found, return the response and include information about the missing druids instead

if another unexpected error occurs, still fail the entire job

~~Hold pending verification of this behavior by Andrew~~ Verified in ticket sul-dlss/argo#2784 

Updated/added tests

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -62,10 +62,8 @@ class ObjectsController < ApplicationController
   # note: this is deliberately allowed to be a POST to allow for a large number of druids to be passed in
   # GET OR POST /v1/objects/checksums?druids[]=druid1&druids[]=druid2&druids[]=druid3
   def checksums
-    checksum_list, missing_druids, errored_druids = generate_checksum_list
-
-    bad_recs_msg = "\nStorage object(s) not found for #{missing_druids.join(', ')}" if missing_druids.any?
-    bad_recs_msg = (bad_recs_msg || '') + "\nProblems generating checksums for #{errored_druids.join(', ')}" if errored_druids.any?
+    checksum_list, missing_druids_list, errored_druids = generate_checksum_list
+    bad_recs_msg = "\nProblems generating checksums for #{errored_druids.join(', ')}" if errored_druids.any?
     if bad_recs_msg.present?
       render(plain: "409 Conflict - #{bad_recs_msg}", status: :conflict)
       return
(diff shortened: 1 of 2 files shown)


ndushay (approved):
LGTM

app/controllers/objects_controller.rb
```diff
@@ -62,21 +62,19 @@ def checksum
   # note: this is deliberately allowed to be a POST to allow for a large number of druids to be passed in
   # GET OR POST /v1/objects/checksums?druids[]=druid1&druids[]=druid2&druids[]=druid3
   def checksums
-    checksum_list, missing_druids, errored_druids = generate_checksum_list
-
-    bad_recs_msg = "\nStorage object(s) not found for #{missing_druids.join(', ')}" if missing_druids.any?
-    bad_recs_msg = (bad_recs_msg || '') + "\nProblems generating checksums for #{errored_druids.join(', ')}" if errored_druids.any?
+    checksum_list, missing_druids_list, errored_druids = generate_checksum_list
+    bad_recs_msg = "\nProblems generating checksums for #{errored_druids.join(', ')}" if errored_druids.any?
     if bad_recs_msg.present?
```
jmartin-sul, line 67: i think this could be simplified to just check `errored_druids.any?` now that `bad_recs_msg` isn't built with two different conditions.

app/controllers/objects_controller.rb
```diff
@@ -62,21 +62,19 @@ def checksum
   # note: this is deliberately allowed to be a POST to allow for a large number of druids to be passed in
   # GET OR POST /v1/objects/checksums?druids[]=druid1&druids[]=druid2&druids[]=druid3
   def checksums
-    checksum_list, missing_druids, errored_druids = generate_checksum_list
-
-    bad_recs_msg = "\nStorage object(s) not found for #{missing_druids.join(', ')}" if missing_druids.any?
-    bad_recs_msg = (bad_recs_msg || '') + "\nProblems generating checksums for #{errored_druids.join(', ')}" if errored_druids.any?
+    checksum_list, missing_druids_list, errored_druids = generate_checksum_list
+    bad_recs_msg = "\nProblems generating checksums for #{errored_druids.join(', ')}" if errored_druids.any?
```
jmartin-sul, line 66: i think the leading newline can go

app/controllers/objects_controller.rb
```diff
@@ -62,21 +62,19 @@ def checksum
...
     if bad_recs_msg.present?
       render(plain: "409 Conflict - #{bad_recs_msg}", status: :conflict)
```
jmartin-sul, line 68: i think this might more appropriately be a 500 now that we really don't expect this to happen (since missing druid was our only known error condition, and iirc the reason we chose 409, despite it not being a perfect fit semantically)

jmartin-sul (commented):
had a couple minor bit of feedback, but will just put up a follow on PR in a moment since this is already merged.

peetucket:
Thanks @jmartin-sul - those all sound great.  Happy to review the follow up PR when ready.

jmartin-sul:
> Thanks @jmartin-sul - those all sound great. Happy to review the follow up PR when ready.

follow up PR:  https://github.com/sul-dlss/preservation_catalog/pull/1739

PR #1739, opened 2021-08-03
Checksum report error handling touchup

jmartin-sul:
## Why was this change made?

small follow on to #1738 implementing the changes i'd requested, plus bonus rubocop touchup (added the new rules that didn't necessitate any other changes)

CI

openapi definition

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -185,8 +185,8 @@ paths:
           description: Bad request (e.g., malformed druid(s))
         '406':
           description: Unsupported response format
-        '409':
-          description: Conflict - one or more missing or error-state objects
+        '500':
+          description: Unexpected error generating checksum report for one or more druids
       parameters:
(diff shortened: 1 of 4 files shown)


peetucket (approved):

openapi.yml
```diff
@@ -185,8 +185,8 @@ paths:
           description: Bad request (e.g., malformed druid(s))
         '406':
           description: Unsupported response format
-        '409':
-          description: Conflict - one or more missing or error-state objects
+        '500':
+          description: Unexpected error generating checksum report for one or more druids
```
peetucket, line 189: thanks for catching this improvement to the documentation too

jmartin-sul: this might even be the first time i remembered to do it before the test suite reminded me that i had to 😆

PR #1750, opened 2021-09-21
Add CatalogRemediator class to prune replication failures from catalog 

### LINKED ISSUE #1748: create a method for cleaning up catalog entries for replication failures
for context, this is the current manual process, which doesn't really scale past more than a few druids at a time:  https://github.com/sul-dlss/preservation_catalog/wiki/Cleaning-up-delivery-failures

see also #1733 (this issue describes the meat of the code for the remediation approach described there -- it'd be nice to have it as a properly tested method, so that it can be called from the rails console, and so that we don't have too much important remediation code in an untested rake task)

I think the new method belongs in a new service class, maybe something like `CatalogRemediator#delete_failed_replication_records!`?  It might require some helper methods?  Or maybe not, up to the implementer.

The goal of the method is to take a druid and version, and for that druid/version combo, delete from the database any `ZippedMoabVersion`s (and any child `ZipPart`s for such ZMVs) for which replication to the ZMV's particular cloud endpoint did not fully succeed.  Additionally, I think we should refrain from cleaning up any ZMVs younger than `zip_cache_expiry_time`, to avoid things that might be retriable.  Note that `zip_cache_expiry_time` is specified in minutes in a format that can be fed easily to the `find` command, e.g.

`zip_cache_expiry_time: '+10080' # this is UNIX `find` speak for "greater than seven days"`

Very roughly, I think this method does something like:
* given a druid and version, query for all such `ZippedMoabVersion`s (ZMVs hang off `PreservedObject`, which has the druid, see schema here: https://github.com/sul-dlss/preservation_catalog/tree/main/db)
* for each ZMV older than `zip_cache_expiry_time`, do what `PartReplicationAuditJob` does, and call the two methods it calls in its ZMV loop:  https://github.com/sul-dlss/preservation_catalog/blob/a213649bf733557731f20d7f21c2cb9eacfb63e2/app/jobs/part_replication_audit_job.rb#L17-L18 (this will require instantiating and passing in an `AuditResults` instance, which i think should probably be instantiated anew for each ZMV; it might also be nice to log the results)
* if the ZMV that was just checked has any errors in the audit results (or if it has no child parts), and it's old enough, remove that `ZippedMoabVersion` and its child `ZipPart`s, if any.
* it's probably best to do the deletion for each ZMV (and its child ZPs) in a transaction.  however, the audit code will reach out to the cloud to check for presence of the expected zip parts (assuming it doesn't detect a problem just by looking at the DB entries), so it's probably best not to wrap the audit calls in a transaction (or at least, i don't see much benefit since we're already filtering for things that should be too told to be in flux, and even a few quick network calls inside a DB transaction seems like it should be avoided if possible).
* return the version number and endpoint name for each ZMV that was removed

We can figure out a way to cause some delivery failures or create some bad DB records for testing on QA or stage.

Totally happy to pair on this, and would be interested in feedback on the suggested approach.  Definitely open to other ideas.  The approach I describe above is sort of an amalgam of the above linked wiki entry and the approach described in 2 on [this comment](https://github.com/sul-dlss/preservation_catalog/issues/1733#issuecomment-900730705).  One thing that comment suggests is a size check on the zip parts, basically an audit version of what `ZipMaker` currently does to confirm that the size of the zip it creates is consistent with the size of the moab version in our on prem storage roots (since we might've made some bad uploads before `ZipMaker` had that check added).  Now that I realize we can just reuse some existing audit code in this new method, I think we should just let the size check happen in the audit code we're calling (e.g. `check_child_zip_part_attributes`?), and I'd implement that new size check under the already filed ticket #1742.

mjgiarlo:
Fixes #1748

And call the class from a new rake task, printing out results.

Also includes:
* Add new Rubocop rules
  * Because rubocop barked at me. And, as part of spinning up the app after a long time, do a little work to make sure `rake` does what we expect usually. That involved moving rspec-rails to the dev/test bundler group and some Rakefile touches.

## Why was this change made?

Fixes #1748 

TBD

None

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -323,3 +323,18 @@ Rails/I18nLocaleAssignment: # (new in 2.11)
   Enabled: true
 Rails/UnusedIgnoredColumns: # (new in 2.11)
   Enabled: true
+
+Layout/LineEndStringConcatenationIndentation: # new in 1.18
+  Enabled: true
+Lint/AmbiguousOperatorPrecedence: # new in 1.21
+  Enabled: true
+Lint/AmbiguousRange: # new in 1.19
+  Enabled: true
+Lint/IncompatibleIoSelectWithFiberScheduler: # new in 1.21
+  Enabled: true
+Style/RedundantSelfAssignmentBranch: # new in 1.19
+  Enabled: true
+Rails/EagerEvaluationLogMessage: # new in 2.11
+  Enabled: false
+Rails/RedundantTravelBack: # new in 2.12
+  Enabled: true
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,7 @@ gem 'druid-tools' # for druid validation and druid-tree parsing
 gem 'moab-versioning' # work with Moab Objects
 
 group :development, :test do
+  gem 'rspec-rails', '~> 4.0'
   # Ruby static code analyzer https://rubocop.readthedocs.io/en/latest/
   gem 'rubocop', '~> 1.0'
   gem 'rubocop-rails'
@@ -40,7 +41,6 @@ end
 group :test do
   gem 'factory_bot_rails', '~> 4.0'
   gem 'rails-controller-testing'
-  gem 'rspec-rails', '~> 4.0'
   gem 'shoulda-matchers'
   # Codeclimate is not compatible with 0.18+. See https://github.com/codeclimate/test-reporter/issues/413
   gem 'simplecov', '~> 0.17.1'
--- a/Rakefile
+++ b/Rakefile
@@ -7,4 +7,7 @@ Rails.application.load_tasks
 require 'rubocop/rake_task'
 RuboCop::RakeTask.new
 
+# clear the default task injected by rspec
+task(:default).clear
+
 task default: [:rubocop, :spec]
--- a/app/services/audit_results.rb
+++ b/app/services/audit_results.rb
@@ -115,13 +115,19 @@ class AuditResults
 
   def report_results
     # Report completed
-    completed_results = result_array.select { |result| status_changed_to_ok?(result) }
     report_completed(completed_results)
 
     # Report errors
-    error_results = result_array - completed_results
     report_errors(error_results)
     result_array
+  end
+
+  def completed_results
+    result_array.select(&result_ok?)
+  end
+
+  def error_results
+    result_array.reject(&result_ok?)
   end
 
   def contains_result_code?(code)
@@ -141,6 +147,10 @@ class AuditResults
   end
 
   private
+
+  def result_ok?
+    proc { |result| status_changed_to_ok?(result) }
+  end
 
   def reporters
     @reporters ||= [
--- a/app/services/audit_results.rb
+++ b/app/services/audit_results.rb
@@ -44,11 +44,11 @@ class AuditResults
 
   RESPONSE_CODE_TO_MESSAGES = {
     ACTUAL_VERS_GT_DB_OBJ => 'actual version (%{actual_version}) greater than ' \
-      '%{db_obj_name} db version (%{db_obj_version})',
+                             '%{db_obj_name} db version (%{db_obj_version})',
     ACTUAL_VERS_LT_DB_OBJ => 'actual version (%{actual_version}) less than ' \
-      '%{db_obj_name} db version (%{db_obj_version}); ERROR!',
+                             '%{db_obj_name} db version (%{db_obj_version}); ERROR!',
     CM_PO_VERSION_MISMATCH => 'CompleteMoab online Moab version %{cm_version} ' \
-      'does not match PreservedObject current_version %{po_version}',
+                              'does not match PreservedObject current_version %{po_version}',
     CM_STATUS_CHANGED => 'CompleteMoab status changed from %{old_status} to %{new_status}',
     CREATED_NEW_OBJECT => 'added object to db as it did not exist',
     DB_OBJ_ALREADY_EXISTS => '%{addl} db object already exists',
@@ -57,33 +57,33 @@ class AuditResults
     FILE_NOT_IN_MANIFEST => 'Moab file %{file_path} was not found in Moab manifest %{manifest_file_path}',
     FILE_NOT_IN_MOAB => '%{manifest_file_path} refers to file (%{file_path}) not found in Moab',
     FILE_NOT_IN_SIGNATURE_CATALOG => 'Moab file %{file_path} was not found in ' \
-      'Moab signature catalog %{signature_catalog_path}',
+                                     'Moab signature catalog %{signature_catalog_path}',
     INVALID_ARGUMENTS => 'encountered validation error(s): %{addl}',
     INVALID_MANIFEST => 'unable to parse %{manifest_file_path} in Moab',
     INVALID_MOAB => 'Invalid Moab, validation errors: %{addl}',
     MANIFEST_NOT_IN_MOAB => '%{manifest_file_path} not found in Moab',
     MOAB_CHECKSUM_VALID => 'checksum(s) match',
     MOAB_FILE_CHECKSUM_MISMATCH => 'checksums or size for %{file_path} version ' \
-      '%{version} do not match entry in latest signatureCatalog.xml.',
+                                   '%{version} do not match entry in latest signatureCatalog.xml.',
     MOAB_NOT_FOUND => 'db CompleteMoab (created %{db_created_at}; last updated ' \
-      '%{db_updated_at}) exists but Moab not found',
+                      '%{db_updated_at}) exists but Moab not found',
     SIGNATURE_CATALOG_NOT_IN_MOAB => '%{signature_catalog_path} not found in Moab',
     UNABLE_TO_CHECK_STATUS => 'unable to validate when CompleteMoab status is %{current_status}',
     UNEXPECTED_VERSION => 'actual version (%{actual_version}) has unexpected ' \
-      'relationship to %{db_obj_name} db version (%{db_obj_version}); ERROR!',
+                          'relationship to %{db_obj_name} db version (%{db_obj_version}); ERROR!',
     VERSION_MATCHES => 'actual version (%{actual_version}) matches %{addl} db version',
     ZIP_PART_CHECKSUM_MISMATCH => 'replicated md5 mismatch on %{endpoint_name}: ' \
-      "%{s3_key} catalog md5 (%{md5}) doesn't match the replicated md5 " \
-      '(%{replicated_checksum}) on %{bucket_name}',
+                                  "%{s3_key} catalog md5 (%{md5}) doesn't match the replicated md5 " \
+                                  '(%{replicated_checksum}) on %{bucket_name}',
     ZIP_PART_NOT_FOUND => 'replicated part not found on %{endpoint_name}: ' \
-      '%{s3_key} was not found on %{bucket_name}',
+                          '%{s3_key} was not found on %{bucket_name}',
     ZIP_PARTS_COUNT_DIFFERS_FROM_ACTUAL => '%{version} on %{endpoint_name}: ' \
-      "ZippedMoabVersion stated parts count (%{db_count}) doesn't match actual " \
-      'number of zip parts rows (%{actual_count})',
+                                           "ZippedMoabVersion stated parts count (%{db_count}) doesn't match actual " \
+                                           'number of zip parts rows (%{actual_count})',
     ZIP_PARTS_COUNT_INCONSISTENCY => '%{version} on %{endpoint_name}: ' \
-      'ZippedMoabVersion has variation in child parts_counts: %{child_parts_counts}',
+                                     'ZippedMoabVersion has variation in child parts_counts: %{child_parts_counts}',
     ZIP_PARTS_NOT_ALL_REPLICATED => '%{version} on %{endpoint_name}: not all ' \
-      'ZippedMoabVersion parts are replicated yet: %{unreplicated_parts_list}',
+                                    'ZippedMoabVersion parts are replicated yet: %{unreplicated_parts_list}',
     ZIP_PARTS_NOT_CREATED => '%{version} on %{endpoint_name}: no zip_parts exist yet for this ZippedMoabVersion'
   }.freeze
 
--- a/app/services/catalog_remediator.rb
+++ b/app/services/catalog_remediator.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+# Remediates catalog entries, e.g., failed replication records
+class CatalogRemediator
+  def self.prune_replication_failures(druid:, version:)
+    new(druid: druid, version: version).prune_replication_failures
+  end
+
+  def initialize(druid:, version:)
+    @druid = druid
+    @version = version
+  end
+
+  def prune_replication_failures
+    zipped_moab_versions_with_errors.map do |zipped_moab_version, audit_results|
+      zip_parts = zipped_moab_version.zip_parts
+      Rails.logger.info(
+        "Replication failure error(s) found with #{druid} (v#{version}): #{audit_results.error_results}\n" \
+        "Destroying zip parts (#{zip_parts.pluck(:id)}) and zipped moab version (#{zipped_moab_version.id})"
+      )
+
+      ActiveRecordUtils.with_transaction_and_rescue(audit_results) do
+        # NOTE: The order here matters! ZipPart instances must be destroyed before the ZippedMoabVersion
+        zip_parts.destroy_all
+        zipped_moab_version.destroy
+      end
+
+      [zipped_moab_version.version, zipped_moab_version.zip_endpoint.endpoint_name]
+    end
+  end
+
+  private
+
+  attr_reader :druid, :version
+
+  delegate :check_child_zip_part_attributes, to: Audit::CatalogToArchive
+
+  def preserved_object
+    PreservedObject.find_by(druid: druid)
+  end
+
+  def zip_cache_expiry_timestamp
+    Settings.zip_cache_expiry_time.to_i.minutes.ago
+  end
+
+  def zipped_moab_versions_beyond_expiry
+    preserved_object.zipped_moab_versions.where(version: version, created_at: ..zip_cache_expiry_timestamp)
+  end
+
+  def zipped_moab_versions_with_errors
+    zipped_moab_versions_beyond_expiry.to_a.filter_map do |zipped_moab_version|
+      audit_results = audit_results_for(zipped_moab_version)
+      # If this check returns false, it indicates a lack of zip parts, in which
+      # case we still want to delete the ZMV, so short-circuit the rest of the
+      # block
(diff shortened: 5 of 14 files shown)


Rakefile
```diff
@@ -7,4 +7,7 @@ Rails.application.load_tasks
 require 'rubocop/rake_task'
 RuboCop::RakeTask.new
 
+# clear the default task injected by rspec
+task(:default).clear
```
mjgiarlo, line 11: This is what we do everywhere to make sure the default rake task is ours.

Gemfile
```diff
@@ -27,6 +27,7 @@ gem 'druid-tools' # for druid validation and druid-tree parsing
 gem 'moab-versioning' # work with Moab Objects
 
 group :development, :test do
+  gem 'rspec-rails', '~> 4.0'
```
mjgiarlo, line 30: Moved because 

1. without this change, the default rake task winds up not invoking RSpec at all;
2. rspec-rails docs suggest it should be in the dev/test bundler group not just test

.rubocop.yml
```diff
@@ -323,3 +323,18 @@ Rails/I18nLocaleAssignment: # (new in 2.11)
...
+Lint/AmbiguousOperatorPrecedence: # new in 1.21
+  Enabled: true
+Lint/AmbiguousRange: # new in 1.19
+  Enabled: true
+Lint/IncompatibleIoSelectWithFiberScheduler: # new in 1.21
+  Enabled: true
+Style/RedundantSelfAssignmentBranch: # new in 1.19
+  Enabled: true
+Rails/EagerEvaluationLogMessage: # new in 2.11
+  Enabled: false
```
mjgiarlo, line 338: `false` so that we don't need to change our tests around logging behavior. 🤷🏻‍♂️

app/services/audit_results.rb
```diff
@@ -57,33 +57,33 @@ class AuditResults
...
-      "ZippedMoabVersion stated parts count (%{db_count}) doesn't match actual " \
-      'number of zip parts rows (%{actual_count})',
+                                           "ZippedMoabVersion stated parts count (%{db_count}) doesn't match actual " \
+                                           'number of zip parts rows (%{actual_count})',
     ZIP_PARTS_COUNT_INCONSISTENCY => '%{version} on %{endpoint_name}: ' \
-      'ZippedMoabVersion has variation in child parts_counts: %{child_parts_counts}',
+                                     'ZippedMoabVersion has variation in child parts_counts: %{child_parts_counts}',
     ZIP_PARTS_NOT_ALL_REPLICATED => '%{version} on %{endpoint_name}: not all ' \
-      'ZippedMoabVersion parts are replicated yet: %{unreplicated_parts_list}',
+                                    'ZippedMoabVersion parts are replicated yet: %{unreplicated_parts_list}',
```
mjgiarlo, line 86: Changes above this line are Rubocop auto-fixes.

app/services/audit_results.rb
```diff
@@ -115,15 +115,25 @@ def remove_db_updated_results
 
   def report_results
     # Report completed
-    completed_results = result_array.select { |result| status_changed_to_ok?(result) }
```
mjgiarlo, line 118: Extract variable to (public) method. This lets collaborators act on error results without needing knowledge of how to discern error results from non-error results.

app/services/catalog_remediator.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+# Remediates catalog entries, e.g., failed replication records
+class CatalogRemediator
```
mjgiarlo, line 4: This class is entirely uncovered right now! I'd like feedback on the approach before writing unit tests, if that's OK, just in case we want to make substantial changes.

app/services/catalog_remediator.rb
```diff
@@ -0,0 +1,62 @@
...
+    preserved_object.zipped_moab_versions.where(version: version, created_at: zip_cache_expiry_timestamp..)
+  end
+
+  def zipped_moab_versions_with_errors
+    zipped_moab_versions_beyond_expiry.to_a.filter_map do |zipped_moab_version|
+      results = AuditResults.new(druid, version, zipped_moab_version.zip_endpoint)
+      next unless check_child_zip_part_attributes(zipped_moab_version, results)
+
+      zipped_moab_version.zip_endpoint.audit_class.check_replicated_zipped_moab_version(zipped_moab_version, results)
+      # TODO: "set check_unreplicated_parts to true"
```
mjgiarlo, line 56: This is a recommendation @jmartin-sul made on the ticket. I'll see to this soon.

Gemfile
jmartin-sul: thanks!  i had noticed that rspec wouldn't run by default for me locally when calling `rake`, but i never made time to figure out why, and would just `bundle exec rubocop && bundle exec rspec`.

.rubocop.yml
jmartin-sul: yeah, i can see the point of this cop, but at present, i don't think we'd gain a ton in this code base by enabling it.

app/services/catalog_remediator.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def zipped_moab_versions_beyond_expiry
+    preserved_object.zipped_moab_versions.where(version: version, created_at: zip_cache_expiry_timestamp..)
+  end
+
+  def zipped_moab_versions_with_errors
+    zipped_moab_versions_beyond_expiry.to_a.filter_map do |zipped_moab_version|
+      results = AuditResults.new(druid, version, zipped_moab_version.zip_endpoint)
+      next unless check_child_zip_part_attributes(zipped_moab_version, results)
```
jmartin-sul, line 53: i believe that if this returns `false`, it indicates a lack of zip parts, in which case we still want to delete the ZMV.  but, as is already the case in this PR (and the similar audit job code), i do think it's correct to skip the `check_replicated_zipped_moab_version` call when `check_child_zip_part_attributes` returns false.

so... maybe this should be `next [zipped_moab_version, results.error_results] unless ...`?  i'm a little rusty on the nuances of ruby `next` usage.

jmartin-sul (commented):
approach so far LGTM!  i like the decomposition in the service class, feels very readable to me.

app/services/catalog_remediator.rb
mjgiarlo: That sounds exactly right to me. Made that change and dropped a comment copied/pasted from this blurb. 🍻

app/services/audit_results.rb
jmartin-sul: as discussed the other day on slack: reminder to @jmartin-sul to file a ticket to:
* add a list of non-error result codes
* and to switch to using that list for determining whether a result is "ok", instead of `status_changed_to_ok?`'s current regexp logic (which i think will allow some non-error result codes to be classified as errors, though i also don't think that'll be an issue for the replication result codes we care about in this PR, hence dealing with it as follow on work)

mjgiarlo: Sounds good!

jmartin-sul:
testing so far on stage looks good:
* the rake task appropriately cleaned up the three missing ZMV records (and child zip parts) for xr565kx3078
* it appropriately left dp188yy3158 alone, as its cloud copies were fine
* zf568gg7790 - 1 bad copy (aws east), remediated correctly
* yv860tq1898 - 1 bad copy (ibm), remediated correctly
* all remediated druids replicated successfully by way of `PreservedObject#create_zipped_moab_versions!` (with manual audit triggered by `#audit_moab_version_replication!` after replication had completed, according to resque web console)

```ruby
druid_list = %w[xr565kx3078 zf568gg7790 yv860tq1898]
ZipPart.joins(zipped_moab_version: [:zip_endpoint, :preserved_object]).where(zipped_moab_version: { preserved_object: PreservedObject.where(druid: druid_list) }).order(:druid, :endpoint_name, :version, :suffix).pluck(:druid, :endpoint_name, :version, :suffix, :status, :created_at, :updated_at)
=> [["xr565kx3078",
  "aws_s3_east_1",
  1,
  ".zip",
  "ok",
  Sat, 25 Sep 2021 00:53:50.617238000 UTC +00:00,
  Sat, 25 Sep 2021 00:55:05.427718000 UTC +00:00],
 ["xr565kx3078",
  "aws_s3_west_2",
  1,
  ".zip",
  "ok",
  Sat, 25 Sep 2021 00:53:50.596238000 UTC +00:00,
  Sat, 25 Sep 2021 00:55:06.337690000 UTC +00:00],
 ["xr565kx3078",
  "ibm_us_south",
  1,
  ".zip",
  "ok",
  Sat, 25 Sep 2021 00:53:50.605899000 UTC +00:00,
  Sat, 25 Sep 2021 00:55:03.326062000 UTC +00:00],
 ["yv860tq1898",
  "aws_s3_east_1",
  1,
  ".zip",
  "ok",
  Wed, 23 Sep 2020 16:17:41.426273000 UTC +00:00,
  Sat, 25 Sep 2021 00:55:11.274536000 UTC +00:00],
 ["yv860tq1898",
  "aws_s3_west_2",
  1,
  ".zip",
  "ok",
  Wed, 23 Sep 2020 16:17:41.170539000 UTC +00:00,
  Sat, 25 Sep 2021 00:55:10.161716000 UTC +00:00],
 ["yv860tq1898",
  "ibm_us_south",
  1,
  ".zip",
  "ok",
  Sat, 25 Sep 2021 00:53:50.600686000 UTC +00:00,
  Sat, 25 Sep 2021 00:55:05.856585000 UTC +00:00],
 ["zf568gg7790",
  "aws_s3_east_1",
  1,
  ".zip",
  "ok",
  Sat, 25 Sep 2021 00:53:50.601529000 UTC +00:00,
  Sat, 25 Sep 2021 00:55:09.631643000 UTC +00:00],
 ["zf568gg7790",
  "aws_s3_west_2",
  1,
  ".zip",
  "ok",
  Wed, 23 Sep 2020 16:17:42.634910000 UTC +00:00,
  Sat, 25 Sep 2021 00:55:09.678616000 UTC +00:00],
 ["zf568gg7790",
  "ibm_us_south",
  1,
  ".zip",
  "ok",
  Wed, 23 Sep 2020 16:17:42.775584000 UTC +00:00,
  Sat, 25 Sep 2021 00:55:06.046305000 UTC +00:00]]
```

note that for zf568gg7790 and yv860tq1898, the expected two of the three zip parts for each druid were created about a year ago, and only the part that was remediated has today as a creation date (xr565kx3078 shows all of its parts as created today, since it hadn't yet replicated to any endpoints successfully).

mjgiarlo:
> testing so far on stage looks good (paired w/ @mjgiarlo on testing for a bit, then did a bit more testing after that):
> * the rake task appropriately cleaned up the three missing ZMV records (and child zip parts) for xr565kx3078
> * it appropriately left dp188yy3158 alone, as its cloud copies were fine
> * zf568gg7790 - 1 bad copy (aws east), remediated correctly
> * yv860tq1898 - 1 bad copy (ibm), remediated correctly
> * all remediated druids replicated successfully by way of `PreservedObject#create_zipped_moab_versions!` (with manual audit triggered by `#audit_moab_version_replication!` after replication had completed, according to resque web console)
> 
> ```ruby
> druid_list = %w[xr565kx3078 zf568gg7790 yv860tq1898]
> ZipPart.joins(zipped_moab_version: [:zip_endpoint, :preserved_object]).where(zipped_moab_version: { preserved_object: PreservedObject.where(druid: druid_list) }).order(:druid, :endpoint_name, :version, :suffix).pluck(:druid, :endpoint_name, :version, :suffix, :status, :created_at, :updated_at)
> => [["xr565kx3078",
>   "aws_s3_east_1",
>   1,
>   ".zip",
>   "ok",
>   Sat, 25 Sep 2021 00:53:50.617238000 UTC +00:00,
>   Sat, 25 Sep 2021 00:55:05.427718000 UTC +00:00],
>  ["xr565kx3078",
>   "aws_s3_west_2",
>   1,
>   ".zip",
>   "ok",
>   Sat, 25 Sep 2021 00:53:50.596238000 UTC +00:00,
>   Sat, 25 Sep 2021 00:55:06.337690000 UTC +00:00],
>  ["xr565kx3078",
>   "ibm_us_south",
>   1,
>   ".zip",
>   "ok",
>   Sat, 25 Sep 2021 00:53:50.605899000 UTC +00:00,
>   Sat, 25 Sep 2021 00:55:03.326062000 UTC +00:00],
>  ["yv860tq1898",
>   "aws_s3_east_1",
>   1,
>   ".zip",
>   "ok",
>   Wed, 23 Sep 2020 16:17:41.426273000 UTC +00:00,
>   Sat, 25 Sep 2021 00:55:11.274536000 UTC +00:00],
>  ["yv860tq1898",
>   "aws_s3_west_2",
>   1,
>   ".zip",
>   "ok",
>   Wed, 23 Sep 2020 16:17:41.170539000 UTC +00:00,
>   Sat, 25 Sep 2021 00:55:10.161716000 UTC +00:00],
>  ["yv860tq1898",
>   "ibm_us_south",
>   1,
>   ".zip",
>   "ok",
>   Sat, 25 Sep 2021 00:53:50.600686000 UTC +00:00,
>   Sat, 25 Sep 2021 00:55:05.856585000 UTC +00:00],
>  ["zf568gg7790",
>   "aws_s3_east_1",
>   1,
>   ".zip",
>   "ok",
>   Sat, 25 Sep 2021 00:53:50.601529000 UTC +00:00,
>   Sat, 25 Sep 2021 00:55:09.631643000 UTC +00:00],
>  ["zf568gg7790",
>   "aws_s3_west_2",
>   1,
>   ".zip",
>   "ok",
>   Wed, 23 Sep 2020 16:17:42.634910000 UTC +00:00,
>   Sat, 25 Sep 2021 00:55:09.678616000 UTC +00:00],
>  ["zf568gg7790",
>   "ibm_us_south",
>   1,
>   ".zip",
>   "ok",
>   Wed, 23 Sep 2020 16:17:42.775584000 UTC +00:00,
>   Sat, 25 Sep 2021 00:55:06.046305000 UTC +00:00]]
> ```
> 
> note that for zf568gg7790 and yv860tq1898, the expected two of the three zip parts for each druid were created about a year ago, and only the part that was remediated has today as a creation date (xr565kx3078 shows all of its parts as created today, since it hadn't yet replicated to any endpoints successfully).

Thank you!

jmartin-sul (changes requested):
two small nitpicky suggestions, and i'd like to do a quick manual test of the no-zip-parts scenario on stage (which i'll save for monday morning).  but otherwise looks good, and i think we can merge and start using it manually on small batches of stuff in prod if the last of the stage testing goes as expected.

lib/tasks/prescat.rake
```diff
@@ -3,6 +3,13 @@
 require 'csv'
 
 namespace :prescat do
+  desc 'Prune failed remediation records from catalog'
+  task :prune_failed_remediation, [:druid, :version] => :environment do |_task, args|
```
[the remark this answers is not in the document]
mjgiarlo: OK.

spec/services/catalog_remediator_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      expect(instance).to have_received(:prune_replication_failures).once
+    end
+  end
+
+  # NOTE: Yes, we are testing a private method here. There's some logic in the
+  #       method that is important enough, given the scope and scale of what
+  #       prescat does, to take extra care to test that it works as expected.
+  #       Why not make it a public method? There aren't any use cases for making
+  #       the method public; there would currently be zero collaborators using
+  #       it. Thus, we'll test the private method here.
```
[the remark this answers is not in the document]
mjgiarlo: 🍻

app/services/catalog_remediator.rb
```diff
@@ -0,0 +1,72 @@
...
+        "Destroying zip parts (#{zip_parts.pluck(:id)}) and zipped moab version (#{zipped_moab_version.id})"
+      )
+
+      ActiveRecordUtils.with_transaction_and_rescue(audit_results) do
+        # NOTE: The order here matters! ZipPart instances must be destroyed before the ZippedMoabVersion
+        zip_parts.destroy_all
+        zipped_moab_version.destroy
+      end
+
+      [zipped_moab_version.version, zipped_moab_version.zip_endpoint.endpoint_name]
```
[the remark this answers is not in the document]
mjgiarlo: Alternatively, what if we jettison ActiveRecordUtils.yadayada and use `ApplicationRecord.transaction`? I'm fine with either. If there are exceptions, should we rescue and print informational messages or should we let it blow up?

spec/services/catalog_remediator_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      expect(Rails.logger).to have_received(:info).twice
+    end
+
+    it 'creates a transaction' do
+      expect(ActiveRecordUtils).to have_received(:with_transaction_and_rescue).twice
+    end
+
+    it 'destroys the affected zipped moab versions and related zip parts' do
+      expect(zipped_moab_version_no_parts.zip_parts).to be_empty
+      expect(zipped_moab_version_no_parts).to be_destroyed
```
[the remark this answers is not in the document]
mjgiarlo: When I wrote this test, I wasn't sure if it was there or not. It seemed like it would be. And, indeed, it was. Live by the rails, die by the rails.

app/services/catalog_remediator.rb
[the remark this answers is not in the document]
jmartin-sul: that works for me, whichever you'd prefer (sorry for the slow reply)

mjgiarlo:
@jmartin-sul 💬 
> two small nitpicky suggestions, and i'd like to do a quick manual test of the no-zip-parts scenario on stage (which i'll save for monday morning). but otherwise looks good, and i think we can merge and start using it manually on small batches of stuff in prod if the last of the stage testing goes as expected.

Any other changes or tweaks I can toss at the branch?

jmartin-sul:
@mjgiarlo 💬 

> @jmartin-sul 💬
> 
> > two small nitpicky suggestions, and i'd like to do a quick manual test of the no-zip-parts scenario on stage (which i'll save for monday morning). but otherwise looks good, and i think we can merge and start using it manually on small batches of stuff in prod if the last of the stage testing goes as expected.
> 
> Any other changes or tweaks I can toss at the branch?

nothing i can think of, @mjgiarlo!

also i just finished manual testing on stage, all looks well there!

mjgiarlo:
> @mjgiarlo speech_balloon
> 
> > @jmartin-sul speech_balloon
> > > two small nitpicky suggestions, and i'd like to do a quick manual test of the no-zip-parts scenario on stage (which i'll save for monday morning). but otherwise looks good, and i think we can merge and start using it manually on small batches of stuff in prod if the last of the stage testing goes as expected.
> > 
> > 
> > Any other changes or tweaks I can toss at the branch?
> 
> nothing i can think of, @mjgiarlo!
> 
> also i just finished manual testing on stage, all looks well there!

Great news!

ndushay (approved):
I did a driveby-review;  LGTM.

jmartin-sul (approved):

PR #1754, opened 2021-09-28
add orphaned ZMV query to db README

jmartin-sul:
## Why was this change made?

i found this useful when looking for test cases on stage for https://github.com/sul-dlss/preservation_catalog/pull/1750

it should be useful for auditing the catalog in the future.

i spot checked a number of druids the query returned on stage, confirming that they in fact had no zip parts for the versions returned by this query.

documentation only change

### DIFF
--- a/db/README.md
+++ b/db/README.md
@@ -415,6 +415,39 @@ druids_with_replication_issues_human = druids_with_replication_issues.map do |re
 end
 ```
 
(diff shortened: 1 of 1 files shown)


jmartin-sul:
new section:

Screen Shot 2021-09-28 at 9 47 36 AM

ndushay (approved):

PR #1769, opened 2021-11-22
fix prod deploy targets

ndushay:
## Why was this change made?

Ops seems to have changed the VM name:

image


mjgiarlo (approved):
LGTM. Can you confirm this is copacetic, @tallenaz @sul-dlss/operations ?

ndushay:
`cap prod deploy` is happy now, in any case.

tallenaz:
Yes, copacetic.

PR #1775, opened 2022-01-07
make capistrano deployment roles consistent between prod, stage, and qa

jmartin-sul:
specifically:
* everything that has the `app` role puts it first, for a bit more readability
* the missing `queue_populator` and `cache_cleaner` roles are added to the worker boxes (-02) for stage and QA, as they apparently went entirely missing at some point (noticed because the zip temp space filled up on stage, and the cache cleaner rake task wasn't in the crontab for -stage-02, where i'd have expected it).

## Why was this change made?

noticed because stage workers couldn't create archive zips, because the zip temp space had filled up

* this branch was deployed to stage at 4:37 pm
* the job was allowed to run at 5:00 pm (it's scheduled for the top of the hour)
* space started freeing up once the job started (at 635 GB of 1 TB available as of 5:34 pm)
* the failed `zipmaker` jobs were retried, and all of the 10 or so that i spot checked succeeded (including a 9.78 GB object, with a 41.1 GB object still chugging along)

cap deploy configs for the QA and stage envs

### DIFF
--- a/config/deploy/stage.rb
+++ b/config/deploy/stage.rb
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 server 'preservation-catalog-stage-01.stanford.edu', user: 'pres', roles: %w[app db web]
-server 'preservation-catalog-stage-02.stanford.edu', user: 'pres', roles: %w[resque app]
+server 'preservation-catalog-stage-02.stanford.edu', user: 'pres', roles: %w[app resque queue_populator cache_cleaner]
 
 Capistrano::OneTimeKey.generate_one_time_key!
 set :rails_env, 'production'
 set :bundle_without, 'deploy test'
 set :deploy_to, '/opt/app/pres/preservation_catalog'
+set :whenever_roles, [:queue_populator, :cache_cleaner]
 append :linked_files, 'config/newrelic.yml', 'config/resque.yml'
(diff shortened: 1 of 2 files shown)


config/deploy/stage.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 server 'preservation-catalog-stage-01.stanford.edu', user: 'pres', roles: %w[app db web]
-server 'preservation-catalog-stage-02.stanford.edu', user: 'pres', roles: %w[resque app]
+server 'preservation-catalog-stage-02.stanford.edu', user: 'pres', roles: %w[app resque queue_populator cache_cleaner]
```
jmartin-sul, line 4: `queue_populator` is what queues up the various weekly audit jobs.  i don't think anyone had noticed this not running on stage, i only noticed it when fixing the `cache_cleaner` issue.

justinlittman (approved):

jmartin-sul:
i'm not sure why codeclimate isn't reporting coverage.  but this PR shouldn't affect coverage at all, as this is functionality/configuration that isn't easily testable in CI (and so is tested manually).

anyway, merging with the approval, will investigate the codeclimate problem if it pops up again on another PR, esp one with regular app code changes.

PR #1787, opened 2022-01-14
rubocop: ruby 3, use new cops and take autocorrect for violations

ndushay:
## Why was this change made?

set rubocop to ruby 3.0, reduce noise for next sprint

running rubocop and circleci


app/models/complete_moab.rb
```diff
@@ -29,7 +29,7 @@ class CompleteMoab < ApplicationRecord
   #  assign the other complete_moab to preserved_objects_primary_moab
   has_one :preserved_objects_primary_moab, dependent: :destroy
 
-  validates :moab_storage_root, :preserved_object, :status, :version, presence: true
```
jcoyne, line 32: Are we dropping these validations intentionally?  Why?

app/models/preserved_object.rb
```diff
@@ -68,7 +67,7 @@ def audit_moab_version_replication!
   # Of those eligible for replication, use whichever was checksum validated most recently.
   # @return [String, nil] The storage location (storage_root/storage_trunk) where the Moab that should be replicated lives.
   def moab_replication_storage_location
-    moabs_eligible_for_replication.joins(:moab_storage_root).order(last_checksum_validation: :desc).limit(1).pluck(:storage_location).first
+    moabs_eligible_for_replication.joins(:moab_storage_root).order(last_checksum_validation: :desc).limit(1).pick(:storage_location)
```
jcoyne, line 70: Won't this now return an Array, where it returned a scalar before?

app/models/zip_part.rb
```diff
@@ -16,7 +16,7 @@ class ZipPart < ApplicationRecord
     'replicated_checksum_mismatch' => 3
   }
 
-  validates :zipped_moab_version, :create_info, presence: true
+  validates :create_info, presence: true
```
jcoyne, line 19: Why are we dropping this validation?

app/models/zipped_moab_version.rb
```diff
@@ -12,7 +12,7 @@ class ZippedMoabVersion < ApplicationRecord
   belongs_to :zip_endpoint, inverse_of: :zipped_moab_versions
   has_many :zip_parts, dependent: :restrict_with_exception, inverse_of: :zipped_moab_version
 
-  validates :preserved_object, :version, :zip_endpoint, presence: true
+  validates :version, presence: true
```
jcoyne, line 15: Why are we dropping these validations?

jcoyne (changes requested):

app/models/preserved_object.rb
ndushay: pick takes the first:  https://til.hashrocket.com/posts/wcwerkn2il-rails-6-new-activerecord-method-pick

app/models/zip_part.rb
ndushay: Rails/RedundantPresenceValidationOnBelongsTo

app/models/zipped_moab_version.rb
ndushay: Rails/RedundantPresenceValidationOnBelongsTo

ndushay (commented):
JCoyne:  I responded to your questions.  I blithely assumed that the rubocop rules know what they are doing, and the validation rule has the word "redundant" in it which gave me reassurance.

https://www.bigbinary.com/blog/rails-5-makes-belong-to-association-required-by-default

We don't seem to have strayed from the default configuration of "required" and these "belongs_to" statements are generally 4 years old ...

jcoyne (approved):
Sounds good, but those validation changes are causing the tests to fail.

.rubocop.yml
```diff
@@ -373,7 +373,7 @@ Rails/CompactBlank: # new in 2.13
 Rails/DurationArithmetic: # new in 2.13
   Enabled: true
 Rails/RedundantPresenceValidationOnBelongsTo: # new in 2.13
-  Enabled: true
+  Enabled: false # NOTE: we want this to be false!
```
jcoyne, line 376: Can you leave a note here about why we want it to be false? "Why" is always the question I wonder about when I look to comments.

ndushay: I switched this back to true and fixed the tests.  In most cases we weren't doing explicit validation in the code, so I only had to change that in one file.

app/models/complete_moab.rb
```diff
@@ -29,7 +31,7 @@ class CompleteMoab < ApplicationRecord
...
+  validates :status, :version, presence: true
```
ndushay, line 34: this is the only spot where I changed the explicit validations in the code, and I added the comment above "belongs_to" for good measure.

ndushay:
@jcoyne ready for one last review of yours (tests green, left the redundant presence validation rule on and fixed the tests)

PR #1792, opened 2022-01-25
Do not register the zips dir OKComputer check on 01 boxes

mjgiarlo:
Because they do not mount that directory, which leads to a required status check failing when actually all is well.

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -120,8 +120,14 @@ end
 workflows_url = "#{Settings.workflow_services_url}/objects/druid:oo000oo0000/workflows"
 OkComputer::Registry.register 'external-workflow-services-url', OkComputer::HttpCheck.new(workflows_url)
 
-# Replication (only) uses zip_storage directory to build the zips to send to zip endpoints
-OkComputer::Registry.register 'feature-zip_storage_dir', OkComputer::DirectoryCheck.new(Settings.zip_storage)
+# For each deployed environment (qa, stage, prod), the host that runs the resque
+# dashboard by convention does not mount the zip-transfers directory, so this
+# check will always fail on those hosts. Instead of failing a check on these
+# hosts, only register the check on hosts that do not run the resque dashboard.
+unless Settings.resque_dashboard_hostnames.include?(Socket.gethostname)
+  # Replication (only) uses zip_storage directory to build the zips to send to zip endpoints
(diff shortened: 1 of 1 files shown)


config/initializers/okcomputer.rb
```diff
@@ -121,7 +121,9 @@ def check
 OkComputer::Registry.register 'external-workflow-services-url', OkComputer::HttpCheck.new(workflows_url)
 
 # Replication (only) uses zip_storage directory to build the zips to send to zip endpoints
-OkComputer::Registry.register 'feature-zip_storage_dir', OkComputer::DirectoryCheck.new(Settings.zip_storage)
+unless Settings.resque_dashboard_hostnames.include?(Socket.gethostname) # the dashboard hosts do not mount the directory
```
ndushay, line 124: this comment is not clarifying things for me.  I wonder if we'd be better off just hard-coding that it's -01 that is the problem child here and explaining why.

ndushay (changes requested):

config/initializers/okcomputer.rb
mjgiarlo: @ndushay I'll replace the one-liner comment with a sentence or two. I'd rather use what's in settings than stuff explicit hostnames in the initializer, personally. This was also what @jmartin-sul was inclined to do.

mjgiarlo: @ndushay Wrote up a little paragraph. Take another look?

config/initializers/okcomputer.rb
```diff
@@ -120,8 +120,14 @@ def check
 workflows_url = "#{Settings.workflow_services_url}/objects/druid:oo000oo0000/workflows"
 OkComputer::Registry.register 'external-workflow-services-url', OkComputer::HttpCheck.new(workflows_url)
 
-# Replication (only) uses zip_storage directory to build the zips to send to zip endpoints
-OkComputer::Registry.register 'feature-zip_storage_dir', OkComputer::DirectoryCheck.new(Settings.zip_storage)
+# For each deployed environment (qa, stage, prod), the host that runs the resque
+# dashboard by convention does not mount the zip-transfers directory, so this
+# check will always fail on those hosts. Instead of failing a check on these
+# hosts, only register the check on hosts that do not run the resque dashboard.
```
mjgiarlo, line 126: @ndushay here ya go

ndushay: much better, thank you.

ndushay (approved):

jmartin-sul (approved):

PR #1797, opened 2022-01-31
fixup okcomputer check for zip temp space

jmartin-sul:
## Why was this change made?

when deploying to stage to test load balancing of pres cat (see e.g. #1795), we were seeing errors like the following in the apache logs:

```
App 658587 output: Error: The application encountered the following error: No matching check (OkComputer::Registry::CheckNotFound)
```

we'll deploy to stage and test

ok computer config


jmartin-sul:
tested on stage, works on both web and non-web hosts

tallenaz (changes requested):
We need to install and use cap-rvm to be able to deploy to the new nodes.

tallenaz (approved):
We'll do a separate PR for cap-rvm.

PR #1799, opened 2022-01-31
Let capistrano report to honeybadger

tallenaz:
## Why was this change made?

Because honeybadger doesn't integrate with capistrano-rvm

This worked for technical-metadata-service

N/A


jmartin-sul (approved):

jmartin-sul:
see also:
* https://github.com/honeybadger-io/honeybadger-ruby/issues/149
* https://github.com/sul-dlss/technical-metadata-service/blob/1befdb3c94bc525e41dca893304dc4478402d7ee/config/deploy.rb#L43-L45

jmartin-sul:
that did the trick
Screen Shot 2022-01-31 at 12 43 01 PM
Screen Shot 2022-01-31 at 12 43 37 PM

PR #1812, opened 2022-02-25
Update dependencies for cocina-models update

justinlittman:


jmartin-sul (approved):

jmartin-sul:
i almost refrained from merging, just in case there was a desire to coordinate merging the cocina-models updates, but i don't think pres cat actually uses cocina-models, so i think this is safe (i believe cocina-models is an indirect dep, because of dor-services-client, and that is only used for pushing events to the DSA log, iirc).

PR #1813, opened 2022-02-25
suggestion for a situation where it'd be good to run integration tests even w/o cross service changes

jmartin-sul:

⚡ ⚠ If this change has cross service impact, ***run [integration test create_preassembly_image_spec.rb](https://github.com/sul-dlss/infrastructure-integration-test) on stage as it tests preservation*** and/or test in stage environment, in addition to specs. ⚡

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,6 +4,7 @@
 
 ## How was this change tested? 🤨
 
-⚡ ⚠ If this change has cross service impact, ***run [integration test create_preassembly_image_spec.rb](https://github.com/sul-dlss/infrastructure-integration-test) on stage as it tests preservation*** and/or test in stage environment, in addition to specs. ⚡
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -4,6 +4,6 @@
 
 ## How was this change tested? 🤨
 
-⚡ ⚠ If this change has cross service impact, ***run [integration test create_preassembly_image_spec.rb](https://github.com/sul-dlss/infrastructure-integration-test) on stage as it tests preservation*** and/or test in stage environment, in addition to specs. ⚡
+⚡ ⚠ If this change has cross service impact, or if it changes [code used in the replication pipeline](https://github.com/sul-dlss/preservation_catalog/blob/main/app/jobs/README.md), ***run [integration test create_preassembly_image_spec.rb](https://github.com/sul-dlss/infrastructure-integration-test) on stage as it tests preservation*** and/or test in stage environment, in addition to specs. ⚡
```
ndushay, line 7: refer to "jobs" and other spots in the code directly, in addition to link?

Also, would like one more blank line after -- my bad for only having 2.   If you want to leave this PR as is, okay to leave it at 2.

ndushay (approved):
approving with comment for your consideration (not sure folks will know what "replication pipeline" is or will click link to find out.

.github/pull_request_template.md
jmartin-sul: ok, i added another blank line, and i fiddled with the text a bit.  i tried to list the most relevant classes, mostly jobs, with a link to the replication overview that encourages a glance if the author is unsure.

initially i kept the examples of what's replication-related in a parenthetical before the testing advice, but it felt like it read a little better as a follow-on explanatory sentence.  other version is in this orphan commit:  https://github.com/sul-dlss/preservation_catalog/blob/5a0a89d25f4f15c2a19bf6dc6fb948acd30e90e2/.github/pull_request_template.md

what this PR has:
Screen Shot 2022-02-25 at 9 25 10 PM


the version i decided not to use:
Screen Shot 2022-02-25 at 9 25 30 PM


tested the links and they work.

merging since you approved and hopefully this addresses the feedback, but wouldn't mind a second look to confirm.

PR #1821, opened 2022-03-09
remove redundant API documentation in the README

### LINKED ISSUE #1751: Reconsider API documentation in README
Prescat already has a fully described API in `openapi.yml` which is openly available at https://sul-dlss.github.io/preservation_catalog/. It might be nice to take any useful bits from the README that are missing from the OpenAPI spec, if there are any, and move them into the spec so we can clean up the README (thinking of brevity, consistency, and reduction of duplication).

peetucket:

Fixes #1751 - remove redundant API documentation in the readme.  I went through and verified it was all documented in openapi, making some small tweaks to openapi to add more descriptions from the readme.

Existing tests


jcoyne (approved):

mjgiarlo (commented):
THANK YOU

PR #1834, opened 2022-03-21
Prune dead resque-pool workers after hot-swapping during deploy

mjgiarlo:

Still occasionally seeing dead workers lingering after deployment. So prune dead workers as part of deployment process.

CI, + a test deploy to stage. Resulted in expected number of workers! (Was previously +1.)

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -58,3 +58,14 @@ task :db_seed do
     end
   end
 end
+
+desc 'Prune dead Resque workers'
+after 'resque:pool:hot_swap', :prune_dead_workers do
+  on roles(:resque) do
+    within release_path do
+      with rails_env: fetch(:rails_env) do
+        execute :rails, 'runner', '"Resque.workers.map(&:prune_dead_workers)"'
+      end
+    end
+  end
+end


justinlittman (approved):

mjgiarlo:
I'd like @jmartin-sul to review before we merge this. (I think we can ignore the coverage ding, once again.)

jmartin-sul (approved):
i'm 👍 on this.  i was initially hesitant, just because the problem has occurred pretty infrequently in the past.  but it did happen twice today, this call seems harmless in the cases where it's not needed, and it should address a bad state that's always confusing when it does occur.  so yeah, :shipit:!

one suggestion: what about a short comment like:
> If a Resque process doesn't stop gracefully, it may leave stale state information in Redis. This task does some garbage collection, checking the current Redis state info against the actual environment, and removing entries from Redis for any workers that aren't actually running.  See Resque::Worker#prune_dead_workers

maybe overkill, feel free to merge without if you think the comment is unnecessary (and/or to re-word if you like the idea but think the text could be better).

jmartin-sul:
@mjgiarlo i went ahead and merged, because i couldn't remember if you'd be allowed to merge as the author with the complaint about the (totally acceptable) coverage drop.  i'll just throw up a quick follow on PR with the comment, and you can approve or reject as you see fit.

mjgiarlo:
@jmartin-sul thanks!

PR #1878, opened 2022-05-23
Downgrade ruby-rails orb to 2.0

jcoyne:
So that the validate-api step will pass

ci


jcoyne:
Superceded by #1879

PR #1907, opened 2022-07-07
Make job locking explicit

### LINKED ISSUE #1899: [refactor] Extract job locking code to a separate subclass
Currently the locking applies to all children of ApplicationJob.  In the future we may add new jobs that do not need this locking behavior, and we may inherit it by default without knowing we are doing so.  This is just about making the locking code explicitly opt-in rather than getting it by default. This will help readers of the code understand that there is some non-default code happening.

jcoyne:

Now every job that is using the locking strategy declares "include UniqueJob".  This means we only have jobs that have opted into this strategy on purpose rather than accidentally inheriting this behavior.

Fixes #1899

Test suite

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -6,8 +6,6 @@ class ApplicationJob < ActiveJob::Base
     ActiveRecord::Base.clear_active_connections!
   end
 
-  include UniqueJob
-
   # Raises if the metadata is incomplete
   # @param [Hash<Symbol => String>] metadata
   # @option metadata [String] :checksum_md5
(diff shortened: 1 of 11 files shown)


ndushay:
Tests failing;   I'm going to suggest @jmartin-sul review this to ensure we only add UniqueJob to those job that need it

jmartin-sul (changes requested):
i suspect the failing test can be fixed by adding an `include UniqueJob` to the definition of `RegularParameterJob` in `spec/jobs/application_job_spec.rb`.

Otherwise, this LGTM!  I'd be in favor of retaining current behavior, where all the current jobs disallow duplicate queue entries.  While I think de-duplication is more important for jobs that write to local or cloud storage (zipmaker and the delivery jobs), I can't think of a benefit to allowing multiple instances of any of the jobs to be enqueued, at present.

And even zipmaker and the delivery jobs are designed to be safe to retry (zipmaker looks for an existing zip file, cloud buckets should be write once).  But file systems typically aren't as good at handling that sort of concurrency as an RDBMS is.

jmartin-sul (approved):

PR #1908, opened 2022-07-08
Decouple the redis useage in the lock from Resque

### LINKED ISSUE #1882: replace job locking via resque for switch to sidekiq
From John in slack on May 11, 2022

>One other thing we'd need to do for sidekiq is port pres cat's job locking, which uses resque's redis methods.  We could either switch to using the redis gem directly (since that's also an indirect dependency of sidekiq anyway).  Or we could try a gem like https://github.com/veeqo/activejob-uniqueness (which at a glance seems maintained, but is just one recommendation from a random blog post found when searching "queue deduplication sidekiq", and the current job locking code has a pretty small footprint anyway).

image

jcoyne:

Fixes #1882

Ran https://github.com/sul-dlss/infrastructure-integration-test/blob/main/spec/features/create_preassembly_image_spec.rb

### DIFF
--- a/config/resque.yml
+++ b/config/resque.yml
@@ -1,1 +1,2 @@
 development: <%= ENV.fetch('REDIS_URL', 'localhost:6379') %>
+test: <%= ENV.fetch('REDIS_URL', 'localhost:6379') %>
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ db/schema_er_diagram.svg.dot
 .railsrc
 
 coverage/*
+spec/examples.txt
 
 config/settings.local.yml
 config/settings/*.local.yml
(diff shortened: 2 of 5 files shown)


ndushay:
I'm requesting that @jmartin-sul review this.

config/resque.yml
```diff
@@ -1 +1,2 @@
 development: <%= ENV.fetch('REDIS_URL', 'localhost:6379') %>
+test: <%= ENV.fetch('REDIS_URL', 'localhost:6379') %>
```
ndushay, line 2: would we need this for prod as well?

ndushay:
this was causing errors on -stage

config/resque.yml
jcoyne: no because config/resque.yml is in shared configs.

jcoyne:
> this was causing errors on -stage

I refactored this so we're not managing the redis connection ourselves and I think things have improved.

jmartin-sul (approved):

PR #1912, opened 2022-07-14
replace expectations on removed dor-services-client event client with expectations on its replacement, Dor::Event::Client

jmartin-sul:

* it seemed like `client` was still referenced even after its definition was removed [here](https://github.com/sul-dlss/preservation_catalog/pull/1909/files#diff-092e0f690b001a6977f6f5b15d9412246f6ce67764588074112ec71dc818d436), so it wasn't clear to me how that spec was passing in Circle.
* and then when i tried to see what would happen for me locally with that branch, i got an error when those tests actually tried to connect to rabbitmq on `localhost` (where i had no rabbitmq instance running).
* it's nice to keep some explicit expectations about what the event service will receive, esp in that event reporter spec.

unit tests (local/CI)

### DIFF
--- a/spec/services/reporters/event_service_reporter_spec.rb
+++ b/spec/services/reporters/event_service_reporter_spec.rb
@@ -12,6 +12,7 @@ RSpec.describe Reporters::EventServiceReporter do
 
   before do
     allow(Socket).to receive(:gethostname).and_return('fakehost')
+    allow(Dor::Event::Client).to receive(:create)
   end
 
   describe '#report_errors' do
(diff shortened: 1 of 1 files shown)


jcoyne (approved):

jcoyne:
@jmartin-sul fabulous! I'm still befuddled by the fact that CI doesn't actually seem to care if the tests work or not.

jmartin-sul:
> @jmartin-sul fabulous! I'm still befuddled by the fact that CI doesn't actually seem to care if the tests work or not.

super weird.  i looked at the CircleCI output and it definitely seemed to be running the tests.

PR #1925, opened 2022-08-02
Upgrade to Rails 7

jcoyne:

Ran `create_preassembly_image_spec.rb` on QA.


app/services/audit/checksum.rb
```diff
@@ -34,6 +34,8 @@ def self.validate_list_of_druids(druid_list_file_path)
 
     # validate objects with a particular status on a particular moab_storage_root
     def self.validate_status_root(status, storage_root_name)
+      raise ArgumentError, "invalid status #{status}" unless CompleteMoab.statuses.key?(status)
```
justinlittman, line 37: AR used to object to querying on an invalid enum value. It no longer does, so explicitly checking and raising.

Gemfile
```diff
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+gem 'rails', '~> 7.0.0'
+
```
mjgiarlo, line 4: I have a slight preference for keeping the gems in alpha order but would not block over this. Unless you found that declaring this dependency above the others was required for some reason. I wouldn't be too surprised...

Gemfile.lock
```diff
@@ -403,11 +407,10 @@ GEM
     websocket-extensions (0.1.5)
     whenever (1.0.0)
       chronic (>= 0.6.3)
-    zeitwerk (2.6.1)
+    zeitwerk (2.6.2)
 
 PLATFORMS
-  ruby
```
mjgiarlo, line 409: I'm not sure if we need this platform or not. Do you recall? If not, what would you think about deploying this branch to QA and running an integration test to confirm we don't need the `ruby` platform in deployed envs?

app/services/audit/checksum.rb
mjgiarlo, line 37: Is this part of the Rails 7 upgrade or extra goodness?

spec/services/audit/checksum_spec.rb
```diff
@@ -91,8 +91,8 @@
       end
     end
 
-    it 'with invalid status, raises StatementInvalid' do
-      expect { described_class.validate_status_root('foo', root_name) }.to raise_error(ActiveRecord::StatementInvalid)
```
mjgiarlo, line 95: Ah, I'm guessing this behavior changed between Rails 6 and 7?

Gemfile
justinlittman: That's where Jcoyne put it. Let me try moving and see what happens.

mjgiarlo (approved):
I have a few minor questions and comments. Feel free to reply & edit or just merge. 😄

Gemfile.lock
justinlittman: Oh wait, I did and it worked.

mjgiarlo: OK, great. 🎉

Gemfile
mjgiarlo: kk, appreciate it

app/services/audit/checksum.rb
mjgiarlo: 🥇

justinlittman: Hmmm. I thought I put a comment about this.

Rails used to validate the values of enums during a query. It no longer does.

There was a test for this, so above change replicates the enum check. I doubt it is crucial, but there was a test.

mjgiarlo: 🚀

PR #1957, opened 2022-10-18
Add unique index to ZippedMoabVersions.

### LINKED ISSUE #1831: add uniqueness constraint to zipped_moab_versions table on preserved_object_id / zip_endpoint_id / version
a `ZippedMoabVersion` represents the collection of one or more `ZipPart`s for one version of one Moab on one cloud archive endpoint.  As such, there should only be one `zipped_moab_versions` row for any combination of `preserved_object_id`/`zip_endpoint_id`/`version` in the `zipped_moab_versions` table.

thankfully, no records seemed to violate this constraint yet when i queried for problematic records on production just now:

```ruby
[8] pry(main)> ZippedMoabVersion.group(:preserved_object_id, :zip_endpoint_id, :version).having('count(id) > 1').pluck(:preserved_object_id)
=> []
```

i think this would only be a problem at present if two processes simultaneously called `PreservedObject#create_zipped_moab_versions!` simultaneously on the same druid, which i doubt is a possibility in the system at present, and it's unlikely that two humans would be remediating the same druid at the same time, so it makes sense that we haven't created bad data like this.

still, it's a cheap safeguard to add, and seems like the right thing in principle.

justinlittman:
closes #1831

Safety.

Verified that no existing zipped moab versions violate this constraint in all environments.

### DIFF
--- a/db/migrate/20221018143200_add_unique_zipped_moab_versions.rb
+++ b/db/migrate/20221018143200_add_unique_zipped_moab_versions.rb
@@ -1,0 +1,5 @@
+class AddUniqueZippedMoabVersions < ActiveRecord::Migration[6.1]
+  def change
+    add_index :zipped_moab_versions, [:preserved_object_id, :zip_endpoint_id, :version], unique: true, name: 'index_unique_on_zipped_moab_versions'
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_10_07_185705) do
+ActiveRecord::Schema.define(version: 2022_10_18_143200) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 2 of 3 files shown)


edsu (approved):

edsu (changes requested):
Do you have any idea why the tests are failing here?

ndushay (approved):

edsu (approved):

ndushay:
Does this need to be deployed, with a full backup, separately from dep updates?

justinlittman:
I don't think that is necessitated by this change. The worst that could happen is creating the index fails.

jmartin-sul:
just in case the migration takes a couple minutes to run, deploying this change now, before dep updates go out (but not doing the snapshot first, since this has an exceedingly low risk of introducing bad data):

- [x] tag de76a0a20cfe6a931dacdefb17e856ff4390565d (includes daily CV/C2A audits for evening load)
  - https://github.com/sul-dlss/preservation_catalog/compare/rel-2022-10-17...zmv-unique-per-druid-endpoint-version
- [x] turn off google books
- [x] stop pres robots
- [x] stop pres cat workers
- [x] query pres cat DB to make sure no contraint violating rows (`ZippedMoabVersion.group(:preserved_object_id, :zip_endpoint_id, :version).having('count(id) > 1').pluck(:preserved_object_id, :zip_endpoint_id, :version)`)
  - returns `[]` as expected
- [ ] deploy tag
- [ ] start pres cat workers
- [ ] start pres robots
- [ ] turn on google books

jmartin-sul:
done, took about 40 seconds for the migration to run:
```
00:32 deploy:migrating
      01 bundle exec rake db:migrate
      01 == 20221018143200 AddUniqueZippedMoabVersions: migrating ======================
      01 -- add_index(:zipped_moab_versions, [:preserved_object_id, :zip_endpoint_id, :version], {:unique=>true, :name=>"ind…
      01    -> 26.5345s
      01 == 20221018143200 AddUniqueZippedMoabVersions: migrated (26.5347s) ============
      01
    ✔ 01 pres@preservation-catalog-prod-02.stanford.edu 39.175s
```

PR #1958, opened 2022-10-19
Refactors checksum validator.

justinlittman:
More maintainable?

Unit


mjgiarlo (approved):
LGTM! Would be good for either @jmartin-sul or @ndushay to take a gander too

app/services/checksum_validator.rb
```diff
@@ -2,86 +2,81 @@
 
 # code for validating Moab checksums
 class ChecksumValidator
-  attr_reader :bare_druid, :results, :complete_moab
+  attr_reader :druid, :results, :complete_moab
```
ndushay, line 5: bare_druid vs. druid:  in some of our apps "druid" is the full druid.  bare_druid makes it clear that in this app we do not want the prefix.  I am fine with the will of the people on this decision.

app/services/checksum_validator.rb
```diff
@@ -2,86 +2,81 @@
...
+    @results = AuditResults.new(druid, nil, moab_storage_root, 'validate_checksums', logger: Audit::Checksum.logger)
     # TODO: fix fragile interdependence, MoabValidator wants AuditResults instance, but we want MoabValidator#moab.current_version_id
     # in that AuditResults instance.  so set AuditResults#actual_version after both instances have been created.
     @results.actual_version = moab.current_version_id
   end
 
   def validate_checksums
     # check first thing to make sure the moab is present on disk, otherwise weird errors later
-    return results.result_array unless check_file_existence!
+    return persist! { moab_validator.mark_moab_not_found } unless moab_exists?
```
ndushay, line 22: in the validate_checksums method, which still has a high cognitive load, would it be clearer as "if not" rather than "unless" ?   Or perhaps have a method named for the opposite case (ew)  `moab_absent?`

app/services/checksum_validator.rb
```diff
@@ -2,86 +2,81 @@
...
 
+    # These will populate the results object
     validate_manifest_inventories
     validate_signature_catalog
 
-    # This is to deal with db connection timeouts.
-    ActiveRecord::Base.clear_active_connections!
-
-    transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
+    persist!(clear_connections: true) do
```
ndushay, line 28: - "persist!" as a method concept is great;  the notion that it's about database transaction is less obvious, tho.  Maybe that's ok?  Or maybe `persist_db_transaction!` ?   I'm open to other opinions.

app/services/checksum_validator.rb
```diff
@@ -2,86 +2,81 @@
...
       end
-      complete_moab.save!
     end
-    results.remove_db_updated_results unless transaction_ok
+  end
 
-    results.report_results
+  # @return [Moab::StorageObjectVersion]
+  def latest_moab_version
+    @latest_moab_version ||= moab.version_list.last
```
ndushay, line 43: presumably there is a test, or already was a test, for when moab.version_list is ... whatever it initializes to ...   Also ... is this a "complete moab"?

Or perhaps you're starting to understand why we started abbreviating ...

app/services/checksum_validator.rb
```diff
@@ -92,141 +87,202 @@ def moab_validator
...
+        )
+      end
+    end
   end
 
-  def paths_from_signature_catalog
-    @paths_from_signature_catalog ||= latest_signature_catalog_entries.map { |entry| signature_catalog_entry_path(entry) }
-  end
+  # Validates files on disk against the signature catalog
+  class SignatureCatalogValidator
```
ndushay, line 186: I think this class is a great idea - but I would like it to be in a separate file with its own tests.   Perhaps that is "phase 2" of this specific refactoring and will be in a separate PR?

app/services/checksum_validator.rb
```diff
@@ -92,141 +87,202 @@ def moab_validator
...
-  def validate_manifest_inventory(moab_version)
-    manifest_file_path = "#{moab_version.version_pathname}/#{MANIFESTS}/#{MANIFESTS_XML}"
-    parse_verification_subentities(moab_version.verify_manifest_inventory)
-  rescue Nokogiri::XML::SyntaxError
-    results.add_result(AuditResults::INVALID_MANIFEST, manifest_file_path: manifest_file_path)
-  rescue Errno::ENOENT
-    results.add_result(AuditResults::MANIFEST_NOT_IN_MOAB, manifest_file_path: manifest_file_path)
-  end
+  # Validates files on disk against the manifest inventory
+  class ManifestInventoryValidator
```
ndushay, line 103: I think this class is a great idea - but I would like it to be in a separate file with its own tests. Perhaps that is "phase 2" of this specific refactoring and will be in a separate PR?

app/services/checksum_validator.rb
```diff
@@ -92,141 +87,202 @@ def moab_validator
...
 
-  def validate_signature_catalog_listing
-    latest_signature_catalog_entries.each { |entry| validate_signature_catalog_entry(entry) }
-  rescue Errno::ENOENT
-    results.add_result(AuditResults::SIGNATURE_CATALOG_NOT_IN_MOAB, signature_catalog_path: latest_signature_catalog_path)
-  rescue Nokogiri::XML::SyntaxError
-    results.add_result(AuditResults::INVALID_MANIFEST, manifest_file_path: latest_signature_catalog_path)
-  end
+  def persist!(clear_connections: false)
+    # This is to deal with db connection timeouts.
```
ndushay, line 91: method comments before the declaration, please.

app/services/checksum_validator.rb
```diff
@@ -2,86 +2,81 @@
...
+    moab_validator.update_status(status)
+  end
+
+  # Moab and complete moab versions match?
+  def versions_match?
+    moab.current_version_id == complete_moab.version
+  end
+
+  def validate_versions
+    # set_status_as_seen_on_disk will update results and complete_moab
```
ndushay, line 64: method comments before the declaration, please.

app/services/checksum_validator.rb
```diff
@@ -92,141 +87,202 @@ def moab_validator
...
     end
-  end
 
-  def validate_signature_catalog_entry(entry)
-    unless entry.signature.eql?(calculated_signature(signature_catalog_entry_path(entry)))
-      mismatch_error_data = { file_path: signature_catalog_entry_path(entry), version: entry.version_id }
-      results.add_result(AuditResults::MOAB_FILE_CHECKSUM_MISMATCH, mismatch_error_data)
+    # @param [Moab::VerificationResult] subentity
+    def parse_verification_subentity(subentity)
+      add_cv_result_for_modified_xml(subentity) if subentity.subsets[MODIFIED]
```
ndushay, line 149: and here the need for abbreviations definitely rears its head.

spec/services/checksum_validator_spec.rb
```diff
@@ -317,42 +317,41 @@
...
 
   describe '#validate_manifest_inventories' do
     it 'calls validate_manifest_inventory for each moab_version' do
-      sov1 = instance_double(Moab::StorageObjectVersion)
-      sov2 = instance_double(Moab::StorageObjectVersion)
-      sov3 = instance_double(Moab::StorageObjectVersion)
-      version_list = [sov1, sov2, sov3]
-      moab = instance_double(Moab::StorageObject, version_list: [sov1, sov2, sov3])
-      allow(cv).to receive(:moab).and_return(moab)
+      storage_object_version1 = instance_double(Moab::StorageObjectVersion)
```
ndushay, line 344: I think `_v1` is sufficiently clear, myself.  But don't care much either way.

ndushay:
You should run pre-assembly integration test on stage for the branch as it does the preservation stuff.

jmartin-sul:
> You should run pre-assembly integration test on stage for the branch as it does the preservation stuff.

i don't think that'll exercise the changes here, as it won't trigger this audit job -- iirc, the `preservationIngestWF` validation happens entirely via `ValidateMoabJob`.  this job only gets triggered by, i think, regularly scheduled audits, manually triggered audits, and detection of a Moab that pres cat wasn't notified about (e.g. the occasional situation where a pres robots `update-catalog` REST call is dropped, and `MoabToCatalog` detects the new/updated Moab).

to test this beyond the unit tests in the app, i'd call `Audit::Checksum.validate_druid(druid)` on some stage object with more than one version.  if you wanted to be really thorough to make sure error detection still worked, you could manually move or corrupt a file in a stage moab and do the same, and confirm that the error is detected (but prob good to give @andrewjbtw a heads up if you do that).

jmartin-sul:
which is a good reminder -- `ValidateMoabJob` was written with an eye towards expediency, and some refactoring to consolidate shared code between `ValidateMoabJob` and `ChecksumValidator` might be a good maintenance ticket, in terms of making things more approachable/maintainable.

app/services/checksum_validator.rb
justinlittman: I agree this is a big mess. However, the attribute on PreservedObject is `druid` so this keeps it consistent. In fact, I'm going to change this to a `delegate`.

app/services/checksum_validator.rb
justinlittman: Since this is only used in one place, I changed to `moab_absent?`.

app/services/checksum_validator.rb
justinlittman: Changed to `persist_db_transaction!`.

app/services/checksum_validator.rb
justinlittman: This is a `Moab::StorageObjectVersion`. Renaming to make clearer.

app/services/checksum_validator.rb
justinlittman: This is a comment about the next line, not about the method.

app/services/checksum_validator.rb
justinlittman: This is a comment about the next line, not the method.

app/services/checksum_validator.rb
justinlittman: Removing `_cv_`. It doesn't add anything here (and I think should be `AuditResult` anyway).

spec/services/checksum_validator_spec.rb
justinlittman: In most contexts I would agree. However, this is a `storage_object_version` so using the full name.

app/services/checksum_validator.rb
ndushay: I dunno if this class might be used in other places?

justinlittman: Me neither. If we find it is useful, it can be pulled into a separate class and decoupled from ChecksumValidator.

app/services/checksum_validator.rb
ndushay: again, I don't know off the top of my head if ManifestInventoryValidator might be used elsewhere.   i suspect we test to see if there is a manifestInventory.xml file in some places, but not sure if that should live in this service.

app/services/checksum_validator.rb
```diff
@@ -2,86 +2,80 @@
...
+    moab_validator.update_status(status)
+  end
 
-    false
+  # Moab and complete moab versions match?
+  def versions_match?
+    moab.current_version_id == complete_moab.version
+  end
+
+  def validate_versions
```
ndushay, line 62: would it be correct/useful to name this `validate_versions_match` ?

ndushay (approved):
approving;  one more comment for your consideration.

app/services/checksum_validator.rb
ndushay: would it be worth making a ticket to find out?

PR #1966, opened 2022-10-20
Refactor AuditResults.

### LINKED ISSUE #1965: refactor AuditResults class
From  [maint/tech debt storytime discussion Wed 10/19](https://docs.google.com/document/d/1vTv9ySljSgJ-zcrH2ZztDpdUtnVCxe_hgfZnIehkPGs/edit#heading=h.a6kqlslw7c5o):

This is one of our 5 biggest classes:

image

John already has some ideas for refactors / rethinking of how this is done, but feel free to branch out on your own ...

justinlittman:
closes #1965

Clearer code? Single responsibility?


app/models/audit_results.rb
```diff
@@ -1,17 +1,6 @@
...
-#    - in the future, this may also include reporting to the provenance database as well
-#
-# All results are kept in the result_array attribute, which is returned by the report_results method.
-#   result_array = [result1, result2]
-#   result1 = {response_code => msg}
-#   result2 = {response_code => msg}
+# AuditResults allows the correct granularity of auditing information to be modeled in various contexts.
 class AuditResults
   ACTUAL_VERS_GT_DB_OBJ = :actual_vers_gt_db_obj
   ACTUAL_VERS_LT_DB_OBJ = :actual_vers_lt_db_obj
```
mjgiarlo, line 6: Is `app/models/` the best place for this new version of the class? I don't know. But I think it's a better fit than `app/services/` and I'm not sure inventing a new `app/` subdirectory is better than stuffing this PORO into `app/models/` so :shipit: ?

app/models/audit_results.rb
```diff
@@ -92,16 +81,14 @@ class AuditResults
     CM_STATUS_CHANGED
   ].freeze
 
-  attr_reader :result_array, :druid, :storage_area, :logger
-  attr_accessor :actual_version, :check_name
+  attr_reader :druid, :storage_area, :actual_version, :check_name
 
-  def initialize(druid, actual_version, storage_area, check_name = nil, logger: nil)
+  def initialize(druid:, actual_version:, storage_area:, check_name: nil)
```
mjgiarlo, line 86: Appreciate moving to keyword args over positional ones!

app/models/audit_results.rb
```diff
@@ -113,31 +100,22 @@ def remove_db_updated_results
...
+    result_array.reject(&result_ok?).freeze
   end
 
   def contains_result_code?(code)
     result_array.detect { |result_hash| result_hash.key?(code) } != nil
   end
 
-  def status_changed_to_ok?(result)
-    /to ok$/.match(result[AuditResults::CM_STATUS_CHANGED]) != nil
-  end
```
mjgiarlo, line 139: Did this method move somewhere or is it just nuked? Asking because I think it's still invoked elsewhere in the `AuditResults` model.

app/models/audit_results.rb
```diff
@@ -148,33 +126,12 @@ def to_json(*_args)
 
   private
 
+  attr_reader :result_array
```
mjgiarlo, line 129: Appreciate being intentional about what's in the public API and what's an implementation detail.

mjgiarlo (commented):
Looks better to me.

app/models/audit_results.rb
ndushay: I think this is how stuff ended up in `app/lib` -- because it wasn't a rails model, which is all I expect to find in the `app/models` directory.   I'm not sure what's best;  Putting it in the services junk drawer feels more comfortable to me.  :shrug:   I would defer to John.

app/models/audit_results.rb
```diff
@@ -1,17 +1,6 @@
...
-#   enable an easy way to provide:
-#    - the correct HTTP response code for ReST calls routed via controller
-#    - error information reported to workflows
-#    - information to Rails log
-#    - in the future, this may also include reporting to the provenance database as well
-#
-# All results are kept in the result_array attribute, which is returned by the report_results method.
-#   result_array = [result1, result2]
-#   result1 = {response_code => msg}
-#   result2 = {response_code => msg}
```
ndushay, line 14: I'm not in love with these comments, but they are informative.  What is the thinking behind removing them ... and not adding this info anywhere else (e.g. in the "beginners guide" that doesn't exists yet ... but there is a ticket that could get this info pasted as a comment)

app/models/audit_results.rb
ndushay: Maybe I'm being slow, but ... who are the callers and are these changes made there?

app/services/audit_results_reporter.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# Reports audit results to the Rails log, Honeybadger, Workflow Service, and Event Service.
+class AuditResultsReporter
```
ndushay, line 4: Here, I like breaking this out ... but isn't it also a service that is only used by one class?  That is the argument used for NOT breaking a service into a separate class in a different PR.   

Can you say why this is a different case, or something?  I am a bit baffled.

app/models/audit_results.rb
mjgiarlo: @ndushay That'd work for me too. 😄

app/models/audit_results.rb
mjgiarlo: You're not being slow; this isn't done yet (AFAICT). I don't think this PR has a passing test suite yet. My vague recollection is J-Litt wanted some early feedback on decisions made so far before proceeding further.

app/models/audit_results.rb
mjgiarlo: Good call.

app/models/audit_results.rb
justinlittman: In our other apps, we have non-Rails models in `app/models/`. At least based on my familiarity with our other apps, that is where I would have looked for it.

app/models/audit_results.rb
justinlittman: Correct. I didn't want to change all of the callers and tests until there was consensus on the refactor/split.

app/models/audit_results.rb
```diff
@@ -113,31 +100,22 @@ def remove_db_updated_results
...
   end
 
   def completed_results
-    result_array.select(&result_ok?)
+    result_array.select(&result_ok?).freeze
   end
 
   def error_results
-    result_array.reject(&result_ok?)
+    result_array.reject(&result_ok?).freeze
```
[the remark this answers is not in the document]
justinlittman: Yes, the previous implementation allowed other code to manipulate the result array. The errs on the other side, and makes sure that other code knows it cannot manipulate the results.

app/models/audit_results.rb
justinlittman: Ooops. You're right.

app/services/audit_results_reporter.rb
justinlittman: 1. This class isn't used by only one class. (It isn't invoked in `AuditResult` at all.)
2. It will have completely separate tests from `AuditResult`.

app/models/audit_results.rb
justinlittman: I put some of this back in.

app/models/audit_results.rb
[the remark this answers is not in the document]
mjgiarlo: Excellent.

app/models/audit_results.rb
mjgiarlo: I think you've convinced me `app/models/` is fine here.

app/models/audit_results.rb
```diff
@@ -113,66 +105,36 @@ def remove_db_updated_results
...
+  attr_reader :result_array
 
-  def report_completed(results)
-    reporters.each do |reporter|
-      results.each do |result|
-        reporter.report_completed(druid: druid, version: actual_version, storage_area: storage_area, check_name: check_name, result: result)
-      end
-    end
+  def status_changed_to_ok?(result)
+    /to ok$/.match(result[AuditResults::CM_STATUS_CHANGED]).present?
```
mjgiarlo, line 137: If we're merely checking for the presence of a match, without caring about the value of the match, I think this method could be:

```suggestion
  def status_changed_to_ok?(result)
    /to ok$/.match?(result[AuditResults::CM_STATUS_CHANGED])
```

mjgiarlo (approved):
Approve pending consideration of one minor tweak.

app/models/audit_results.rb
justinlittman: True, but I would rather the method return a boolean instead of a Matcher.

mjgiarlo: @justinlittman `#match?` returns a boolean. (`#match` doesn't.)

justinlittman: Super. I'll make that change.

PR #1972, opened 2022-10-20
Refactor zip delivery

### LINKED ISSUE #1922: [refactor for clarity] make a generic S3 delivery class, and have all 3 endpoint delivery classes inherit from that
Currently, `S3WestDeliveryJob` is both the delivery class for the AWS S3 West archive endpoint, and the base class from which the S3 East and IBM US South delivery classes inherit most of their logic.  See:
https://github.com/sul-dlss/preservation_catalog/blob/53b961df0761fb8b89368952dd82bc9cd35c6c69/app/jobs/s3_west_delivery_job.rb
https://github.com/sul-dlss/preservation_catalog/blob/53b961df0761fb8b89368952dd82bc9cd35c6c69/app/jobs/s3_east_delivery_job.rb
https://github.com/sul-dlss/preservation_catalog/blob/53b961df0761fb8b89368952dd82bc9cd35c6c69/app/jobs/ibm_south_delivery_job.rb

`#bucket` is the only method the children override at present.

The current inheritance structure makes it slightly confusing to read Honeybadger error alerts, as occasional delivery failures (e.g. due to network errors) almost always come from the `#perform` method, leading to HB alerts that point at first glance to `S3WestDeliveryJob#perform` as the source of the error, even if the error occurred when a subclass was trying to deliver to an endpoint other than AWS West.  If you go to the HB alert's ["Params" section](https://app.honeybadger.io/projects/54415/faults/64624826#notice-params), you can see the actual endpoint on which the failure occurred via the `job_class` field.  The Resque failure queues also capture the failures in the appropriate endpoint-specific failure queue.  But this isn't always obvious at first glance for first responders and ops folks in a hurry, or people who aren't familiar with this slightly confusing part of the pres cat codebase.

mjgiarlo:

Fixes #1922

This commit refactors zip delivery in the following ways:

* Move zip delivery into its own small service class
* Invoke zip delivery service from a new abstract delivery job
* Base all three delivery jobs on the abstract delivery job

CI

### DIFF
--- a/app/jobs/ibm_south_delivery_job.rb
+++ b/app/jobs/ibm_south_delivery_job.rb
@@ -1,8 +1,7 @@
 # frozen_string_literal: true
 
-# Different provider than parent class
-# @note The IBM delivery job inherits from an AWS job because the mechanics only differ in bucket name lookup
-class IbmSouthDeliveryJob < S3WestDeliveryJob
+# @see PreservationCatalog::Aws for how S3 credentials and bucket are configured
+class IbmSouthDeliveryJob < AbstractDeliveryJob
   queue_as :ibm_us_south_delivery
 
   def bucket
--- a/app/jobs/s3_east_delivery_job.rb
+++ b/app/jobs/s3_east_delivery_job.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-# Same as parent class, just a different queue.
+# @see PreservationCatalog::Aws for how S3 credentials and bucket are configured
 # @note This name is slightly misleading, as this class solely deals with AWS US East 1 endpoint
-class S3EastDeliveryJob < S3WestDeliveryJob
+class S3EastDeliveryJob < AbstractDeliveryJob
   queue_as :s3_us_east_1_delivery
 
   def bucket
--- a/app/jobs/s3_west_delivery_job.rb
+++ b/app/jobs/s3_west_delivery_job.rb
@@ -1,31 +1,9 @@
 # frozen_string_literal: true
 
-# Posts zips to S3, if needed.
-# Notify ResultsRecorderJob, if posted.
 # @see S3::Aws for how S3 credentials and bucket are configured
 # @note This name is slightly misleading, as this class solely deals with AWS US West 2 endpoint
-class S3WestDeliveryJob < ZipPartJobBase
+class S3WestDeliveryJob < AbstractDeliveryJob
   queue_as :s3_us_west_2_delivery
-  # NOTE: base class gives us `zip`, `dvz_part` (but not an instance of zip with the path to the moab used for creation)
-
-  before_enqueue { |job| job.zip_info_check!(job.arguments.fourth) }
-
-  # @param [String] druid
-  # @param [Integer] version
-  # @param [String] part_s3_key
-  # @param [Hash<Symbol => String, Integer>] metadata Zip info
-  # @see PlexerJob#perform warning about why metadata must be passed
-  def perform(druid, version, part_s3_key, metadata)
-    s3_part = bucket.object(part_s3_key) # ::Aws::S3::Object
-    return if s3_part.exists?
-
-    fresh_md5 = dvz_part.read_md5
-    given_md5 = metadata[:checksum_md5]
-    raise "#{part_s3_key} MD5 mismatch: passed #{given_md5}, computed #{fresh_md5}" unless fresh_md5 == given_md5
-
-    s3_part.upload_file(dvz_part.file_path, metadata: stringify_values(metadata))
-    ResultsRecorderJob.perform_later(druid, version, part_s3_key, self.class.to_s)
-  end
 
   def bucket
     S3::AwsProvider.new(
@@ -33,12 +11,5 @@ class S3WestDeliveryJob < ZipPartJobBase
       access_key_id: Settings.zip_endpoints.aws_s3_west_2.access_key_id,
       secret_access_key: Settings.zip_endpoints.aws_s3_west_2.secret_access_key
     ).bucket
-  end
-
-  # coerce size int to string (all values must be strings)
-  # @param [Hash<Symbol => #to_s>] metadata
-  # @return [Hash<Symbol => String>] metadata
-  def stringify_values(metadata)
-    metadata.merge(size: metadata[:size].to_s, parts_count: metadata[:parts_count].to_s)
   end
 end
--- a/app/services/zip_delivery_service.rb
+++ b/app/services/zip_delivery_service.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+# Posts zips to S3, if needed.
+class ZipDeliveryService
+  def self.deliver(...)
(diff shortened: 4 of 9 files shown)


mjgiarlo:
**NOTE**: ❌ is caused by a single uncovered line: `AbstractDeliveryJob#bucket`. This method raises an error and should never be invoked as this class has zero children that fail to override this method. I could add a new spec for this class that tests only this condition but I'm on the fence about whether that's useful or not.

justinlittman (approved):

app/jobs/s3_east_delivery_job.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
-# Same as parent class, just a different queue.
+# @see PreservationCatalog::Aws for how S3 credentials and bucket are configured
 # @note This name is slightly misleading, as this class solely deals with AWS US East 1 endpoint
-class S3EastDeliveryJob < S3WestDeliveryJob
+class S3EastDeliveryJob < AbstractDeliveryJob
```
ndushay, line 5: How about S3AWSEastDeliveryJob?  Or even `AWSEastDeliveryJob` to make it more like IBM?

app/services/zip_delivery_service.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+# Posts zips to S3, if needed.
+class ZipDeliveryService
```
ndushay, line 4: if this class doesn't warrant its own spec file, would it make sense to 
- put class in abstract_delivery_job.rb file?
- decompose it into methods in abstract_delivery_job?

ndushay (changes requested):
I added some comments for consideration.

I'm not sure how this relates to #1973 in terms of merge order.

app/jobs/ibm_south_delivery_job.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
-# Different provider than parent class
-# @note The IBM delivery job inherits from an AWS job because the mechanics only differ in bucket name lookup
-class IbmSouthDeliveryJob < S3WestDeliveryJob
+# @see PreservationCatalog::Aws for how S3 credentials and bucket are configured
```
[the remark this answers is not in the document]
mjgiarlo: I think `git` will help us out and I'll deal with any merge conflicts when they arise.

app/jobs/s3_west_delivery_job.rb
```diff
@@ -1,31 +1,11 @@
 # frozen_string_literal: true
 
-# Posts zips to S3, if needed.
-# Notify ResultsRecorderJob, if posted.
 # @see PreservationCatalog::Aws for how S3 credentials and bucket are configured
 # @note This name is slightly misleading, as this class solely deals with AWS US West 2 endpoint
-class S3WestDeliveryJob < ZipPartJobBase
+class S3WestDeliveryJob < AbstractDeliveryJob
```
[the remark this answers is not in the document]
mjgiarlo: This is straight-up refactoring, so I'd prefer not to rename existing jobs as part of this PR.

app/jobs/s3_east_delivery_job.rb
mjgiarlo: See above.

app/services/zip_delivery_service.rb
mjgiarlo: It probably should have its own spec file broken out of the S3West job's yucky spec, but I didn't care enough to include lots of testing changes in this refactoring.

app/jobs/s3_west_delivery_job.rb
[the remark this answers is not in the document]
ndushay: I disagree.  The existing names are confusing, and the renames will help.   Or is that because it makes it harder to grok the changes in github UI?

(p.s. that's something JCoyne would say)

app/services/zip_delivery_service.rb
ndushay: yeah, I'm uncomfortable with that.

mjgiarlo: Of course, totally happy to write up a ticket. Or even just do the work immediately after.

ndushay:
@jmartin-sul I'm wondering if this does what you hoped with ticket #1922;  also would love you to weigh in on whether the new DeliveryService class should have its own specs, and if so, is a separate ticket fine for that?

ndushay (approved):

jmartin-sul (approved):

jmartin-sul:
> @jmartin-sul I'm wondering if this does what you hoped with ticket #1922;

i think so, yeah.  definitely seems like a readability improvement, especially for quick glances at honeybadger errors when deliveries fail (i suspect we'll be a lot more likely to remember to look for the queue/subclass for which the failure actually occurred if the error message comes from an abstract base class instead of a base class named for a specific endpoint).

> also would love you to weigh in on whether the new DeliveryService class should have its own specs, and if so, is a separate ticket fine for that?

sure.  or... i actually might be more inclined to take what @mjgiarlo moved into the `describe 'integration test'` block in `s3_west_delivery_job_spec.rb`, and make that a delivery job integration test in a separate spec file... and not give the service its own direct spec file?  i do somewhat like that more integration-y test of the interplay between `ZipDeliveryService.deliver` and `ResultsRecorderJob.perform_later` for `AbstractDeliveryJob#perform`.  as opposed to separate specs for the service and the abstract base class.  i don't feel strongly, but i do agree that putting that set of tests in `s3_west_delivery_job_spec.rb` feels a bit odd, and makes less sense outside the historical inheretence relationship this PR moves away from.

i do also like the idea of naming renaming the AWS specific job classes to not have generic S3 names, but i think that's fine for a follow-on ticket or PR.

for ease of review, i appreciated the relatively minimal spec changes (and feels like a good sign that `spec/jobs/integration_spec.rb` needed no adjustment).

i think it'd be good to deploy this branch and run `preasssembly_image_accessioning_spec.rb` from the integration test suite, since that confirms replication works when accessioning something with more than one version.

jmartin-sul:
no objection to merging if integration test succeeds

mjgiarlo:
@jmartin-sul @ndushay OK, sounds good. Here's my plan:

1. Deploy to stage
2. Run the pre-assembly integration test
3. (Iterate if any bugs found)
4. Have another hack at improving the test structure

spec/jobs/abstract_delivery_job_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  # NOTE: Test a concrete impl of the abstract job so the call to `#bucket` does not raise
+  subject(:job_implementation) do
+    Class.new(described_class) do
+      def bucket
+        Struct.new(:unused) do
+          def object(arg); end
+        end.new
+      end
+    end
+  end
```
mjgiarlo, line 15: I moved the testing out of the S3 West job and split the tests into the abstract job and the new service, @jmartin-sul @ndushay

jmartin-sul (approved):

PR #1973, opened 2022-10-20
Improve names and locations of code in app/lib/

### LINKED ISSUE #1968: refactor:  move app/lib/audit classes to app/services
From  [maint/tech debt storytime discussion Wed 10/19](https://docs.google.com/document/d/1vTv9ySljSgJ-zcrH2ZztDpdUtnVCxe_hgfZnIehkPGs/edit#heading=h.a6kqlslw7c5o):

**Goal:**  code in "standard" places makes it easier to find.  In rails, `app/services` is now more standard than `app/lib`

Move the `app/lib/audit` classes to `app/services`.   

- It may still make sense to have an audit folder under `app/services`.  
- It may make sense to rename the classes to end with `Service` 
- don't forget to move the specs to the new appropriate spot, too.

### LINKED ISSUE #1969: refactor:  move app/lib/preservation_catalog to app/services/s3
From  [maint/tech debt storytime discussion Wed 10/19](https://docs.google.com/document/d/1vTv9ySljSgJ-zcrH2ZztDpdUtnVCxe_hgfZnIehkPGs/edit#heading=h.a6kqlslw7c5o):

**Goal:**  code in "standard" places makes it easier to find.  In rails, `app/services` is now more standard than `app/lib`

Move the `app/lib/preservation_catalog` classes to `app/services/s3`.   

- It may still make sense to have a folder under `app/services`, but "s3" is a better name for this than "preservation_catalog"  
- It may make sense to rename the classes to end with `Service` 
- don't forget to move the specs to the new appropriate spot, too.

See also #1922

mjgiarlo:

Fixes #1968
Fixes #1969

This commit relocates all the code that was previously in app/lib/, moving all of it into app/services/, and renames some modules to be more accurate and more descriptive. Why? To more closely apply the principle of least surprise.

CI


mjgiarlo:
I think the coverage ding here is bogus.

app/services/audit.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# @example Usage
+#  Audit::Checksum.validate_druid(druid)
+
+# Holds audit tools
+module Audit
```
justinlittman, line 7: Why does the empty module exist?

mjgiarlo: Nuked the file and all is well.

spec/services/active_record_utils_spec.rb
```diff
@@ -2,9 +2,6 @@
 
 require 'rails_helper'
 
-require 'active_record_utils'
-require 'audit_results'
-
```
mjgiarlo, line 7: This is dead code, vestiges of when this code used to live in `lib/`. Why? Because code in `lib/` is not auto-loaded by Rails apps, and all code in `app/` is.

justinlittman (approved):

PR #1987, opened 2022-10-26
Refactor check existence service.

justinlittman:
More readable?

### DIFF
--- a/app/services/complete_moab_service/base.rb
+++ b/app/services/complete_moab_service/base.rb
@@ -115,5 +115,22 @@ module CompleteMoabService
     def report_results!
       AuditResultsReporter.report_results(audit_results: results)
     end
+
+    def complete_moab_exists?
+      CompleteMoab.by_druid(druid).by_storage_root(moab_storage_root).exists?
+    end
+
+    def validation_errors?
+      moab_validator.moab_validation_errors.present?
+    end
+
+    def record_missing
+      results.add_result(AuditResults::DB_OBJ_DOES_NOT_EXIST, 'CompleteMoab')
+      if validation_errors?
+        create_db_objects('invalid_moab')
+      else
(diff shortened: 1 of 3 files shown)


app/services/complete_moab_service/base.rb
```diff
@@ -115,5 +115,9 @@ def create_db_objects(status, checksums_validated: false)
     def report_results!
       AuditResultsReporter.report_results(audit_results: results)
     end
+
+    def complete_moab_exists?
```
justinlittman, line 119: This is going to be re-used by other services, so putting in base.

app/services/complete_moab_service/check_existence.rb
```diff
@@ -60,5 +33,38 @@ def update_complete_moab_preserved_object_or_set_status
...
+        else # incoming_version < complete_moab.version
+          moab_validator.set_status_as_seen_on_disk(false)
+          results.add_result(AuditResults::ACTUAL_VERS_LT_DB_OBJ, db_obj_name: 'CompleteMoab', db_obj_version: complete_moab.version)
+        end
+        complete_moab.update_audit_timestamps(moab_validator.ran_moab_validation?, true)
+        complete_moab.save!
+      end
+    end
+
+    def report_missing
```
edsu, line 61: I noticed that three of the other subclasses have similar logic around adding AuditResults? Maybe this is another candidate for refactoring as part of another task?

edsu (approved):
I agree it looks more readable, and gets rid of a disabled cop too!

edsu:
Whoops I just noticed this was a draft, lol. Sorry for the premature approve.

app/services/complete_moab_service/check_existence.rb
justinlittman: You're right. Let me get that.

Issue #1240, opened 2019-11-18
figure out how to re-enable zip part audit codes in AuditResults::WORKFLOW_REPORT_CODES


jmartin-sul:
see https://github.com/sul-dlss/preservation_catalog/blob/master/app/services/audit_results.rb#L92-L98

this way, replication errors are surfaced in argo the same way as other errors are.

i wasn't sure whether to count this as a bug, or tech debt, so i counted it as both 😆

jmartin-sul:
https://github.com/sul-dlss/preservation_catalog/issues/1249 might obviate the need for this.

jmartin-sul:
i think we still want this, and iirc, the reason we disabled the codes was because the error messages would exceed some field length or other when workflow service's database was still in oracle (it was later migrated to postgres).

removing [HOLD]

jmartin-sul:
The easiest way to test this is probably to just deploy the potential fix to QA or stage, and in rails console there, run: `PreservedObject.find_each(&:audit_moab_version_replication!)`.

that'll kick off replication for all objects.  you may want to double-check, but as of a few days ago, each had about 39k druids, which is not a terribly huge amount of objects on which to run replication audit.  those environments are a bit of a mess, so the only downside is that those envs are a mess, so a lot of errors will probably be reported.

actually, now that i think about it, we might want to wait for https://github.com/sul-dlss/preservation_catalog/issues/1986 to be finished before doing this.

ok, marking blocked.

jmartin-sul:
this warrants a storytime: do we want to halt accessioning on cloud audit errors?  we should also unpack how this audit reporting relates to DSA: e.g., do workflow errors still wedge accessioning?  should they?

justinlittman:
@jmartin-sul Given that we're planning to deprecate reporting errors to the workflow service, should this be closed?

PR #1993, opened 2022-10-26
Log when total zip part size is less than the moab size

### LINKED ISSUE #1742: check total size of zip parts in C2A
Similar to what's done when `DruidVersionZip` creates a zip:  https://github.com/sul-dlss/preservation_catalog/blob/db92f223fe1e4c43775770aaf1e43e6355653faf/app/models/druid_version_zip.rb#L162-L164

The new check would confirm that the zip parts for a given moab version on a given endpoint total to at least the expected minimum size (i.e. greater than the size of that moab version's directory on disk, since we zip without compression).

I believe the new check would be a helper method called from somewhere in here:  https://github.com/sul-dlss/preservation_catalog/blob/main/app/services/s3/s3_audit.rb
Though of course i'd be open to other ideas.

Since
* the size check on zip creation was added in 2020
* we started sending stuff to S3 in 2018
  * A read of the [pre-#1302](https://github.com/sul-dlss/preservation_catalog/issues/1302) `ZipmakerJob` code implies to me that if a bad zip file were created, as described in #1633, and if the job that created the zip was retried manually before the bad zip file had aged out of `/sdr-transfers`, it seems like we could have delivered that bad archive copy to the cloud.
  * Related: hopefully the implementation of #1302 means that we have nothing like this from after 2020!
* It seems from querying like we do have at least some replication copies that are too small.

This should help us more easily find druid versions that were mis-replicated before the size check was added.  There is a query in the db/README.md that attempts to query across the whole catalog for instances of this problem.  but it's a bit complex, and i suspect 1) that it's not quite accurate and 2) that making it fully accurate would lead to a very expensive query that might take hours to return.  whereas the logic for doing this per-object is more straight-forward, and incorporating it into the existing cloud audit seems like a natural enhancement to make.

see also:
https://github.com/sul-dlss/preservation_catalog/issues/1302
https://github.com/sul-dlss/preservation_catalog/issues/1637
https://github.com/sul-dlss/preservation_catalog/issues/1642
https://github.com/sul-dlss/preservation_catalog/issues/1733

mjgiarlo:

Fixes #1742

This commit adds a new audit result to the catalog-to-archive audit service such that, when the total zip part size for a zipped moab version is less than the size of the moab on disk, this condition is reported. Reporting this information will help us find and remediate replication errors.

As part of this work, as a convenience, add a new `#total_part_size` method to the `ZippedMoabVersion` model.

CI. I'll have to defer to @jmartin-sul for advice on whether this is sufficient, or whether poking about in a deployed environment may be warranted.


app/services/audit/catalog_to_archive.rb
```diff
@@ -15,6 +15,15 @@ def self.check_child_zip_part_attributes(zmv, results)
         return false # everything else relies on checking parts, nothing left to do
       end
 
+      total_part_size = zmv.total_part_size
+      moab_version_size = zmv.preserved_object.complete_moab&.size || 0
```
mjgiarlo, line 19: @jmartin-sul I'm not entirely confident I've got the right sizes here, but maybe I'm not far off?

spec/services/audit/catalog_to_archive_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe Audit::CatalogToArchive do
-  let(:zmv) { create(:zipped_moab_version) }
+  let(:zmv) { create(:zipped_moab_version, preserved_object: create(:preserved_object_fixture, druid: 'bz514sm9647')) }
```
mjgiarlo, line 6: Without this the zipped moab version's preserved object does not have a complete moab.

app/services/audit/catalog_to_archive.rb
jmartin-sul: @mjgiarlo I think we probably want a method like `CompleteMoab#version_size_from_disk(version)`, that'll use the `complete_moab.moab_storage_root.storage_location` and the druid (from the parent `PreservedObject`) to get a druid tree (for which i think you can use `DruidTools::Druid.new(druid, storage_location).path`).  And then this new method would get the size of e.g. the `v0003` dir in the particular Moab.  Sorry, when we were pairing on this yesterday I mentioned thinking that the sizes in the DB could be trusted, but a bit of nuance is that the `CompleteMoab` encompasses all versions, so its size field isn't at the right level of granularity for comparison with one `ZippedMoabVersion`, I don't think. (perhaps replacing `CompleteMoab#size` with a `CompleteMoab#version_sizes` array is worth ticketing and considering, since that might be useful info for querying here and in general)

app/services/audit/catalog_to_archive.rb
```diff
@@ -15,6 +15,15 @@ def self.check_child_zip_part_attributes(zmv, results)
...
+      if total_part_size < moab_version_size
```
jmartin-sul, line 20: And so, if `moab_version_size`'s calculation changes as suggested above, then like the corresponding check when building the zip file from the Moab version, I think we'd want to confirm that `total_part_size > moab_version_size`, instead of less than.  The assumption being that the zips are created without compression (presumably with a bit of additional zip file metadata), and that the (hopefully past 😅) failure mode we're worried about was the silent/uncaught omission of some files when zips were created before that size check was added.

jmartin-sul (changes requested):
whoops, left individual comments, but forgot to do the official "request changes" thing

app/services/audit/catalog_to_archive.rb
mjgiarlo: @jmartin-sul Sounds good, and rings a bell, even. 😄 I'll tackle that next.

app/services/audit/catalog_to_archive.rb
mjgiarlo: @jmartin-sul Maybe I still have this backwards but instead of running with this, I'd like to dig in so I can hopefully get this in my brain better. Here's what I was thinking... let me know where I go off the rails:

* `total_part_size` is the total size of zips on an endpoint for a given druid + version
* `moab_version_size` is the size on disk for a given druid + version (once fixed per your suggestion above)
* the error condition we want to flag now is when we have stuff missing in the cloud
* the way we flag that is by comparing `total_part_size` to `moab_version_size` and if the former is smaller than the latter, then it's likely missing content
* so isn't the conditional for adding a size inconsistency error result when `total_part_size < moab_version_size`?

app/models/druid_version_zip.rb
```diff
@@ -166,6 +166,10 @@ def zip_split_size
     '10g'
   end
 
+  def moab_version_size
+    moab_version_files.sum { |f| File.size(f) }
+  end
```
mjgiarlo, line 171: Since `DruidVersionZip` already has a method to get the size on a disk of a given druid + version + storage location, I decided to use that method instead of writing a new one. And as part of that, I moved this method into the public API (and added a test below).

app/models/preserved_object.rb
```diff
@@ -61,6 +61,12 @@ def audit_moab_version_replication!
     MoabReplicationAuditJob.perform_later(self)
   end
 
+  def total_size_of_moab_version(version)
+    return 0 unless complete_moab
+
+    DruidVersionZip.new(druid, version, complete_moab.moab_storage_root.storage_location).moab_version_size
+  end
```
mjgiarlo, line 68: OK to have this responsibility defined on the PreservedObject?

spec/models/druid_version_zip_spec.rb
```diff
@@ -101,7 +101,7 @@
     after { FileUtils.rm_rf('/tmp/bj') } # cleanup
 
     context 'when zip size is less than the moab size' do
-      let(:moab_version_size) { dvz.send(:moab_version_size) }
+      let(:moab_version_size) { dvz.moab_version_size }
```
mjgiarlo, line 104: This method moved from private to public.

spec/models/preserved_object_spec.rb
```diff
@@ -199,4 +199,39 @@
...
+    let(:current_version) { 3 }
+
+    context 'when complete moab is nil' do
+      it 'returns 0' do
+        expect(preserved_object.total_size_of_moab_version(current_version)).to eq(0)
+      end
+    end
+
+    context 'when complete moab exists' do
+      let!(:cm1) { create(:complete_moab, preserved_object: preserved_object, version: current_version, moab_storage_root: msr) } # rubocop:disable RSpec/LetSetup
```
mjgiarlo, line 215: Maybe I failed to drink enough 🍵 before working on this, but I couldn't tell why Rubocop was dinging me for `let!` on this line when this file uses plenty of `let!` already. I found 0 other references to `RSpec/LetSetup` in the codebase, when I expected to see this cop disabled or excluded elsewhere. Alas. 🤷🏻‍♂️

jmartin-sul (approved):

mjgiarlo:
@jmartin-sul do you think we should test this in a deployed env or is it safe to merge? I've unheld the PR for now.

jmartin-sul:
> How was this change tested? 🤨
> 
> CI. I'll have to defer to @jmartin-sul for advice on whether this is sufficient, or whether poking about in a deployed environment may be warranted.

probably a good idea to deploy this to stage or QA and let the nightly replication audit run the new check, just as a basic smoke test.  with data in our test envs being generally iffy, but much smaller in volume than prod, i'm not sure whether to expect the detection of inconsistencies.

alternatively, we could deploy to QA or stage, and manually trigger `#audit_moab_version_replication!` on a handful of random `PreservedObject`s (instead of waiting for the nightly run).

actually, i just looked, and QA and stage each have just over 39k druids... maybe we should just `PreservedObject.find_each(&:audit_moab_version_replication!)` on one of those envs, and see what happens when everything is checked?  those environments don't use ceph and so are unlikely to have the issue with silent omission of e.g. a random datastream file or two.  but there are probably other size inconsistencies because those envs are such a mess.  and if we really wanted to be thorough, maybe we could hand-edit a `ZipPart` or two to have a smaller size, as a test?  but maybe that's overkill.

i think any sort actual test in stage or QA, whether it was on known bad data or not, would suffice, just to make sure the happy path is working as expected.  i'd be fine to merge with even cursory testing in stage or QA.

mjgiarlo:
@jmartin-sul 💬 
> actually, i just looked, and QA and stage each have just over 39k druids... maybe we should just `PreservedObject.find_each(&:audit_moab_version_replication!)` on one of those envs, and see what happens when everything is checked?

OK, I did this on QA and Resque is slowly working down the queue of ~39K jobs. What should I be looking for when this finishes? Honeybadger exceptions? Nothing so far, anyway.

mjgiarlo:
> @jmartin-sul speech_balloon
> 
> > actually, i just looked, and QA and stage each have just over 39k druids... maybe we should just `PreservedObject.find_each(&:audit_moab_version_replication!)` on one of those envs, and see what happens when everything is checked?
> 
> OK, I did this on QA and Resque is slowly working down the queue of ~39K jobs. What should I be looking for when this finishes? Honeybadger exceptions? Nothing so far, anyway.

@jmartin-sul This is now complete. There were a handful of part replication audit job runs that raised exceptions: https://app.honeybadger.io/projects/54415/faults/83422358 But none of them appear to be due to size inconsistency at first blush.

jmartin-sul:
> @jmartin-sul 💬
> 
> > actually, i just looked, and QA and stage each have just over 39k druids... maybe we should just `PreservedObject.find_each(&:audit_moab_version_replication!)` on one of those envs, and see what happens when everything is checked?
> 
> OK, I did this on QA and Resque is slowly working down the queue of ~39K jobs. What should I be looking for when this finishes? Honeybadger exceptions? Nothing so far, anyway.

i think i'd mostly look for failed jobs.  if any HB alerts come up, i'd probably just give one or two a look to confirm they're indeed pointing out something mis-replicated.

and... i just realized, thinking about this again after writing that last sentence, i overlooked something in my review:  i think the new result code needs to be added to the applicable lists of handled codes in the specific reporter classes (under `app/services/reporters`).  i'd just put it next to `ZIP_PARTS_COUNT_INCONSISTENCY` in those classes.  so, `handled_single_codes` in `honeybadger_reporter.rb`, and `handled_merge_codes` in `event_service_reporter.rb` and `audit_workflow_reporter.rb`.  for that last one, i notice also that all the zip part codes are commented out -- i think you could do the same here.  iirc, those were commented out because they exceeded some field size limit in the WFS DB before it was migrated from oracle to PG?  i think there's a ticket to revisit that, but if not, i'll file one.

sorry for the lag and the confusion!  not my finest review 😬 

also: i'm gonna merge this, and we can deal with the reporter result codes in a separate issue/PR.  i'm looking at `c2a.log` on the QA worker box (-02), and it seems to be recording the usual flurry of `not all ZippedMoabVersion parts are replicated yet` (presumably from failed deliveries that were never retried), and i saw endpoint specific part audit jobs get queued, which wouldn't happen if the `CatalogToArchive.check_child_zip_part_attributes` method modified by this PR was blowing up.

and then i went ahead and did the test where i manually futzed with a part size, and the updated audit caught the issue 🎉   i grabbed a random zip part (happened to be for `ry530ff4694` on QA), set its size to 1, ran the audit on the druid, and saw this result when i grepped the logs for the new message: `log/c2a.log:E, [2022-11-03T16:21:56.205400 #1205410] ERROR -- : PartReplicationAuditJob(ry530ff4694, aws_s3_east_1) 1 on aws_s3_east_1: Sum of ZippedMoabVersion child part sizes (1) is less than what is in the Moab: 51025`

so yeah, i think this works!

mjgiarlo:
> > @jmartin-sul speech_balloon
> > > actually, i just looked, and QA and stage each have just over 39k druids... maybe we should just `PreservedObject.find_each(&:audit_moab_version_replication!)` on one of those envs, and see what happens when everything is checked?
> > 
> > 
> > OK, I did this on QA and Resque is slowly working down the queue of ~39K jobs. What should I be looking for when this finishes? Honeybadger exceptions? Nothing so far, anyway.
> 
> i think i'd mostly look for failed jobs. if any HB alerts come up, i'd probably just give one or two a look to confirm they're indeed pointing out something mis-replicated.
> 
> and... i just realized, thinking about this again after writing that last sentence, i overlooked something in my review: i think the new result code needs to be added to the applicable lists of handled codes in the specific reporter classes (under `app/services/reporters`). i'd just put it next to `ZIP_PARTS_COUNT_INCONSISTENCY` in those classes. so, `handled_single_codes` in `honeybadger_reporter.rb`, and `handled_merge_codes` in `event_service_reporter.rb` and `audit_workflow_reporter.rb`. for that last one, i notice also that all the zip part codes are commented out -- i think you could do the same here. iirc, those were commented out because they exceeded some field size limit in the WFS DB before it was migrated from oracle to PG? i think there's a ticket to revisit that, but if not, i'll file one.
> 
> sorry for the lag and the confusion! not my finest review grimacing
> 
> also: i'm gonna merge this, and we can deal with the reporter result codes in a separate issue/PR. i'm looking at `c2a.log` on the QA worker box (-02), and it seems to be recording the usual flurry of `not all ZippedMoabVersion parts are replicated yet` (presumably from failed deliveries that were never retried), and i saw endpoint specific part audit jobs get queued, which wouldn't happen if the `CatalogToArchive.check_child_zip_part_attributes` method modified by this PR was blowing up.
> 
> and then i went ahead and did the test where i manually futzed with a part size, and the updated audit caught the issue tada i grabbed a random zip part (happened to be for `ry530ff4694` on QA), set its size to 1, ran the audit on the druid, and saw this result when i grepped the logs for the new message: `log/c2a.log:E, [2022-11-03T16:21:56.205400 #1205410] ERROR -- : PartReplicationAuditJob(ry530ff4694, aws_s3_east_1) 1 on aws_s3_east_1: Sum of ZippedMoabVersion child part sizes (1) is less than what is in the Moab: 51025` (i undid my size tweak when i was done testing)
> 
> so yeah, i think this works!

Thank you, John! 🍻

jmartin-sul:
existing issue about commented result codes: https://github.com/sul-dlss/preservation_catalog/issues/1240

spec/models/preserved_object_spec.rb
ndushay: I *think* it might have been complaining because `cm1` is never explicitly called.  If this is done in a before block, without using let, I think it would work.  but no biggie.

PR #2012, opened 2022-11-03
add debug gem and binstub for running tests in vscode

ndushay:

For the good of all ... who use vscode


jmartin-sul (approved):

Gemfile
```diff
@@ -41,6 +41,7 @@ group :development do
 end
 
 group :test do
+  gem 'debug'
```
mjgiarlo, line 44: Can you say more about how `debug` relates to e.g. `byebug`? I wonder if `pry-byebug` can be removed now?

PR #2035, opened 2022-11-10
Refactor: CatalogRemediator slight refactors, ditto S3Audit

ndushay:

I was having trouble grokking the code; I believe these changes, largely comments, make it easier.  

CodeClimate is out of its gourd.

Wingnut approves this PR.

image

specs

### DIFF
--- a/app/services/catalog_remediator.rb
+++ b/app/services/catalog_remediator.rb
@@ -11,6 +11,8 @@ class CatalogRemediator
     @version = version
   end
 
+  # prunes ZipParts and ZippedMoabVersion from database for any ZippedMoabVersions that have replication errors for
+  #   the druid and version (so remediation can then be kicked off with PreservedObject.create_zipped_moab_versions!)
   def prune_replication_failures
     zipped_moab_versions_with_errors.map do |zipped_moab_version, audit_results|
       zip_parts = zipped_moab_version.zip_parts
@@ -20,7 +22,7 @@ class CatalogRemediator
       )
 
       ApplicationRecord.transaction do
-        # NOTE: The order here matters! ZipPart instances must be destroyed before the ZippedMoabVersion
+        # NOTE: ZipPart instances must be destroyed _before_ the ZippedMoabVersion
         zip_parts.destroy_all
         zipped_moab_version.destroy
       end
@@ -49,12 +51,13 @@ class CatalogRemediator
 
   def zipped_moab_versions_with_errors
     zipped_moab_versions_beyond_expiry.to_a.filter_map do |zipped_moab_version|
-      audit_results = audit_results_for(zipped_moab_version)
-      # If this check returns false, it indicates a lack of zip parts, in which
-      # case we still want to delete the ZMV, so short-circuit the rest of the
-      # block
+      audit_results = empty_audit_results(zipped_moab_version)
+      # when Audit::CatalogToArchive.check_child_zip_part_attributes() returns true, we have ZipParts for the
+      #   ZippedMoabVersion, so continue to the next check
       next [zipped_moab_version, audit_results] unless check_child_zip_part_attributes(zipped_moab_version, audit_results)
 
+      # S3Audit.check_replicated_zipped_moab_version creates audit_results.error_results if problem with existence or
+      #   checksum of replicated ZipPart files
       endpoint_audit_class_for(zipped_moab_version).check_replicated_zipped_moab_version(zipped_moab_version, audit_results, true)
       next if audit_results.error_results.empty?
 
@@ -62,7 +65,7 @@ class CatalogRemediator
     end
   end
 
-  def audit_results_for(zipped_moab_version)
+  def empty_audit_results(zipped_moab_version)
     AuditResults.new(druid: druid, actual_version: version, moab_storage_root: zipped_moab_version.zip_endpoint)
   end
 
--- a/app/services/s3/s3_audit.rb
+++ b/app/services/s3/s3_audit.rb
@@ -3,6 +3,10 @@
 module S3
   # Base class for AWS and IBM audit classes
   class S3Audit
+    def self.check_replicated_zipped_moab_version(zmv, results, check_unreplicated_parts = false)
+      new(zmv, results, check_unreplicated_parts).check_replicated_zipped_moab_version
+    end
+
     delegate :bucket, :bucket_name, to: :s3_provider
 
     attr_reader :zmv, :results, :check_unreplicated_parts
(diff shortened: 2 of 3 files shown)


app/services/catalog_remediator.rb
```diff
@@ -11,6 +13,8 @@ def initialize(druid:, version:)
     @version = version
   end
 
+  # prunes ZipParts and ZippedMoabVersion from database for any ZippedMoabVersions that have replication errors for
+  #   the druid and version
```
mjgiarlo, line 17: Can you say more about having similar but slightly different documentation on lines 5-6 and 16-17 (and also line 3)? I might be biased in favor of a single comment at the class-level since this service object only does one thing.

app/services/catalog_remediator.rb
```diff
@@ -49,20 +53,21 @@ def zipped_moab_versions_beyond_expiry
...
+      #   checksum of replicated files
       endpoint_audit_class_for(zipped_moab_version).check_replicated_zipped_moab_version(zipped_moab_version, audit_results, true)
       next if audit_results.error_results.empty?
 
       [zipped_moab_version, audit_results]
     end
   end
 
-  def audit_results_for(zipped_moab_version)
+  def empty_audit_results(zipped_moab_version)
```
mjgiarlo, line 70: Are they necessarily empty? I'm not sure I get this change.

app/services/catalog_remediator.rb
```diff
@@ -49,20 +53,21 @@ def zipped_moab_versions_beyond_expiry
 
   def zipped_moab_versions_with_errors
     zipped_moab_versions_beyond_expiry.to_a.filter_map do |zipped_moab_version|
-      audit_results = audit_results_for(zipped_moab_version)
-      # If this check returns false, it indicates a lack of zip parts, in which
-      # case we still want to delete the ZMV, so short-circuit the rest of the
-      # block
+      audit_results = empty_audit_results(zipped_moab_version)
+      # when Audit::CatalogToArchive.check_child_zip_part_attributes() returns true, we have all ZipParts for the
+      #   ZippedMoabVersion, so continue to the next check
```
mjgiarlo, line 58: I think I prefer the prior wording here but it could be because I probably wrote it. Would be good to get more @sul-dlss/infrastructure-team  👀 on this one.

app/services/catalog_remediator.rb
```diff
@@ -49,20 +53,21 @@ def zipped_moab_versions_beyond_expiry
...
       next [zipped_moab_version, audit_results] unless check_child_zip_part_attributes(zipped_moab_version, audit_results)
 
+      # S3Audit.check_replicated_zipped_moab_version creates audit_results.error_results if problem with existence or
+      #   checksum of replicated files
```
mjgiarlo, line 62: I do like surfacing that this method mutates the audit results, which better connects it to the following line.

app/services/s3/s3_audit.rb
```diff
@@ -3,6 +3,10 @@
 module S3
   # Base class for AWS and IBM audit classes
   class S3Audit
+    def self.check_replicated_zipped_moab_version(zmv, results, check_unreplicated_parts = false)
+      new(zmv, results, check_unreplicated_parts).check_replicated_zipped_moab_version
+    end
```
mjgiarlo, line 8: Thanks for moving this to the top!

app/services/catalog_remediator.rb
ndushay: happy to have it on the instance method only;  not sure the class will continue to do only one thing.

app/services/catalog_remediator.rb
ndushay: they're empty until the checks are done, no?  thats the impression I got from John when we were speaking earlier.

app/services/catalog_remediator.rb
ndushay: I tried to grok the previous wording, and I just couldn't (the unless is part of the problem) so this was one of the big reasons for this PR.

app/services/catalog_remediator.rb
```diff
@@ -49,20 +53,21 @@ def zipped_moab_versions_beyond_expiry
 
   def zipped_moab_versions_with_errors
     zipped_moab_versions_beyond_expiry.to_a.filter_map do |zipped_moab_version|
-      audit_results = audit_results_for(zipped_moab_version)
-      # If this check returns false, it indicates a lack of zip parts, in which
-      # case we still want to delete the ZMV, so short-circuit the rest of the
-      # block
+      audit_results = empty_audit_results(zipped_moab_version)
+      # when Audit::CatalogToArchive.check_child_zip_part_attributes() returns true, we have all ZipParts for the
```
jmartin-sul, line 57: small quibble -- i think it'd be more accurate to say that "we have at least some ZipParts" or just "we have ZipParts" in this sentence.  the condition checked in `check_child_zip_part_attributes` is `unless zmv.zip_parts.count.positive?`.  because if there are any ZipParts, then there are other useful consistency checks that can be done, both just within the DB and by reaching out to the cloud (e.g. if we're missing ZipParts or we detect a count inconsistency, this method will still return `true`, i believe).

jmartin-sul (commented):
i have one small comment quibble, but i think this is an improvement.  but i am also the wrong person to judge whether this is an improvement, so i will let fresher eyes approve either way 🙂

app/services/catalog_remediator.rb
jmartin-sul: the method that's being renamed does, i believe, return an empty data structure, which is then populated by the subsequent checks its passed to.  i don't know that either of these names is as clear as i'd like, but i also don't have a better suggestion.

jmartin-sul:
yeah, i don't get the codeclimate complaint either, all the changed lines of code appear to be covered.

wingnut also looks skeptical of the total-coverage ding.

app/services/catalog_remediator.rb
ndushay: made it "we have ZipParts"

app/services/catalog_remediator.rb
ndushay: I just double-checked, and it is empty.

app/services/catalog_remediator.rb
ndushay: now on instance method only.  thanks for suggestion.

ndushay:
@mjgiarlo @jmartin-sul I believe I addressed your concerns and took your most excellent suggestions.

mjgiarlo (approved):

mjgiarlo:
@jmartin-sul Happy for this to be merged if @jmartin-sul approves

jmartin-sul (approved):

PR #2053, opened 2022-11-21
Rename MoabStorageRootReporter

justinlittman:
Reduce confusion by only using "reporters" in one way -- see the reporters module.

### DIFF
--- a/app/services/moab_storage_root_report_service.rb
+++ b/app/services/moab_storage_root_report_service.rb
@@ -5,11 +5,10 @@ require 'csv'
 ##
 # run queries and produce reports from the results, for consumption
 # by preservation catalog maintainers
-class MoabStorageRootReporter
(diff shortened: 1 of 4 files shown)


app/services/moab_storage_root_report_service.rb
```diff
@@ -5,11 +5,10 @@
 ##
 # run queries and produce reports from the results, for consumption
```
ndushay, line 6: is this comment still correct?

ndushay (approved):
looks good;  one question for your consideration.

app/services/moab_storage_root_report_service.rb
justinlittman: As far as I know it is. I think?

jmartin-sul: yup, i think so

PR #2061, opened 2022-11-22
Refactor catalog to moab

justinlittman:
Readability

### DIFF
--- a/app/services/audit/catalog_to_moab.rb
+++ b/app/services/audit/catalog_to_moab.rb
@@ -1,45 +1,33 @@
 # frozen_string_literal: true
 
 module Audit
-  # Catalog to Moab existence check code
+  # Service for checking versions, validating moab on storage, and updating catalog.
   class CatalogToMoab
-    attr_reader :complete_moab, :druid, :logger
+    attr_reader :complete_moab, :logger
 
     def initialize(complete_moab)
       @complete_moab = complete_moab
-      @druid = complete_moab.preserved_object.druid
       @logger = Logger.new(Rails.root.join('log', 'c2m.log'))
     end
 
-    # shameless green implementation
+    # Check the catalog version (complete moab version) against versions of the preserved object and the moab on storage,
+    # possibly validate the moab on storage, and possibly update the complete moab.
     def check_catalog_version
-      unless complete_moab.matches_po_current_version?
-        results.add_result(AuditResults::CM_PO_VERSION_MISMATCH,
-                           cm_version: complete_moab.version,
-                           po_version: complete_moab.preserved_object.current_version)
-        return report_results!
-      end
+      return report_results! unless check_preserved_object_and_complete_moab_versions_match
 
-      if moab_on_storage.nil?
-        transaction_ok = ActiveRecordUtils.with_transaction_and_rescue(results) do
-          status_handler.update_status('online_moab_not_found')
-          complete_moab.save!
-        end
-        results.remove_db_updated_results unless transaction_ok
+      return report_results! unless check_for_moab_on_storage
 
-        results.add_result(AuditResults::MOAB_NOT_FOUND,
-                           db_created_at: complete_moab.created_at.iso8601,
-                           db_updated_at: complete_moab.updated_at.iso8601)
-        return report_results!
-      end
-
+      # Complete moab status is not currently 'invalid_checksum'
       return report_results! unless moab_on_storage_validator.can_validate_current_comp_moab_status?(complete_moab: complete_moab)
 
-      compare_version_and_take_action
-    end
+      # An expected outcome if nothing changes on storage.
+      return report_versions_match if complete_moab.version == moab_on_storage_version
 
-    def moab_on_storage
-      @moab_on_storage ||= MoabOnStorage.moab(druid: druid, storage_location: storage_location)
+      # An expected outcome if new version added to storage. Validate and update the complete moab version.
+      return validate_moab_on_storage_and_report_results if complete_moab.version < moab_on_storage_version
+
+      # An error when complete_moab.version > moab_on_storage_version
+      report_complete_moab_greater_than_moab_on_storage_version
     end
 
     def results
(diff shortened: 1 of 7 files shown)


app/services/status_handler.rb
```diff
@@ -12,7 +12,7 @@ def initialize(audit_results:, complete_moab:)
     @complete_moab = complete_moab
   end
 
-  def update_status(new_status)
+  def update_complete_moab_status(new_status)
```
justinlittman, line 15: Also did some renaming of StatusHandler methods to make them clearer.

app/services/status_handler.rb
```diff
@@ -40,24 +40,24 @@ def mark_moab_not_found
   # @caller_validates_checksums [Boolean] defaults to false.  was this called by code that re-computes checksums to confirm that they match the
   #   values listed in the manifests?
   # @return [void]
-  def set_status_as_seen_on_disk(found_expected_version:, moab_on_storage_validator:, caller_validates_checksums: false)
+  def validate_moab_on_storage_and_set_status(found_expected_version:, moab_on_storage_validator:, caller_validates_checksums: false)
```
justinlittman, line 43: Took me a long time to realize that validation was happening in catalog to moab and this is where it was happening.

app/services/checksum_validator.rb
```diff
@@ -43,19 +43,18 @@ def moab_on_storage_absent?
...
   # Moab on storage and complete moab versions match?
   def versions_match?
     moab_on_storage.current_version_id == complete_moab.version
   end
 
   def validate_versions
-    # set_status_as_seen_on_disk will update results and complete_moab
-    status_handler.set_status_as_seen_on_disk(found_expected_version: versions_match?, moab_on_storage_validator: moab_on_storage_validator,
-                                              caller_validates_checksums: true)
+    # validate_moab_on_storage_and_set_status will update results and complete_moab
```
ndushay, line 54: is this a method comment?  Ruby practice is outside the block.

app/services/complete_moab_service/update_version.rb
```diff
@@ -44,7 +44,7 @@ def update_complete_moab_to_expected_version(status:, checksums_validated:)
 
       complete_moab.upd_audstamps_version_size(moab_on_storage_validator.ran_moab_validation?, incoming_version, incoming_size)
       complete_moab.last_checksum_validation = Time.current if checksums_validated && complete_moab.last_checksum_validation
-      status_handler.update_status(status) if status
+      status_handler.update_complete_moab_status(status) if status
```
ndushay, line 47: bleah.  For another PR, but ... `update_moab_catalog_status` or something?

app/services/status_handler.rb
ndushay: super!  But is this better named 
- update_moab_catalog_status
- update_catalog_moab_status
- update_db_moab_status 

or something?

ndushay (commented):
Reviewed.  Some comments.  None are blockers -- I think what's here is better than what was.  Just wondering if any of the tweaks I suggest would be simple and improve this.

Setting the status in results.add_result ... is not "record_result"?   So we have multiple kinds of results?  I'm confused.  This can be addressed in a second refactor.

app/services/audit/catalog_to_moab.rb
```diff
@@ -3,43 +3,30 @@
...
 
     def initialize(complete_moab)
       @complete_moab = complete_moab
-      @druid = complete_moab.preserved_object.druid
       @logger = Logger.new(Rails.root.join('log', 'c2m.log'))
     end
 
-    # shameless green implementation
+    # Check the catalog version (complete moab version) against versions of the preserved object and the moab on storage
     def check_catalog_version
```
mjgiarlo, line 14: I don't have specific qualms with any of the changes in this class, but in total, the class is iceberg-like: it's bottom-heavy with the public API doing little other than invoking private methods*. We are guilty of this in lots of places, I concede.

* I do like the intention-revealing names of these methods! I'm more wondering if we shouldn't inline more of the code lurking in the private methods.

app/services/status_handler.rb
mjgiarlo: 🤷🏻‍♂️

mjgiarlo (approved):
Dropped a couple specific comments but they're non-blocking. LGTM.

app/services/status_handler.rb
justinlittman: So it is updating the complete moab, which is the current name. If in the future the complete moab get renamed, then this method should get renamed accordingly.

app/services/audit/catalog_to_moab.rb
```diff
@@ -49,6 +36,36 @@ def results
 
     private
 
+    def druid
+      @druid ||= complete_moab.preserved_object.druid
+    end
+
+    def check_preserved_object_and_complete_moab_versions_match
```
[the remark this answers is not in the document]
justinlittman: The intent of calling it "check_*" instead of "*?" is to indicate that there is more happening than just an equality check. I'm open to other suggestions for a follow-up PR.

app/services/audit/catalog_to_moab.rb
```diff
@@ -49,6 +36,36 @@ def results
...
+      return true if complete_moab.matches_po_current_version?
+
+      results.add_result(AuditResults::CM_PO_VERSION_MISMATCH,
+                         cm_version: complete_moab.version,
+                         po_version: complete_moab.preserved_object.current_version)
+      false
+    end
+
+    def check_moab_on_storage_present
```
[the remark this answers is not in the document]
justinlittman: Going with "check_for_moab_on_storage".

app/services/checksum_validator.rb
justinlittman: No, it is a comment about the following line not the entire method.

app/services/audit/catalog_to_moab.rb
justinlittman: Yes, this is quite deliberate and I've been using this strategy throughout prescat. The goal is to provide a grokkable version of what actions the class is doing. This is particularly important because many of the prescat classes involve many steps / conditionals. With the previous inline code you'd lose track of what was happening before you got through the main method.

This could also be accomplished with comments, but comments always end up out of sync.

Open to suggestions about how to adjust this strategy and/or specific tweaks.

app/services/audit/catalog_to_moab.rb
```diff
@@ -3,43 +3,30 @@
...
       return report_results! unless moab_on_storage_validator.can_validate_current_comp_moab_status?(complete_moab: complete_moab)
 
-      compare_version_and_take_action
-    end
+      # An expected outcome.
+      return report_versions_match if complete_moab.version == moab_on_storage_version
 
-    def moab_on_storage
-      @moab_on_storage ||= MoabOnStorage.moab(druid: druid, storage_location: storage_location)
+      # An expected outcome. Validate and update the complete moab version.
```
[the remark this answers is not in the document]
justinlittman: Enhanced comment.

app/services/audit/catalog_to_moab.rb
mjgiarlo: OK, then let's roll with it. 😄

PR #2062, opened 2022-11-22
Rename moab to catalog

justinlittman:
Very, very misleading name.


ndushay (approved):
Approving.  Leaving in case someone else wants to review.   We should notify @julianmorley of this change in case he needs to change his scripts ...

mjgiarlo (approved):

Issue #2059, opened 2022-11-22
Audit for Replicated files - put job classes in namespace, rename queues to match


ndushay:
Replication audits include what is currently called "CatalogToArchive" which is a service class, and some other service classes as well.  

Here's a proposal for refactoring the relevant classes and redis queue names - if it doesn't make sense, float an alternative to the team.

Put these classes in the Audit::Replication namespace:
- services:
    - [ ] CatalogToArchive  (currently in services/audit - move to services/audit/replication and rename to Audit::Replication::ZipPartsReplicatedService 
    - [ ] S3Audit (currently in services/s3) - move to services/audit/replication and rename Audit::Replication::ZippedMoabVersionToEndpoint
- jobs:
    - [ ] MoabReplicationAuditJob becomes Audit::Replication::MoabToAllEndpointsJob  and redis queue becomes audit_replication_moab_to_all_endpoints
    - [ ] PartReplicationAuditJob becomes Audit::Replication::MoabVersionToEndpointJob and redis queue(s) become audit_replication_moab_version_to_(endpoint)  where (endpoint) is one of aws_east_1, aws_west_2 or ibm_us_south?

Your lowly ticket writer can't figure out which classes correspond to these redis queue names and exactly how it is done:
        - part_audit_aws_s3_east_1
        - part_audit_aws_s3_west_2
        - part_audit_ibm_us_south

From Services::S3::AWS, we have audit.rb:

```
module S3
  module Aws
    # Methods for auditing the state of a ZippedMoabVersion on an AWS S3 endpoint.
    # @note this class name appears in the configuration for the endpoints for which it audits content.
    #   Please update the configs for the various environments if it's renamed or moved.
    class Audit < S3::S3Audit
      def s3_provider_class
        ::S3::AwsProvider
      end
    end
  end
end
```

which implies there is db data to be migrated for those endpoint renames.  Confer with @jmartin-sul on that point.

ndushay:
closing this as PR #2157 is taking a different, better course for audit replication

PR #2070, opened 2022-11-23
Renames replication related services.

justinlittman:
HOLD BECAUSE THIS PR IS GOING TO REQUIRE SHARED CONFIG CHANGES.

also HOLD:  wait for sidekiq to replace redis (#1984) in _production_ before this change is **merged** to prod!

refs #2059

Less squinting.

HOLD BECAUSE THIS PR IS GOING TO REQUIRE SHARED CONFIG CHANGES.


justinlittman:
image

app/services/audit/base_replication_to_endpoint.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-module S3
+module Audit
   # Base class for AWS and IBM audit classes
-  class S3Audit
+  class BaseReplicationToEndpoint
```
ndushay, line 5: Isn't Base usually the last bit of the compound name?   I know it's long here.  If we do Audit::Replication::ToEndpointBase is that worse or better?

app/services/audit/replication_support.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 module Audit
-  # Catalog to cloud archive provider (what ZipEndpoint, ZippedMoabVersion etc represent) audit code
-  class CatalogToArchive
+  # Methods to support auditing replication.
+  class ReplicationSupport
```
ndushay, line 5: Why not Audit::ReplicatonService ... like every other service class?  "Support" is a new term being introduced, and I'm open to it if there's a reason and consistency.

app/services/replication/aws_provider.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module S3
+module Replication
```
ndushay, line 3: just spitballing here:  S3 has some meaning, since everything we use is built around it.  Replication has better meaning ... but should we drop S3?   Is there a utility in S3Base class name?

app/services/replication/failure_remediator.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+module Replication
+  # Remediates catalog entries, e.g., prunes failed replication records
+  # Currently this is only called from rake.
+  class FailureRemediator
```
ndushay, line 6: love this name.

config/settings/test.yml
```diff
@@ -20,15 +20,15 @@ zip_endpoints:
     endpoint_node: 'us-west-2'
     storage_location: 'sul-sdr-aws-us-west-2-test'
     delivery_class: 'S3WestDeliveryJob'
-    audit_class: 'S3::Aws::Audit'
+    audit_class: 'Audit::ReplicationToAws'
```
ndushay, line 23: it certainly reads well in this context.

app/services/replication/zip_delivery_service.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Replication
+  # Posts zips to S3, if needed.
+  class ZipDeliveryService
```
ndushay, line 5: yay!  this is a perfect was that the namespace helps.

ndushay (changes requested):
Justin - thanks for this.  There's a lot of good stuff and it's very interesting.  I think we should discuss our refactoring notions a bit more before I buy into this as it stands.

wondering about splitting the refactor of the replication jobs and services from the  AUDIT of those services, to make the PR(s) easier to review.   

Re:  rename of Audit::CatalogToArchive, see ticket #2059 for my thoughts on this.  See also the description now added to the top of the dashboard (if you click the button):

image

The notion of the "catalog" still applies to replication.  We should brainstorm together on names.

app/services/audit/base_replication_to_endpoint.rb
justinlittman: In my head it is always first, but if there is a convention happy to move it.

I didn't add a `Replication` namespace to `Audit` because none of the other audits have namespaces and didn't think it helped much.

app/services/audit/replication_support.rb
justinlittman: The reason I used `Support` was because in general (with the exception of `ActiveRecordUtils`) `Utils` in the codebase are standalone helpers for one-off work (e.g., via the rails console).

ndushay (commented):
approving modulo BaseReplicationxxx becoming ReplicationxxxxBase

and rebase.

ndushay (approved):

PR #2092, opened 2022-12-01
ZipDeliveryService notifies Honeybadger if s3 part exists already

### LINKED ISSUE #2041: ZipDeliveryService: complain loudly if the part already exists
### TODO

Currently, we just return silently if the part already exists in the cloud:  https://github.com/sul-dlss/preservation_catalog/blob/ffb18f660cb651c258032ba8746deeec15375deb/app/services/zip_delivery_service.rb#L18-L20

However, we may want to update this behavior a bit: as far as I know, it'd be pretty unexpected to actually re-attempt delivery of something that'd been successfully uploaded, and I'm unaware of a situation where we queue multiple jobs to deliver the same part to the same endpoint.  Additionally, we want to enable automated backfilling for missing `ZippedMoabVersion`s (https://github.com/sul-dlss/preservation_catalog/issues/2036).  More on the relevance of that in "Context".

I'd suggest a helpful alert message like `"WARNING: attempting to push a druid version zip part to an S3 location that already has content.  Perhaps a replication failure was pruned from the database, but still needs to be cleaned up from the cloud.  Prune the failure again and ask ops to delete the bad replicated content."`  And then I'd include the druid/version/endpoint in the HB alert context.  However I wouldn't `raise` or cause the job to fail, and would still be sure to return a falsey value in this case, because there's no use in retrying the job, and we also don't want the calling `AbstractDeliveryJob` to proceed to calling `ResultsRecorderJob.perform_later`, since nothing was delivered.

### Context

We'll also be trying to clean up known partial replications (see #1733), and we've just enabled a new audit error about sanity checks on size (see #1993) that will likely alert us to a few more replications that need re-doing.

Since we want to restrict deletion and overwrite of cloud archive content as much as possible, we have to request that ops handle deletion of partial replications.

So, we could occasionally run into a corner case where:
* we've cleaned up the database records for a mis-replicated or partially replicated druid version on a given endpoint, using `CatalogRemediator.prune_replication_failures` or its `rake` task (`prune_failed_replicationprune_failed_replication`)
* we've given ops the druid/version/endpoint combos for deletion from the DB records that `prune_replication_failures` determined needed a re-push (it'll be fed from `CatalogToArchive` audit results).
* ops hasn't had a chance to actually delete the bad cloud content
* `CatalogToArchive` runs for a druid in this situation, tries to backfill the missing content, it can't push over the yet-to-be-removed bad replicated content.

Since replication audit only runs every 3 months on a given druid, it seems unlikely that we'll run into this situation much if at all, so starting out with a simple HB alert to let us know this happened, and that we need to re-run `prune_replication_failures` on the druid version, seems like a better approach to me than trying to do something more automated.

ndushay:

These errors must be surfaced so we can address them.  Automatic retries won't fix them, and allowing the `CatalogToArchive` audit to backfill missing parts ... will never address this either.   So let's report the error.

Fixes #2041 

new unit test

### DIFF
--- a/app/services/zip_delivery_service.rb
+++ b/app/services/zip_delivery_service.rb
@@ -15,8 +15,22 @@ class ZipDeliveryService
     @metadata = stringify_values(metadata)
   end
 
-  def deliver
-    return if s3_part.exists?
+  def deliver # rubocop:disable Metrics/AbcSize
+    if s3_part.exists?
+      msg = 'WARNING: S3 location already has content. ' \
+            'Perhaps a failed replication was pruned from the database, but was not removed from the cloud. ' \
+            "If so, prune this content from the database again (there's a rake task), ask ops to delete the bad replicated content, and try again."
+      hb_context = {
+        druid: dvz_part.druid.druid, # dvz_part.dvz.druid is a DruidTools::Druid object
+        version: dvz_part.dvz.version,
+        endpoint: s3_part.bucket_name
+      }
+      Honeybadger.notify(msg, context: hb_context)
+      Rails.logger.warn("msg: #{msg}, context: #{hb_context}")
+
+      # retries will fail and as nothing was delivered, we do NOT want to proceed with replication job sequence.
+      return
+    end
 
     raise "#{s3_part.key} MD5 mismatch: passed #{given_md5}, computed #{fresh_md5}" if fresh_md5 != given_md5
 
--- a/spec/services/zip_delivery_service_spec.rb
+++ b/spec/services/zip_delivery_service_spec.rb
@@ -8,7 +8,8 @@ describe ZipDeliveryService do
   let(:version) { 1 }
   let(:dvz) { Replication::DruidVersionZip.new(druid, version) }
   let(:dvz_part) { Replication::DruidVersionZipPart.new(dvz, part_s3_key) }
-  let(:s3_part) { instance_double(::Aws::S3::Object, exists?: part_exists, upload_file: true, key: part_s3_key) }
+  let(:bucket_name) { 's3-bucket-shop' }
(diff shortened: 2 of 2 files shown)


ndushay:
tagging @jmartin-sul so he at some point reviews this and is comfortable with it fulfilling ticket #2041 -- squawk when s3 part already exists for ZipDeliveryService

app/services/zip_delivery_service.rb
```diff
@@ -15,8 +15,22 @@ def initialize(s3_part:, dvz_part:, metadata:)
...
 
-  def deliver
-    return if s3_part.exists?
+  def deliver # rubocop:disable Metrics/AbcSize
+    if s3_part.exists?
+      msg = 'WARNING: S3 location already has content. ' \
+            'Perhaps a failed replication was pruned from the database, but was not removed from the cloud. ' \
+            "If so, prune this content from the database again (there's a rake task), ask ops to delete the bad replicated content, and try again."
+      hb_context = {
+        druid: dvz_part.druid.druid, # dvz_part.dvz.druid is a DruidTools::Druid object
```
justinlittman, line 24: Does this given enough context to know what the S3 location and part is that already has content? I know you're trying to provide additional information (thank you!), but I'm still not following what the S3 location is and what the part is.

Happy to discuss more because you can't be expected to guess what I don't understand.

ndushay: dvz_part is the DruidVersionZipPart so we can get the druid and the version in question (what would need to be redone) and s3_part is an aws-sdk  gem object and the bucket_name should give us the endpoint.  That should be enough (and is what John put in the ticket.)

mjgiarlo (approved):

jmartin-sul:
> tagging @jmartin-sul so he at some point reviews this and is comfortable with it fulfilling ticket #2041 (squawk when s3 part already exists for ZipDeliveryService)

looks great to me, thanks!

PR #2096, opened 2022-12-05
dashboard uses turbo-frames to render while waiting for non-performant pieces.

### LINKED ISSUE #2083: dashboard lazy loading:  create additional routes, etc. to utilize turbo-frames
in addition to `/dashboard`, we need routes for anything going inside a turbo-frame

~~These routes should be used by the dashboard, but that is blocked by #2082, so if this is done first, #2082 will have to do it.~~

ndushay:

We want to render the dashboard and then wait for long loading bits.   This will also help us isolate which bits aren't performant so we can improve them.

Fixes #2083


ndushay (commented):
switched to a controller spec instead of a request spec as we didn't need the request spec overhead for what we are testing.

app/components/spinner.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
```
justinlittman, line 2: Why not just use a partial?

app/views/layouts/application.html.erb
```diff
@@ -8,6 +8,9 @@
     <title>SUL - Preservation Catalog</title>
     <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.2/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-Zenh87qX5JnK2Jl0vWa8Ck2rdkQ2Bzep5IDxbcnCeuOxjzrPF/et3URy9Bv1WTRi" crossorigin="anonymous">
     <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.2.2/dist/js/bootstrap.bundle.min.js" integrity="sha384-OERcA2EqjJCMA+/3y+gxIOqMEjwtxJY7qPCqsdltbNJuaOe923+mo//f6V8Qbsw3" crossorigin="anonymous"></script>
+    <script type="module">
```
justinlittman, line 11: This seems like an atypical place to do this import. How do we handle it in other apps?

config/routes.rb
```diff
@@ -2,12 +2,25 @@
...
-Rails.application.routes.draw do
+Rails.application.routes.draw do # rubocop:disable Metrics/BlockLength
   mount Resque::Server.new,
         at: '/resque',
         constraints: ->(req) { Settings.resque_dashboard_hostnames.include?(req.host) }
 
   get 'dashboard', to: 'dashboard#index', defaults: { format: 'html' }
+  namespace :dashboard do
+    # routes for turbo-frame partials
+    get 'moab_storage_status', to: '/dashboard#moab_storage_status', defaults: { format: 'html' }
```
justinlittman, line 13: Are the defaults necessary here?

spec/routing/routing_spec.rb
```diff
@@ -44,4 +44,64 @@
       expect(get: dashboard_path).to route_to(controller: 'dashboard', action: 'index', format: 'html')
     end
   end
+
+  describe 'dashboard/moab_storage_status' do
+    it 'routes to dashboard controller moab_storage_status action' do
+      expect(get: dashboard_moab_storage_status_path).to route_to(controller: 'dashboard', action: 'moab_storage_status', format: 'html')
+    end
+  end
```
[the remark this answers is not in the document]
justinlittman: Now that they are set up correctly, are they worth keeping around? It is just more code to maintain.

justinlittman (approved):
Comments and questions, but nothing blocking.

app/components/spinner.rb
ndushay: I did what was in Argo.

PR #2114, opened 2022-12-12
dashboard: sort zip part suffixes numerically

### LINKED ISSUE #2027: Dashboard:  ZipPart suffix counts need to be ordered numerically, not lexically
We would expect, for example, `.z100` to be after `.z99`

image

ndushay:

Fixes #2027

To put `.z100` after `.z99`, for example:

### Before

image

### After

(no image because need to deploy to prod for zips that are split into 100 parts)

unit test, runnning locally and on stage

### DIFF
--- a/app/components/dashboard/zip_parts_suffixes_component.html.erb
+++ b/app/components/dashboard/zip_parts_suffixes_component.html.erb
@@ -1,6 +1,7 @@
 <div>
   <h4>ZipPart suffix counts</h4>
   <p>The first ZipPart suffix will always be .zip. If a zip is over <%= Replication::DruidVersionZip::ZIP_SPLIT_SIZE %>, the suffixes of the parts will be .zip, .z01, .z02, etc.</p>
+  <p>The count column should be descending.</p>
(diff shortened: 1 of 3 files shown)


justinlittman (approved):
Out of curiosity, why are having the zip part counts on the dashboard helpful?

ndushay:
> Out of curiosity, why are having the zip part counts on the dashboard helpful?

Not sure they are.  They are at the bottom and with turbo-frame they don't hold up rendering.  

I guess with my initial dashboard implemetation, I started with the db schema and let it inform what I displayed.

ndushay:
> Out of curiosity, why are having the zip part counts on the dashboard helpful?

Also ... we're just getting going with the dashboard ... we'll find out what is missing, what needs tweaking, and maybe jettison some of it over time.

justinlittman:
I would make the pitch for putting less on the dashboard. It adds cognitive overhead for making use of it.

ndushay:
I think more iterations on the dashboard will be warranted;  I think they should be part of a workcycle.  :-)

Issue #2084, opened 2022-11-29
improve dashboard display performance in production


ndushay:
Note that after a prescat reboot, it is many minutes to load, but after that, from the same browser tab, it's much faster.

-------------------------------------

This will be more tractable once #2082 and #2083 are done, as it will be easier to determine where the performance is the worst

It takes many minutes for iteration 1 of the dashboard to load in prod.  Some thoughts:
- figure out where the performance is slowest (audit / catalog / replication)
- figure out which component is the slowest
- figure out what code is the problem (db queries?)

benchmarking tools exist for ruby ...

thoughts on improvements:
- memoize data used in display?
- index fields in database?
- cache some displays?
- (I'm sure you'll have better ones)

justinlittman:
And the queries taking > 500ms are:
```
{"time":"2022-12-20T09:55:58-08:00","duration_ms":1688,"source":"Dashboard::AuditInformationComponent.num_moab_expired_checksum_validation (line 129)","query":"SELECT COUNT(*) FROM 'moab_records' INNER JOIN 'preserved_objects' ON 'preserved_objects'.'id' = 'moab_records'.'preserved_object_id' WHERE ((last_checksum_validation + (7776000 * INTERVAL '1 SECOND')) < CURRENT_TIMESTAMP OR last_checksum_validation IS NULL)"}
{"time":"2022-12-20T09:56:00-08:00","duration_ms":1980,"source":"Dashboard::AuditInformationComponent.num_replication_errors (line 51)","query":"SELECT COUNT(*) FROM 'zip_parts' WHERE 'zip_parts'.'status' != $1","bindings":[0]}
{"time":"2022-12-20T09:56:04-08:00","duration_ms":569,"source":"Dashboard::MoabRecordsComponent.moab_record_total_size (line 110)","query":"SELECT SUM('moab_records'.'size') FROM 'moab_records'"}
{"time":"2022-12-20T09:56:04-08:00","duration_ms":579,"source":"Dashboard::MoabOnStorageStatusComponent.num_moab_records (line 166)","query":"SELECT COUNT(*) FROM 'moab_records'"}
{"time":"2022-12-20T09:56:04-08:00","duration_ms":1365,"source":"Dashboard::MoabStorageRootsComponent.storage_root_info (line 28)","query":"SELECT SUM('moab_records'.'size') FROM 'moab_records' WHERE 'moab_records'.'moab_storage_root_id' = $1","bindings":[25]}
[...]

PR #2115, opened 2022-12-12
dashboard services get memoization for perf improvement

ndushay:

I started a PR for dashboard performance improvemets a while back.  These are simple memoizations that may help performance a bit.  It seemed better to do this than throw away the work.

Part of #2084.

CI

### DIFF
--- a/app/services/dashboard/replication_service.rb
+++ b/app/services/dashboard/replication_service.rb
@@ -6,8 +6,8 @@ require 'action_view' # for number_to_human_size
 module Dashboard
   # methods pertaining to replication (cloud storage) tables in database for dashboard
   module ReplicationService
-    include ActionView::Helpers::NumberHelper # for number_to_human_size
     include MoabOnStorageService
+    include ActionView::Helpers::NumberHelper # for number_to_human_size
 
     def replication_and_zip_parts_ok?
       zip_parts_ok? && replication_ok?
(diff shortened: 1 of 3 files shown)


app/services/dashboard/replication_service.rb
```diff
@@ -25,19 +25,23 @@ def endpoint_replication_count_ok?(endpoint_replication_count)
...
-      endpoint_data = {}
-      ZipEndpoint.all.each do |zip_endpoint|
-        endpoint_data[zip_endpoint.endpoint_name] =
-          {
-            delivery_class: zip_endpoint.delivery_class,
-            replication_count: ZippedMoabVersion.where(zip_endpoint_id: zip_endpoint.id).count
-          }
+      # called multiple times, so memoize to avoid db queries
+      @endpoint_data ||= begin
+        endpoint_data = {}
```
justinlittman, line 30: This would be slightly cleaner as:
```
@endpoint_data ||= {}.tap do |endpoint_data|
  ...
end

justinlittman (approved):
Approve pending consideration of non-blocking comment.

app/services/dashboard/replication_service.rb
ndushay: thanks for this - I forgot about `.tap`

PR #2117, opened 2022-12-13
Fix to rake hb whenever job type.

justinlittman:
I didn't intend to hijack the rake job type.


ndushay (approved):
so this ... doesn't require anything else but this change?  and rake_rb is part of whenever gem?

justinlittman:
No other change required. `rake_rb` is not part of the whenever gem, but is a new whenever job type being defined here.

PR #2130, opened 2022-12-16
Adjust table layout.

### LINKED ISSUE #2124: dashboard UI:  some tables bleed off the right side of the browser window
Note the table on the right is off the page:

image

vs

image

Same for replication files:

image

justinlittman:
closes #2124

Local.

Tables with side-by-side layout changed to vertical layout since tables don't play well in columns.
image

Wide tables made responsive so they get scroll bars when exceed viewport width.
image

### DIFF
--- a/app/views/dashboard/_moab_information.html.erb
+++ b/app/views/dashboard/_moab_information.html.erb
@@ -2,20 +2,14 @@
 <p><strong class="fs-4">PreservedObject</strong> records store the druid and links to db records about both the Moab on local storage and all the replicated instances in the cloud.</p>
 <p><strong class="fs-4">MoabRecord</strong> records are about the Moab on storage.</p>
 <hr/>
-<div class="row row-cols-2">
-  <div class="column col-5">
-    <turbo-frame id="moab-record-versions" src="<%= dashboard_moab_record_versions_path %>" loading="lazy">
-      <%= render Spinner.new %>
-    </turbo-frame>
-  </div>
-  <div class="column col-7">
-    <turbo-frame id="moab-record-info" src="<%= dashboard_moab_record_info_path %>" loading="lazy">
-      <%= render Spinner.new %>
-    </turbo-frame>
-  </div>
-</div>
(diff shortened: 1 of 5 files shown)


app/views/dashboard/_moab_information.html.erb
```diff
@@ -2,20 +2,14 @@
...
-    </turbo-frame>
-  </div>
-</div>
+<turbo-frame id="moab-record-versions" src="<%= dashboard_moab_record_versions_path %>" loading="lazy">
+  <%= render Spinner.new %>
+</turbo-frame>
+<hr/>
+<turbo-frame id="moab-record-info" src="<%= dashboard_moab_record_info_path %>" loading="lazy">
+  <%= render Spinner.new %>
+</turbo-frame>
```
ndushay, line 11: hrm.  I thought it was better use of screen real estate to have this tables beside each other.  Maybe it doesn't work with the left nav-tab stuff.

app/views/dashboard/_replication_information.html.erb
```diff
@@ -6,18 +6,13 @@
...
-    </turbo-frame>
-  </div>
-</div>
+<turbo-frame id="replication-endpoints" src="<%= dashboard_replication_endpoints_path %>" loading="lazy">
+  <%= render Spinner.new %>
+</turbo-frame>
+<hr/>
+<turbo-frame id="replicated-files" src="<%= dashboard_replicated_files_path %>" loading="lazy">
+  <%= render Spinner.new %>
+</turbo-frame>
```
ndushay, line 15: ditto here.

app/views/dashboard/_moab_information.html.erb
ndushay: can we have responsive tables and the side by side, or is that yucky?

ndushay (approved):
approving with a question

app/views/dashboard/_moab_information.html.erb
justinlittman: It turns out yucky because then you end up with a scroll bar on a small table.

ndushay: gotcha.

PR #2133, opened 2022-12-20
Adds query performance instrumentation.

justinlittman:
refs #2084

Support targeted performance instrumentation (in particular, for the dashboard).

Tested on stage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,3 +432,33 @@ Cron check-ins are configured in the following locations:
 2. `config/settings.yml`: Stubs out a check-in key for each cron job. Since we may not want to have a check-in for all environments, this stub key will be used and produce a null check-in.
 3. `config/settings/production.yml` in shared_configs: This contains the actual check-in keys.
 4. HB notification page: Check-ins are configured per project in HB. To configure a check-in, the cron schedule will be needed, which can be found with `bundle exec whenever`. After a check-in is created, the check-in key will be available. (If the URL is `https://api.honeybadger.io/v1/check_in/rkIdpB` then the check-in key will be `rkIdpB`).
+
+## Query Performance Instrumentation
+To log slow queries to `logs/slow_query.log`, enable the slow query instrumentation in `config/setting/production.yml`:
+```
+slow_queries:
+  enable: true
+  threshold: 500 # milliseconds
+```
+
+To instrument a query, include `InstrumentationSupport` and then annotate the query. For example:
+```
+module ReplicationService
+  include InstrumentationSupport
+
+  def num_replication_errors
+    ZipPart.where.not(status: 'ok').annotate(caller).count
+  end
+  ...
+```
+
+Note that `.annotate(caller)` should be placed where it receives and returns an `ActiveRecord::Relation`.
+```
+# Good
+ZipPart.where.not(status: 'ok').annotate(caller).count
+ZipPart.all.annotate(caller).where.not(status: 'ok').count
+
+# Bad
+ZipPart.annotate(caller).where.not(status: 'ok').count
+ZipPart.where.not(status: 'ok').count.annotate(caller)
+```
--- a/app/services/dashboard/audit_service.rb
+++ b/app/services/dashboard/audit_service.rb
@@ -5,6 +5,7 @@ module Dashboard
   # methods pertaining to audit functionality for dashboard
   module AuditService
     include MoabOnStorageService
+    include InstrumentationSupport
 
     # MoabRecord.last_version_audit is the most recent of 3 separate audits:
     #   moab_to_catalog - all MoabRecords are queued for this on the 1st of the month
@@ -25,12 +26,12 @@ module Dashboard
 
     def validate_moab_audit_ok?
       # NOTE: unsure if there needs to be more checking of MoabRecord.status_details for more statuses to figure this out
-      (MoabRecord.invalid_moab.count + MoabRecord.moab_on_storage_not_found.count).zero?
+      (MoabRecord.invalid_moab.annotate(caller).count + MoabRecord.moab_on_storage_not_found.annotate(caller).count).zero?
     end
 
     def catalog_to_moab_audit_ok?
       # NOTE: unsure if there needs to be more checking of MoabRecord.status_details for more statuses to figure this out
-      !MoabRecord.exists?(status: %w[moab_on_storage_not_found unexpected_version_on_storage])
+      !MoabRecord.all.annotate(caller).exists?(status: %w[moab_on_storage_not_found unexpected_version_on_storage])
     end
 
     def moab_to_catalog_audit_ok?
@@ -41,11 +42,11 @@ module Dashboard
 
     def moab_checksum_validation_audit_ok?
       # NOTE: unsure if there needs to be more checking of MoabRecord.status_details for more statuses to figure this out
-      MoabRecord.invalid_checksum.count.zero?
+      MoabRecord.invalid_checksum.annotate(caller).count.zero?
     end
 
     def catalog_to_archive_audit_ok?
-      !ZipPart.where.not(status: 'ok').exists?
+      !ZipPart.where.not(status: 'ok').annotate(caller).exists?
     end
 
     def moab_audit_age_threshold
@@ -53,7 +54,7 @@ module Dashboard
     end
 
     def num_moab_audits_older_than_threshold
-      MoabRecord.version_audit_expired(moab_audit_age_threshold).count
+      MoabRecord.version_audit_expired(moab_audit_age_threshold).annotate(caller).count
     end
 
     def moab_audits_older_than_threshold?
@@ -65,7 +66,7 @@ module Dashboard
     end
 
     def num_replication_audits_older_than_threshold
-      PreservedObject.archive_check_expired.count
+      PreservedObject.archive_check_expired.annotate(caller).count
     end
 
     def replication_audits_older_than_threshold?
--- a/app/services/dashboard/moab_on_storage_service.rb
+++ b/app/services/dashboard/moab_on_storage_service.rb
@@ -7,6 +7,7 @@ module Dashboard
   # methods pertaining to PreservedObject and MoabRecord database data for dashboard
   module MoabOnStorageService # rubocop:disable Metrics/ModuleLength
     include ActionView::Helpers::NumberHelper # for number_to_human_size
+    include InstrumentationSupport
 
     def moabs_on_storage_ok?
       moab_on_storage_counts_ok? && !any_moab_record_errors?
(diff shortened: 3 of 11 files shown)


README.md
```diff
@@ -432,3 +432,22 @@ Cron check-ins are configured in the following locations:
 2. `config/settings.yml`: Stubs out a check-in key for each cron job. Since we may not want to have a check-in for all environments, this stub key will be used and produce a null check-in.
 3. `config/settings/production.yml` in shared_configs: This contains the actual check-in keys.
 4. HB notification page: Check-ins are configured per project in HB. To configure a check-in, the cron schedule will be needed, which can be found with `bundle exec whenever`. After a check-in is created, the check-in key will be available. (If the URL is `https://api.honeybadger.io/v1/check_in/rkIdpB` then the check-in key will be `rkIdpB`).
+
+## Query Performance Instrumentation
+To log slow queries to `logs/slow_query.log`, enable the slow query instrumentation:
+```
+slow_queries:
+  enable: true
+  threshold: 500 # milliseconds
```
mjgiarlo, line 441: Should we specify where to make this change or do you think it's obvious?

README.md
```diff
@@ -432,3 +432,22 @@ Cron check-ins are configured in the following locations:
...
+```
+
+To instrument a query, include `InstrumentationSupport` and then annotate the query. For example:
+```
+module ReplicationService
+  include InstrumentationSupport
+
+  def num_replication_errors
+    ZipPart.annotate(sa).where.not(status: 'ok').count
```
mjgiarlo, line 450: Can you say more about `sa` vs `source_annotation`? It's a trivial delta but thought it worth asking given the semi-recent work to replace acronym usage in prescat with more explicit naming.

mjgiarlo, line 450: And, just curious: where does the `#annotate` method come from? Is that a Rails thing?

app/services/dashboard/audit_service.rb
```diff
@@ -41,19 +42,19 @@ def moab_to_catalog_audit_ok?
 
     def moab_checksum_validation_audit_ok?
       # NOTE: unsure if there needs to be more checking of MoabRecord.status_details for more statuses to figure this out
-      MoabRecord.invalid_checksum.count.zero?
+      MoabRecord.invalid_checksum.annotate(sa).count.zero?
     end
 
     def catalog_to_archive_audit_ok?
-      !ZipPart.where.not(status: 'ok').exists?
+      !ZipPart.where.not(status: 'ok').annotate(sa).exists?
```
mjgiarlo, line 49: Can you say more about where in the chain we generally want to place the `annotate` call? Given what's in the README, I expected `annotate` earlier but maybe that's not appropriate here?

app/services/instrumentation_support.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# Methods to support instrumentation
+module InstrumentationSupport
+  # Provides a source annotation for logging queries.
+  # Example: storage_root.moab_records.annotate(source_annotation).sum(:size)
+  # See lib/slow_query_logger.rb
+  def source_annotation
+    return unless Settings.slow_queries.enable
```
mjgiarlo, line 9: And I guess if the `annotate` method gets a `nil`, it no-ops and does the right thing?

lib/slow_query_logger.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'json'
+require 'digest'
+require 'logger'
```
mjgiarlo, line 5: Presumably these are needed here because this class is loaded via an initializer which is before all the app's bundled gems are loaded?

lib/slow_query_logger.rb
```diff
@@ -0,0 +1,60 @@
...
+
+# Logs slow queries to a file as reported by ActiveSupport::Notifications
+# Only queries annotated with a source will be logged.
+# For example, PreservedObject.annotate('source=MoabOnStorageService.num_preserved_objects').count
+# Adapted from https://sosedoff.com/2020/02/18/rails-log-slow-queries.html
+class SlowQueryLogger
+  def initialize(output:, threshold:)
+    @threshold = threshold
+
+    @logger = Logger.new(output || $stdout)
```
mjgiarlo, line 15: Worth considering moving the default value here (`$stdout`) into the method definition?

```ruby
def initialize(output: $stdout, threshold:)
```

❓

lib/slow_query_logger.rb
```diff
@@ -0,0 +1,60 @@
...
+    @logger.formatter = method(:formatter)
+  end
+
+  def call(_name, start, finish, _id, payload)
+    sql, source = parse_sql(payload[:sql])
+    return unless source
+    # Skip transaction start/end statements
+    return if sql.match(/BEGIN|COMMIT|SET|SHOW/)
```
mjgiarlo, line 23: If we're not using the match object, we could consider `#match?` here also

spec/spec_helper.rb
```diff
@@ -7,6 +7,7 @@
 SimpleCov.start :rails do
   add_filter '/spec/'
   add_filter '/vendor/'
+  add_filter '/lib/'
```
mjgiarlo, line 10: 👍🏻

app/services/instrumentation_support.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    location = Kernel.caller_locations(2, 1).first
+    path = location.path.delete_prefix("#{Rails.root}/")
+
+    "source=#{path}:#{location.lineno} in #{location.base_label}"
+  end
+
+  # Alias for source_annotation
```
ndushay, line 17: there is an alias feature in ruby - could we use that?

app/services/dashboard/moab_on_storage_service.rb
```diff
@@ -143,7 +144,7 @@ def num_moab_record_not_ok
 
     def num_preserved_objects
       # used multiple times, so memoize to avoid db queries
-      @num_preserved_objects ||= PreservedObject.count
+      @num_preserved_objects ||= PreservedObject.all.annotate(sa).count
```
ndushay, line 147: is `.all` useful?

spec/spec_helper.rb
ndushay: I kinda disagree - if SlowQueryLogger is code we wrote ... why wouldn't we want tests?

ndushay (changes requested):

README.md
justinlittman: Easy enough do add. Done.

README.md
justinlittman: No, `sa` is meaningless to me too. The assumption is that this would be a cut / paste and that the `annotate` provided the meaning. The use of `sa` was to make this as short as possible so that the query itself wasn't obscured.

That being said, I think `caller` is a good balance between the two considerations, so I'll go with that.

spec/spec_helper.rb
mjgiarlo: In this case we're implementing an interface we don't control, but I think you make a good point that it is worth testing `#call` is doing the right thing (there are ~4 branches in this class that'd be good to test).

README.md
justinlittman: Sidenote: `annotate` is an ActiveRecord method, not mine.

justinlittman: https://edgeapi.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-annotate

app/services/dashboard/audit_service.rb
justinlittman: Great question. I've added this to README:

Note that `.annotate(caller)` should be placed where it receives and returns an `ActiveRecord::Relation`.
```
# Good
ZipPart.where.not(status: 'ok').annotate(caller).count
ZipPart.all.annotate(caller).where.not(status: 'ok').count

# Bad
ZipPart.annotate(caller).where.not(status: 'ok').count
ZipPart.where.not(status: 'ok').count.annotate(caller)
```

app/services/instrumentation_support.rb
justinlittman: Correct.

lib/slow_query_logger.rb
justinlittman: Correct.

lib/slow_query_logger.rb
justinlittman: Done.

spec/spec_helper.rb
justinlittman: We tend to put things in `lib` that are not the sort of things that we test, e.g., rake tasks. In this case, I'm not sure I see the value in testing it -- really the only effective way to test it is to run it; everything else is "make believe" testing.

app/services/dashboard/moab_on_storage_service.rb
justinlittman: Yes, it is required for `.annotate`. This is clarified in README update.

app/services/instrumentation_support.rb
ndushay: since this was a question, is it worth a comment or a test?

config/settings/development.yml
```diff
@@ -23,3 +23,7 @@ zip_endpoints:
     access_key_id: 'foo'
     secret_access_key: 'bar'
 workflow_services_url: 'https://sul-lyberservices-test.stanford.edu/workflow/'
+
+slow_queries:
+  enable: true
```
ndushay, line 28: i wonder if we do want this on by default - how often will we have prescat running dev mode without caring about this, vs. actually caring about this?  but I don't feel strongly.

lib/slow_query_logger.rb
ndushay: again - this question in the review may imply a comment is warranted (or not).

spec/spec_helper.rb
ndushay: I think a class we write could be considered different from a rake task, which is, more or less, a script we write with a DSL.   We don't write test code fro scripts, but we do write test code for classes.   That said, a comment as to why there are no tests might be considered.

app/services/dashboard/audit_service.rb
```diff
@@ -5,6 +5,7 @@ module Dashboard
   # methods pertaining to audit functionality for dashboard
   module AuditService
     include MoabOnStorageService
+    include InstrumentationSupport
```
ndushay, line 8: add comment here and in other places `# for caller method for annotating slow queries` or some such?  It might help us know where "caller" is defined.  Or maybe others would be comfy searching on "caller" to figure out where it's defined?  Oy.

ndushay (approved):
dumb question:  can the `annotate` method infer the caller with a stacktrace?   It's a little hinky to be using an argument that is defined somewhere else in the code.

justinlittman:
@ndushay To answer your question: `annotate()` is a Rails method, so it doesn't infer anything. However, caller uses a stacktrace (`Kernel.caller_locations`) to find the location.

app/services/dashboard/audit_service.rb
mjgiarlo: Happy to ask my IDE to figure this stuff out too

PR #2134, opened 2022-12-20
Tweaks slow query.

justinlittman:
refs #2084

Faster dashboard.

Rails console.


edsu:
Nice, I guess doing the math outside of the SQL lets Postgres use the index?

edsu (approved):

justinlittman:
Yes, and the removal of the unnecessary join.

PR #2138, opened 2023-01-04
Replication error query

justinlittman:
Faster dashboard

Rails / db console

### DIFF
--- a/app/services/dashboard/replication_service.rb
+++ b/app/services/dashboard/replication_service.rb
@@ -48,11 +48,16 @@ module Dashboard
     end
 
     def num_replication_errors
-      @num_replication_errors ||= ZipPart.where.not(status: 'ok').annotate(caller).count
+      # This is faster than querying .where.not(status: 'ok')
+      ZipPart.where(status: replication_error_statuses).annotate(caller).count
+    end
+
+    def replication_error_statuses
+      ZipPart.statuses.keys.excluding('ok')
     end
 
     def zip_parts_ok?
-      num_replication_errors.zero?
+      !ZipPart.where(status: replication_error_statuses).annotate(caller).exists?
     end
 
     def zip_parts_unreplicated?


app/services/dashboard/replication_service.rb
```diff
@@ -48,11 +48,16 @@ def zip_parts_total_size
...
+      ZipPart.where(status: replication_error_statuses).annotate(caller).count
+    end
+
+    def replication_error_statuses
+      ZipPart.statuses.keys.excluding('ok')
     end
 
     def zip_parts_ok?
-      num_replication_errors.zero?
+      !ZipPart.where(status: replication_error_statuses).annotate(caller).exists?
```
ndushay, line 60: I'd be surprised if this was faster than, eg  doing a lookup against a memoized value, OR just ZipParts.ok.count

ndushay (changes requested):

jmartin-sul (approved):

app/services/dashboard/replication_service.rb
jmartin-sul: i assumed the change was the result of some profiling and testing?

did the memoized value actually get re-used?  if so, still seems memoizable.

if the `ok` count is used for this purpose, it'd need to be compared to the total count, and both counts would have to happen in the same query or transaction to keep from possibly spurious drift if replication is happening while the dashboard is being loaded.  and then i'm not sure you'd come out ahead anymore.

as it is in this PR, this is basically a count of `status IN ['unreplicated', 'not_found', 'replicated_checksum_mismatch']`.  which i'd expect to be pretty similar in performance to a count of `status = 'ok'`.  (i may not have exactly the right SQL syntax for that `IN` clause, but hopefully that makes sense

jmartin-sul: this is a bit beyond my postgres performance intuition, but it wouldn't surprise me greatly if counting `status IN ['unreplicated', 'not_found', 'replicated_checksum_mismatch']` was more performant than counting `status != ok`.

justinlittman: This was based on profiling and testing using postgres query plans.

Memoized values don't buy anything due to Rails query caching.

`ZipParts.ok.count` isn't the same; it would need to be checked against the the total ZipParts count.

`status != ok` resulted in a table scan; this approach was able to use the index.

ndushay: ok, whatevs.  

@jmartin-sul it's okay if it's a little off - it's just the dashboard.  Especially if a refresh doesn't take forever.

ndushay (approved):

app/services/dashboard/replication_service.rb
jmartin-sul: > Memoized values don't buy anything due to Rails query caching.

TIL a new bit of rails magic ✨ 

> @jmartin-sul it's okay if it's a little off - it's just the dashboard. Especially if a refresh doesn't take forever.

i figured it's nice to not have the error indicator flap once things are ok, so as not to alarm a first responder who's just peeking in.

PR #2153, opened 2023-01-06
For dashboard, do not load any tabs by default.

justinlittman:
Performance.

QA


ndushay (approved):
Approving with question for consideration:  will it eventually load one of the tabs when it's ready?   (could be whichever is fastest)

justinlittman:
No it won't. No reason to hit the DB loading the wrong tab and slowing down the tab rendering..

justinlittman:
P.S. If in the future we find one of the tabs to be good default based on usage, we can make active.

PR #2155, opened 2023-01-08
audit_results_reporter moved inside reporters namespace

ndushay:

I would look for a class with a name ending in Reporter ... in the reporters folder/namespace.  

Specs 


ndushay:
After discussion this PR is not going in the right direction.

PR #2157, opened 2023-01-09
Refactors audit replications jobs.

### LINKED ISSUE #2151: Refactor MoabReplicationAuditJob
* PartReplicationAuditJob is inlined instead of being separate jobs.
* Audit functionality is moved into a service class so that it can be invoked directly.

justinlittman:
closes #2151

Simpler code.

Refactors from primary job that invokes multiple child jobs to a single job that uses a service class. This reduces complexity and allows an audit to be invoked synchronously as part of manual troubleshooting/remediation.


justinlittman:
Once this is merged, the sidekiq shared configs will need to be cleaned up, removing:
```
  - ["part_audit_aws_s3_east_1", 1]
  - ["part_audit_aws_s3_west_2", 1]
  - ["part_audit_ibm_us_south", 1]
 ```

File: spec/services/audit/replication_spec.rb [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+  before do
+    # allow(Audit::ReplicationSupport).to receive(:logger).and_return(logger)
+    # creation of MoabRecord triggers archive zip creation, as archive zips are created from moabs
+    create(:moab_record, preserved_object: preserved_object, version: preserved_object.current_version)
+    allow(Audit::ReplicationSupport).to receive(:check_child_zip_part_attributes).and_return(false)
+    allow(audit_class1).to receive(:check_replicated_zipped_moab_version)
+    allow(audit_class2).to receive(:check_replicated_zipped_moab_version)
+  end
+
+  describe '#results' do
```
ndushay, line 28: I would expect to see somewhere (and maybe it's not here???) that the results are as expected for different cases:  when the result is at the endpoint;  when the result is not at the endpoint;   when the result is at the endpoint but doesn't match.

The naming of "services/audit/replication" has me expecting this to be here.

app/services/audit/replication.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module Audit
+  # Audit the replication of a single PreservedObject.
```
ndushay, line 4: ... "which may have one or more ZipParts each replicating to multiple configured ZipEndpoints" ?

Now that You are among the chosen few that "get it" ... please try to keep "beginners mind" ... in mind.

ndushay (changes requested):
This looks good.  I wonder if we're missing some test cases, and I'm also wondering if we can do any better with some of the naming/namespaces.  Happy to discuss

spec/services/audit/replication_spec.rb
justinlittman: All changes to the audit results are performed by `Audit::ReplicationSupport.check_child_zip_part_attributes` and `zip_endpoint.audit_class`. That these are invoked is tested here; the actions that they perform are not.

spec/jobs/moab_replication_audit_job_spec.rb
```diff
@@ -31,28 +41,13 @@
...
+      it 'creates missing zipped moab versions and logs a warning' do
         expect(preserved_object).to receive(:create_zipped_moab_versions!).and_call_original
         expect(logger).to receive(:warn)
           .with(/backfilled 4 ZippedMoabVersions: 1 to aws_s3_west_2; 1 to ibm_us_south; 2 to aws_s3_west_2; 2 to ibm_us_south/)
-        expect(PartReplicationAuditJob).not_to receive(:perform_later)
-        expect { job.perform(preserved_object) }.not_to change { preserved_object.reload.last_archive_audit }
-      end
-    end
-
-    it 'updates last_archive_audit timestamp' do
```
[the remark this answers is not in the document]
justinlittman: It is important and tested here: https://github.com/sul-dlss/preservation_catalog/pull/2157/files#diff-efde1fffe70d6ebcd1d624c267bdc7801ac7e8dfa6535eeea6d795bd0f20f4c3R35

spec/jobs/moab_replication_audit_job_spec.rb
```diff
@@ -10,15 +10,25 @@
...
     allow(Settings.replication).to receive(:audit_should_backfill).and_return(true) # enable for tests
+    allow(Audit::Replication).to receive(:results).and_return([])
     # creation of MoabRecord triggers archive zip creation, as archive zips are created from moabs
     create(:moab_record, preserved_object: preserved_object, version: preserved_object.current_version)
   end
 
   describe '#perform' do
+    let(:audit_results) { instance_double(AuditResults) }
+
     context 'when there are no zipped moab versions to backfill' do
```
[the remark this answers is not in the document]
justinlittman: This code doesn't do checksum matching; I'm assuming it is performed by `Audit::ReplicationSupport.check_child_zip_part_attributes`  or `zip_endpoint.audit_class`.

app/services/audit/replication.rb
justinlittman: Agree that this makes clearer. Changing.

spec/jobs/moab_replication_audit_job_spec.rb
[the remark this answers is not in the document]
justinlittman: P.S. Adding a code comment to Audit::Replication class to make this delegation clearer.

ndushay (approved):

Issue #2162, opened 2023-01-09
Create replication flow wiki page


justinlittman:
Remove the replication flow section of the dashboard and create a wiki page. This new page should be added to the "About PresCat" section of the sidebar and linked to from the dashboard.

https://preservation-catalog-web-prod-01.stanford.edu/dashboard/replication_flow

ndushay:
FYI, this is how it currently looks in dashboard;  it doesnt have to be a table in the wiki -- but the information on what triggers each step, etc, should be skimmablew.

![image](https://user-images.githubusercontent.com/96775/211415147-3e29ca09-f571-4408-a220-e49f439161f0.png)

ndushay:
@justinlittman @jmartin-sul I sent you both slack messages re: my hope to replace the OLD wiki page https://github.com/sul-dlss/preservation_catalog/wiki/Replication-flow with my NEW wiki page  https://github.com/sul-dlss/preservation_catalog/wiki/Replication-Flow-WIP.   Let me know what you think.

justinlittman:
Completely support

jmartin-sul:
also support

PR #2169, opened 2023-01-11
Replication code - documentation tweaks

ndushay:

To make it easier to maintain this code.

Part of #2162

Specs.
Integration test on stage.


app/jobs/replication/delivery_dispatcher_job.rb
```diff
@@ -26,15 +27,9 @@ class DeliveryDispatcherJob < Replication::ZipPartJobBase
     # @param [String] druid
     # @param [Integer] version
     # @param [String] part_s3_key, e.g. 'ab/123/cd/4567/ab123cd4567.v0001.z03'
-    # @param [Hash<Symbol => String>] metadata Zip info
-    # @option metadata [Integer] :parts_count
-    # @option metadata [Integer] :size
-    # @option metadata [String] :checksum_md5
-    # @option metadata [String] :suffix
-    # @option metadata [String] :zip_cmd
-    # @option metadata [String] :zip_version
```
ndushay, line 35: these are never used, and there is no `options` argument, so ... removing cruft.

app/jobs/replication/results_recorder_job.rb
```diff
@@ -27,17 +25,12 @@ class ResultsRecorderJob < ApplicationJob
...
       part.ok!
 
-      # log to event service if this part upload is the last one for the endpoint
+      # log to event service if all ZipParts are replicated for this endpoint
       create_zmv_replicated_event(druid) if zmv.reload.all_parts_replicated?
-
-      # only publish result if all of the parts replicated for all zip_endpoints
-      return unless zmvs.reload.all?(&:all_parts_replicated?)
-
-      publish_result(message(druid, version).to_json)
```
ndushay, line 40: there is no consumer for these, and hasn't been for 4 years, so removing this code.

app/jobs/replication/delivery_dispatcher_job.rb
mjgiarlo: @ndushay Doesn't this unpack the `metadata` argument? Or is this better documented elsewhere, since all this job does with `metadata` is pass it on elsewhere?

app/jobs/replication/ibm_south_delivery_job.rb
```diff
@@ -2,11 +2,13 @@
 
 module Replication
   # @see Replication::IbmProvider for how S3 credentials and bucket are configured
-  # @note this class name appears in the configuration for the endpoints for which it delivers content.
+  # @note this class name appears in config files for the endpoints for which it delivers content.
   #   Please update the configs for the various environments if it's renamed or moved.
   class IbmSouthDeliveryJob < Replication::AbstractDeliveryJob
     queue_as :ibm_us_south_delivery
 
+    # perform method is defined in AbstractDeliveryJob
```
mjgiarlo, line 10: I'm 50/50 on whether this comment is more helpful than it is noisy. Could go either way.

mjgiarlo (approved):
Have a couple minor questions but approving for now

app/jobs/replication/delivery_dispatcher_job.rb
ndushay: I may have misinterpreted the `@option` yard command -- as I have seen it in the past as @options.  I can leave in.

app/jobs/replication/ibm_south_delivery_job.rb
ndushay: agree that this comment isnt necessary, but personally I found it weird that there was no perform method in these classes and had to go look in the superclass.   I could also go either way.

app/jobs/replication/delivery_dispatcher_job.rb
ndushay: indeed, I am in the wrong. https://rubydoc.info/gems/yard/file/docs/Tags.md#option.  Leaving in.

app/jobs/replication/ibm_south_delivery_job.rb
mjgiarlo: K, let's leave them in then

mjgiarlo (approved):

PR #2184, opened 2023-01-24
Add an option to skip verification of expiration date to prescat rake…

### LINKED ISSUE #2171: Allow Replication::FailureRemediator to override zip guard
Currently, when pruning DB records, `Replication::FailureRemediator` limits zipped moab versions based on the zip cache expiration date. The intent of this guard is to avoid pruning records for zipped moab versions that might still be in flight. (With Sidekiq retries this is less likely, but still a reasonable guard.)

There are cases in which you might want to prune DB records for zipped moab versions that have not yet expired. To support this:
1. Allow this guard to be bypassed / forced, such that all zipped moab versions are checked for pruning. (See L55).
2. Make this force option available from the `prescat:prune_failed_replication` rake task.
3. Document in the https://github.com/sul-dlss/preservation_catalog/wiki/Replication-errors

aaron-collier:
… task

Fixes #2171  to allow skipping expiration verification from cron (defaulting to existing behavior)


app/services/replication/failure_remediator.rb
```diff
@@ -4,13 +4,14 @@ module Replication
   # Remediates catalog entries, e.g., prunes failed replication records
   # Currently this is only called from rake.
   class FailureRemediator
-    def self.prune_replication_failures(druid:, version:)
-      new(druid: druid, version: version).prune_replication_failures
+    def self.prune_replication_failures(druid:, version:, verify_expiration:)
```
justinlittman, line 7: Would you consider making this default to `true`? (Current behavior).

justinlittman (approved):
Approve pending consideration of request.

app/services/replication/failure_remediator.rb
ndushay: defaulting to current behavior would be megaspiffy!

aaron-collier: Sure, is this called from anywhere else but the rake task? It is currently defaulted there, so will always get sent true from the task. I ran into an issue or two testing the default here, but can work through it. Thoughts @justinlittman ?

justinlittman: I suspect that we may want to use this elsewhere in the future (e.g., for remediation via a web interface) so I would have a slight preference for providing a default.

PR #2188, opened 2023-01-25
Move AuditResults model to service class

### LINKED ISSUE #2123: AuditResult model ... should it be clearer it's not an ActiveRecord model?
AuditResults is a model class that is NOT an ActiveRecord model.   Should we put it in an "Audit" namespace as we did with the replication models?   Should it be a service??

mjgiarlo:

Fixes #2123

This commit moves what was the AuditResults model---a PORO, not an ActiveRecord model---into `app/services/`, which we admittedly have low and unclear standards around (what is a service object? etc.). But we will be less surprised by finding this class among other junk-drawer-type classes than we will in the directory where its siblings are all ActiveRecord-based models. This is a purely mechanical refactoring.

CI


app/services/audit/results.rb
```diff
@@ -0,0 +1,166 @@
+# frozen_string_literal: true
+
+module Audit
+  # Audit::Results allows the correct granularity of auditing information to be modeled in various contexts.
+  # All results are kept in the result_array attribute, which is returned by the report_results method.
+  #   result_array = [result1, result2]
+  #   result1 = {response_code => msg}
+  #   result2 = {response_code => msg}
+  class Results
```
mjgiarlo, line 9: This entire PR follows from:

```shell
$ git mv app/models/audit_results.rb app/services/audit/results.rb
```

justinlittman:
We have other non-AR models under `models/`. So now some non-AR models go under `models/` and others go under `services/`? Not seeing how this is an improvement.

ndushay (commented):
There are 2 non-AR models under models/replication;  I wonder if they should also move to services, or if this should move to models/audit.

If they should also move to services, then we should make a ticket for that.

If there's a clearer way to have non-AR models within the model directory, I'd be up for that.

mjgiarlo:
@ndushay 💬 
> There are 2 non-AR models under models/replication; I wonder if they should also move to services, or if this should move to models/audit.

I implemented one of the suggested directions in #2123 with a rationale. I don't know that putting this elsewhere is going to be any better.

> If they should also move to services, then we should make a ticket for that.

No opinion on that.
 
> If there's a clearer way to have non-AR models within the model directory, I'd be up for that.

🤷🏻‍♂️

ndushay (approved):

PR #2201, opened 2023-02-08
move Audit::Results::DB_OBJ_ALREADY_EXISTS from audit workflow report codes to honeybadger report codes

jmartin-sul:

this can happen when pres robots alerts pres cat to a moab version that's already been cataloged, as might happen when the update-catalog workflow step sits after having failed, and the moab version is detected by e.g. moab-to-catalog audit before the step is retried.

it's unusual, and worth a glance, but shouldn't wedge preservation or accessioning workflows.

will take this out of draft if/when https://github.com/sul-dlss/preservation_robots/pull/469 is merged.

once this and the pres robots PR are merged and deployed, we should (hopefully) be able to easily clear the druids wedged at the `update-catalog` step atm, by using argo to reset the workflow step en masse, instead of manually marking the step complete for individual druids (whether via argo show page or WFS rails console/rake task)

### DIFF
--- a/app/services/audit_reporters/audit_workflow_reporter.rb
+++ b/app/services/audit_reporters/audit_workflow_reporter.rb
@@ -14,7 +14,6 @@ module AuditReporters
     def handled_merge_codes
       [
         Audit::Results::ACTUAL_VERS_LT_DB_OBJ,
-        Audit::Results::DB_OBJ_ALREADY_EXISTS,
         Audit::Results::DB_UPDATE_FAILED,
         Audit::Results::DB_VERSIONS_DISAGREE,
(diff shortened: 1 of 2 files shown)


jmartin-sul:
actually, taking this out of draft -- no harm in merging it before the pres robots PR (or the pres client PR that blocks it)

jmartin-sul:
also note: the much more serious `ACTUAL_VERS_LT_DB_OBJ` (where the incoming moab version on disk appears to be lower than the version pres cat last saw) is still handled by reporting to `preservationAuditWF`, as that does warrant wedging accessioning for such a druid, till things can be manually investigated (and fixed if need be).

PR #2215, opened 2023-03-27
Upgrade to Sidekiq7.

justinlittman:
Latest, greatest.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,8 +1,8 @@
 GEM
   remote: https://gems.contribsys.com/
   specs:
-    sidekiq-pro (5.5.8)
-      sidekiq (~> 6.0, >= 6.5.6)
+    sidekiq-pro (7.0.8)
+      sidekiq (>= 7.0.0, < 8)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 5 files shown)


Gemfile.lock
```diff
@@ -385,10 +388,11 @@ GEM
...
       activesupport (>= 5.2.0)
-    sidekiq (6.5.8)
-      connection_pool (>= 2.2.5, < 3)
-      rack (~> 2.0)
-      redis (>= 4.5.0, < 5)
+    sidekiq (7.0.7)
+      concurrent-ruby (< 2)
+      connection_pool (>= 2.3.0)
+      rack (>= 2.2.4)
+      redis-client (>= 0.11.0)
```
edsu, line 395: Is it weird we are using redis-client and redis? It looks like sidekiq may use redis-client now?

edsu (approved):

Gemfile.lock
jmartin-sul: i was wondering the same thing, but it looks like it allows us to keep the `UniqueJob` concern without having to change much (since previously the redis connection was obtained via `Sidekiq.redis`): https://github.com/sul-dlss/preservation_catalog/pull/2215/files#diff-1a156c36692545c253a32a57b532d04830c70b502d3511076177406a2317929c

jmartin-sul (approved):

Gemfile.lock
justinlittman: Yes, Sidekiq7 switches to using redis-client. However, we (and Rails) use redis. So in many apps we need both.

PR #2216, opened 2023-03-31
[HOLD] okcomputer: Sidekiq worker count check

jmartin-sul:

fail the health check if a worker box has an unexpected number of sidekiq worker processes or threads


jmartin-sul:
rebased on latest `main`, and deployed the rebased branch to stage for some quick testing to make sure all was still working (happy path works, check fails if i manually kill some sidekiq processes).

justinlittman (commented):
Curious why you're checking thread counts (instead of just process counts)?

jmartin-sul:
> Curious why you're checking thread counts (instead of just process counts)?

some combination of... it was pretty easy to get at that info (both the actual and expected thread counts), it seemed potentially useful at a glance for infra FR and ops folks, and we've seen occasional sidekiq worker issues but not enough to have a great idea of what the most common misbehavior modes are.

so i figured (a little more) info == better for triaging alerts.  i also was thinking that the sidekiq API is pretty stable and pretty well documented, and this is unlikely to need changes in the near future, so even though thread counts adds a bit of complexity over just doing process counts, it didn't seem like complexity we'd have to pay much for going forward?  and then if e.g. sidekiq 8 or 9 breaks this, we can simplify if fixing doesn't seem worth the effort?

justinlittman:
My concern is that this is just making work for our future selves to troubleshoot. Do we have any indication that there is any value in monitoring this or that Sidekiq does not repair any damaged threads? Do we know this won't set off false alarms? How will our future selves know why we were monitoring threads in the first place if it breaks?

Might be worth getting some additional opinions on this.

jmartin-sul:
backburnered this since it was reviewed.  haven't seen any issues with sidekiq processes running the wrong number of worker threads compared to what's configured in the year or so this change has hung out.  we did run into [an issue in hydra_etd](https://github.com/sul-dlss/hydra_etd/issues/1687) that was along the lines of what justin worried about above.  after discussing the hydra_etd issue in slack with aaron and laura, i was convinced to simplify the hydra_etd check to just look for expected process count (and to give a bit of helpful guidance about when to worry if there's an alert for too many workers).

next unscheduled week, i'll circle back to this PR and simplify it in the same way as https://github.com/sul-dlss/hydra_etd/pull/1689

PR #2222, opened 2023-04-17
Do not bundle exec when using runner in whenever.

justinlittman:
Executing with the default runner was failing.

```
0 20 * * * /bin/bash -l -c 'cd /Users/jlittman/data/sdr/preservation_catalog && bin/rails runner -e production "PreservedObject.order(last_archive_audit: :asc).limit(PreservedObject.daily_check_count).find_each(&:audit_moab_version_replication!)" >> /dev/null 2>> log/c2a-err.log'

0 22 * * * /bin/bash -l -c 'cd /Users/jlittman/data/sdr/preservation_catalog && bin/rails runner -e production "MoabRecord.order(last_checksum_validation: :asc).limit(MoabRecord.daily_check_count).find_each(&:validate_checksums!)" >> /dev/null 2>> log/cv-err.log'``
```

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -13,6 +13,8 @@ Config.load_and_set_settings(Config.setting_files('config', 'production'))
 # If changing the schedule of one of these jobs, also update at https://app.honeybadger.io/projects/54415/check_ins
 job_type :rake_rb, 'cd :path && :environment_variable=:environment bundle exec rake --silent ":task" :output && curl --silent https://api.honeybadger.io/v1/check_in/:check_in'
 job_type :runner_hb, 'cd :path && bin/rails runner -e :environment ":task" :output && curl --silent https://api.honeybadger.io/v1/check_in/:check_in'
+# Overriding default runner job_type to remove invoking bundle exec.
+job_type :runner, 'cd :path && bin/rails runner -e :environment ":task" :output'
 
 # 11 am on the 1st of every month
 # If changing schedule, also change for HB checkin


ndushay (commented):
are you intentionally overriding whenever's runner?  If so, a comment explaining why would be a good idea.

ndushay (changes requested):
something along the lines of "#overwriting whenever's runner so we don't bundle exec

lwrubel:
Just curious: how was `bundle exec` was getting added to begin with? The [whenever docs](https://github.com/javan/whenever#define-your-own-job-types) docs say the default `runner` is defined as:

`job_type :runner,  "cd :path && bin/rails runner -e :environment ':task' :output"`

ndushay:
> Just curious: how was `bundle exec` was getting added to begin with? The [whenever docs](https://github.com/javan/whenever#define-your-own-job-types) docs say the default `runner` is defined as:
> 
> `job_type :runner, "cd :path && bin/rails runner -e :environment ':task' :output"`

Look in https://github.com/javan/whenever/blob/main/lib/whenever/setup.rb -- there's a :bundle_command in there.

mjgiarlo:
@justinlittman Where were you seeing these failures? In your inbox or somewhere else? I'm just curious. Happy for this to move forward otherwise.

justinlittman:
@mjgiarlo They were in the logs on the server.

config/schedule.rb
```diff
@@ -13,6 +13,8 @@
 # If changing the schedule of one of these jobs, also update at https://app.honeybadger.io/projects/54415/check_ins
 job_type :rake_rb, 'cd :path && :environment_variable=:environment bundle exec rake --silent ":task" :output && curl --silent https://api.honeybadger.io/v1/check_in/:check_in'
 job_type :runner_hb, 'cd :path && bin/rails runner -e :environment ":task" :output && curl --silent https://api.honeybadger.io/v1/check_in/:check_in'
+# Overriding default runner job_type to remove invoking bundle exec.
+job_type :runner, 'cd :path && bin/rails runner -e :environment ":task" :output'
```
mjgiarlo, line 17: Perusing https://github.com/javan/whenever/issues/732 and the whenever code, the other option would be to use:

```ruby
set :runner_command, 'rails runner'
```

and let whenever handle the rest. I'm not sure that's better than redefined the `:runner` job type, though? 🤷🏻‍♂️

justinlittman: I'd suggest that the more explicit specification makes it clear what the `runner` is invoking (without having to look at whenever code).

mjgiarlo (approved):

PR #2239, opened 2023-06-01
add link to replication errors wiki in dashboard

ndushay:

Because I think the wiki page is now stable enough to warrant linking from the part of the dashboard where we would most like to see it.


justinlittman (approved):
Is anyone really not going to know what we have a wiki?

ndushay:
> Is anyone really not going to know what we have a wiki?

It saves me, and perhaps others, some clicking around after looking at the dashboard

PR #2263, opened 2023-08-31
Remove unnecessary json suffix from endpoint.

justinlittman:
# Why was this change made? 🤔

⚡ ⚠ If this change has cross service impact, or if it changes code used internally for cloud replication, **_run [integration test preassembly_image_accessioning_spec.rb](https://github.com/sul-dlss/infrastructure-integration-test/blob/main/spec/features/preassembly_image_accessioning_spec.rb) against stage as it tests preservation (including cloud replication)_**, and/or test in stage environment, in addition to specs. The main classes relevant to replication are `ZipmakerJob`, `DeliveryDispatcherJob`, `*DeliveryJob`, `ResultsRecorderJob`, and `DruidVersionZip`; [see here for overview diagram of replication pipeline](https://github.com/sul-dlss/preservation_catalog/wiki/Replication-Flow).⚡

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -143,7 +143,7 @@ paths:
           example: 'druid:bc123df4567'
           schema:
             $ref: '#/components/schemas/Druid'
-  /v1/objects/{id}/checksum.json:
+  /v1/objects/{id}/checksum:
     get:
       tags:
         - objects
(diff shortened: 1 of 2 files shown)


edsu (approved):

openapi.yml
```diff
@@ -143,7 +143,7 @@ paths:
           example: 'druid:bc123df4567'
           schema:
             $ref: '#/components/schemas/Druid'
-  /v1/objects/{id}/checksum.json:
+  /v1/objects/{id}/checksum:
```
edsu, line 146: It looks like there are other endpoints that end in .json ?

justinlittman: Took a closer look. I think this is the only one that ends in `.json`. However, there is some odd use of `.json` appended to druids. Not sure why it is done that way.

PR #2310, opened 2024-03-01
A tool to identify which GIS items have a data.zip

edsu:
# Why was this change made? 🤔

This is a temporary utility to help in GIS remediation. Currently it is only checking to see if a `data.zip` can be found in preservation. For context see: https://github.com/sul-dlss/gis-robot-suite/issues/829

The utility will be modified to extract files from the data.zip and write them to an external location. Once it has been used, and remediation is complete it can be deleted.

It was run in prescat stage.

### DIFF
--- a/bin/gis_remediation
+++ b/bin/gis_remediation
@@ -1,0 +1,38 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# This is a temporary utility created to help remediate accessioned GIS SDR 
+# items. It was adapted from prescat's create_bag utility.
+# See: https://github.com/sul-dlss/gis-robot-suite/issues/813
+
+require 'rubygems'
+require 'bundler/setup'
+require 'moab/stanford'
+require 'yaml'
+
+# setup our preservation storage environment
+
+settings = YAML.load_file(File.join(__dir__, '..', 'config', 'settings', 'production.yml'))
+
+Moab::Config.configure do
+  storage_roots settings['storage_root_map']['default'].values.sort
+  storage_trunk 'sdr2objects'
+  deposit_trunk 'deposit'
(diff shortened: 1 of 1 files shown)


thatbudakguy (approved):

bin/gis_remediation
```diff
@@ -0,0 +1,38 @@
...
+File.readlines(ARGV[0], chomp: true).each do |druid|
+  storage_object = Stanford::StorageServices.find_storage_object(druid)
+
+  object_version = storage_object.current_version
+
+  begin
+    data_zip = object_version.find_filepath('content', 'data.zip')
+    puts "#{druid} #{data_zip}"
+  rescue Moab::FileNotFoundException
+    puts "#{druid} data.zip not found"
```
thatbudakguy, line 36: not sure the ultimate destination for this info, but you could send the `data.zip` paths to stdout and the "not found" messages to stderr, so that you could redirect them to separate files if desired. just an idea

peetucket:
Just to note that in Rails apps like this you can also take advantage of the "rails runner" command to write a class and then execute in the context of the Rails environment of choice (and it will have the full context of the app,  including any settings, libraries, gems etc.).  Lets you skip some of that boilerplate setup code in a one off script like this

https://guides.rubyonrails.org/command_line.html#bin-rails-runner

bin/gis_remediation
edsu: Yeah, good thought. For now I just grepped the output for `not found` lines. But I'll keep this in mind for the next iteration which is going to be copying out of the data.zip.

edsu:
Thanks for that doc link @peetucket -- I'll give that a try in the next iteration of this.

Issue #830, opened 2018-06-05
Replicate some stuff from stage (using test credentials)


atz:
On stage, for some known druid/versions, basically do:
```ruby
ZipMaker.perform_later(druid, version)
```
NOTE: you will need to establish the TEST credentials for your console process, as documented [in the Wiki](https://github.com/sul-dlss/preservation_catalog/wiki/AWS-credentials,-S3-configuration).  Test S3 bucket objects are automatically culled on the backend, so long-term consistency should not be expected.

atz:
Blocked by various bugs (Zipmaker and resque related)

atz:
Blocked by #855

atz:
This has been accomplished

PR #2312, opened 2024-03-07
Add a utility to export GIS items from preservation

edsu:
# Why was this change made? 🤔

The `bin/gis_exporter.rb` utility reads a file of druids and retrieves the latest version from preservation, and exports the files to a druid directory. Care is taken to read original files from the GIS item's `data.zip`, and to ignore metadata files that are created as part of the gis-robot-suite.

The PR includes a lot of files but the majority are just new files for testing in `spec/fixtures/storage_root_gis`. Once the remediation is complete all these files can be completey removed as well as:

* bin/gis_exporter.rb
* app/services/gis_exporter.rb
* spec/services/gis_exporter_spec.rb

References #830

specs and running bin/gis_exporter on stage


edsu:
After discussing this with @andrewjbtw we decided to extract this functionality so that it operates on bag output from SDR-GET instead. This way it doesn't clutter up preservation_catalog, and we can rely on SDR-GET which uses symlinks to make files available (no impact on storage), and has seen a lot of use.

edsu:
Instead of merging here this code was moved over to https://github.com/sul-dlss-labs/gis_exporter

PR #2360, opened 2024-10-15
Upgrade to Rails 7.2

aaron-collier:
# Why was this change made? 🤔

Tested in stage against integration tests

Integration tests.


config/application.rb
```diff
@@ -4,9 +4,10 @@
 
 # Select only parts we need from rails/all
 require 'rails'
-require 'active_record/railtie'
 require 'action_controller/railtie'
 require 'active_job/railtie'
+require 'active_record/railtie'
+require 'action_cable/engine'
```
aaron-collier, line 10: This was require to avoid https://github.com/sul-dlss/preservation_catalog/pull/2360/files#diff-053aa3ea46cb7c6a649e0d0fc592ce729b66e1d44be475516fd58857ba695862L6 raising an error in CI (does not get raised when running the tests locally).

config/environments/production.rb
jmartin-sul, line 1: definitely wouldn't hold up the PR over this, especially since i don't think any of the controllers use `only`/`except` options for `before_action` (we only seem to have one `before_action` for checking the auth token).  but i wonder whether we should add `config.action_controller.raise_on_missing_callback_actions = true` to the prod env config.  seems like a good safeguard there too maybe?

jmartin-sul (approved):
all looks fine to me, thanks!  i tested the `clear_active_connections!` tweak by queuing the 25 largest moabs on stage for checksum validation.  all went through successfully, no HB alerts about a closed connection at the end of the job, and a few of those jobs ran for at least a half an hour (largest couple objects were about 1 TB each), if not over an hour, according to my glances at sidekiq console.  and that's the sort of activity where we saw PG connection errors pop up before that `clear_active_connections!` call was added, so that seems like a decent test.  i also saw the expected logs getting generated, which seems like good confirmation the `BroadcastLogger` change is working.

one very small question, but i'm going to merge, and if adding that bit of config to the production env seems like a good idea, i'm happy to put up a tiny follow up PR.

PR #2385, opened 2025-02-17
Update dependencies

sul-devops-team:


jmartin-sul (approved):
Approved by automated merge script

.rubocop.yml
```diff
@@ -2,9 +2,11 @@
 require:
   - rubocop-capybara
   - rubocop-factory_bot
+  - rubocop-rspec_rails
+
+plugins:
```
jmartin-sul, line 7: gets rid of a deprecation warning that was added in 1.72.x.  i confirmed that this works by making this change first, and continuing to get rubocop rails errors before making the fixes rubocop suggested (see this PR's commit history, where i made those changes as separate commits in order to better show that).

app/models/application_record.rb
```diff
@@ -21,7 +21,5 @@ def self.provlog
     @provlog ||= Logger.new Rails.root.join('log', 'prov.log')
   end
 
-  def provlog
-    self.class.provlog
-  end
+  delegate :provlog, to: self
```
jmartin-sul, line 24: 1. this feature is not enabled at the moment (see shared_configs for each env, as well as base settings here -- the setting is explicitly false in all of those places).
2. i tested that this would behave in the same way by starting and seeding my local instance, and running:
```ruby
preservation-catalog(dev)> MoabStorageRoot.take.provlog == ApplicationRecord.provlog
  MoabStorageRoot Load (2.0ms)  SELECT "moab_storage_roots".* FROM "moab_storage_roots" LIMIT $1  [["LIMIT", 1]]
=> true
preservation-catalog(dev)> 
```

PR #2394, opened 2025-03-31
Attempt to fix flaky tests

peetucket:
# Why was this change made? 🤔

The `spec/jobs/application_job_spec.rb` specs are flaky and now fail repeatedly in CI (though pass locally).  This is an attempt to address it.

### DIFF
--- a/spec/jobs/application_job_spec.rb
+++ b/spec/jobs/application_job_spec.rb
@@ -1,20 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
-
-# A very simple job class that extends ApplicationJob to allow testing of basic queue locking behavior
-class RegularParameterJob < ApplicationJob
-  include UniqueJob
-  def self.lock_timeout
-    1
-  end
-
-  def perform(param_a, param_b, should_raise)
-    Rails.logger.info("simulate a running job to better test queue locking: (#{param_a}, #{param_b}, #{should_raise})")
-
-    raise 'oops' if should_raise
-  end
-end
+require 'regular_parameter_job'
 
 RSpec.describe ApplicationJob do
   include ActiveJob::TestHelper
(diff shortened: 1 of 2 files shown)


jmartin-sul (commented):
I slightly prefer the current code organization, since this class is only used in the one spec file where it's declared in line.  But I think the re-org in this PR is still totally readable.  I think I'm just hesitant to merge, because I'm wondering what the rationale is for why this would improve test reliability, since the error messages in the test failures don't seem to be about inability to find the class, or similar.  And since I have a long running suspicion that flaky tests in CircleCI tend to pass more reliably when the whole workflow is re-run from the start, which this PR would do.

But also, I have no idea why "re-run workflow from start" would behave differently from "re-run workflow from failed", besides some vague handwaving at caching or something (but anecdotally it has seemed to work more reliably in my experience with dep updates).

I did notice just now that the retries on the dep update PR were from failed, so i just retriggerd from start, we'll see what happens...

justinlittman:
Or we could decide that testing unique jobs is not worth it / too difficult and just trust that they work.

jmartin-sul:
> > Or we could decide that testing unique jobs is not worth it / too difficult and just trust that they work.
> 
> yeah, it didn't strike me as a super important test when i looked at it

I would prefer to keep it, since in the distant past [the locking behavior did break due to drift between dependencies](https://github.com/sul-dlss/preservation_catalog/issues/1726).  Though in that case, the problem was with [an unmaintained unique jobs dependency](https://github.com/defunkt/resque-lock) drifting from the Resque dependency (or something 🤷).  And now of course this is implemented with just the plain Redis dependency.  So, maybe there's less room for breakage now (though I'm not sure we'd easily notice this mechanism breaking otherwise, and we did enhance these tests as part of the in-lining of the resque-lock code).

Anyway, that's my vote.  Not a thing I test I'd fight hard to keep, if the consensus is to remove.

mjgiarlo:
@peetucket @justinlittman @jmartin-sul Here's a different approach to less flappy job specs: https://github.com/sul-dlss/preservation_catalog/pull/2400

PR #2398, opened 2025-04-01
Extract directory checksum validator so we can audit uncataloged moabs

### LINKED ISSUE #2396: create an easy way to run checksum validation on an arbitrary Moab directory
If there's not currently a documented way to run checksum validation on a Moab that's not already housed on a "storage root" and tracked in the DB, create one (I couldn't find one when doing a quick search of the front page of the wiki, or a quick grep on my local git clone of it).

This could be wiki entry documenting a script to run via rails console or rake; or it could be a new utility class and/or rake task.  A rake task with a bit of usage description as a thin wrapper for a utility class with tests would be ideal, but it also might be overkill.

The building blocks are in `ChecksumValidator`, and these calls in `#validate_checksums`: https://github.com/sul-dlss/preservation_catalog/blob/be7679ff0a0da5dd7a341afcdc62e393680e78df/app/services/audit/checksum_validator.rb#L21-L23

_might_ be close to usable for the purposes of uncataloged Moab fixity checking, but there might also be some DB tendrils lodged below those calls, the author of this ticket is unsure at the time of its creation.

This would be useful because it'll help spot checksum validate some randomly chosen Moabs from after replicating them to the new GCP endpoint.

part of https://github.com/sul-dlss/preservation_catalog/issues/2237

mjgiarlo:
# Why was this change made?

Fixes #2396

# How was this change tested?

Unit tests (so far)


app/services/audit/checksum_validator.rb
```diff
@@ -8,6 +8,8 @@ class ChecksumValidator
     delegate :moab_storage_root, :preserved_object, to: :moab_record
     delegate :storage_location, to: :moab_storage_root
     delegate :druid, to: :preserved_object
+    delegate :moab_on_storage_absent?, :latest_moab_storage_object_version, :object_dir, :moab_on_storage_validator,
+             to: :directory_checksum_validator
```
mjgiarlo, line 12: Move storage/directory concerns into newly extracted class to validate moab storage objects whether they are cataloged or not.

app/services/audit/directory_checksum_validator.rb
```diff
@@ -0,0 +1,165 @@
+# frozen_string_literal: true
+
+module Audit
+  # Service for validating Moab checksums in a directory. This class's primary collaborator
+  # is Audit::ChecksumValidator, which depends on Audit::DirectoryChecksumValidator to
+  # validate manifest inventories on disk and update the passed-in `results` object. It is
+  # also invoked by the `prescat:audit:validate_uncataloged` rake task which expects the
+  # class to emit validation results and errors to STDOUT.
```
mjgiarlo, line 8: Try to be clear about the intended usage of this class and why it behaves differently sometimes.

app/services/audit/checksum_validation_orchestrator.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+module Audit
+  # Service for validating Moab checksums on storage, updating the MoabRecord db record, and reporting results.
+  class ChecksumValidationOrchestrator
```
mjgiarlo, line 5: I don't know that I love this name, but with most of the actual validation moved out of this class, it seemed like a good time to rename it. I also considered `ChecksumValidationService` 🤷🏻‍♂️ Suggestions?

app/services/audit/checksum_validator.rb
```diff
@@ -1,296 +1,82 @@
 # frozen_string_literal: true
 
 module Audit
-  # Service for validating Moab checksums on storage, updating the MoabRecord db record, and reporting results.
+  # Service for validating Moab checksums in a directory. This class's primary collaborator
+  # is Audit::ChecksumValidationOrchestrator, which depends on Audit::ChecksumValidator to
+  # validate manifest inventories on disk and update the passed-in `results` object. It is
+  # also invoked by the `prescat:audit:validate_uncataloged` rake task which expects the
+  # class to emit validation results and errors to STDOUT.
   class ChecksumValidator
```
mjgiarlo, line 9: This class no longer depends on the app DB. It *only* handles validating a Moab on disk.

app/services/audit/manifest_inventory_validator.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+module Audit
+  # Validates files on storage against the manifest inventory
+  class ManifestInventoryValidator
```
mjgiarlo, line 5: Broke both validators out into their own files.

jmartin-sul (changes requested):
love this!  this was actually sort of what i was hoping we'd get to with the initial implementation for the checksum validation job, something where the fixity checking isn't so tightly coupled with the DB status updating.

i did run into what seemed like a small bug when testing the rake task, for which i put up https://github.com/sul-dlss/preservation_catalog/pull/2399

i sort of wanted to add a simple test for that, just a call to `Audit::ChecksumValidator#validate` using some of the existing fixture objects i tested #2399 with manually on my laptop, but i need to step away for a moment, and i didn't want to hold this up if you wanted to push it forward.  will try to add said tests later this evening or tomorrow if you don't beat me to it, or will file a ticket for follow on work, if neither of us adds tests in these PRs.

app/services/audit/checksum_validation_orchestrator.rb
mjgiarlo: @jmartin-sul I do like the idea of this class's name being more explicitly tied to its responsibilities. I actually quite like the name `CatalogedMoabFixityChecker` but I'd be hesitant to use it here for two reasons:

1. I sense that checking fixities in prescat is (somewhat) consistently referred to as either "auditing" or "checksum validation" and leaning into "fixity checking" might muddy the domain model (leading to possible "young people / companion"[^1] confusion)
2. This class does very little if any validation or fixity checking itself. It makes sure validation happens, and then it updates records and logs results. 

But we *do* want to make sure we continue to use this class as the sort of "front door" for auditing, so I dunno, maybe I've talked myself in enough circles to believe that `ChecksumValidationService` seems good enough for now.

[^1]: [!Mr Show Young People and Companions Clip](https://img.youtube.com/vi/jp83JZg0hU4/0.jpg)](https://www.youtube.com/watch?v=jp83JZg0hU4) (some lightly NSFW language)

jmartin-sul (approved):

PR #2400, opened 2025-04-04
Deflappify application job spec

mjgiarlo:
# Why was this change made?

Flappy specs make for grumpy developers.

Closes #2394 

I used an RSpec seed from a failing build in CI and analyzed which tests were running when. From there I was able to determine that the application job spec *always* fails when run after the replication integration spec. That led me to removing all the manual queue adapter jazz and making job specs consistently use the test adapter. I have now run the job specs in the same order as used to reproducibly fail about thirty times in a row with no failures, and run the whole suite a bunch of times with other random seeds. Looks better to me now, at least.

I suspect that this used to work just fine, but then the job testing harness and adapter defaults changed circa Rails 5-to-6 or 6-to-7.

# How was this change tested?

See above.

### DIFF
--- a/spec/jobs/application_job_spec.rb
+++ b/spec/jobs/application_job_spec.rb
@@ -17,15 +17,6 @@ class RegularParameterJob < ApplicationJob
 end
 
 RSpec.describe ApplicationJob do
-  include ActiveJob::TestHelper
-
-  around do |example|
-    old_adapter = ActiveJob::Base.queue_adapter
-    ActiveJob::Base.queue_adapter = :sidekiq
-    example.run
-    ActiveJob::Base.queue_adapter = old_adapter
-  end
-
   it 'queues start empty' do
     expect(enqueued_jobs.size).to eq 0
   end
--- a/spec/jobs/replication/integration_spec.rb
+++ b/spec/jobs/replication/integration_spec.rb
@@ -23,13 +23,6 @@ describe 'the whole replication pipeline' do # rubocop:disable RSpec/DescribeCla
   let(:ibm_provider) { instance_double(Replication::IbmProvider, bucket: ibm_bucket) }
   let(:moab_storage_root) { MoabStorageRoot.find_by!(name: 'fixture_sr1') }
 
-  around do |example|
-    old_adapter = ApplicationJob.queue_adapter
-    ApplicationJob.queue_adapter = :inline
-    example.run
-    ApplicationJob.queue_adapter = old_adapter
-  end
-
   before do
     allow(Settings).to receive(:zip_storage).and_return(Rails.root.join('spec', 'fixtures', 'zip_storage'))
     allow(Replication::AwsProvider).to receive(:new).and_return(aws_provider)
@@ -65,7 +58,9 @@ describe 'the whole replication pipeline' do # rubocop:disable RSpec/DescribeCla
     )
 
     # creating or updating a MoabRecord should trigger its parent PreservedObject to replicate any missing versions to any target endpoints
-    create(:moab_record, preserved_object: preserved_object, version: version, moab_storage_root: moab_storage_root)
+    perform_enqueued_jobs do
+      create(:moab_record, preserved_object: preserved_object, version: version, moab_storage_root: moab_storage_root)
+    end
   end
 
   context 'updating an existing moab' do
@@ -98,9 +93,11 @@ describe 'the whole replication pipeline' do # rubocop:disable RSpec/DescribeCla
         data: a_hash_including({ host: 'fakehost', version: 3, endpoint_name: 'ibm_us_south' })
       )
 
-      # updating the MoabRecord#version and its PreservedObject#current_version should trigger the replication cycle again, on the new version
-      MoabRecordService::UpdateVersion.execute(druid: druid, incoming_version: next_version, incoming_size: 712,
-                                               moab_storage_root: moab_storage_root, checksums_validated: true)
+      perform_enqueued_jobs do
+        # updating the MoabRecord#version and its PreservedObject#current_version should trigger the replication cycle again, on the new version
+        MoabRecordService::UpdateVersion.execute(druid: druid, incoming_version: next_version, incoming_size: 712,
+                                                 moab_storage_root: moab_storage_root, checksums_validated: true)
+      end
     end
   end
 end
--- a/spec/support/job_config.rb
+++ b/spec/support/job_config.rb
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 RSpec.configure do |config|
+  config.include ActiveJob::TestHelper, type: :job
+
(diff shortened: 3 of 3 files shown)


spec/support/job_config.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 RSpec.configure do |config|
+  config.include ActiveJob::TestHelper, type: :job
```
mjgiarlo, line 4: Make enqueued jobs methods available across all job specs.

spec/support/job_config.rb
```diff
@@ -1,10 +1,11 @@
...
+
   config.before do
     next unless RSpec.current_example.metadata[:type] == :job
 
-    ActiveJob::Base.queue_adapter = :test
```
mjgiarlo, line 7: This is default in modern Rails.

spec/jobs/replication/integration_spec.rb
```diff
@@ -98,9 +93,11 @@
...
       )
 
-      # updating the MoabRecord#version and its PreservedObject#current_version should trigger the replication cycle again, on the new version
-      MoabRecordService::UpdateVersion.execute(druid: druid, incoming_version: next_version, incoming_size: 712,
-                                               moab_storage_root: moab_storage_root, checksums_validated: true)
+      perform_enqueued_jobs do
+        # updating the MoabRecord#version and its PreservedObject#current_version should trigger the replication cycle again, on the new version
+        MoabRecordService::UpdateVersion.execute(druid: druid, incoming_version: next_version, incoming_size: 712,
+                                                 moab_storage_root: moab_storage_root, checksums_validated: true)
+      end
```
mjgiarlo, line 100: Make sure the enqueued jobs are run.

spec/jobs/replication/integration_spec.rb
```diff
@@ -65,7 +58,9 @@
     )
 
     # creating or updating a MoabRecord should trigger its parent PreservedObject to replicate any missing versions to any target endpoints
-    create(:moab_record, preserved_object: preserved_object, version: version, moab_storage_root: moab_storage_root)
+    perform_enqueued_jobs do
+      create(:moab_record, preserved_object: preserved_object, version: version, moab_storage_root: moab_storage_root)
+    end
```
mjgiarlo, line 63: Make sure the enqueued jobs are run.

spec/jobs/replication/integration_spec.rb
```diff
@@ -23,13 +23,6 @@
   let(:ibm_provider) { instance_double(Replication::IbmProvider, bucket: ibm_bucket) }
   let(:moab_storage_root) { MoabStorageRoot.find_by!(name: 'fixture_sr1') }
 
-  around do |example|
-    old_adapter = ApplicationJob.queue_adapter
-    ApplicationJob.queue_adapter = :inline
-    example.run
-    ApplicationJob.queue_adapter = old_adapter
-  end
```
mjgiarlo, line 31: No adapter switcheroo shenanigans are needed if we lean on Rails' built-in enqueued jobs helper.

spec/jobs/application_job_spec.rb
```diff
@@ -17,15 +17,6 @@ def perform(param_a, param_b, should_raise)
...
 
 RSpec.describe ApplicationJob do
-  include ActiveJob::TestHelper
-
-  around do |example|
-    old_adapter = ActiveJob::Base.queue_adapter
-    ActiveJob::Base.queue_adapter = :sidekiq
-    example.run
-    ActiveJob::Base.queue_adapter = old_adapter
-  end
```
mjgiarlo, line 27: I don't think we really need to be testing Sidekiq. Best to keep sidekiq out of the test suite.

spec/jobs/application_job_spec.rb
```diff
@@ -17,15 +17,6 @@ def perform(param_a, param_b, should_raise)
 end
 
 RSpec.describe ApplicationJob do
-  include ActiveJob::TestHelper
```
mjgiarlo, line 20: This is now included on all job specs.

spec/jobs/application_job_spec.rb
```diff
@@ -17,15 +17,6 @@ def perform(param_a, param_b, should_raise)
...
-
-  around do |example|
-    old_adapter = ActiveJob::Base.queue_adapter
-    ActiveJob::Base.queue_adapter = :sidekiq
```
mjgiarlo, line 24: An _ex post facto_ thought after I've merged the PR: It seems the intent of this test was partly to make sure the unique job concern is working as expected. Does that mean we care about using Sidekiq here? Or does it not matter what the queue adapter is, since UniqueJob is not coupled to Sidekiq but to ActiveJob?

jmartin-sul: i don't think the rationale was documented anywhere, but what i recall from in-person/slack discussions way back when...

there were two partially overlapping concerns that motivated these machinations:
1. we wanted to test the unique job logic as realistically as we could, and it was more directly tied to resque because it was done via resque plugin at first.
2. we wanted at least a little testing using the real queue adapter since replication and audit depend so heavily on queueing.

thinking about it more, i think 1 is less of a concern now than it was then because i'm pretty sure that the unique job logic is now tested a bit more thoroughly in the wake of #1729, and the infra integration test suite checks for replication, which is an even better smoke test of the ActiveJob being wired up correctly because it's in a deployed environment.  i think the latter thing also makes 2 less of a concern.

so i think i'm not practically that worried about this change.

spec/jobs/application_job_spec.rb
jmartin-sul: yeah, agree, ideally

spec/jobs/replication/integration_spec.rb
jmartin-sul: fwiw, my memory is that we did this to make tests more reliable.  but my memory is also that we did this right around when we were finishing up replication and replication audit, at the end of the first set of work cycles.  and of course the job coordination and job de-duplication were just a little shakier then.  so it's possible that this was actually papering over small issues that've been fixed over the years.  anyway, more than happy to drop this and see if it works out.

spec/jobs/replication/integration_spec.rb
jmartin-sul: or maybe we didn't realize we had to do this, and that's what the `:inline` stuff was making up for? 🤷

jmartin-sul (commented):
i was worried at first glance, but you talked me out of it 🙂

spec/jobs/replication/integration_spec.rb
mjgiarlo: Probably? I'm guessing this code pre-dated the `:test` adapter becoming the default, which works hand in hand with the `*enqueued_jobs` helpers.

PR #2452, opened 2025-07-30
[WIP] enable Committee::Middleware::ResponseValidation

### LINKED ISSUE #1407: Enable response validation per OpenAPI spec
1. Uncomment `# config.middleware.use Committee::Middleware::ResponseValidation, schema_path: 'openapi.yml'` in `config/application.rb`
1. Run the specs and watch the failures
1. Fix the failures

jmartin-sul:
# Why was this change made? 🤔

closes #1407

this probably has to stay WIP/draft until https://github.com/interagent/committee/issues/300 is fixed.

when revisiting this, also consider the suggestion from @justinlittman on the pres cat ticket: https://github.com/sul-dlss/preservation_catalog/issues/1407#issuecomment-1297371479

> For what it is worth, I believe that for other apps, we only do response validation during tests.
> See https://github.com/sul-dlss/dor-services-app/blob/fe8a3a024571092b2e0d8290a90f36f8aa69a285/config/application.rb#L61-L72

⚡ ⚠ If this change has cross service impact, or if it changes code used internally for cloud replication, **_run [integration test preassembly_reaccessioning_spec.rb](https://github.com/sul-dlss/infrastructure-integration-test/blob/main/spec/features/preassembly_reaccessioning_spec.rb) against stage, as it tests preservation (including cloud replication)_**, and/or test manually in stage environment, in addition to specs.

The main classes relevant to replication are `ZipmakerJob`, `DeliveryDispatcherJob`, `*DeliveryJob`, `ResultsRecorderJob`, and `DruidVersionZip`; [see here for overview diagram of replication pipeline](https://github.com/sul-dlss/preservation_catalog/wiki/Replication-Flow).⚡

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -448,9 +448,6 @@ components:
       description: A preserved object instance
       type: object
       properties:
-        id:
-          type: integer
-          example: 123
         druid:
           $ref: '#/components/schemas/Druid'
         current_version:
@@ -465,7 +462,6 @@ components:
           format: date-time
           example: '2020-02-21T16:36:28.075Z'
       required:
-        - id
         - druid
         - current_version
         - created_at
--- a/config/application.rb
+++ b/config/application.rb
@@ -63,11 +63,8 @@ module PreservationCatalog
       strict_reference_validation: true,
       parameter_overwrite_by_rails_rule: false
     )
-    # TODO: we can uncomment this at a later date to ensure we are passing back
-    #       valid responses. Currently, uncommenting this line causes 24 spec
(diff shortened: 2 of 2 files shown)


openapi.yml
```diff
@@ -448,9 +448,6 @@ components:
       description: A preserved object instance
       type: object
       properties:
-        id:
-          type: integer
-          example: 123
```
jmartin-sul, line 453: this doesn't actually get returned in the response, and i don't think consumers are actually looking for it (because it's just the internal DB PK, but everything does lookup and matching by druid in other systems)

PR #2489, opened 2025-12-03
[HOLD] Allow scheduled tasks to respect the specified DB order

### LINKED ISSUE #2479: fix ordering for nightly checksum validation and catalog-to-archive audits
The intent is that the two nightly cron jobs for, respectively, checksum validation and catalog to archive audit, examine the least recently audited druids:

https://github.com/sul-dlss/preservation_catalog/blob/b171b00d72e566b56b4e5ea287c215e616edbec4/config/schedule.rb#L39-L44

https://github.com/sul-dlss/preservation_catalog/blob/b171b00d72e566b56b4e5ea287c215e616edbec4/config/schedule.rb#L54-L59

However, we've noticed that we still have bursts where a significant fraction of the catalog is audited by the weekly checksum validation and catalog to archive audits, which look for things where those checks are expired (anything that hasn't had the audit in question for >= 90 days).

It turns out, we're probably rarely looking at the least recently audited fraction on any given night: the queries used by the cron jobs have both an `order` statement and a `find_each` loop.  But `find_each` ignores ordering and orders by primary key, under the assumption it'll be monotonically increasing (since `find_each` grabs batches, and this approach lets it move through everything deterministically).  And sure enough, if you run the `find_each` on rails console on stage, you see this warning (in the deployed cron job, the output is sent to `/dev/null`):

```
3.4.0 :001 > PreservedObject.order(last_archive_audit: :asc).limit(PreservedObject.daily_check_count).find_each(&:audit_moab_version_replication!)
Scoped order is ignored, use :cursor with :order to configure custom order.
```

We should take the error message's advice, so that we'll actually get the originally desired behavior.

mjgiarlo:
Fixes #2479

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -36,7 +36,7 @@ end
 # Any that are not validated but hit fixity TTL will be validated by weekly audit below.
 every :day, at: '8pm', roles: [:queue_populator] do
   set :output, standard: nil, error: 'log/c2a-err.log'
-  runner 'PreservedObject.order(last_archive_audit: :asc).limit(PreservedObject.daily_check_count).find_each(&:audit_moab_version_replication!)'
+  runner 'PreservedObject.order(last_archive_audit: :asc).limit(PreservedObject.daily_check_count).each_instance(&:audit_moab_version_replication!)'
 end
 
 every :wednesday, roles: [:queue_populator] do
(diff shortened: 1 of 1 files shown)


jmartin-sul (commented):
replied to the comment on the issue (good find in the docs, thanks!) with a suggestion

jmartin-sul:
@mjgiarlo 💬 
> I ran a version of the second test in stage just now:
> 
> ...snip...
> 
> I'll deploy to stage and QA now and check back on Friday.
> 
> Thanks for the helping hand w/ the test plan!

nice, that all looks great, thanks!  well, i can't explain the things that haven't been checked since april in the second set of query results, since we're supposed to check everything every 90 days.  but it's stage, so i'm sorta inclined to 🙉 (but happy to investigate if you think ignoring would be negligent)

(and maybe it could be explained by #2498 🤷 -- we will find out in the maintenance work cycle next year, i guess)

(and to be totally clear, i'm pretty sure the issue is something with stage data, not with the approach in this PR)

mjgiarlo:
@jmartin-sul 💬 
> nice, that all looks great, thanks! well, i can't explain the things that haven't been checked since april in the second set of query results, since we're supposed to check everything every 90 days. but it's stage, so i'm sorta inclined to 🙉 (but happy to investigate if you think ignoring would be negligent)

I don't think it's negligent since we're talking about stage data. Thanks!

PR #2516, opened 2026-01-07
Add options to create_bag script

### LINKED ISSUE #2474: Add new options to "create_bag" script (retrieve old version, retrieve only certain files)
**Background**

There is a script deployed in the `bin` directory called `create_bag` that takes the following input:

- a list of druids
- a destination

and produces the following output:

- in the destination, create a BagIt bag containing symlinks to the content in the latest version of each druid

This script is the primary tool SDR admins use to fulfill requests for SDR data that requires manual retrieval.

This script works great when retrieving all files from the latest version of an item, but we get requests that can't be filled so easily:

- Retrieve a specific older version
- Retrieve only certain files (e.g. only PDFs) 

In those cases, we either need to modify the script or find an alternative strategy for fulfilling the request.

For more details, there is [documentation on how this script is used](https://consul.stanford.edu/spaces/SDRdocs/pages/1646529897/Fulfilling+an+SDRGET+request+instructions+for+SDR+staff) on Consul. Some of the exception cases are covered in that documentation.

**Request**

Add two options to the `create_bag` script that can be invoked on the command line:

- Retrieve a previous version: pass in the version number with this option
- Retrieve only certain files: pass in an extension with this option

An alternative for the file type selection could be to use mimetype, but I'm not sure we have that data in the moab manifests. Matching on the extension is also more straightforward and escapes some of the issues that come up when mimetype detection is incorrect.

mjgiarlo:
# Why was this change made?

Fixes #2474

New CLI:

```
Usage: bin/create_bag DRUID_LIST_FILE TARGET_DIRECTORY [options]
    -v, --version VERSION_NUMBER     Create a bag for the specified version of a given object, e.g., 2 (default: current version)
    -e, --extension FILE_EXTENSION   Include only files that match a given file extension, e.g., pdf (default: all)
        --[no-]dry_run               Simulate the bag creation (defaults to false)
        --[no-]quiet                 Do not output progress information (defaults to false)
    -h, --help                       Displays help.
```

# How was this change tested?

Ran the script with various options on QA

### DIFF
--- a/bin/create_bag
+++ b/bin/create_bag
@@ -19,12 +19,58 @@ ENV["RAILS_ENV"] ||= "production"
 
 require_relative '../config/environment'
 
-File.foreach(ARGV[0], chomp: true).each do |druid|
+require 'optparse'
+
+options = { version: nil, extension: nil, dry_run: false, quiet: false }
+
+parser = OptionParser.new do |option_parser|
+  option_parser.banner = "Usage: #{$PROGRAM_NAME} DRUID_LIST_FILE TARGET_DIRECTORY [options]"
+  option_parser.on('-v', '--version VERSION_NUMBER',
+                   'Create a bag for the specified version of a given object, e.g., 2 (default: current version)')
+  option_parser.on('-e', '--extension FILE_EXTENSION',
+                   'Include only files that match a given file extension, e.g., pdf (default: all)')
+  option_parser.on('--[no-]dry_run', 'Simulate the bag creation (defaults to false)')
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 
+require_relative 'boot'
```
mjgiarlo: This is what modern Rails does, presumably because of bootsnap.

jmartin-sul (approved):

PR #2568, opened 2026-01-21
Model refactors

justinlittman:
# Why was this change made?
Less code.

# How was this change tested?


app/models/moab_storage_root.rb
```diff
@@ -22,8 +22,8 @@ def validate_expired_checksums!
   end
 
   # Use a queue to check all associated MoabRecord objects for C2M
-  def c2m_check!(last_checked_b4_date = Time.current)
-    moab_records.version_audit_expired(last_checked_b4_date).find_each do |moab_rec|
+  def c2m_check!
```
justinlittman: Never called with argument.

PR #2574, opened 2026-01-22
Refactor cloud provider classes to DRY up and clarify source of config

aaron-collier:
# Why was this change made?

This is a possible take on refactoring the Cloud Provider classes (AwsProvider, GcpProvider, IbmProvider) because they are almost identical with a single client config value difference.

# How was this change tested?


app/services/replication/base_provider.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Replication
+  # The Application's configured general interface to AWS S3 compatible resources.
+  class BaseProvider
```
aaron-collier: This is mostly the original AwsProvider class as this code was almost entirely shared in each class.

app/services/replication/base_provider.rb
```diff
@@ -0,0 +1,37 @@
...
+    def bucket_name
+      endpoint_settings.storage_location
+    end
+
+    # @return [::Aws::S3::Resource]
+    def resource
+      ::Aws::S3::Resource.new(aws_client_args)
+    end
+
+    def endpoint_settings
```
aaron-collier: This most impactful change for all of this is here, instead of munging the "region" value into a the key for settings, we use the actual endpoint_name for more robustness and clarity.

jmartin-sul (approved):
two small suggestions.  feel free to merge as is and be done with it, to take either of the suggestions and then to merge, or to merge and do a small follow on with either/both of the suggestions.

PR #2577, opened 2026-01-22
Adds help text to dashboard.

justinlittman:
# Why was this change made?
So devs can understand the dashboard better.

# How was this change tested?


justinlittman:
image

PR #2618, opened 2026-04-02
[DRAFT | HOLD] Switch to SolidQueue from Sidekiq/Redis

### LINKED ISSUE #2501: [RESEARCH] what would it take to switch preservation_catalog from sidekiq to solidqueue?
Here is a (possibly non-exhaustive) list of the custom functionality we've built for our jobs on Sidekiq and/or Redis:
- [ ] "queue locking": an approach for preventing the queuing of duplicate jobs (same job type with the same salient parameters withing a specified window, by default one hour).  this is a safety measure, since we really never want duplicate jobs running at the same time, and since the job should clean up its lock (whether it succeeds or fails), allowing the job to be run immediately again upon completion, if need be.  see the `UniqueJob` concern.
- [ ] reservation of workers (threads or processes or whatever does the work in solidqueue) for different queues.  this allows us to know that there will always be at least N workers for `validate-moab` (which gates accessioning), X workers for replication, etc.  doing this prevents any one category of work from usurping all the resources and blocking another category of work (e.g. prevent lots of long running checksum validation jobs from slowing accessioning or replication).  the cost is of course that we often have idle workers.
- [ ] "hotswap" the ability to let a job that has WIP wind down gracefully after the workers receive a shutdown signal, whether because the workers are being restarted manually, because the app was deployed and restarted workers as part of that, etc.

Since we expect this conversion likely takes more time than we have in the upcoming maintenance work cycle, the goal of this ticket is just to research and prototype and get a rough idea of where the pitfalls are and how much effort switching ActiveJob back-ends might take.

aaron-collier:
Fixes #2501 

### Status

This currently replaced sidekiq/resque with solid_queue. This is a relatively straight forward swap with a couple of outstanding concerns:

1 - Is the stopping/restarting approach in solid_queue sufficient in place of existing hot swapping?
2 - Are the deployed machines able to be configured for specific recurring task roles (as is currently used in `whenever`)
3 - How will/should we determine if this will be sufficient? Can we use qa and/or stage and rollback database changes if necessary.
4 - This is currently configured as a single database deployment, which is likely not ideal, especially with the number of jobs that can be enqueued for preservation, should we start the process with ops to setup and configure a dedicated queue db, at least for production?
5 - There are likely puppet and shared_configs changes that will need to be coordinated.

** NOTE: ** this may be an incomplete list of concerns, I will add/edit as necessary.

### Removed

- sidekiq, sidekiq-pro, redis, connection_pool, whenever gems
- config/initializers/sidekiq.rb
- config/initializers/redis.rb
- app/jobs/concerns/unique_job.rb
- redis settings from config/settings.yml
- sidekiq_systemd_* from config/deploy.rb
- whenever_identifier from config/deploy.rb

### Added
- solid_queue, mission_control-jobs gems
- config/initializers/mission_control.rb
- config/queue.yml
- bin/jobs
- db/queue_schema.rb
- jobs: entry in Procfile.dev

### Updated:
- config/database.yml (queue alias)
- config/environments/production.rb (SolidQueue adapter)
- config/routes.rb (MissionControl mount)
- .circleci/config.yml (drop Redis executor)
- all 6 job files (limits_concurrency replacing UniqueJob)
- spec/support/job_config.rb (removed Redis flush)
- app/views/dashboard/dashboard/index.html.erb (updated link)
- README.md


app/jobs/audit/catalog_to_moab_job.rb
```diff
@@ -6,7 +6,7 @@ module Audit
   class CatalogToMoabJob < ApplicationJob
     queue_as :c2m
 
-    include UniqueJob
+    limits_concurrency to: 1, key: ->(job) { job.arguments.first }, duration: 1.hour, on_conflict: :discard
```
aaron-collier, line 9: `limits_concurrency` replaces `UniqueJob` here (and below)

config/deploy/prod.rb
```diff
@@ -2,11 +2,10 @@
 
 server 'preservation-catalog-web-prod-01.stanford.edu', user: 'pres', roles: %w[app web]
 server 'preservation-catalog-web-prod-02.stanford.edu', user: 'pres', roles: %w[app web]
-server 'preservation-catalog-worker-prod-01.stanford.edu', user: 'pres', roles: %w[app db worker queue_populator]
+server 'preservation-catalog-worker-prod-01.stanford.edu', user: 'pres', roles: %w[app db worker scheduler]
```
aaron-collier: `workers` should maintain the same role as before and `scheduler` adds the role to enqueue jobs. Note, one thing we lose here is the granularity of `whenever_roles` for only running certain recurring jobs on one box. This is currently still being investigated.

config/application.rb
```diff
@@ -46,12 +46,17 @@ class Application < Rails::Application
...
+    # Add timestamps to all loggers
     config.log_formatter = proc do |severity, datetime, _progname, msg|
       "[#{datetime.to_fs(:iso8601)}] [#{severity}] #{msg}\n"
     end
 
-    # accept_request_filter omits OKComputer & Sidekiq routes
+    # Allow running jobs to finish before SolidQueue workers are killed during deployment.
+    # Override with JOB_SHUTDOWN_TIMEOUT env var (value in seconds). Default: 345600 (96 hours).
+    # Jobs that exceed this timeout will be retried via retry_on in ApplicationJob.
+    config.solid_queue.shutdown_timeout = ENV.fetch('JOB_SHUTDOWN_TIMEOUT', 345_600).to_i
```
aaron-collier, line 57: Bumps the existing timeout to 96 hours (for now) from the 48 hours here: https://github.com/sul-dlss/puppet/blob/21451e995a3ceb1cf0d546722b8af752614b56a3/hieradata/node/preservation-catalog-worker-prod-01.stanford.edu.yaml#L95 -- may revert this back to 48 after discussion.

config/queue.yml
```diff
@@ -0,0 +1,26 @@
+default: &default
+  dispatchers:
+    - polling_interval: 1
+      batch_size: 500
+  workers:
```
aaron-collier, line 5: We _may_ be able to setup this configuration to match the `whenever_roles`, will follow up.

app/jobs/audit/moab_to_catalog_job.rb
```diff
@@ -6,7 +6,7 @@ module Audit
   class MoabToCatalogJob < ApplicationJob
     queue_as :m2c
 
-    include UniqueJob
+    limits_concurrency to: 1, key: ->(job) { job.arguments.first }, duration: 1.hour, on_conflict: :discard
```
jmartin-sul, line 9: i think for this one, we want to key on both the args, `root` and `druid`.

config/application.rb
jmartin-sul: i like 96, personally (this is mostly gut feeling atm, but i believe i've seen the very largest multi-TB replication and `validate-moab` jobs take right around a couple days.  @andrewjbtw might have a more accurate gut feeling, or even concrete stats.

config/database.yml
```diff
@@ -13,6 +13,23 @@ development:
...
+    database: test
+
+production:
+  primary:
+    <<: *default
+    database: <%= ENV.fetch('DATABASE_NAME', 'preservation_catalog') %>
+
+  queue:
+    <<: *default
+    database: <%= ENV.fetch('DATABASE_NAME', 'preservation_catalog') %>
```
jmartin-sul, line 35: i thought i saw somewhere that the advice was to use a different database for SolidQueue in production, but now i'm having trouble finding a reference for that.  happy feels right.  i guess i might lean towards a separate database (even if stored in the same PG instance), unless that's un-idiomatic or unnecessary or counterproductive.

config/application.rb
jmartin-sul, line 1: given that network flakiness is a problem we somewhat regularly see, i wonder if we should consider upping `process_alive_threshold` from the default 5 minutes, to maybe 10.  our jobs are pretty reliable generally, and often long running.  don't feel strongly at all about this, happy to roll with the default and see what happens, but thought it might warrant discussion.

config/deploy.rb
```diff
@@ -24,8 +24,6 @@
 
 set :honeybadger_env, fetch(:stage)
 
-set :whenever_identifier, -> { "#{fetch(:application)}_#{fetch(:stage)}" }
```
jmartin-sul, line 27: did we end up taking out `whenever`?  is this cruft from backing out of the removal?  can't clearly recall what was decided.

config/settings.yml
```diff
@@ -50,14 +50,19 @@ api_jwt:
 
 filesystem_delay_seconds: 0.05 # the default value is artificially low to keep test suite fast
 
+# Controls how jobs are retried after being killed or pruned by SolidQueue.
+# ProcessExitError is raised when a worker is terminated (e.g. KILL signal or OOM killer).
+# ProcessPrunedError is raised when a worker's heartbeat expires and SolidQueue prunes it.
+job_retry:
+  wait: 1      # minutes to wait before re-queuing a killed or pruned job
```
jmartin-sul, line 57: is it possible to do the exponential backoff thing sidekiq does?  doesn't feel essential, but feels like a nice to have, esp for failed network calls to external services.

this is also definitely something where i'm happy to deploy like this and tune as needed.

README.md
```diff
@@ -115,9 +113,9 @@ curl -H 'Authorization: Bearer eyJhbGcxxxxx.eyJzdWIxxxxx.lWMJ66Wxx-xx' http://lo
 
 - The PostgreSQL database is the catalog of metadata about preserved SDR content, both on premises and in the cloud.  Database-level constraints are used heavily for keeping data clean and consistent.
 
-- Background jobs (using ActiveJob/Sidekiq/Redis) perform the audit and replication work.
+- Background jobs (using ActiveJob/SolidQueue) perform the audit and replication work. SolidQueue uses the existing PostgreSQL database — no separate Redis instance is required.
 
-- The whenever gem is used for writing and deploying cron jobs. Queueing of weekly audit jobs, temp space cleanup, etc are scheduled using the whenever gem.
+- Cron jobs (defined in `config/schedule.rb`) handle periodic scheduling of audit jobs, temp space cleanup, etc. Recurring job configuration is in `config/recurring.yml`.
```
jmartin-sul, line 118: last sentence might be cruft now?

README.md
```diff
@@ -48,11 +48,9 @@ First get the Ruby dependencies:
 bundle install
 ```
 
-The credentials for SideKiq Pro must be provided (e.g., in `.bash_profile`): `export BUNDLE_GEMS__CONTRIBSYS__COM=xxxx:xxxx` (available from shared_configs).
-
-Use `docker compose` to start supporting services (PostgreSQL and Redis)
+Use `docker compose` to start supporting services (PostgreSQL):
 ```sh
-docker compose up -d db redis
+docker compose up -d db
```
jmartin-sul, line 53: could maybe take out the name of the service, since it's the only one now?  iirc, we originally listed specific services in the readme, because the ER diagram generator was still defined in the main `compose.yaml` file.

also, i think the `compose.yaml` could be updated to drop redis?

spec/jobs/application_job_spec.rb
```diff
@@ -2,97 +2,35 @@
...
-      expect { Audit::CatalogToMoabJob.perform_later(moab_rec) }
-        .not_to change(enqueued_jobs, :size).from(1)
+    it 'retries when a worker process is pruned due to a missed heartbeat (ProcessPrunedError)' do
+      expect(registered_handler_names).to include('SolidQueue::Processes::ProcessPrunedError')
     end
 
-    it 'but adds novel messages' do
-      expect { Audit::CatalogToMoabJob.perform_later(moab_rec2) }
-        .to change(enqueued_jobs, :size).from(1).to(2)
+    it 'uses the wait period from Settings' do
```
jmartin-sul, line 31: the title seems to not quite match the test logic here?

config/queue.yml
jmartin-sul, line 1: i compared this setup to the `sidekiq*.yml` files in shared_configs for stage, QA, and prod, and i think the SolidQueue version for each env would look something like this, since the process/thread architecture is pretty similar.  comments adapted and cleaned up from current shared_configs `sidekiq*.yml`.  also `validate_moab` count was upped here to make the comment true, and match what pres bots allocates, since current shared_configs allocates 3 threads instead of pres bots' 5.

the concision and clarity of controlling this all in one file instead of a separate `sidekiq.yml` file for each separately launched process is very nice.  i think this means we can get rid of some custom systemd/puppet code for sidekiq process management, once we switch everything that uses it to SolidQueue.

# actual prod
```yaml
production:
  dispatchers:
    - polling_interval: 1
      batch_size: 500
  workers:
    # validate_moab gets dedicated workers, since it blocks accessionWF (via
    # preservationWF validate-moab) from proceeding. Match the number in
    # preservation_robots/config/sidekiq.yml (which is directly in app, not shared_configs).
    - queues: [validate_moab]
      threads: 5
      processes: 1
      polling_interval: 0.1
    # c2m and m2c pretty much always get run in bulk by weekly cron jobs at different times, and
    # so can share a set of workers.
    - queues: [m2c, c2m]
      threads: 1
      processes: 1
      polling_interval: 0.1
    - queues: [replication, moab_replication_audit]
      threads: 9
      processes: 3
      polling_interval: 0.1
    # Dedicated CV workers, since CV can be resource intensive and since we sometimes manually run
    # batches of druids through CV while regularly scheduled audits are happening.
    # This re-creates the former Sidekiq setup, with 5 separate single-threaded worker processes.
    - queues: [checksum_validation]
      threads: 1
      processes: 5
      polling_interval: 0.1
```

# stage and QA
```yaml
production:
  dispatchers:
    - polling_interval: 1
      batch_size: 500
  workers:
    # validate_moab gets dedicated workers, since it blocks accessionWF (via
    # preservationWF validate-moab) from proceeding. Match the number in
    # preservation_robots/config/sidekiq.yml (which is directly in app, not shared_configs).
    - queues: [validate_moab]
      threads: 5
      processes: 1
      polling_interval: 0.1
    # c2m and m2c pretty much always get run in bulk by weekly cron jobs at different times, and
    # so can share a set of workers.
    - queues: [m2c, c2m]
      threads: 1
      processes: 1
      polling_interval: 0.1
    - queues: [moab_replication_audit]
      threads: 8
      processes: 1
      polling_interval: 0.1
    - queues: [replication]
      threads: 9
      processes: 2
      polling_interval: 0.1
    # Dedicated CV workers, since CV can be resource intensive and since we sometimes manually run
    # batches of druids through CV while regularly scheduled audits are happening.
    # This re-creates the former Sidekiq setup, with 5 separate single-threaded worker processes.
    - queues: [checksum_validation]
      threads: 1
      processes: 3
      polling_interval: 0.1
```

i dropped the use of `JOB_CONCURRENCY` as an env var, because i think it existed for an okcomputer check once, but i think we dropped that?  i couldn't find it when i grepped puppet and pres cat.

so i guess those would override the default `config/queue.yml` from the repo, by going in shared_configs branches and being specified in `linked_files` in `config/deploy.rb`?

and then for the default `config/queue.yml` from the repo for dev and test, maybe a pared down and simplified version of the above, with thread counts that i think add up to close to what you have in your version of this config file (9 threads across 5 processes here, 10 threads across 3 processes in yours):

```yaml
default: &default
  dispatchers:
    - polling_interval: 1
      batch_size: 500
  workers:
    - queues: [validate_moab]
      threads: 3
      processes: 1
      polling_interval: 0.1
    - queues: [m2c, c2m]
      threads: 1
      processes: 1
      polling_interval: 0.1
    - queues: [replication, moab_replication_audit]
      threads: 3
      processes: 1
      polling_interval: 0.1
    - queues: [checksum_validation]
      threads: 1
      processes: 2
      polling_interval: 0.1

development:
  <<: *default

test:
  <<: *default
```

one other thought on this front, about something not introduced by this PR, just highlighted by it: i understand the differing thread and process counts for stage/QA and prod, given the differences in CPU (5 cores on QA/stage, 8 on prod) and memory (4.8 GB on QA/stage, 15.6 GB on prod).  but the different approach to shared processes or not for `moab_replication_audit` and `replication` queues seems like it might be unintentional?  i feel like we should check with ops and see if they have a preference for the allocation of threads/processes for that work on those boxes, and make that part consistent among all envs, even if the process and thread counts and resourcing differ.

app/jobs/validate_moab_job.rb
```diff
@@ -8,7 +8,7 @@
 class ValidateMoabJob < ApplicationJob
   queue_as :validate_moab
 
-  include UniqueJob
+  limits_concurrency to: 1, key: ->(job) { job.arguments.first }, duration: 1.hour, on_conflict: :discard
```
jmartin-sul, line 11: this should also probably be a 2 or 3 day duration, since it checksum validates the incoming version.

app/jobs/audit/checksum_validation_job.rb
```diff
@@ -4,10 +4,10 @@ module Audit
   # Confirm checksum for one Moab on storage and update MoabRecord in database
   # @see Audit::ChecksumValidationService
   class ChecksumValidationJob < ApplicationJob
-    include UniqueJob
-
     queue_as :checksum_validation
 
+    limits_concurrency to: 1, key: ->(job) { job.arguments.first }, duration: 1.hour, on_conflict: :discard
```
jmartin-sul, line 9: i'd give this a day or two duration, since this can take quite a while for large objects

jmartin-sul (changes requested):
some small suggestions and comments, but this seems great overall, pending tweaks and testing.

for testing, ii would vote for something like:
* integration tests
  * fixity check the replicated objects from integration tests (may happen automatically in at least `preassembly_reaccession_spec.rb` if https://github.com/sul-dlss/infrastructure-integration-test/pull/1002 is merged)
* manually checksum validate something giant, deploy while that checksum validation job is running, make sure the long running job winds down gracefully and is allowed the configured hotswap period.
* manually checksum validate something large enough that an attempt can be made to enqueue the same job again while the first is running on the given druid, confirm no second job is queued before the first one finishes, confirm that the job can be run again on the same druid once the first run completes.
* work with andrew to accession and/or version something using more than 10 GB of unique content, so that a multi-part zip is replicated
  * watch the job console, and deploy pres cat while the replication job is running for this druid.  confirm that the WIP job is allowed to complete gracefully despite the deploy
  * run `bin/fixity_check_replicated_moabs` against GCP the druid from this test (script has usage notes)

PR #2622, opened 2026-04-07
decommission_endpoint becomes decommission_endpoint_in_db, and defaults to doing the DB removal in a piecemeal fashion

jmartin-sul:
# Why was this change made?

https://stanfordlib.slack.com/archives/C0RK5EM9N/p1775058777691269

# How was this change tested?

### DIFF
--- a/lib/tasks/decommission_endpoint.rake
+++ b/lib/tasks/decommission_endpoint.rake
@@ -14,8 +14,8 @@ task :decommission_endpoint_in_db, [:endpoint_name, :dry_run, :batch_size, :limi
   args.with_defaults(dry_run: 'true', batch_size: 50_000, limit: 50_000)
   endpoint_name = args[:endpoint_name]
   dry_run = args[:dry_run] != 'false'
-  limit = args[:limit]
-  batch_size = args[:batch_size]
+  limit = args[:limit].to_i
+  batch_size = args[:batch_size].to_i
 
   logger.info "===== #{task.name}: interpreted args: #{{ endpoint_name:, dry_run:, batch_size:, limit: }}"
   zip_endpoint = ZipEndpoint.find_by(endpoint_name:)
--- a/lib/tasks/decommission_endpoint.rake
+++ b/lib/tasks/decommission_endpoint.rake
@@ -24,6 +24,16 @@ task :decommission_endpoint_in_db, [:endpoint_name, :dry_run, :batch_size, :limi
     exit 1
   end
 
+  endpoint_stats =
+    ZipPart.joins(
+      zipped_moab_version: [:zip_endpoint]
+    ).group(
+      :endpoint_name
+    ).pluck(
+      :endpoint_name, 'COUNT(DISTINCT(zipped_moab_versions.id))', 'COUNT(DISTINCT(zip_parts.id))'
+    )
+  logger.info "endpoint stats (name, zipped_moab_versions count, zip_parts count): #{endpoint_stats}"
+
   unless dry_run
     puts 'Are you sure you want to proceed? (y/n)'
     answer = $stdin.gets.chomp.downcase
@@ -53,9 +63,21 @@ task :decommission_endpoint_in_db, [:endpoint_name, :dry_run, :batch_size, :limi
     end
   end
 
-  # if it's not a dry run and all the zipped_moab_versions are gone, remove the endpoint
-  unless dry_run || zip_endpoint.reload.zipped_moab_versions.exist?
+  # only remove the endpoint if it's not a dry run and all its zipped_moab_versions are gone
+  if dry_run || zip_endpoint.reload.zipped_moab_versions.exist?
+    logger.info "dry run, or zipped_moab_versions remain on #{endpoint_name}, cannot delete zip_endpoints row"
+  else
     logger.info "Deleting ZipEndpoint: #{endpoint_name}"
     zip_endpoint.destroy!
   end
+
+  endpoint_stats =
+    ZipPart.joins(
+      zipped_moab_version: [:zip_endpoint]
+    ).group(
+      :endpoint_name
+    ).pluck(
+      :endpoint_name, 'COUNT(DISTINCT(zipped_moab_versions.id))', 'COUNT(DISTINCT(zip_parts.id))'
+    )
+  logger.info "endpoint stats (name, zipped_moab_versions count, zip_parts count): #{endpoint_stats}"
 end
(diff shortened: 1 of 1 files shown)


lib/tasks/decommission_endpoint.rake
```diff
@@ -22,21 +26,24 @@ task :decommission_endpoint, [:endpoint_name, :access_key, :secret_access_key, :
     end
   end
 
-  zip_endpoint.zipped_moab_versions.find_each do |zipped_moab_version|
-    zip_info = "#{zipped_moab_version.druid} (#{zipped_moab_version.version})"
+  # this should be stable from run to run, as #in_batches orders on primary key
+  zip_endpoint.zipped_moab_versions.limit(limit).in_batches(of: batch_size) do |zipped_moab_versions_relation|
+    zip_parts_relation = ZipPart.where(zipped_moab_version: zipped_moab_versions_relation)
```
justinlittman, line 49: A `where` with 50K conditions?

aaron-collier: They're small-ish records, so I wouldn't think 50K would be a big deal... But I wonder if we could benchmark that.

jmartin-sul: thanks for this nudge @justinlittman, and for the standup discussion.  by default, this actually did put all the `zipped_moab_versions` IDs in a big `IN` clause when i tested with small batch size (20), and so the `zip_parts` query in turn used this explicitly listed `IN`.  turns out there's an ActiveRecord param that uses ranges instead, and [the default is not to use ranges](https://apidock.com/rails/ActiveRecord/Batches/in_batches) unless the batching is across the full table with no `WHERE` clause.  setting the `use_ranges` param to true fixed things.

from a bit of manual testing of the parts in the terminal on stage...

old:
```
3.4.0 :012 > zip_endpoint.zipped_moab_versions.limit(limit).in_batches(of: batch_size) do |zipped_moab_versions_relation|
  zip_parts_relation = ZipPart.where(zipped_moab_version: zipped_moab_versions_relation)
  zip_info = "#{zip_parts_relation.count} zip parts for #{zipped_moab_versions_relation.count} zipped moab versions " \
             "(zipped_moab_versions id range: #{zipped_moab_versions_relation.pluck('MIN(id)')} to #{zipped_moab_versions_relation.pluck('MIN(id)')})"
 
  puts "zipped_moab_versions_relation.to_sql=#{zipped_moab_versions_relation.to_sql}"
  puts "zip_parts_relation.to_sql=#{zip_parts_relation.to_sql}"
  puts "[Dry run] Deleting: #{zip_info}"
end

zipped_moab_versions_relation.to_sql=SELECT "zipped_moab_versions".* FROM "zipped_moab_versions" WHERE "zipped_moab_versions"."zip_endpoint_id" = 4 AND "zipped_moab_versions"."id" IN (113995, 113996, 114000, 114004, 114005, 114006, 114007, 114008, 114009, 114010) LIMIT 10
zip_parts_relation.to_sql=SELECT "zip_parts".* FROM "zip_parts" WHERE "zip_parts"."zipped_moab_version_id" IN (SELECT "zipped_moab_versions"."id" FROM "zipped_moab_versions" WHERE "zipped_moab_versions"."zip_endpoint_id" = 4 AND "zipped_moab_versions"."id" IN (113995, 113996, 114000, 114004, 114005, 114006, 114007, 114008, 114009, 114010) LIMIT 10)
[Dry run] Deleting: 10 zip parts for 10 zipped moab versions (zipped_moab_versions id range: [113995] to [113995])
```

`use_ranges: true`:
```
3.4.0 :022 > zip_endpoint.zipped_moab_versions.limit(limit).in_batches(of: batch_size, use_ranges: true) do |zipped_moab_versions_relation|
  zip_parts_relation = ZipPart.where(zipped_moab_version: zipped_moab_versions_relation)
  zip_info = "#{zip_parts_relation.count} zip parts for #{zipped_moab_versions_relation.count} zipped moab versions " \
             "(zipped_moab_versions id range: #{zipped_moab_versions_relation.pluck('MIN(id)')} to #{zipped_moab_versions_relation.pluck('MIN(id)')})"
 
  puts "zipped_moab_versions_relation.to_sql=#{zipped_moab_versions_relation.to_sql}"
  puts "zip_parts_relation.to_sql=#{zip_parts_relation.to_sql}"
  puts "[Dry run] Deleting: #{zip_info}"
end
zipped_moab_versions_relation.to_sql=SELECT "zipped_moab_versions".* FROM "zipped_moab_versions" WHERE "zipped_moab_versions"."zip_endpoint_id" = 4 AND "zipped_moab_versions"."id" <= 114010
zip_parts_relation.to_sql=SELECT "zip_parts".* FROM "zip_parts" WHERE "zip_parts"."zipped_moab_version_id" IN (SELECT "zipped_moab_versions"."id" FROM "zipped_moab_versions" WHERE "zipped_moab_versions"."zip_endpoint_id" = 4 AND "zipped_moab_versions"."id" <= 114010)
[Dry run] Deleting: 10 zip parts for 10 zipped moab versions (zipped_moab_versions id range: [113995] to [113995])
```

jmartin-sul: no idea how it got the range limits vs the individual values, and whether it had to do another query beforehand to get that info in either mode.

mjgiarlo: Fine by me to ship this. Don't want to do so unless/until @justinlittman has a chance to sound off when he's back.

jmartin-sul: sorry if this is restating the obvious, but just to be clear since it was kind of a long thread and it languished for a while (and since i may not be watching closely in the coming days): what ended up in this PR used a range query instead of many individual IDs, and was reasonably performant -- iirc it would delete batches of 100k to 500k ZMVs and associated zip parts in 10ish minutes per batch, but i would have to refer back to slack conversation at the time to refresh my memory for sure on exact timing.  either way, doing batches with this approach got it so we only had to pause replication for the last few thousand and the endpoint removal, which only took a few minutes.

the S3 bucket was cleared by generating a list of S3 keys from another zip endpoint later on, and [giving that list to ops](https://github.com/sul-dlss/operations-tasks/issues/4321#issuecomment-4226960122) to do the deletions from IBM storage, before shutting off service.

jmartin-sul: that's not an argument for merging this, just a clarification for when others are hashing that out 🙂

PR #2652, opened 2026-07-16
When updating the catalog is requested for the currently recorded version, treat as a no-op.

### LINKED ISSUE #2651: Update catalog / m2c race condition
There is currently a race condition between the preservationIngestWF and Moab To Catalog jobs:
1. Update Moab robot adds a new version to the moab for an object.
2. Moab To Catalog runs for the object. It finds a new version in the moab and creates a Moab Record for the version.
3. Update Catalog robot calls Pres Cat update for the object. Since the moab version is not greater than the Moab Record version, it raises a 500:
```
update-catalog : Preservation::Client.Preservation::Client::VersionedApiService#http_response got 500 from Preservation at v1/catalog/druid:bb066fp1347: {"druid":"bb066fp1347","results":[{"unexpected_version":"actual version (4) has unexpected relationship to MoabRecord db version (4); ERROR!"},{"version_matches":"actual version (4) matches MoabRecord db version"}]} 
```

justinlittman:
closes #2651

# Why was this change made?
See ticket for description of race condition.

# How was this change tested?
Unit, stage

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -28,7 +28,7 @@ class CatalogController < ApiController
     results = MoabRecordService::UpdateVersion.execute(druid: bare_druid, incoming_version: incoming_version, incoming_size: incoming_size,
                                                        moab_storage_root: moab_storage_root, checksums_validated: checksums_validated)
     status_code =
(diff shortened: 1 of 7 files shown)


lwrubel (approved):
Approved pending integration testing

PR #2656, opened 2026-07-22
Prevent collision of empty directory and stale file cache cleaners.

justinlittman:
# Why was this change made?
Avoid errors like:
```
find: ‘/sdr-transfers/bc’: No such file or directory
```
and
```
find: ‘/sdr-transfers/bz’: No such file or directory
find: ‘/sdr-transfers/fw’: No such file or directory
```

# How was this change tested?

### DIFF
--- a/lib/tasks/prescat.rake
+++ b/lib/tasks/prescat.rake
@@ -1,5 +1,18 @@
 # frozen_string_literal: true
 
+# Prevents cache_cleaner:stale_files and cache_cleaner:empty_directories from walking/deleting
+# in the same tree concurrently, which otherwise causes intermittent
+# "find: '...': No such file or directory" errors when one task deletes an entry the other
+# has already listed but not yet visited.
+CACHE_CLEANER_LOCK_FILE = '/tmp/prescat_cache_cleaner.lock'
+
+def with_cache_cleaner_lock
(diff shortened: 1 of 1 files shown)


lib/tasks/prescat.rake
```diff
@@ -1,5 +1,18 @@
...
+# "find: '...': No such file or directory" errors when one task deletes an entry the other
+# has already listed but not yet visited.
+CACHE_CLEANER_LOCK_FILE = '/tmp/prescat_cache_cleaner.lock'
+
+def with_cache_cleaner_lock
+  File.open(CACHE_CLEANER_LOCK_FILE, File::CREAT) do |lock_file|
+    lock_file.flock(File::LOCK_EX)
+    yield
+  end
+end
```
mjgiarlo, line 14: Should this method also be cleaning up the lock file or is that not necessary?

mjgiarlo (approved):
Approve pending consideration of question

lib/tasks/prescat.rake
justinlittman: I asked Claude the same thing. The file will continue to exist in `/tmp/` which seems fine. If the server crashes mid-cleanup the lock will be released. So I think this approach is OK as is.

mjgiarlo: I ran a quick smoke test locally and I can confirm that this works as intended.

